PR #7, opened 2020-02-12
Add view for Gnosis Safes

tschubotz:
I've checked that:

### DIFF
--- a/schema/gnosis_safe/view_safes.sql
+++ b/schema/gnosis_safe/view_safes.sql
@@ -1,15 +1,15 @@
 create materialized view gnosis_safe.view_safes as
     with safe_setups as (
-        select call_tx_hash as tx_hash, _threshold as threshold, LENGTH(REPLACE(text(_owners),'\\x','~'))-LENGTH(REPLACE(text(_owners),'\\x','')) as owners
+        select call_tx_hash as tx_hash, _threshold as threshold, array_length(_owners, 1) as owners
         from gnosis_safe."Safev0.1.0_call_setup"
         union all
-        select call_tx_hash as tx_hash, _threshold as threshold, LENGTH(REPLACE(text(_owners),'\\x','~'))-LENGTH(REPLACE(text(_owners),'\\x','')) as owners
+        select call_tx_hash as tx_hash, _threshold as threshold, array_length(_owners, 1) as owners
         from gnosis_safe."Safev1.0.0_call_setup"
         union ALL
-        select call_tx_hash as tx_hash, _threshold as threshold, LENGTH(REPLACE(text(_owners),'\\x','~'))-LENGTH(REPLACE(text(_owners),'\\x','')) as owners
+        select call_tx_hash as tx_hash, _threshold as threshold, array_length(_owners, 1) as owners
         from gnosis_safe."Safev1.1.0_call_setup"
         union ALL
-        select call_tx_hash as tx_hash, _threshold as threshold, LENGTH(REPLACE(text(_owners),'\\x','~'))-LENGTH(REPLACE(text(_owners),'\\x','')) as owners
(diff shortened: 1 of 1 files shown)


schema/gnosis_safe/view_safes.sql
```diff
@@ -0,0 +1,66 @@
+create materialized view gnosis_safe.view_safes as
+    with safe_setups as (
+        select call_tx_hash as tx_hash, _threshold as threshold, LENGTH(REPLACE(text(_owners),'\\x','~'))-LENGTH(REPLACE(text(_owners),'\\x','')) as owners
```
mewwts: Are you calculating the length of the `_owners` array here? I wonder if it would be better to just use `array_length(_owners, 1)`, unless there's a reason not to that I can't spot?

schema/gnosis_safe/view_safes.sql
```diff
@@ -0,0 +1,66 @@
...
+        left outer join safe_setups ss
+            on et.tx_hash = ss.tx_hash
+        left outer join ethereum.traces et2  -- for the potential proxy calls
+            on et.tx_hash = et2.tx_hash and et2.call_type = 'call' and left(text(et2."input"),10) in ('\x1688f0b9', '\x61b69abd') -- createProxyWithNonce or createProxy method call
+        where et.tx_success = True
+            AND left(text(et."input"),10) in ('\x0ec78d9e', '\xa97ab18a', '\xb63e800d') -- setup methods of v0.1.0, v1.0.0, v1.1.0=v.1.1.1
+            AND et.call_type = 'delegatecall' -- the delegate call to the master copy is the Safe address
+            AND et.to in ('\x8942595A2dC5181Df0465AF0D7be08c8f23C93af', '\xb6029ea3b2c51d09a50b53ca8012feeb05bda35a', '\xae32496491b53841efb51829d6f886387708f99b', '\x34cfac646f301356faa8b21e94227e3583fe3f5f') -- mastercopy address v0.1.0, v1.0.0, v1.1.0, v1.1.1
+    )
+    select creation_block_number, creation_time, address, "type"
```
mewwts: Consider using `safe_type` instead of `"type"` here?

schema/gnosis_safe/view_safes.sql
tschubotz: nope, no reasons, just changed it

tschubotz:
@mewwts Thanks for the suggestions - just fixed both

schema/gnosis_safe/view_safes.sql
```diff
@@ -0,0 +1,66 @@
...
+                else 'unknown'
+                end
+             as safe_type
+        from ethereum.traces et -- for the setup calls
+        left outer join safe_setups ss
+            on et.tx_hash = ss.tx_hash
+        left outer join ethereum.traces et2  -- for the potential proxy calls
+            on et.tx_hash = et2.tx_hash and et2.call_type = 'call' and left(text(et2."input"),10) in ('\x1688f0b9', '\x61b69abd') -- createProxyWithNonce or createProxy method call
+        where et.tx_success = True
+            AND left(text(et."input"),10) in ('\x0ec78d9e', '\xa97ab18a', '\xb63e800d') -- setup methods of v0.1.0, v1.0.0, v1.1.0=v.1.1.1
```
mewwts: I'm sorry I didn't see this on my first pass, but I think we can speed this up by using
`substring(et."input" for 4)` instead of `left(text(et."input"), 10)` throughout the query.

tschubotz: No worries at all. If you think this makes it faster, I'm happy to change it :)
I also chanced it in the 2 other places inside the select statement of the "safes" subquery

mewwts:
I merged this manually while adding a small change! Thanks @tschubotz. This should go live shortly

PR #9, opened 2020-02-25
Compund views - first iteration

hagaetc:
I've checked that:

* [x] the schemaname exists in Dune


mewwts:
https://youtu.be/vtEL0uKO4tA?t=58

PR #23, opened 2020-04-10
add numerai and erasure views

milkyklim:
I've checked that:


milkyklim:
I also want to add erasure bay views to track requests, burns and fulfillment. But it seems this table `erasure_v130.CountdownGriefingEscrow_call_initialize` doesn't contain all calls. 🤔 

This sql returns only 3 test requests but there definitely should be more:  
```
SELECT convert_from(metadata, 'UTF8')
FROM erasure_v130."CountdownGriefingEscrow_call_initialize"
LIMIT 10; 
```

Can anyone from Dune check the table actually has all data?

mewwts:
Thanks @milkyklim, I've checked and it seemed we missed this updated here https://github.com/erasureprotocol/erasure-protocol/pull/358/files. I've added it now!

milkyklim:
@mewwts can you also check that all of the following pattern
- `erasure_v130.CountdownGriefingEscrow_evt...` 
- `erasure_v130.CountdownGriefingEscrow_call...`

tables are actually up-to-date? 

E.g. it is strange that [`initialize`](https://github.com/erasureprotocol/erasure-protocol/blob/b2e586daec0014214f54ad9827e357ffec269586/contracts/escrows/CountdownGriefingEscrow.sol#L85) call is tracked but [`Initialized`](https://github.com/erasureprotocol/erasure-protocol/blob/b2e586daec0014214f54ad9827e357ffec269586/contracts/escrows/CountdownGriefingEscrow.sol#L144) event is not.

mewwts:
Thanks @milkyklim, working with the Erasure-team to get the correct ABI for CountdownGriefingEscrow.

hagaetc:
Thanks for doing this @milkyklim :) Note that the views should produce tables at per event level, i.e. not grouped by week or round. The idea is that those insights/groupings can be easily created on explore.duneanalytics.com once the lower-level data is cleaned.

milkyklim:
@hagaetc gotcha! I'll update current views to match event/call level today and add Erasure Bay views once `CountdownGriefingEscrow` ABI problem is fixed.

hagaetc:
@milkyklim also note that I'm not sure all relevant events are added in the current abstractions, I just asked Stephan from Erasure about it. Will let you know as soon as I hear from him.

milkyklim:
Force pushed to overwrite irrelevant views and converted PR to draft for now. 
I've collected all necessary events/views for _erasure bay_ but still have to refactor naming. Erasure Quant is next.

milkyklim:
@hagaetc this one https://github.com/duneanalytics/abstractions/pull/23/commits/21b7ba212eefe0651966fa99c6272087d446b6eb fully covers Erasure Bay flow. 

Takes care of 
- stake deposit;
- reward deposit;
- burn;
- data request;
- data submission. 

Will add numerai tourney related stuff once thegostep updates abi. Till then, can you have a look at the above mentioned commit?

hagaetc:
Hey @milkyklim thanks for this! 

* I suggest you name s.symbol as token_symbol 
* Could add token address as well from erc20.tokens table (could be useful for some joining with other tables)
* Might be worth adding the USD value of dai in addition to the DAI amount. [Here's an example](https://github.com/duneanalytics/abstractions/blob/master/schema/uniswap/view_eth_purchase.sql)
* A little bit of a shame that DAI is hardcoded, but unless there is a way to map tokenID to a token address onchain (?) I guess this is the best solution for now

Beyond that I think this is looking very nice!

milkyklim:
> map tokenID to a token address onchain (?)

@hagaetc some events don't have _any_ reference to token, in this case, DAI. I don't see clean way to introduce mapping there, therefore, will leave hardcoded values.

... 

Will update everything else today!

milkyklim:
Hey @mewwts @hagaetc this should be the final version for Erasure Bay, Erasure Quant and Numerai Tournament views. 

Only problem, it is impossible to track Erasure Quant rewards cause they are payed directly from the treasury hot wallet. Filtering by Quant stakers doesn't give a result that matches [leaderboard](https://erasurequant.com/leaderboard). 

Since thegostep also says that Quant "future is unclear", I leave progress on it as it is.

milkyklim:
Figured out the problem and added rewards for Erasure Quant!

Richard ruins some of the stats cause he is Quant participant, treasury and has some unrelated transfers: `0xb7cf1206e3003d7d3f2c7590a12f11b6a35ff0d9`.
 
Everything else matches Quant leaderboard!

milkyklim:
@mewwts updated the naming as discussed.

mewwts:
I merged this locally while adding a missing `AS`. Thanks a bunch for this - it's gonna be great! :+1:

PR #25, opened 2020-04-21
Materialized views

anxolin:
This PR:

* Renames the schema from `gnosis_dfusion` to `gnosis_protocol`
* Makes the tokens view a Materialized View
* Adds a new Materialized view for the trades, big bottleneck for all dashboards
* The SQL query for the trades is quite complex, but it was simplified a bit taking advantage of the new feature of reporting already the batch number, log index, and date in the events and calls

@mewwts I think this can be a great performance change. 
How can we ensure this views are updated? If you want, for the deployment we scheduled a periodic `REFRESH MATERIALIZED VIEW CONCURRENTLY <view>`

Then we can easily experiment with triggers, cause for example `view_tokens` only changes if `BatchExchange_call_addToken` changes. And that's doesn't happen so often


mewwts:
Hey @anxolin, this looks good!

You can add a
`SELECT cron.schedule('0 0 * * *', 'REFRESH MATERIALIZED VIEW gnosis_protocol.X')`
To the bottom of the files for now, and they will refresh midnight UTC.

anxolin:
Cool!

I've done in my last commit, I've used `CONCURRENTLY`, not an expert in postgresql, but my understanding is that it'll smart update the table using the Id. 

We only add rows to the trades on every solution submission, so most of the data won't change in a refresh. Also, I understood this won't block read operations

Let me know if I got `CONCURRENTLY` flag right and it's good to add it.

schema/gnosis_protocol/view_trades.sql
```diff
@@ -0,0 +1,85 @@
...
+)
+SELECT
+    trades.batch_id,
+    trades.owner as trader_hex,
+    trades.order_id,
+    trades.block_time,
+    trades.block_number,
+    reverts.block_time as revert_time,
+    sell_token."token" as sell_token,
+    sell_token."token" as sell_token,
```
mewwts, line 45: Duplicate column definition here.

schema/gnosis_protocol/view_trades.sql
```diff
@@ -0,0 +1,85 @@
...
+    sell_token."symbol" as sell_token_symbol,
+    sell_token."decimals" as sell_token_decimals,
+    trades.executed_sell_amount as sell_amount_atoms,
+    trades.executed_sell_amount / 10^(sell_token.decimals) as sell_amount,
+    buy_token.token as buy_token,
+    buy_token.symbol as buy_token_symbol,
+    buy_token.decimals as sell_token_decimals,
```
mewwts, line 52: this should be `buy_token_decimals`?

schema/gnosis_protocol/view_trades.sql
```diff
@@ -0,0 +1,85 @@
...
+    trades.order_id;
+
+
+CREATE UNIQUE INDEX IF NOT EXISTS view_trades_id ON gnosis_protocol.view_trades (batch_id, trader_hex, order_id) ;
+CREATE INDEX view_trades_1 ON gnosis_protocol.view_trades (batch_id);
+CREATE INDEX view_trades_2 ON gnosis_protocol.view_trades (sell_token_symbol);
+CREATE INDEX view_trades_3 ON gnosis_protocol.view_trades (sell_token);
+CREATE INDEX view_trades_4 ON gnosis_protocol.view_trades (buy_token_symbol);
+CREATE INDEX view_trades_5 ON gnosis_protocol.view_trades (buy_token);
+CREATE INDEX view_trades_6 ON gnosis_protocol.view_trades (trader_hex);
```
mewwts, line 81: This index is covered by the one on line 82 so can be ommitted.

schema/gnosis_protocol/view_trades.sql
anxolin: good catch

schema/gnosis_protocol/view_trades.sql
anxolin: I was not sure if it would use the index if you just filter by trader

PR #29, opened 2020-04-28
Add Aave aTokens view

mrdavey:
I've checked that:

* [x ] the view name starts with `view_`
* [x ] the filename matches the defined view and ends with .sql
* [x ] each file has only one view defined  


mrdavey:
Note that for aETH, the underlying asset is just ETH, not WETH.

mewwts:
I merged this with some formatting changes! https://github.com/duneanalytics/abstractions/blob/master/schema/aave/view_atokens.sql Thanks for your contribution :raised_hands:

PR #35, opened 2020-05-08
Create new view for price estimation in USD

anxolin:
Adds a new important view for Gnosis Protocol
The query is making an estimation on the USD price for every solution submitted by a solver.
For that it follows a strategy that will always give a price estimate.

It takes into account the solver prices, the external prices (prices.usd), and it has a strategy on selecting the best OWL price for every batch depending on the importance of the token.

**Regarding the name of the view**
I think for the other views, I should have named the table using a singular form, most people agree that is better. Also most of the tables defined in dune are also singular.

**Regarding the Index**
Regarding index, I added `token_id` althout it's in the PK, because as I understand, the index will not be used if you don't use the index in order, and I expect queries filtering by `token_id` but not `batch_id`.
https://www.postgresql.org/message-id/87F42982BF2B434F831FCEF4C45FC33E4207F755@EXCHANGE.corporate.connx.com

### DIFF
--- a/schema/gnosis_protocol/view_price_batch.sql
+++ b/schema/gnosis_protocol/view_price_batch.sql
@@ -1,0 +1,163 @@
+CREATE MATERIALIZED VIEW gnosis_protocol.view_price_batch AS
+WITH token_priorities AS (
+  SELECT * FROM (VALUES
+        (100, 7), -- DAI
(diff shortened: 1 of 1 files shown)


mewwts:
merged manually :+1:

PR #36, opened 2020-05-08
Add event information solution

anxolin:
This PR improves the trade view:

It still return the same number of registries but it:
* Adds a trade date
* Fixes an issue with the batch ID: It was not the correct number
* Return the events index for the trade and the solution. Becomes relevant to JOIN with the new table for prices
* Restructure the query a bit: Simplifies the last query, moves some parts to the temp "trades"

Aditionally, regarding the indexes, updates `view_trades_1`:
* Removes the index by `batch_id` since it's not needed (it's derived by the PK index)
* Adds an index by `trade_date`

### DIFF
--- a/schema/gnosis_protocol/view_trades.sql
+++ b/schema/gnosis_protocol/view_trades.sql
@@ -16,63 +16,66 @@ JOIN gnosis_protocol."BatchExchange_call_submitSolution" solution
 ),
 trades as (
 SELECT
-    solution."batchId" as batch_id,
-    trades.evt_block_time as block_time,
+	-- id
+    --solution."batchId" as batch_id,
+    floor(extract(epoch from solution.evt_block_time) / 300) - 1 AS batch_id, -- The event time tells us the batch. Between minute 0-4 is resolved batch N-1
+    trades."owner" AS trader_hex,    
     trades."orderId" as order_id,
-    trades."owner",
-    trades.evt_block_number as block_number,
-    trades.evt_index,
-    trades."buyToken" as buy_token_id,
-    trades."sellToken" as sell_token_id,
-    trades."executedBuyAmount" as executed_buy_amount,
-    trades."executedSellAmount" as executed_sell_amount,
-    trades.evt_tx_hash as tx_hash,
     RANK() OVER (
-        PARTITION BY solution."batchId", trades.owner, trades."orderId"
+        PARTITION BY solution.evt_block_time, trades.owner, trades."orderId"
         ORDER BY trades.evt_block_number, trades.evt_index
-    ) as trade_sub_id
-FROM gnosis_protocol."BatchExchange_evt_Trade" trades
-JOIN gnosis_protocol."BatchExchange_call_submitSolution" solution
-    ON solution.call_tx_hash=trades.evt_tx_hash
-    AND solution.call_success = true
-)
-SELECT
-    trades.batch_id,
-    trades.owner as trader_hex,
-    trades.order_id,
-    trades.trade_sub_id,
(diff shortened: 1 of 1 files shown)


schema/gnosis_protocol/view_trades.sql
```diff
@@ -16,72 +16,75 @@ JOIN gnosis_protocol."BatchExchange_call_submitSolution" solution
...
+
+
 CREATE UNIQUE INDEX IF NOT EXISTS view_trades_id ON gnosis_protocol.view_trades (batch_id, trader_hex, order_id, trade_sub_id);
-CREATE INDEX view_trades_1 ON gnosis_protocol.view_trades (batch_id);
+CREATE INDEX view_trades_1 ON gnosis_protocol.view_trades (trade_date);
 CREATE INDEX view_trades_2 ON gnosis_protocol.view_trades (sell_token_symbol);
 CREATE INDEX view_trades_3 ON gnosis_protocol.view_trades (sell_token);
 CREATE INDEX view_trades_4 ON gnosis_protocol.view_trades (buy_token_symbol);
 CREATE INDEX view_trades_5 ON gnosis_protocol.view_trades (buy_token);
 CREATE INDEX view_trades_6 ON gnosis_protocol.view_trades (trader_hex, order_id);
```
mewwts, line 86: Can you add a `_idx`-suffix to these?

schema/gnosis_protocol/view_trades.sql
```diff
@@ -16,72 +16,75 @@ JOIN gnosis_protocol."BatchExchange_call_submitSolution" solution
...
+    trades.trader_hex,
     trades.order_id,
     trade_sub_id;
 
 
+
+
 CREATE UNIQUE INDEX IF NOT EXISTS view_trades_id ON gnosis_protocol.view_trades (batch_id, trader_hex, order_id, trade_sub_id);
-CREATE INDEX view_trades_1 ON gnosis_protocol.view_trades (batch_id);
+CREATE INDEX view_trades_1 ON gnosis_protocol.view_trades (trade_date);
```
mewwts, line 81: Postgres complains trade_date does not exist here.

schema/gnosis_protocol/view_trades.sql
```diff
@@ -16,72 +16,75 @@ JOIN gnosis_protocol."BatchExchange_call_submitSolution" solution
...
     sell_token."decimals" as sell_token_decimals,
-    trades.executed_sell_amount as sell_amount_atoms,
-    trades.executed_sell_amount / 10^(sell_token.decimals) as sell_amount,
-    buy_token.token as buy_token,
+    -- sell amounts
+    trades."executedSellAmount" as sell_amount_atoms,
+    trades."executedSellAmount" / 10^(sell_token.decimals) as sell_amount,
+    -- buy token
+    trades."buyToken" as buy_token_id,
+    buy_token.token as buy_token,
```
mewwts, line 44: If not to much trouble would be great to be consistent and use uppercase `AS` throughout the file

schema/gnosis_protocol/view_trades.sql
anxolin: Strange, I see the field returned by the query:
image

Could it be that we need to use `DROP INDEX view_trades_1` first, since the index was previously defined?

mewwts: Yeah, I was surprised to, but will try now.
(WOW intellij does not like those bytearrays!)

mewwts: This was just a flaw in my testing setup (the index referenced the old view)

PR #39, opened 2020-05-14
Add erc20 tokens used by bancor

milkyklim:
I've checked that:


mewwts:
Thanks for sorting!

PR #46, opened 2020-05-28
Add Gnosis Protocol balance movements

anxolin:
Add `view_movement`, a new materialized view for Gnosis Protocol. 
The goal is to keep tracks of the user balances changes.

If a user deposit, withdraw or trade in a batch, it will show the amount of the operations (positive or negative) and the balance as a result

This table is basic for reproducing the balance in any given moment in time for each user and token.


mewwts:
Merged manually with change on line 111 to index `token_symbol`

PR #48, opened 2020-05-29
Add 0x views

alexkroeger:
Added several views to the `zeroex` schema as well as adding 0x to the dex trades table.

I've checked that:


mewwts:
Thanks @alexkroeger, I merged manually with some stylistic changes to the dex view.

PR #49, opened 2020-06-02
update kyber volume query

halink0803:
I've checked that:


mewwts:
Apart from some minor formatting that I can fix this looks goo to me. @hagaetc did you have a chance to check?

PR #50, opened 2020-06-03
Create known addresses table

anxolin:
Address the issue described in https://gnosisinc.slack.com/archives/CPPKMC5HN/p1591118391007000

--- 

For all dashboards that show info related to “organic trading” happening in Gnosis Protocol, we need to exclude some addresses from Market makers, and internal accounts.

For that, we start all our queries with the list of excluded addresses. This list is copy pasted over and over to dozens of queries.

This have 3 issues:
* It makes it very hard to maintain: Every time we need to add an address, we need to update all the queries
* Error prone: Since is copy & paste based, there could be differences in the queries
* Queries are unnecessarily complex

This is an example:
https://explore.duneanalytics.com/queries/3427/source#6765

If there’s no a better option, can we just move it to a public one?


schema/gnosis_protocol/known_address.sql
```diff
@@ -0,0 +1,422 @@
...
+  SELECT * FROM (VALUES
+    (decode('5a30ff01dc11223cff4e99d4263cb6f3aaa69e70', 'hex'), 'Solver'),
+    (decode('665316dabde5c5bc57ad0b2eed523447c5d2a570', 'hex'), 'Solver'),
+    (decode('36f7c5b5b8d66cec99e1282dd1c069648a0d4637', 'hex'), 'Solver'),
+    (decode('9ee11fccd3f692d1ddb281d81403b7e08b964c76', 'hex'), 'Solver'),
+    (decode('453ad119f26128034d3b5c2b6179b8b7f63ae1c7', 'hex'), 'Solver'),
+    (decode('0f833795b7597fcb7f22d8278c91ef63d441e949', 'hex'), 'Solver'),
+    (decode('f419c46fe10914cf5605be63b803a90cb3d4af82', 'hex'), 'Solver'),
+    (decode('E5a93F2Ffbc70c39154B09d6511Fc612d2f16DE4', 'hex'), 'Solver'),
+    (decode('5Ce64e024D54Eee4E6409829CF0fF8529E5da170', 'hex'), 'market maker'),
```
mewwts, line 14: Hi @anxolin, would you mind following the same format as we use here https://github.com/duneanalytics/abstractions/blob/90f8e803d4a15d25130e590f3447d92f2564a754/schema/erc20/tokens.sql#L1-L17

It has some benefits, among them being performant and easy to change and redeploy.

anxolin: Done

mewwts:
I made some changes, notably remove duplicates, add tabs, and fix index stmt before rebasing and merging locally. deployed this now. Thanks!

PR #51, opened 2020-06-03
update kyber volume query

halink0803:
I've checked that:

### DIFF
--- a/schema/dex/view_trades_alpha.sql
+++ b/schema/dex/view_trades_alpha.sql
@@ -100,30 +100,43 @@ FROM (
     UNION
 
     -- Kyber
-    SELECT
+    SELECT -- trade from Token - ETH
         t.evt_block_time AS block_time,
         'Kyber' AS "project",
         NULL AS version,
         trader AS trader_a,
         NULL::bytea AS trader_b,
-        "dstAmount" AS token_a_amount_raw,
-        "srcAmount" token_b_amount_raw,
-        CASE WHEN t.dest = '\xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee' THEN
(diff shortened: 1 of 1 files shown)


schema/dex/view_trades_alpha.sql
```diff
@@ -100,30 +100,43 @@ FROM (
...
+        NULL::bytea AS trader_b,
+        t."ethWeiValue" as token_a_amount_raw,
+        CASE WHEN t.dest IN ('\x5228a22e72ccc52d415ecfd199f99d0665e7733b') THEN 0 --ignore volume of token PT
+        ELSE t."dstAmount" END AS token_b_amount_raw,
+        '\xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2' AS token_a_address,
+        t.dest AS token_b_address,
+        t.contract_address exchange_contract_address,
+        t.evt_tx_hash AS tx_hash,
+        NULL::integer[] AS trace_address,
+        t.evt_index+1
```
mewwts, line 136: Curious why you're adding 1 to the `evt_index` here @halink0803? Feel free to elaborate on tg if more convenient.

halink0803: @mewwts Because here I want to separate the event into 2 parts, trade from token-token would be separate into token-(w)eth and (w)eth-token, but when do aggregate the volume, you guys will use `(tx_hash, trace_address and evt_index)` to group the trade: https://explore.duneanalytics.com/queries/4141/source. 

```sql
https://explore.duneanalytics.com/queries/4141/source
```

Therefore, I think the best option to separate the trade here to increase the evt_index. Do you have any other suggestion I should do to separate them here?

mewwts:
made some stylistic changes, rebased, and merged! Thanks :raised_hands:

PR #58, opened 2020-06-05
dex: Use lateral join for oasis 

mewwts:


hagaetc:
Benchmarking with Oasis' own volume stats this looks good!

PR #69, opened 2020-06-19
Make dex.trades

mewwts:
Created a table to replace the materialized view. 

- Added `trade_id` which increments for each trade made in a single `evt_index` or `trace_address`. This allows us to have unique indexed and we can be avoid inserting duplicate rows.


schema/dex/trades.sql
```diff
@@ -0,0 +1,476 @@
+DROP TABLE dex.trades;
+CREATE TABLE dex.trades (
+    block_time timestamptz NOT NULL, 
+    token_a_symbol text,
+    token_b_symbol text,
+    token_a_amount text,
+    token_b_amount text,
```
hagaetc, line 7: These should be `numeric`?

PR #71, opened 2020-06-22
Move bancor views to bancornetwork

milkyklim:
Resolves Bancor contract hell and updates all views to match new contract structure.


milkyklim:
@mewwts updated conversion as discussed – `prices` table is not left joined any more. Ready for merge.

PR #73, opened 2020-06-23
Add bancornetwork to dex.trades

mewwts:


schema/dex/trades.sql
```diff
@@ -446,7 +446,26 @@ WITH rows AS (
...
-        FROM gnosis_protocol."view_trades"
+        FROM gnosis_protocol.view_trades
+
+        UNION
+
+        -- Bancor Network
+        SELECT
+			block_time,
+			'Bancor Network' AS project,
+			NULL AS version,
```
mewwts, line 457: Used NULL here as the view contains all versions

schema/dex/trades.sql
```diff
@@ -446,7 +446,26 @@ WITH rows AS (
...
-        FROM gnosis_protocol."view_trades"
+        FROM gnosis_protocol.view_trades
+
+        UNION
+
+        -- Bancor Network
+        SELECT
+            block_time,
+            'Bancor Network' AS project,
+            NULL AS version,
```
mewwts, line 457: Used NULL as version here as view selects from multiple versions

mewwts:
@milkyklim WDYT

milkyklim:
LGTM!

schema/dex/trades.sql
```diff
@@ -446,7 +446,26 @@ WITH rows AS (
...
+            trader AS trader_a,
+            NULL::bytea AS trader_b,
+            target_token_amount_raw AS token_a_amount_raw,
+            source_token_amount_raw AS token_b_amount_raw,
+            target_token_address AS token_a_address,
+            source_token_address AS token_b_address,
```
milkyklim, line 463: Just realized that these might be wrong: should not `token_a` be the `source` and `token_b` be the `target`?

mewwts: @hagaetc wdyt?

mewwts:
Close in favor of #82

PR #74, opened 2020-06-23
Gnosis Protocol: divide amounts by 2

W3stside:
Divide amounts by 2 as we count 2 people in a trade as $1 each

I've checked that:


mewwts:
Thanks @W3stside

PR #79, opened 2020-07-01
Highlight deprecated sUSD and SNX as "old"

c3rnst:
As per Synthetix blog post, there is one legacy for sUSD and SNX, which might be useful to highlight for users. Please see : https://blog.synthetix.io/proxy-contract-cutover-on-may-10/


mewwts:
Hey @c3rnst, I think this is a great usecase for our new labels feature! Reach out on TG if you have any questions!

PR #84, opened 2020-07-17
Added MTA and mUSD tokens

alfetopito:
Hello Dune team,

I'm adding the MTA and mUSD tokens to the list.
There'll be the MTA IDO on Gnosis Protocol and would be super nice to show the token symbols in Dune dashboards.

For reference, this is the MTA token https://etherscan.io/token/0xa3bed4e1c75d00fa6f4e5e6922db7261b5e9acd2
And this is the mUSD token https://etherscan.io/token/0xe2f2a5c287993345a840db3b0845fbc70f5935a5

---

I don't think the bellow apply since this is not a new schema/table/etc. Simply adding 2 new tokens.

I've checked that:


mewwts:
deployed!

PR #86, opened 2020-07-20
add Yearn token

milkyklim:
Ref: https://ygov.finance

Also submitted 4 related contracts for parsing, hit me on telegram if anything is unclear.


mewwts:
thanks :raised_hands:

PR #93, opened 2020-07-30
Fix issue calculating the price when there's no price reference

anxolin:
Corrects an error in the price calculation for the case when there's no external price reference.
For that it uses the OWL price submitted by the solver, and assumes every OWL is worth 0.8$

This PR fixes an issue where it was multiplying the price 2 times for this case, making prices to be very off.
Also, it changes the OWL estimation to 0.95 since now it trades almost 1:1

Example of one case:
image

 Correction:
image

cc/ @fleupold


mewwts:
Deployed :+1:

anxolin:
Thanks Mats!

PR #99, opened 2020-08-05
dex: get synthetix trades

mewwts:
This PR adds a `usd_amount` to `dex.trades` to accomodate for dexes that calculate their USD-volume outside of joining with prices.usd. We also add Synthetix to `dex.trades` which is a notable such exchange, that uses their own oracles with prices synthesized in `synthetix.rates` to calculate USD amounts.

The volume queries in the dex dashboard should be altered to something like 
```sql
SUM(COALESCE(
        usd_amount,
        (SELECT token_a_amount * price FROM prices.USD WHERE date_trunc('minute', block_time) = minute AND token_a_address=contract_address),
        (SELECT token_b_amount * price FROM prices.USD WHERE date_trunc('minute', block_time) = minute AND token_b_address=contract_address)
    ))) AS usd_volume  -- Use token A when there's an USD price for it, else token B
```

WDYT @hagaetc @dvd-schwrtz?


hagaetc:
LGTM! Except volume queries are now written like 

```
WITH prices AS (                                                                                       
    SELECT minute,                                                         
        contract_address,                                                                              
        price
    FROM prices.usd                                                                                    
    WHERE minute > now() - interval '1 days'
)           

SELECT
    SUM(
        CASE WHEN token_a_address = a.contract_address THEN token_a_amount * a.price -- Use token A when there's USD price for it                    
        ELSE token_b_amount * b.price -- Else use token b                             
        END   
        ) AS usd_volume                                                                                
FROM dex."trades" t                                                                             
LEFT JOIN prices a ON date_trunc('minute', block_time) = a.minute AND token_a_address = a.contract_address 
LEFT JOIN prices b ON date_trunc('minute', block_time) = b.minute AND token_b_address = b.contract_address 
WHERE block_time > now() - interval '1 days'
;
```

So I guess changing it to

```
WITH prices AS (                                                                                       
    SELECT minute,                                                         
        contract_address,                                                                              
        price
    FROM prices.usd                                                                                    
    WHERE minute > now() - interval '1 days'
)           

SELECT
    SUM(
        CASE WHEN usd_amount IS NOT NULL THEN usd_amount
        ELSE token_a_address = a.contract_address THEN token_a_amount * a.price -- Use token A when there's USD price for it                    
        ELSE token_b_amount * b.price -- Else use token b                             
        END   
        ) AS usd_volume                                                                                
FROM dex."trades" t                                                                             
LEFT JOIN prices a ON date_trunc('minute', block_time) = a.minute AND token_a_address = a.contract_address 
LEFT JOIN prices b ON date_trunc('minute', block_time) = b.minute AND token_b_address = b.contract_address 
WHERE block_time > now() - interval '1 days'
;
```

dvd-schwrtz:
I added another PR for a view with the synth symbols tied to addresses that @mewwts asked for. Not 100% sure that a view is the format you wanted. Let me know!

In terms of the prices table, will the rates be added to the prices table? Otherwise, I don't understand the questions since prices.usd is used everywhere above, but doesn't have any of the rates data unless it has been added to that table?

mewwts:
Thanks @dvd-schwrtz.
As I mentioned on Telegram, ideally we'd have the rates in `prices.usd`-partitions, however doing that is slightly complex. The somewhat implicit "contract" of `prices.usd` is that you should be able to join with it is using `date_trunc('minute', my_time_variable) = prices.usd.minute`, which means that there are to things we would need to think about:

### Normalization 

 `prices.usd` should have an entry for **each** minute (`2020-01-01T00:00:00Z`, `2020-01-01T00:01:00Z`, `2020-01-01T00:03:00Z` etc) so the rates from the events would have to be normalized onto minutes. I.e. if the last `RatesUpdated`-event occured at `2020-01-01T00:06:36Z` with rate `x` , and the next at `2020-01-01T00:41:22Z` with rate `y`, that would have to be expanded into 

minute | price
--------- | -------
... | x
... | x
00:06 | x
00:07| y
00:08| y
... | y
00:41 | y

Addressing this in a plpgsql-function should not be that hard.

### Up to date
Another implicit detail of `prices.usd` is that it has prices for timestamps up to only a minute or two ago. I.e. if the time is now 00:40, we'd expect to find prices for a given asset for 00:38. This is hard because the `RatesUpdated` event arrives at spurious intervals, which means that if we did nothing to address this, user would either have to do a special join in order to use the last price for a given asset (quite complex) or live with the lack of recent prices. The only way I can think of that we could mitigate this, would be to run a function each minute using pg_cron that would 1. insert the new rates from a new `RatesUpdated`-event, and then fill in the last observed rate for all assets up until this minute - 3.

### All in all

All in all I think it would be cool to have synth rates in prices.usd, but it's a slightly larger process, and adding a `usd_amount` to `dex.trades` would simply be a quicker fix. It's also not unreasonable to think that other projects would want to rely on some other oracles to get prices instead of joining with our off-chain sources.

PR #120, opened 2020-09-09
Fix issue with price estimation

anxolin:
The VIEW `view_price_batch` has an issue estimating the USD price for tokens that are not in `erc20.tokens` table

For example, in this batch, since XSTAR token was not added, then it was not showing the estimation:
image

This PR adds 18 as a default for the calculation.


mewwts:
Merged this here https://github.com/duneanalytics/abstractions/commit/5104e1c8050c09e64662d93eedbc811da53be47d

PR #129, opened 2020-09-28
Add Curve hbtc and 3pool

pengiundev:


mewwts:
I'm sorry that this is hanging @pengiundev, I was looking to merge #127 first before checking this.

schema/curvefi/view_trades.sql
```diff
@@ -598,16 +598,72 @@ SELECT
         --change address back to renBTC's, right now Dune only tracks WBTC price
         WHEN bought_id = 0 THEN '\x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599'::bytea
         WHEN bought_id = 1 THEN '\x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599'::bytea
-        WHEN bought_id = 2 THEN '\xfE18be6b3Bd88A2D2A7f928d00292E7a9963CfC6'::bytea
+        WHEN bought_id = 2 THEN '\x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599'::bytea
```
mewwts, line 601: Does this have any effect backwards in time? e.g. will historical trades be misrepresented?

pengiundev: I looked at Dune prices tables and couldn't see the sbtc address there anymore so I changed with WBTC one, if the sbtc price table is there, better to keep it that way, but shouldn't have difference as both WBTC and sBTC are 1:1 with BTC

mewwts: Ah got it. Looks good to me then, I will merge and deploy in a batch with a few others later this week hopefully.

mewwts:
Really sorry about the delay here @pengiundev

mewwts:
image

This turned out not to work well. Any ideas on why curve volume is suddenly in trillions @pengiundev ? perhaps different number of decimals?

PR #139, opened 2020-10-27
Update tokens.sql

overanalyser:
Addition of REPv2 as not currently listed.

Added at line 398 

Based on https://etherscan.io/token/0x221657776846890989a759ba2973e427dff5c9bb

Note coingecko (and other ?) use REP for the same symbol https://www.coingecko.com/en/coins/augur,  but I;m using etherscan symbol as reference, as I don't want to change the mapping for REP.

I've checked that:


mewwts:
merged here https://github.com/duneanalytics/abstractions/commit/392d6a889b6a7d40860bdee96a19d744cc9691db and added tabs instead of spaces

PR #145, opened 2020-11-03
Hagaetc patch 30

mewwts:


hagaetc:
Might be worth adding a new column `token_symbol` so we can easily join/filter `prices.usd` with those symbols. There's currently some metadata (the A and B) in the details columns so this is not ideal to use for the above purpose.

mewwts:
You shouldn't use symbol to join with `prices.usd`.

PR #162, opened 2020-11-16
Add category `DEX` and `Aggregator`

hagaetc:
I've checked that:


mewwts (approved):

schema/dex/trades.sql
```diff
@@ -333,6 +346,7 @@ WITH rows AS (
             evt_block_time AS block_time,
             '0x' AS project,
             '3' AS version,
+            'DEX' AS category,
```
markusbkoch, line 349: Shouldn't this be 'Aggregator'? cc @hagaetc

hagaetc: This is a bit of a hard one - figuring out where all trades are originating from. We have Matcha as aggregator currently. Thanks for flagging we'll think about how we can make this even more precise and fair.

sui414: @markusbkoch  this should be the 'DEX' portion as this (v3) and the one above (v2) are only the volume settled onchain on our p2p exchange contracts that anyone can use and decodes the `Fill()` event only (just like uniswap's `swap()` ) , instead of the aggregator (0x API) layer.

the aggregator layer will include:
v2 v3 exchange contracts
+ our direct trades into sushi/uniswap
(see this [0x api fills](https://github.com/duneanalytics/abstractions/blob/master/ethereum/zeroex/view_0x_api_fills.sql) )

and matcha is just one of the affiliates under 0x api.

markusbkoch: Thank you, @sui414. I think I see the problem: I started this conversarion in a commit where the problem didn't exist yet 🤦‍♂️ 

It used to be the case that the 0x/v3 entries in this table referred to the exchange only. 
https://github.com/duneanalytics/abstractions/blob/63f7cff2dc4b208f403fe3e20c0154b4d9c11536/schema/dex/trades.sql#L361

But IIUC in ef6b5bb3 it now consolidates all API fills
https://github.com/duneanalytics/abstractions/blob/ef6b5bb326a6cebe1b8a4859d14d823f6c3122df/ethereum/dex/trades.sql#L365

It seems the majority of the trades in 0x/v3 entries are being double counted: once for 0x/v3 and once for the underlying project supplying liquidity
https://explore.duneanalytics.com/queries/19220/source

So I believe this line should be changed to 'Aggregator'
https://github.com/duneanalytics/abstractions/blob/ef6b5bb326a6cebe1b8a4859d14d823f6c3122df/ethereum/dex/trades.sql#L353

Does that make sense?

sui414: yeah thanks for pointing that out. I will align this with our team & make changes soon 👍

mewwts: Hey @markusbkoch and @sui414, do let me or @milkyklim know if we can help with anything! :)

PR #175, opened 2020-12-09
balancer pools labels

markusbkoch:
Adding balancer pool labels (tokens and their weights)

### DIFF
--- a/labels/queries/balancer_pools.sql
+++ b/labels/queries/balancer_pools.sql
@@ -41,7 +41,7 @@ SELECT
   pool as address, 
   CONCAT(string_agg(symbol, '/'), ' ', string_agg(cast(norm_weight as text), '/')) AS label,
   'balancer_pool' AS type,
-  'markusbkoch' as author
+  'balancerlabs' as author
 
 FROM   (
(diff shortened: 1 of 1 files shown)


labels/queries/balancer_pools.sql
```diff
@@ -0,0 +1,52 @@
...
+    where call_success
+),
+state_with_gaps as (
+    select events.block_number, events.pool, events.token, events.denorm,
+    LEAD(cast(events.block_number as text), '1', '99999999') over (partition by events.pool, events.token order by events.block_number) as next_block_number
+    from events 
+), 
+settings as (
+    select pool, 
+    coalesce(t.symbol,'???') as symbol,
```
mewwts, line 32: Would it be better to inner join (i.e. only have labels with both symbols), or do you think ???/YFI is fine for now?

mewwts (approved):

labels/queries/balancer_pools.sql
markusbkoch: I think it's fine for now. It helps distinguish between 2/3/4... token pools.

mewwts:
Expecting to have this deployed in ~ a wekk @markusbkoch

mewwts:
Running this in dev now @markusbkoch

PR #183, opened 2020-12-14
dex.trades: add usd prices

milkyklim:
I've checked that:


mewwts (approved):

hagaetc:
This might be too late I know, but would be great if `usd_amount` was all the way on top after block_time so it is easy to see when doing `SELECT * FROM dex.trades`

PR #209, opened 2021-01-12
fix Balancer trader_a

markusbkoch:
For consistency with other DEXs - caller is most often an intermediate smart contract, not the actual trader

I've checked that:

### DIFF
--- a/ethereum/dex/trades.sql
+++ b/ethereum/dex/trades.sql
@@ -657,7 +657,7 @@ WITH rows AS (
             'Balancer' AS project,
             '1' AS version,
             'DEX' AS category,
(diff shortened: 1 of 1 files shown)


markusbkoch:
Balancer doesn't log the actual trader in the event, only the msg.sender, which is usually some aggregator. So getting the actual trader would require a join with the transactions table. This sets trader_a to NULL as a quick fix, since the outer query coalesces to tx.from, thus avoiding an additional join with the transactions table in the core Balancer query. Kyber_V2 seems to be an equivalent case.

milkyklim:
Dropped in favor: https://github.com/duneanalytics/abstractions/pull/217

PR #229, opened 2021-02-03
Add token prices for a number of tokens

tschubotz:
I've 

- [x] Found the asset on coinpaprika. Noted the symbol
- [x] Noted the ID of the asset from coinpaprikas URL. 
- [x] If the asset is a token on Ethereum, find the contract address and number of decimals e.g. through etherscan.


mewwts (approved):

mewwts:
Thanks @tschubotz, deployed this now :+1:

PR #232, opened 2021-02-03
prices: derive token prices from dex.trades

markusbkoch:
There's a long tail of tokens for which there's no price feed in coinpaprika and therefore are not available via `prices.usd`. The goal with this materialized view is to provide performant access to estimated prices of tokens based on trades occurring in multiple DEXes (`dex.trades`). 

It indirectly uses the curated list of tokens in `erc20.tokens` because it only selects trades where `token_amount>0` (`token_amount` is computed from the `decimals` in `erc20.tokens` and will be `NULL` if the token is not among the curated ones).

Prices are aggregated based on the median so as to be more resistant to the influence of extreme prices. An additional field `number_of_trades` can help users eliminate infrequently traded tokens from their queries that make use of this materialized view, as for those the median would also risk being quite innacurate.

We have deployed a version of this materialized view in `dune_user_generated.view_token_prices` and [reimplemented the Balancer TVL](https://explore.duneanalytics.com/queries/19109) query making use of it as a complement to the information coming from `prices.usd`. This will be particularly relevant once wPE is added to `erc20.tokens` (#231).

Future iterations of this view might include estimating prices for hours where no trades occurred (forward filling? linear interpolation?) and adding different measures as price estimators (time weighted average, volume weighted average, etc). But in its current version I believe it will be a value-add to the Dune community.

I've checked that:

### DIFF
--- a/ethereum/dex/view_token_prices.sql
+++ b/ethereum/dex/view_token_prices.sql
@@ -31,7 +31,7 @@ CREATE MATERIALIZED VIEW dex.view_token_prices AS (
         date_trunc('hour', block_time) as hour,
         contract_address,
         (PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY price)) AS median_price,
-        count(1) AS number_of_trades
+        count(1) AS sample_size
     FROM dex_trades
     GROUP BY 1, 2
     HAVING count(1) >= 2
--- a/ethereum/dex/view_token_prices.sql
+++ b/ethereum/dex/view_token_prices.sql
@@ -3,7 +3,11 @@ BEGIN;
 DROP MATERIALIZED VIEW IF EXISTS dex.view_token_prices;
 
 CREATE MATERIALIZED VIEW dex.view_token_prices AS (
-    WITH dex_trades AS (
+    WITH tokens_in_prices_usd AS (
+        SELECT DISTINCT contract_address 
+        FROM prices.usd
+        WHERE minute > now() - interval '7 days'
+    ), dex_trades AS (
         SELECT 
             token_a_address as contract_address, 
             usd_amount/token_a_amount as price,
@@ -13,7 +17,7 @@ CREATE MATERIALIZED VIEW dex.view_token_prices AS (
         AND usd_amount  > 0
         AND category = 'DEX'
         AND token_a_amount > 0
-        AND token_a_address NOT IN (SELECT DISTINCT contract_address FROM prices.usd)
+        AND token_a_address NOT IN (SELECT contract_address FROM tokens_in_prices_usd)
         
         UNION ALL 
         
@@ -26,7 +30,7 @@ CREATE MATERIALIZED VIEW dex.view_token_prices AS (
         AND usd_amount  > 0
         AND category = 'DEX'
         AND token_b_amount > 0
-        AND token_b_address NOT IN (SELECT DISTINCT contract_address FROM prices.usd)
+        AND token_b_address NOT IN (SELECT contract_address FROM tokens_in_prices_usd)
     ) 
 
     SELECT 
--- a/ethereum/dex/view_token_prices.sql
+++ b/ethereum/dex/view_token_prices.sql
@@ -34,7 +34,6 @@ CREATE MATERIALIZED VIEW dex.view_token_prices AS (
         count(1) AS sample_size
     FROM dex_trades
     GROUP BY 1, 2
-    HAVING count(1) >= 2
 );
 
 CREATE UNIQUE INDEX IF NOT EXISTS dex_token_prices_unique ON dex.view_token_prices (hour, contract_address);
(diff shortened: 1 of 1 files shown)


0xBoxer:
I made the effort of building a query to compare the deltas within assets that are actually listed in the pricefeed. See this [dashboard](https://explore.duneanalytics.com/dashboard/dex-trades-delta?p_w15016_ADDRESS=0x0bc529c00c6401aef6d220be8c6ea1667f6ad93e&p_w15015_ADDRESS=0x0bc529c00c6401aef6d220be8c6ea1667f6ad93e&p_w15014_ADDRESS=0x0bc529c00c6401aef6d220be8c6ea1667f6ad93e&p_w15013_ADDRESS=0x0bc529c00c6401aef6d220be8c6ea1667f6ad93e) and try it out with different listed assets.

The results are reasonable from what I can tell and I will start working with the pricefeed more broadly if nobody raises any concerns.

Thanks for coming up with this Markus 👍

mewwts:
Looks good to me, I will just create some supporting indexes before merging this.

markusbkoch:
Thanks @mewwts. I've made a few changes after valuable feedback from @sui414 🙏

hagaetc:
Very nice @markusbkoch - thanks for doing this!

I did some testing. Seems like it's most performant to query for contract addresses once:

```
WITH price_address AS (

SELECT DISTINCT contract_address FROM prices.usd

)

 SELECT 
            token_a_address as contract_address, 
            usd_amount/token_a_amount as price,
            block_time
        FROM dex.trades
        WHERE 1=1
        AND usd_amount  > 0
        AND category = 'DEX'
        AND token_a_amount > 0
        AND token_a_address NOT IN (SELECT contract_address FROM price_address)

        UNION ALL 

        SELECT 
            token_b_address as contract_address, 
            usd_amount/token_b_amount as price,
            block_time
        FROM dex.trades
        WHERE 1=1
        AND usd_amount  > 0
        AND category = 'DEX'
        AND token_b_amount > 0
        AND token_b_address NOT IN (SELECT contract_address FROM price_address)
       
```
        
Since tokens are seldom dropped from `prices.usd` it might even be worth to only look at last 24 h or last 7 days, makes the query even more performant:
        
```
        WITH price_address AS (

SELECT DISTINCT contract_address FROM prices.usd WHERE minute > now() - interval '7 days'

)
```

ethereum/dex/view_token_prices.sql
```diff
@@ -0,0 +1,46 @@
...
+DROP MATERIALIZED VIEW IF EXISTS dex.view_token_prices;
+
+CREATE MATERIALIZED VIEW dex.view_token_prices AS (
+    WITH dex_trades AS (
+        SELECT 
+            token_a_address as contract_address, 
+            usd_amount/token_a_amount as price,
+            block_time
+        FROM dex.trades
+        WHERE 1=1
```
hagaetc, line 12: what's the purpose of this line? @markusbkoch

hagaetc:
Also, since it excludes prices in `prices.usd` users will need to do 2 joins for USD prices which is not ideal... might it be worth including all tokens in `erc20.tokens` in this table directly? 🤔 Not sure what makes most sense

ethereum/dex/view_token_prices.sql
markusbkoch: merely aesthetics 🙈 😄 don't mind dropping it

markusbkoch:
> Also, since it excludes prices in `prices.usd` users will need to do 2 joins for USD prices which is not ideal... might it be worth including all tokens in `erc20.tokens` in this table directly? 🤔 Not sure what makes most sense

Good point. As a reference point, this is how I'm using it: I `UNION` this with `prices.usd` before joining it in my actual query.

```
WITH prices AS (
        SELECT date_trunc('day', minute) AS day, contract_address AS token, AVG(price) AS price
        FROM prices.usd
        GROUP BY 1, 2
        
        union all
        
        select date_trunc('day', hour) AS day, contract_address AS token, AVG(median_price) AS price
        FROM dune_user_generated.view_token_prices
        WHERE number_of_trades > 8
        GROUP BY 1, 2
    )
SELECT [...]
```

markusbkoch:
> I did some testing. Seems like it's most performant to query for contract addresses once:
> Since tokens are seldom dropped from `prices.usd` it might even be worth to only look at last 24 h or last 7 days, makes the query even more performant:

Awesome suggestions, thank you!

mewwts:
@markusbkoch @hagaetc let me know when we should merge this!

hagaetc:
I guess the `UNION` thinking makes sense indeed - better to have this table as more of a backup/addon than default for now @markusbkoch :) Feel free to merge @mewwts @milkyklim

mewwts (approved):

mewwts:
@milkyklim feel free to merge at convenience

PR #238, opened 2021-02-10
dex.trades: add 1inch Liquidity Protocol swaps to dex.trades table

k06a:
I've checked that:

### DIFF
--- a/ethereum/dex/trades/insert_1inch_lp.sql
+++ b/ethereum/dex/trades/insert_1inch_lp.sql
@@ -1,4 +1,4 @@
-CREATE OR REPLACE FUNCTION dex.insert_mooniswap(start_ts timestamptz, end_ts timestamptz=now(), start_block numeric=0, end_block numeric=9e18) RETURNS integer
+CREATE OR REPLACE FUNCTION dex.insert_1inch_lp(start_ts timestamptz, end_ts timestamptz=now(), start_block numeric=0, end_block numeric=9e18) RETURNS integer
 LANGUAGE plpgsql AS $function$
 DECLARE r integer;
 BEGIN
@@ -55,28 +55,28 @@ WITH rows AS (
         evt_index,
         row_number() OVER (PARTITION BY tx_hash, evt_index, trace_address) AS trade_id
     FROM (
-        -- Mooniswap
+        -- 1inch LP
         SELECT
-            evt_block_time as block_time,
-            'Mooniswap' AS project,
+            evt_block_time,
+            '1inch LP' AS project,
             '1' AS version,
             'DEX' AS category,
-            account AS trader_a,
+            sender AS trader_a,
             NULL::bytea AS trader_b,
             result AS token_a_amount_raw,
             amount AS token_b_amount_raw,
             NULL::numeric AS usd_amount,
-            CASE WHEN dst = '\x0000000000000000000000000000000000000000' THEN 
-                '\xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'::bytea ELSE dst
+            CASE WHEN "dstToken" = '\x0000000000000000000000000000000000000000' THEN 
+                '\xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'::bytea ELSE "dstToken"
             END AS token_a_address,
-            CASE WHEN src = '\x0000000000000000000000000000000000000000' THEN 
-                '\xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'::bytea ELSE src
+            CASE WHEN "srcToken" = '\x0000000000000000000000000000000000000000' THEN 
+                '\xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE'::bytea ELSE "srcToken"
             END AS token_b_address,
             contract_address AS exchange_contract_address,
             evt_tx_hash AS tx_hash,
             NULL::integer[] AS trace_address,
             evt_index
-        FROM mooniswap."MooniSwap_evt_Swapped"
+        FROM onelp."Mooniswap_evt_Swapped"
     ) dexs
     INNER JOIN ethereum.transactions tx
         ON dexs.tx_hash = tx.hash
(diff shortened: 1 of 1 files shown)


ethereum/dex/trades.sql
```diff
@@ -655,6 +655,27 @@ WITH rows AS (
...
+            result AS token_a_amount_raw,
+            amount AS token_b_amount_raw,
+            NULL::numeric AS usd_amount,
+            dstToken AS token_a_address,
+            srcToken AS token_b_address,
+            contract_address AS exchange_contract_address,
+            evt_tx_hash AS tx_hash,
+            NULL::integer[] AS trace_address,
+            evt_index
+        FROM onelp."MooniSwap_evt_Swapped"
```
markusbkoch, line 675: FYI the table name is Mooniswap, lower case s

ethereum/dex/trades.sql
```diff
@@ -655,6 +655,27 @@ WITH rows AS (
...
+            evt_block_time,
+            '1inch LP' AS project,
+            '1' AS version,
+            'DEX' AS category,
+            sender AS trader_a,
+            NULL::bytea AS trader_b,
+            result AS token_a_amount_raw,
+            amount AS token_b_amount_raw,
+            NULL::numeric AS usd_amount,
+            dstToken AS token_a_address,
```
markusbkoch, line 669: camelCase variable names must be in quotes

ethereum/dex/trades.sql
```diff
@@ -655,6 +655,27 @@ WITH rows AS (
 
         UNION ALL
 
+        -- 1inch LP
+        SELECT
+            evt_block_time,
+            '1inch LP' AS project,
+            '1' AS version,
+            'DEX' AS category,
+            sender AS trader_a,
```
markusbkoch, line 664: You might want to keep this as NULL so that the outer query coalesces it back to transaction sender. As is this will get you the 1inch router as `trader_a` most of the time. See:
https://github.com/duneanalytics/abstractions/blob/e486226de2019b13aab0734f64167bcc06dc1d19/ethereum/dex/trades.sql#L685

ethereum/dex/trades.sql
k06a: Thx, fixed!

ethereum/dex/trades.sql
k06a: Thx, fixed!

ethereum/dex/trades.sql
k06a: 1inch Router is not only called of 1inch Liquidity Protocol swap, wdyt?

ethereum/dex/trades.sql
markusbkoch: Also, one of "dstToken" and "srcToken" seems to be always `0`. Why is that? Maybe you need to preprocess the events somehow to conform them to this table?

ethereum/dex/trades.sql
markusbkoch: What I mean is that as is this will get you a lower count of unique actual traders if one was to base that count on the unique addresses in the `trader_a` field, because from the events's POV the `sender` is the router, not the EOA that started the trasaction. And you probably want the latter in this table.

ethereum/dex/trades.sql
```diff
@@ -655,6 +655,27 @@ WITH rows AS (
...
+            NULL::bytea AS trader_b,
+            result AS token_a_amount_raw,
+            amount AS token_b_amount_raw,
+            NULL::numeric AS usd_amount,
+            "dstToken" AS token_a_address,
```
markusbkoch, line 669: One of `dstToken` or `srcToken` seems to be always `0`. Why is that? Maybe you need to preprocess the events somehow to conform them to this table? The goal is that `token_a` is the token the trader is buying and `token_b` is the token the trader is selling to the pool.

markusbkoch: Oh I see, the pools hold ETH, not WETH

markusbkoch: In this case, I would suggest an analogous approach to what is done for eg Uniswap v1 and Kyber in this table: replace `0` with `\xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 ` (WETH)

k06a: `0` means ETH as well as `0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee`

markusbkoch: Yea, implementing the same `CASE` that was done for Bancor should be enough to make sure the 1inch data is properly accounted for in `dex.trades` and all the queries and dashboards that use it and expect WETH in the `token_address` fields.

k06a: Fixed for both 1LP and Mooniswap

ethereum/dex/trades.sql
k06a: But this is true num of users when 1inch uses 1LP - it is single user. Else outer statistics should rely on tx_from. I will build separate dashboard for nested volumes.

markusbkoch (approved):

k06a:
Can we have ETH for 1LP and Mooniswap because they support it natively? Don't wanna loose this stats.

markusbkoch:
> Can we have ETH for 1LP and Mooniswap because they support it natively? Don't wanna loose this stats.

I agree the stats would be clearer in some ways. But in other ways we would loose information unless further changes were made. For example, the `usd_amount` is determined by joining the trades subquery with the `prices.usd` table on the `contract_address`. If we have ETH instead of WETH that information is lost and volume is (slightly?) underestimated.
https://github.com/duneanalytics/abstractions/blob/657e7add78c2b47f47e5a9a7a60f7bca0a089315/ethereum/dex/trades.sql#L836-L843

Not there aren't other ways to solve this and get the `usd_amount` for trades between ETH and some unpriced asset, but until something else was implemented I'd say the least bad option is classifying the trades as WETH even though we know that not to be the case. But that's just my 2 cents.

k06a:
Check out #206 for ETH prices

markusbkoch:
I don't agree with all the changes proposed on #206, but I agree there's value in joining with the price of ETH when applicable

mewwts:
Hey, just merged changes to the structure of dex. Can you rebase / add changes on top of master? Thanks :pray:

k06a:
@mewwts done!

milkyklim (approved):

PR #247, opened 2021-02-22
balancer: add view balances

mendesfabio:
I'm adding this view to make it easier to calculate Balancer TVL. This is the first materialized view I've done and all feedback is greatly welcome!

I've checked that:


milkyklim (approved):
Fixed indices cause none of them is actually unique. :)

PR #262, opened 2021-03-08
add a contract label from contract creations

sui414:
I've checked that:

### DIFF
--- a/labels/ethereum/account_type.sql
+++ b/labels/ethereum/account_type.sql
@@ -1,7 +1,7 @@
 SELECT DISTINCT
     address,
-    'Contract' AS label,
-    'Account Type' AS type,
+    'contract' AS label,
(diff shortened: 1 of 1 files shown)


sui414:
if we can get Genesis transfer data on dune -- then we can complete this type of label with 2 more values:

- genesis address
- eoa

sui414:
another comment about the time filter -- ideally it would be more efficient to be an appending style with timefilter, but based on my observations, 1 contract address can be created for multiple times with or without suicide in between.

to avoid duplicate records, adding a time filter will require adding another check to the `labels.labels` table to check if that contract address already exist or not -- not sure if this will be efficient.

mewwts:
thanks @sui414, making some changes here https://github.com/duneanalytics/abstractions/pull/268

PR #308, opened 2021-04-12
adding $ to decentral games ticker

josephd8:
I've checked that:


josephd8:
The official ticker for decentral games token is $DG, not just DG. Now there is another DG token that is causing issues with some of our queries.

mewwts (approved):

PR #324, opened 2021-04-27
Added prices.layer1_usd_btc view

va3093:
I've checked that:


mewwts (changes requested):
This needs a second look

mewwts (approved):
LGTM, perhaps be consisten in the use of UPPER or lower case for SQL keywords?

PR #327, opened 2021-04-28
Adding token balances script

Vasilije1990:
I've checked that:


ethereum/token_balances/token_balances.sql
```diff
@@ -0,0 +1,202 @@
...
+
+
+DO $$
+DECLARE
+   hour_   text;
+   arr varchar[];
+BEGIN
+  select array(select generate_series('2021-03-21', '2021-03-23', '1 hour'::interval)::timestamptz)
+    into arr;
+   FOREACH hour_  IN   array arr
```
mewwts, line 26: Can this `DO` just be a `SELECT token_balances.insert_hourly(start_ts, end_ts)`?

mewwts (commented):
Nice!

ethereum/token_balances/token_balances.sql
```diff
@@ -0,0 +1,202 @@
...
+		, "historical_balances" as (
+			select
+				s.address as address
+				, s.contract_address as token_address
+				, s.rawAmount
+		    from sandbox.token_balances_proposal_1 s
+		    join "transfer_events" te
+		    on te.address = s.address
+		    and s.contract_address = te.token_address
+		    where s.ts < hour_::timestamptz
```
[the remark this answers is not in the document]
Vasilije1990: Yup, makes sense. The filter is too broad and the query would get slower over time.

ethereum/token_balances/token_balances.sql
Vasilije1990: Makes sense, I thought along the same lines.

PR #348, opened 2021-05-14
Add DEX Liquidity

mendesfabio:
This PR adds a new materialized view for querying the liquidity of DEX protocols. Like `dex.trades` help us to compare volume between exchanges, `dex.liquidity` aims to make it easier to query another important DeFi metric: TVL.

Balances are calculated based on incoming and outcoming ERC20 transfers of the pool addresses of each protocol. Then, the balances are aggregated by pool and token, and the data is truncated per day. 

Both `prices.usd` and `view_token_prices` are used as price feed to make results more accurate. Also, token weight has the same intention, since we can use it to estimate the pool's liquidity even if we only have the balance in USD for one token.

All feedback is greatly welcomed!

I've checked that:

\* I've checked the query works using dune_user_generated then I changed all functions/tables names manually. I hope I didn't leave some typo. 😅

### DIFF
--- a/ethereum/dex/trades/insert_1inch.sql
+++ b/ethereum/dex/trades/insert_1inch.sql
@@ -223,7 +223,7 @@ SELECT dex.insert_1inch(
     '2021-01-01',
     now(),
     (SELECT max(number) FROM ethereum.blocks WHERE time < '2021-01-01'),
-    (SELECT max(number) FROM ethereum.blocks)
+    SELECT MAX(number) FROM ethereum.blocks where time < now() - interval '20 minutes'
 )
 WHERE NOT EXISTS (
     SELECT *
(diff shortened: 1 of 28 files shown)


ethereum/dex/liquidity/insert_uniswap.sql
```diff
@@ -0,0 +1,209 @@
...
+        token_symbol,
+        0.5 AS token_weight,
+        delta_token_amount,
+        delta_token_amount_raw,
+        delta_usd_amount,
+        token_amount,
+        token_amount_raw,
+        usd_amount,
+        'Uniswap' AS project,
+        '2' AS version,
```
Stefenon, line 165: Replace with `version` only.

0xBoxer:
Hey @mendesfabio and @Stefenon, thanks for taking up the initiative on this.

We are working on an easier way to query balances behind the scenes currently and therefore will not merge this until that is deployed in prod. We don't want to incur tech debt on ourselves that is easily avoidable.

Sorry for the delay, but we will for sure you this as starting point and are immensely thankful for your efforts 👍

markusbkoch:
Thanks @0xBoxer . In that case we'll move the pieces of this logic that apply to Balancer to the `balancer` schema if that's ok, so as to speed up the queries that involve monitoring TVL across V1 and V2 and across pools. Sounds good?

0xBoxer:
Sure thing 👍

PR #377, opened 2021-06-07
Update prices.ini with 2 new tokens

oceanbluu:
Both tokens are trading on uniswap, and be great if their dex prices become available in Dune

I've checked that:


0xBoxer (changes requested):
both of the ids do not exist on coinpaprika.

PR #382, opened 2021-06-12
include HOPR in price

QYuQianchen:
I've checked that:


0xBoxer (changes requested):
Hopr doesn't have price data on Coinpaprika and the erc20 address is wrong.

Please contact coinpaprika to include price data for hopr and change the erc20 address.

PR #398, opened 2021-06-24
Add views for token balances

aalan3:
I've checked that:

Results from the views can be found here: https://duneanalytics.com/queries/68306 https://duneanalytics.com/queries/68312

### DIFF
--- a/ethereum/erc20/view_token_balances.sql
+++ b/ethereum/erc20/view_token_balances.sql
@@ -4,8 +4,8 @@ wallet_address,
 token_address,
 t.symbol as token_symbol,
 amount_raw,
-amount_raw / 10^coalesce(t.decimals, 0) amount_formatted,
-amount_raw / 10^coalesce(t.decimals, 0) * p.price amount_usd,
+amount_raw / 10^coalesce(t.decimals, null) amount,
+amount_raw / 10^coalesce(t.decimals, null) * p.price amount_usd,
 generate_series(min(timestamp), date_trunc('hour', NOW()), '1 hour') as timestamp
 FROM erc20.token_balances
 left join erc20.tokens t on t.contract_address = token_address
--- a/ethereum/erc20/view_token_balances.sql
+++ b/ethereum/erc20/view_token_balances.sql
@@ -1,0 +1,13 @@
+CREATE OR REPLACE VIEW erc20.view_token_balances AS
+SELECT 
+wallet_address,
+token_address,
+t.symbol as token_symbol,
+amount_raw,
+amount_raw / 10^coalesce(t.decimals, 0) amount_formatted,
+amount_raw / 10^coalesce(t.decimals, 0) * p.price amount_usd,
+generate_series(min(timestamp), date_trunc('hour', NOW()), '1 hour') as timestamp
+FROM erc20.token_balances
+left join erc20.tokens t on t.contract_address = token_address
+left join prices.usd p on p.contract_address = token_address and p.minute = timestamp
+group by 1, 2, 3, 4, 5, 6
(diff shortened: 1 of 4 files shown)


ethereum/erc20/view_token_balances_latest.sql
```diff
@@ -0,0 +1,13 @@
...
+wallet_address,
+token_address,
+t.symbol as token_symbol,
+amount_raw,
+amount_raw / 10^coalesce(t.decimals, 0) amount_formatted,
+amount_raw / 10^coalesce(t.decimals, 0) * p.price amount_usd,
+timestamp
+FROM erc20.token_balances
+left join erc20.tokens t on t.contract_address = token_address
+left join prices.usd p on p.contract_address = token_address and p.minute = date_trunc('hour', NOW())
```
aalan3, line 12: Note: here the price for `NOW()` is fetched for the current hour, so it becomes for example `2021-06-24 15:00:00`

ethereum/erc20/view_token_balances.sql
```diff
@@ -0,0 +1,13 @@
+CREATE OR REPLACE VIEW erc20.view_token_balances AS
+SELECT 
+wallet_address,
+token_address,
+t.symbol as token_symbol,
+amount_raw,
+amount_raw / 10^coalesce(t.decimals, 0) amount_formatted,
```
hagaetc, line 7: Should maybe be `amount_raw / 10^coalesce(t.decimals, NULL) amount_formatted,`?

Think it is a bit weird when `amount_formatted` is actually raw 
image

Would also maybe name this column just `amount`

ethereum/erc20/view_token_balances_latest.sql
hagaetc: The timestamp is for the last transfer while the price is current? That seems a little weird to me. Might be better to remove timestamp or have it as the same as the price?

https://duneanalytics.com/queries/68312

ethereum/erc20/view_token_balances.sql
aalan3: Yea that makes sense I'll update

ethereum/erc20/view_token_balances_latest.sql
aalan3: Probably good idea to just remove it. Was thinking about naming it something else but I think since price is there it will be confusing, can be added later if needed.

hagaetc: hmm I think it could be useful. Could maybe name it `last_transfer_timestamp` and then `current_usd_value`. Maybe @0xBoxer has any thoughts?

0xBoxer: `last_transfer_timestamp` seems like something people can make good use of.
IF we can agree on that name I don't think there is a need to rename the `amount_usd` column.

ethereum/erc20/view_token_balances.sql
aalan3: Updated this, only thing is that `amount` could be `null` if decimals don't exist, `amount_raw` is always set.

jorijnsmit: I don't think `coalesce(var, NULL)` makes any sense. Should just be `amount_raw / 10^t.decimals` imo; indeed resulting in `NULL` value for `amount` in case no decimals are known.

ethereum/erc20/view_token_balances_latest.sql
gsteele178: I'd name it usd#1

0xBoxer (approved):

PR #403, opened 2021-06-30
Update token list

aalan3:
Update to the the token list.

I used two lists, on top of our current list.
https://tokenlists.org/token-list?url=https://zapper.fi/api/token-list 
https://tokenlists.org/token-list?url=https://tokens.coingecko.com/uniswap/all.json

The formatting of the symbols are the following: current list > zapper > coingecko


0xBoxer:
This is great and much needed. 
Good thinking on keeping the current symbols alive, that's the only issue that could arise here from my POV.

0xBoxer (approved):

aalan3:
This is deployed

PR #404, opened 2021-06-30
Add foundation user_names table and labels sync query

va3093:
## Context
The foundation team proposed the following approach to tag addresses with their user names (https://github.com/duneanalytics/abstractions/pull/360). This approach is problematic as it will be run every 10 mins and result in unnecessary load on our sync workers and database espescially if `foundation` grows its user base.

## Proposed solution
The solution proposed here is to create a table called `foundation.user_names` that maps addresses to user names. There exists a column that tracks the time that the row was inserted. Additionally I have added a query that will be used by the lable sync workers that will look for new user names since the last successful sync. If this table has not been updated then each run of the workers will finsih immediatly as there is no work to be done. Whenever this list is updated the whole thing will get synced again.

## Questions for foudnation team
- What would help even more is if either `address` or `name` was unique. We could then alter this slightly to only sync newly added user names.

T-3994


0xBoxer (approved):
Seems reasonable to me, I can't think of a better way to solve this.

PR #407, opened 2021-07-01
Set call trace address on paraswap to avoid duplicates

va3093:
## Context
Currently there are two unique indexes on the dex.trades table.
- `CREATE UNIQUE INDEX IF NOT EXISTS dex_trades_tr_addr_uniq_idx ON dex.trades (tx_hash, trace_address, trade_id);`
- `CREATE UNIQUE INDEX IF NOT EXISTS dex_trades_evt_index_uniq_idx ON dex.trades (tx_hash, evt_index, trade_id);`

This means in order to avoid duplicates at least one of these tuples need to contain non-null values.

## The problem
In the paraswap query the call tables are not setting the `trace_address` which results in the unique index not catching duplicates.

## The solution
Set the `trace_address` to `call_trace_address`.


mewwts (approved):

mewwts:
Nice catch

PR #408, opened 2021-07-06
erc20.tokens: update tokens list

bh2smith:
Based on the following query for Gnosis Protocol v2 - the recent updated list of tokens seems to exclude these. 

See here: https://duneanalytics.com/queries/37241/73825

Note that one of the tokens is the unicode character 🦄. Not sure if this will work as intended.

I noticed that the original list was sorted by name, but that the most recent update #403 didn't respect this format, so I appended the tokens to the end. Please let me know if you'd prefer it sorted.


0xBoxer (approved):

bh2smith:
@0xBoxer - I see that this was approved a week ago, is there a schedule for merge?

0xBoxer:
@aalan3  please take a look at this. 
Mind the Emoji character in the last line of the PR, no idea how our database will handle that 😓

bh2smith:
In the meantime I have added a few more tokens, but left the Unicorn emoji on the last line so not to disturb your comment. 🥂

bh2smith:
> Mind the Emoji character in the last line of the PR, no idea how our database will handle that 😓

If the emoji doesn't work out we can always replace it with the token name (Unicorn)

Issue #410, opened 2021-07-08
historical prices missing from dex.view_token_prices when erc20.decimals is added at a later point


jorijnsmit:
When a token is not known from inception of a dex pool, `erc20.decimals` is not known and `dex.trades.token_a_amount` (or `token_b_amount`) cannot be calculated, resulting in `NULL`.

Problem is that when the token decimals are added at a later date, due to using `INSERT`, `token_a_amount` is never recalculated for the older rows.

https://github.com/duneanalytics/abstractions/blob/e2b1a66bcc5d021d9ac807085aaa79d8c669a9a6/ethereum/dex/trades/insert_sushi.sql#L6

This in turn influences `dex.view_token_prices` since it depends and filters on this column
https://github.com/duneanalytics/abstractions/blob/e2b1a66bcc5d021d9ac807085aaa79d8c669a9a6/ethereum/dex/view_token_prices.sql#L19

Due to this, a lot of price history is not present in `dex.view_token_prices` even though it can be calculated quite easily.

Is there any way to rebuild some of the price tables from scratch, using the now known decimals?

An easy use case is oBTC. Compare for example

```
SELECT *
FROM dex.trades
WHERE token_a_address = '\x8064d9Ae6cDf087b1bcd5BDf3531bD5d8C537a68'
ORDER BY 1 DESC
```
with
```
SELECT *
FROM dex.view_token_prices 
WHERE contract_address = '\x8064d9Ae6cDf087b1bcd5BDf3531bD5d8C537a68'
```
Redefining `token_a_symbol` and `token_a_amount` would be all that's needed.

jorijnsmit:
OK so one solution is to join on `erc20` in the `view_token_prices`, I think this actually makes more sense. However, I don't know what other queries depend on `token_a_amount`. I'm writing a PR which should keep the original columns intact but increase historical price points.

bh2smith:
I think it would also be completely reasonable to include the decimals column directly on this this view instead of having to join. If prices are known, this must also mean that decimals are known.

masquot:
This seems to have been resolved with the creation of `prices.prices_from_dex_data` as a replacement for `view_token_prices` especially after we applied @bh2smith's excellent patch #530.

PR #411, opened 2021-07-08
Fix issue410

jorijnsmit:
Please double check. I used `coalesce` to prioritise the original calculation of `usd_amount/token_a_amount`. `LEFT JOIN` ensures the view is still functional even in cases where token is not yet present in `erc20.tokens`.


jorijnsmit:
(Quick) fix for #410.

mewwts:
@va3093 can you have a look?

va3093:
Hey @jorijnsmit. Apologies for the delay in reviewing this. It is unfortunate that this type of solution is necessary. As you can see table dependencies need some work. This is one of the big pieces of work we are currently undertaking.

In the short term, this solution looks valid and I will merge it now.

va3093 (approved):

PR #414, opened 2021-07-09
nft: add NFT trades for ethereum

masquot:
## What?

This repo collects trade data of NFT marketplaces in a single table `nft.trades`. 

Projects currently integrated:
- [x] OpenSea 
- [x] Foundation 
- [x] CryptoPunks LarvaLabs Contract 
- [x] SuperRare
- [x] Rarible

## How?

The structure is similar to that of `dex.trades`. `trades.sql` creates the table `nft.trades` and relevant indexes.

For every project (see list above), there is a file with commands to load historic data and a cron job to load new records.

## Note for the Reviewer

Some things that can use a review as this is my first bigger project :)
- Choice of `INDEXES` in `trades.sql` especially also the `UNIQUE INDEX`
- Does the row number in the `insert` queries make sense. I have it composed of:         `(PARTITION BY platform, tx_hash, evt_index, category)`
- For the cron job, i choose to have it run every hour at `53 * * * *`. I chose the number 53 as it is prime number sand so won't interfere with jobs running every x minutes. 
 
 Thank you

I've checked that:


milkyklim (approved):

ethereum/nft/trades/insert_cryptopunks.sql
```diff
@@ -0,0 +1,165 @@
...
+        '\xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'::bytea AS currency_contract,
+        trades.contract_address AS nft_contract_address,
+        trades.contract_address AS exchange_contract_address,
+        trades.evt_tx_hash AS tx_hash,
+        trades.evt_block_number AS block_number,
+        tx."from" AS tx_from,
+        tx."to" AS tx_to,
+        NULL::integer[] AS trace_address,
+        trades.evt_index,
+        row_number() OVER (PARTITION BY 4, 18, 23, 6) AS trade_id
```
va3093, line 58: As far as I am aware. You don't need trade id because I don't think multiple trades are ever grouped in a trace or a log. Am I wrong on this? If you can remove trade id it will be good because it is a source of many bugs.

va3093: Also can you refer to the columns by name and not index. It is very hard to review the code when one needs to count the column index etc. Also makes it very susceptible to bugs caused by updating the record set list.

ethereum/nft/trades/insert_foundation.sql
```diff
@@ -0,0 +1,106 @@
...
+WHERE NOT EXISTS (
+    SELECT *
+    FROM nft.trades
+    WHERE block_time > '2021-01-01'
+    AND block_time <= now() - interval '20 minutes'
+    AND platform = 'Foundation'
+);
+
+INSERT INTO cron.job (schedule, command)
+VALUES ('53 * * * *', $$
```
va3093, line 99: This runs ever hour at xx:53. Is that what you want. I think dex.trades runs every 10 mins? If this schedule is fine, then I would maybe recommend spreading them out slightly so that the database isn't smacked all at the same time. Unless you want them to update at roughly the same time, which I guess makes sense too.

ethereum/nft/trades/trades.sql
```diff
@@ -0,0 +1,33 @@
...
+    tx_hash bytea NOT NULL,
+    block_number integer,
+    tx_from bytea NOT NULL,
+    tx_to bytea,
+    trace_address integer[],
+    evt_index integer,
+    trade_id integer
+);
+
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS nft_trades_platform_tx_hash_evt_index_uniq_idx ON nft.trades (platform, tx_hash, evt_index);
```
va3093, line 28: Is there a reason you don't have `project, tx_hash, trace_address, trade_id` uniquely indexed?

va3093 (changes requested):

va3093:
@masquot Apologies for the late review :(. Just a few comments.

ethereum/nft/trades/insert_foundation.sql
masquot: Updated `insert_opensea.sql` to run every hour at 47. OpenSea is the biggest marketplace right now so it will run every hour on separate timing from the others but close to it.

ethereum/nft/trades/trades.sql
masquot: No I do not.You make a great point, added in new commit.

ethereum/nft/trades/insert_cryptopunks.sql
masquot: This is a really though call for me. From what i've seen, multiple trades are not grouped in a trace or log. Discussed this with @0xBoxer. Aggregators are coming in the nft space, and this seems a strong argument for keeping trade_id in there.

masquot: I absolutely get your point about referring to the column names by name and not by index. The reason why i did not do so, is because i had bugs while testing with columns names in the web interface. The only way to get consistent behavior was to use indexes. If you insist on this change, i will happily oblige, but it is not a quick replace edit but more a matter of refactoring and testing.

va3093: Thanks for the explanation. What bugs did you have during testing? It would help the readability and future debugging of this a lot if we used the names instead of indexes. Thank you!

ethereum/nft/trades/insert_foundation.sql
va3093: 👍

ethereum/nft/trades/trades.sql
va3093: 👍

ethereum/nft/trades/insert_rarible.sql
```diff
@@ -0,0 +1,362 @@
...
+        owner AS buyer,
+        amount,
+        "sellToken",
+        CASE
+            WHEN "sellToken" = '\x0000000000000000000000000000000000000000' THEN '\xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'
+            ELSE "sellToken"
+        END,
+        'ExchangeV1_evt_Buy' as category
+    FROM rarible."ExchangeV1_evt_Buy"
+    where "sellTokenId" = 0
```
va3093, line 154: Is it possible for both the sellTokenId and the buyTokenId to be 0?

ethereum/nft/trades/insert_rarible.sql
```diff
@@ -0,0 +1,362 @@
...
+        substring(data FROM 77 FOR 20) AS seller,
+        substring(data FROM 109 FOR 20) AS buyer,
+        bytea2numericpy(substring(data FROM 129 FOR 32)) original_amount_raw,
+        '\x0000000000000000000000000000000000000000'::bytea as original_currency_contract,
+        '\xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'::bytea as currency_contract,
+        'Buy' AS category
+    FROM ethereum."logs" 
+    WHERE "contract_address" = '\x9757f2d2b135150bbeb65308d4a91804107cd8d6' 
+    AND topic1 = '\x268820db288a211986b26a8fda86b1e0046281b21206936bb0e61c67b5c79ef4'
+    AND length(data) = 512
```
va3093, line 178: Hmm. Do you know why we don't have a contract table for this? In other words, why are we extracting this directly from the logs table instead of the a contract table?

ethereum/nft/trades/insert_rarible.sql
```diff
@@ -0,0 +1,362 @@
...
+        CAST(bytea2numericpy(substring(data FROM 513 FOR 32)) AS TEXT) AS nft_token_id,
+        substring(data FROM 109 FOR 20) AS seller,
+        substring(data FROM 77 FOR 20) AS buyer,
+        bytea2numericpy(substring(data FROM 161 FOR 32)) AS original_amount_raw,
+        substring(data FROM 365 FOR 20) AS original_currency_contract,
+        substring(data FROM 365 FOR 20) AS currency_contract,
+        'Offer Accepted' AS category
+    FROM ethereum."logs"
+    WHERE "contract_address" = '\x9757f2d2b135150bbeb65308d4a91804107cd8d6'
+    AND topic1 = '\x268820db288a211986b26a8fda86b1e0046281b21206936bb0e61c67b5c79ef4'
```
va3093, line 201: The above 2 selects look similar. The main problem with the trade_ids is that if any of these selects in this union chain return the same trace of event log it will get recorded twice which is bad.

ethereum/nft/trades/insert_cryptopunks.sql
va3093, line 58: Something else just struck me. You need to add an order by clause inside of this partition ([see here](https://github.com/duneanalytics/abstractions/blob/master/ethereum/dex/trades/insert_1inch_lp.sql#L56)). We have had issues where this partition forms a record set of > 1 but the order of the record set was random. That means that if the function was run twice on the same set of data it could insert a given entry multiple times because its `trade_id` is different.

va3093 (changes requested):
🙈 🙈 Since my first review a few more things have come to mind. Apologies for the double review. But I have had so many problems with this type of insert mechanism in dex.trades that I would like to avoid as many of them as possible.

ethereum/nft/trades/insert_rarible.sql
masquot: Missing ABI. @0xBoxer also looked into this.

ethereum/nft/trades/insert_cryptopunks.sql
masquot: I had inconsistencies between using indexes and names, but upon further testing it turns out that using indexes does not produce accurate results.Thank you so much for insisting on the usage of names. Refactored and upcoming commit will use names.

va3093 (approved):
Nice. Thanks for this work. I am sure the community are going to love it :)

PR #421, opened 2021-07-22
Fix balancer v2 pools lower case labels

va3093:
We currently have a constraint that the label must be lowercase in an effort to normalise labels that refer to the same thing but have different casing.

The one branch of the label case condition does not lowercase the result. This PR fixes that.


va3093:
@markusbkoch This cool with you?

va3093:
I am going to merge this but let me know if this doesn't suit your use case and I can revert and we can come up with a different solution.

milkyklim (approved):

PR #422, opened 2021-07-23
Fill missing nft prices helper query

va3093:
## Context
There are cases where the nft.trades table can have a null `usd_amount` even though a relevant price for the row exists in the prices.usd table. One reason for this is that a given token was added after a given row had already been inserted into the table. The new price does not fill historically relevant rows in other tables.

It can be cumbersome to refill the table when all you want is to update the price.

## What was done
This PR just adds a query that can be run to find the rows that have null prices even though a price exists and then updates it with that price.


milkyklim (approved):

va3093:
@milkyklim Gonna close this. @masquot is busy working through some changes. If he ends up needing this again we can re-open.

PR #427, opened 2021-07-26
[Easy] Deduplicate TRU token

bh2smith:
Both of these tokens have quite some trading activity and currently can't be distinguished from each other by symbol. This PR simply uses their name in place of symbol (`TrueBit` and `TrueFi`) instead.


milkyklim:
Merged in https://github.com/duneanalytics/abstractions/pull/423

PR #438, opened 2021-08-12
cryptopunks: fix Cryptopunks cron insert

va3093:
## What was done

The insert functions had the platfrom for cryptopunks set to `CryptoPunks` when infact it is `LarvaLabs Contract`;


masquot (approved):
Thank you very much. Great catch @va3093

milkyklim (approved):

PR #446, opened 2021-08-17
remove filter which excludes tokens already present in prices.usd

jorijnsmit:
The price feed from Coinpaprika starts the day the token got listed on their site, and does not fill in data points retroactively. This makes them pretty useless in a lot of cases. Compare for example the first transaction of renBTC with the starting point of the price feed:
```
SELECT min(evt_block_time)
FROM erc20."ERC20_evt_Transfer"
WHERE contract_address = '\xEB4C2781e4ebA804CE9a9803C67d0893436bB27D'
UNION ALL
SELECT min(minute)
FROM prices.usd
WHERE contract_address = '\xEB4C2781e4ebA804CE9a9803C67d0893436bB27D'
```
```
min
--
2020-05-13 05:44
2021-03-15 10:40
```
As a quick solution I suggest to make it able to fall back onto the dex market prices, by also including tokens into `dex.view_token_prices` that already have a cex price feed (`prices.usd`).


0xBoxer (approved):

masquot:
Thank you for this @jorijnsmit. The proposed change makes sense.

For the implementation, my proposal is to replace the `MATERIALIZED VIEW` with a `TABLE`. Running the modified query without extra parameters times out. A table can be updated incrementally via a function to decrease DB load.

Also, I suggest to change the cron job to run once every hour at most.

@mendesfabio and @markusbkoch. Added you here so you can stay in the loop, as Balancer's `view_liquidity.sql` abstractions rely on `view_token_prices.sql`.

jorijnsmit:
being a table instead of a materialised view makes sense i think. price feeds in this table are a core element of a lot queries currently i believe, and the more inclusive and faster it is the better.

having said that, im not sure if i am the person to write it; i currently dont really have the time to figure out the updater function. is this not something you guys could pick up with the core team @masquot @0xBoxer?

markusbkoch:
Thank you all. Is this PR still relevant considering coinpaprika is being replaced with coingecko? I mean, won't this sort out the lack of retroactive prices?

masquot:
Thank you for your input. @0xBoxer has been thinking a lot about how to best solve this problem. We will get back to you next week after he returns from vacation.

@jorijnsmit in general, yes of course we can help wrapping code in a function or similar :)

masquot:
Please see #527.

PR #449, opened 2021-08-19
[Easy] update erc20 tokenlist

bh2smith:
Just a good old Erc20.tokens update (list taken from Gnosis Protocol V2 trades table https://dune.xyz/queries/37241/73825) 

@milkyklim - May I ask what technique you used to sort the token list here so effectively (for example in #408). I find my self stumbling with this and so far the best way I've found is to put all this into a python list of tuples and sort by key...


antonio-mendes (approved):
Implemented a small change in the `1-UP` token. Check if it looks good for you @bh2smith 🙏 
Otherwise approved, and I already updated the table!

bh2smith:
Thanks! I've actually got another list incoming soon since this one took a while. I'll wait until it get bigger though

PR #450, opened 2021-08-20
Remove stETH from prices.ini

MSilb7:
Removing stETH from the prices table. Lack of historical data from coinpaprika was causing existing dashboards to be out of date. By removing this, we can revert to pulling it's price from `dex.view_token_prices`.

I've checked that:

* [ x] the folder name matches the schema name
* [ x] the schema name exists in Dune
* [ x] views are prefixed with `view_`, functions with `fn_`.
* [ x] the filename matches the defined view, table or function and ends with .sql
* [ x] each file has only one view, table or function defined  
* [ x] column names are `lowercase_snake_cased`

### DIFF
--- a/prices/prices.ini
+++ b/prices/prices.ini
@@ -1762,9 +1762,3 @@ id = mona-monavale
 symbol = mona
 address = 0x275f5Ad03be0Fa221B4C6649B8AeE09a42D9412A
 decimals = 18
-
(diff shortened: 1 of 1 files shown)


MSilb7:
Canceling this.

PR here solves the issue: https://github.com/duneanalytics/abstractions/pull/446

PR #453, opened 2021-08-23
Add usd backfill for nft trades

masquot:
## Data source:

Builds on the work of @va3093:
https://github.com/duneanalytics/abstractions/pull/422/files

## Test query

https://dune.xyz/queries/114733

## Checks 

I've checked that:

### DIFF
--- a/ethereum/nft/trades/backfill_usd_prices.sql
+++ b/ethereum/nft/trades/backfill_usd_prices.sql
@@ -1,0 +1,50 @@
+CREATE OR REPLACE FUNCTION nft.backfill_usd_amount() RETURNS boolean
+LANGUAGE plpgsql AS $function$
+BEGIN
+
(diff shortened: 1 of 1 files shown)


ethereum/nft/trades/backfill_usd_prices.sql
```diff
@@ -0,0 +1,50 @@
...
+
+RETURN TRUE;
+END
+$function$;
+
+-- historical fill
+SELECT nft.backfill_usd_amount();
+
+INSERT INTO cron.job (schedule, command)
+VALUES ('31 * * * *', $$
```
va3093, line 47:
```suggestion
VALUES ('12 12 * * *', $$
```
Lets just run this once a day. This case shouldn't happen that often and as a result it would be wastful to run this expensive query that often.

va3093 (approved):
Approving with suggestion.

PR #465, opened 2021-09-07
Create dex.trades on Polygon

mendesfabio:
This PR creates `dex.trades` for 3 DEXes: Balancer, Quickswap, and Sushiswap. 

I've checked that:


antonio-mendes (approved):
Thanks for this PR @mendesfabio 💪  , it has been deployed and the tables created and backfilled. 

There is however still a gap in the data after the 15th of September since Polygon was a bit behind when I started the backfill. I will correct this today and we should finally have full data in `dex.trades`.

Let me know in the meanwhile if you run into any issues!

PR #471, opened 2021-09-12
Creating an hourly rebase and APY table for olympus

EdgeCaser:
There are several queries dependent on having an available table for rebase rate hour on the hour. Rather than having several team members write the query over and over, this query generates a materialized view for everyone to use.


EdgeCaser (commented):
Realized the query was unnecessarily bloated. This optimizes the query. Basically, I didn't realize the generated time series was using intervals of one second.

PR #497, opened 2021-10-05
Add Balancer views

mendesfabio:
I've checked that:


ethereum/balancer_v2/view_transfers_bpt.sql
```diff
@@ -0,0 +1,18 @@
+CREATE OR REPLACE VIEW balancer_v1.view_transfers_bpt AS
```
masquot, line 1: This probably has to be `balancer_v2` instead of `balancer_v1`.

ethereum/balancer_v1/view_pools_fees.sql
```diff
@@ -0,0 +1,8 @@
+CREATE OR REPLACE VIEW balancer_v1.view_pools_fees AS
```
masquot, line 1: This `VIEW` times out wen I try to run it.

ethereum/balancer/view_trades.sql
```diff
@@ -139,27 +140,21 @@ WITH rows AS (
...
         )
-        LEFT JOIN balancer_v2."StablePool_evt_SwapFeePercentageChanged" s2 ON s2.contract_address = SUBSTRING(t."poolId" from 0 for 21)
-        AND s2.evt_block_time = (
-            SELECT MAX(evt_block_time)
-            FROM balancer_v2."StablePool_evt_SwapFeePercentageChanged"
-            WHERE evt_block_time <= t.evt_block_time
-            AND contract_address = SUBSTRING(t."poolId" from 0 for 21)
-        )
-    ) dexs
+    ) dexss
```
masquot, line 157: `dexs` ?

masquot (changes requested):
Thank you @mendesfabio. Looks great. Some typos. Also, one of the new views times out.

Is it necessary for your use case that the cron job runs every ten minutes? If not, could you please adapt it to have it run less frequently?

For `balancer.view_trades` - is my assumption correct that you want us to delete the current entries from this table and then perform a backfill with the code in the updated file?

ethereum/balancer_v1/view_pools_fees.sql
mendesfabio: Are you sure? I can run the statements below - and it returns <5k rows, shouldn't be an expensive operation.

masquot: Sorry, my bad @mendesfabio. It is not this view, but the `view_transfers_bpt.sql` that is very slow / times out when I try to run in the UI: https://dune.xyz/queries/196397

masquot (approved):

ethereum/balancer_v1/view_pools_fees.sql
mendesfabio: no problem! in fact this view has a lot of rows, so maybe trying to select all its results through the UI won't work. I just created it using dune_user_generated and it went well: https://dune.xyz/queries/197448. what do you think?

PR #500, opened 2021-10-07
Update Token List (139 new ERC20s)

bh2smith:
Just some standard maintenance to keep the erc20 token list updated. As a reference, I am taking these tokens from the Gnosis Protocol v2 dashboard that displays the trade history. We are manually updating this list until it becomes so long that the info should be inserted into the correct location (i.e. here).

https://dune.xyz/queries/37241/73825


ethereum/erc20/tokens.sql
```diff
@@ -5404,6 +5404,145 @@ COPY erc20.tokens (contract_address, symbol, decimals) FROM stdin;
...
+\\x60f63b76e2fc1649e57a3489162732a90acf59fe	FLURRY	18
+\\x7ae1d57b58fa6411f32948314badd83583ee0e8c	PAPER	18
+\\x85f17cf997934a597031b2e18a9ab6ebd4b9f6a4	NEAR	24
+\\xba7970f10d9f0531941dced1dda7ef3016b24e5b	BGLD	18
+\\xbdb4a359a506b0997a257da767308436c7340245	ACE	18
+\\xcb84d72e61e383767c4dfeb2d8ff7f4fb89abc6e	VEGA	18
+\\xcfbd04b3cef2cf1527f143a49e5dc1e19941d254	LOOTL	18
+\\xf009f5531de69067435e32c4b9d36077f4c4a673	UNV	18
+\\x2610f0bfc21ef389fe4d03cfb7de9ac1e6c99d6e	SKYRIM	18
+\\x32353a6c91143bfd6c7d363b546e62a9a2489a20	AGLD	18
```
masquot, line 5498: Could you please remove this line? This is a duplicate entry.

masquot: Yes sorry, that's right only the last one for `AGLD`. I checked your file locally and the tabs were there.

bh2smith: Good eye! Fixed the duplicate record.

PR #518, opened 2021-11-02
Update Prices.USD table with Top 50 unvalued tokens on GPv2

bh2smith:
_**Update: We have added only the top 50 tokens here since such a big change may not be scalable**_

Note that there are still some "well known" tokens still excluded (e.g. xSUSHI, wstETH, BTC2x-FLI). If possible, I would like to additionally include at least xSUSHI.
 
cc @masquot 

Based on the trade history table for Gnosis Protocol V2, we have noticed that there are about 1568 tokens traded which are not listed in `prices.usd`. After stumbling upon [this PR](https://github.com/duneanalytics/abstractions/pull/513) it became clear that the price feed for this table is coming from Coin Paprika whose token list can be queried [here](https://api.coinpaprika.com/v1/coins). 

Based on the two data sources

1. Coin Paprika API tokens https://api.coinpaprika.com/v1/coins and this
2. Dune [Query](https://dune.xyz/queries/224239) for Tokens Traded on GPv2 which are not part of `prices.usd` table

I whipped up [this script](https://github.com/gnosis/gp-v2-dune/pull/116)

which does the following:

For each token without price that is:

1. listed on coin paprika 
2. uniquely identifiable by `symbol` (note that coinpaprika doesn't include token addresses in their API response)
3. is active
4. is not new

build the appropriate yaml entry found.

I've checked that:

* [ ] the query produces the intended results - based on direct comparison with your primary source for prices, I imagine it would.
* [x] column names are `lowercase_snake_cased` (this is the replace statement in the script which instead of name from coin paprika, we use the snake case version of the id field)

Note that I have currently limited the list to only active tokens which **are not new**. If you would also accept new tokens, then more could be added - however I would be very happy to get these included as a first step.


aalan3 (approved):
Thank you for this and nice script! This is running in prod now

PR #521, opened 2021-11-07
add decimals and symbol to dex.view_prices

bh2smith:
Considering the decimals are known at the time this view is generated, this will make some dashboard much more performant if they don't have to join on the erc20 table every time to get decimals.

I've checked that:

* [x] the query produces the intended results (cf https://dune.xyz/queries/230676)


masquot:
Please see #527.

PR #522, opened 2021-11-07
Add 105 tokens to erc20 list

bh2smith:
Usual update similar to those in the past, these are tokens that have been traded on GPv2 which aren't part of the erc20 token list yet.


masquot (approved):

masquot:
Thank you for this update @bh2smith.

PR #526, opened 2021-11-10
Adding ENS Token (among a few others)

bh2smith:
Given that ENS has become so mainstream so quickly, I thought it would be good to push a token list update ASAP. Included are a few other tokens that have recently popped up.


masquot (approved):

masquot:
Thanks a lot @bh2smith. New tokens added :rocket:

PR #527, opened 2021-11-10
Add USD prices from DEX swap data

masquot:
## Problem solved

Create a table with hourly price data based on DEX swap data that replaces the currently existing view `dex.view_token_prices.sql` and makes it more performant.

This approach to creating USD price tables has the advantage that a token just needs to be included in `erc20.tokens` for it to have hourly price data.

To make this table easy to use, the query fills those hours where there was no trades with the price of the previous hour.

## Credits

- First saw this approach used by Balancer @markusbkoch and @mendesfabio 
- Improvements and feedback by:
  - @jorijnsmit #446 
  - @bh2smith #521 
  - @0xBoxer

## Test queries

- https://dune.xyz/queries/235418 (3 days of data)
- https://dune.xyz/queries/235657 (showcases for one token how the query fills those hours where there was no trades with the price of the previous hour) 

## Checks 

I've checked that:

### DIFF
--- a/ethereum/prices/insert_prices_from_dex_data.sql
+++ b/ethereum/prices/insert_prices_from_dex_data.sql
@@ -7,6 +7,7 @@ WITH trades_with_usd_amount AS (
     SELECT
         token_a_address as contract_address,
         token_a_symbol as symbol,
+        decimals,
         usd_amount/(token_a_amount_raw/10^decimals) AS price,
         block_time
     FROM dex.trades
@@ -22,6 +23,7 @@ WITH trades_with_usd_amount AS (
     SELECT
         token_b_address as contract_address,
         token_b_symbol as symbol,
+        decimals,
         usd_amount/(token_b_amount_raw/10^decimals) AS price,
         block_time
     FROM dex.trades
@@ -37,10 +39,11 @@ grouped_by_hour AS (
         date_trunc('hour', block_time) as hour,
         contract_address,
         symbol,
+        decimals,
         (PERCENTILE_CONT(0.5) WITHIN GROUP (ORDER BY price)) AS median_price,
         count(1) AS sample_size
     FROM trades_with_usd_amount
-    GROUP BY 1, 2, 3
+    GROUP BY 1, 2, 3, 4
     -- Get previous entries from the table
     -- This is necessary to provide price continuity for tokens with low swap volume
     -- when performing historic backfill.
@@ -49,6 +52,7 @@ grouped_by_hour AS (
         hour,
         contract_address,
         symbol,
+        decimals,
         median_price,
         sample_size
     FROM prices.prices_from_dex_data
@@ -63,6 +67,7 @@ leaddata as
         hour,
         contract_address,
         symbol,
+        decimals,
         median_price,
         sample_size,
         lead(hour, 1, now() ) OVER (PARTITION BY contract_address ORDER BY hour asc) AS next_hour
@@ -79,6 +84,7 @@ add_data_for_all_hours as
     gen.hour as hour,
     contract_address,
     symbol,
+    decimals,
     median_price,
     CASE WHEN gen.hour = data.hour THEN sample_size ELSE 0 END AS sample_size
     FROM leaddata data
@@ -92,7 +98,8 @@ rows AS (
         hour,
         median_price,
         sample_size,
-        symbol
+        symbol,
+        decimals
     )
 
     SELECT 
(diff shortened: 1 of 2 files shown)


masquot:
@bh2smith. You requested the have the fields `symbol` and `decimals` added in here as well in #521. We included `symbol` but at first sight don't see many use cases for adding `decimals`.

Could you explain why you want to have `decimals` included as well if that's still the case?

bh2smith:
> Could you explain why you want to have `decimals` included as well if that's still the case?

Yes absolutely, here (https://dune.xyz/queries/37241/73825) we have this table "Trade History" for which we display the Units Bought, Units Sold and Fee Amount.  Without decimals, we would have to show atom amounts (in WEI)

Furthermore, this table also shows the USD values of each trade computed as `units * price` where `units = atoms / 10^decimals`.

In brief, most Trade, Swap events emitted from contracts show the raw atom amounts (in wei) - So in fact, I can't really see how anyone could work without the decimals being here.

masquot:
Thank you for explaining your use case @bh2smith. Now it makes sense to me. Will add token decimals in new commit.

ethereum/prices/insert_prices_from_dex_data.sql
```diff
@@ -0,0 +1,200 @@
...
+    SELECT
+    gen.hour as hour,
+    contract_address,
+    symbol,
+    decimals,
+    median_price,
+    CASE WHEN gen.hour = data.hour THEN sample_size ELSE 0 END AS sample_size
+    FROM leaddata data
+    INNER JOIN generate_hours gen ON data.hour <= gen.hour
+    AND gen.hour < data.next_hour -- Yields an observation for every hour after the first transfer until the next hour with transfer
```
bh2smith, line 92: This 1 hour now is more than the 10 minute window that was used in `dex.view_token_prices`. Do we expect that this window is wide enough so not to result in missing data from `prices.usd`?

bh2smith (approved):
This looks absolutely fantastic! I see that you have gone with a different table name here. Does this mean that `dex.view_token_prices` will continue to exist as is and that those who are aware of this new table's existence will just move to use `prices.prices_from_dex_data`?

So excited to start using this! Oddly enough, it looks like I have access to formally approve this PR. Cheers 🥂

ethereum/prices/insert_prices_from_dex_data.sql
masquot: I expect this to be fine given that the script starts running at minute 14 of the next hour. Also, every run of the script goes back three days in time and performs retroactive updates where applicable. If there is more complete price data in `dex.trades` since the previous run, the rows in question will be updated.

masquot:
> This looks absolutely fantastic! I see that you have gone with a different table name here. Does this mean that `dex.view_token_prices` will continue to exist as is and that those who are aware of this new table's existence will just move to use `prices.prices_from_dex_data`?
> 
> So excited to start using this! Oddly enough, it looks like I have access to formally approve this PR. Cheers clinking_glasses

Great question. Ideally, everybody starts using `prices.prices_from_dex_data` and we phase out `dex.view_token_prices`.

To give everyone time to transition, we suggest to keep `dex.view_token_prices` for one more week. Is this timeframe and approach fine with all of you?

ethereum/prices/insert_prices_from_dex_data.sql
bh2smith: Great! My only question now is wen merge?

0xBoxer (approved):

ethereum/prices/insert_prices_from_dex_data.sql
masquot: Merging ongoing. Expect this to be finished in a few hours time. Will run some tests first thing tomorrow.

ethereum/prices/insert_prices_from_dex_data.sql
```diff
@@ -0,0 +1,200 @@
...
+BEGIN
+
+WITH trades_with_usd_amount AS (
+    SELECT
+        token_a_address as contract_address,
+        token_a_symbol as symbol,
+        decimals,
+        usd_amount/(token_a_amount_raw/10^decimals) AS price,
+        block_time
+    FROM dex.trades
```
bh2smith, line 13: There seems to be a lot of missing symbols (3265) when taking the symbol from `dex.trades`. Do you think it would be more appropriate if we could use `symbol` here instead of token_a_symbol?

See this query filling the missing symbol from erc20.tokens: 

https://dune.xyz/queries/237271

I will create a PR.

masquot: Thank you. PR merged

PR #530, opened 2021-11-11
Get symbol from erc20.tokens

bh2smith:
This PR #527 introduced some excellent changes, although it turns out that using on dex.trades for token symbol is unreliable in comparison with erc20.tokens

This query shows that there are 3265 distinct missing token symbols in dex.trades that are all contained in erc20.token

https://dune.xyz/queries/237271

### DIFF
--- a/ethereum/prices/insert_prices_from_dex_data.sql
+++ b/ethereum/prices/insert_prices_from_dex_data.sql
@@ -6,7 +6,7 @@ BEGIN
 WITH trades_with_usd_amount AS (
     SELECT
         token_a_address as contract_address,
-        token_a_symbol as symbol,
+        symbol,
         decimals,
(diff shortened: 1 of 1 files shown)


masquot:
Great catch @bh2smith!

masquot (approved):

bh2smith:
> Great catch @bh2smith!

Glad you like it. Is there anything particular holding this up for merge? We could also back fill the symbols in dex trades with this info, but this will continue to go out of sync as the erc 20 list becomes outdated.

masquot:
> > Great catch @bh2smith!
> 
> Glad you like it. Is there anything particular holding this up for merge? We could also back fill the symbols in dex trades with this info, but this will continue to go out of sync as the erc 20 list becomes outdated.

The update has run overnight and is completed now.

PR #541, opened 2021-11-18
Update function to address price gaps

masquot:


0xBoxer (approved):
Nice catch, looks good to me.

PR #549, opened 2021-11-22
Adding OVM_ETH

MSilb7:
OVM_ETH is used for bridging events '\xdeaddeaddeaddeaddeaddeaddeaddeaddead0000' but we should keep it's symbol as ETH since that's how it's labeled in the token contract (placeholder for ETH).

I've checked that:


masquot (approved):
Thank you @MSilb7!

PR #559, opened 2021-11-25
change contract_address handling

0xBoxer:
**Problem** 
With our current logic in nft.trades users are prone to make wrong calculations based of nft.trades since the ``nft_contract_address`` column can report wrong data. IMO it does not make sense to report a ``nft_contract_address`` when there is multiple collections involved.
However if a bundle trade consists of entirely the same collection we can report a contract address so that at least volume calculations include these sales.

**Solution proposed**
I suggest we only report a ``nft_contract_address`` when all items in a given trade belong to the same collection.

**Documentation**
Query that shows the logic of my proposed change.

[query1](https://dune.xyz/queries/261905)

Nft.trades after these changes

[query2](https://dune.xyz/queries/262342)

**Further thoughts:**
This will at the same time probably hinder people from including bundle trades into their calculations by accident, albeit we now explictly choose to keep some ``nft_contract_address`` values for the sake of calculating volumes.

Volumes can be calculated with:  
``where nft_contract_address = xxxx``

Prices should be calculated with
``where nft_contract_address = xxxx and trade_type = 'Single Item Trade' ``

### DIFF
--- a/ethereum/nft/trades/insert_opensea.sql
+++ b/ethereum/nft/trades/insert_opensea.sql
@@ -6,7 +6,7 @@ BEGIN
 WITH wyvern_calldata AS (
     SELECT
         call_tx_hash,
-        addrs [5] AS nft_contract_address,
+        addrs [5] AS nft_contract_address, -- only works in cases of single item trades, otherwise will call WyvernAtomicizer
         addrs [2] AS buyer,
         addrs [9] AS seller,
         addrs [7] AS original_currency_address,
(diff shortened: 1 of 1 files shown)


masquot:
Thank you @0xBoxer. Great proposal.

Will integrate your code changes as part of one bigger `nft.trades` update as OpenSea backfill is resource intensive.

masquot (approved):

masquot:
Is integrated in `nft.trades` v2. https://dune.height.app/T-6843

PR #564, opened 2021-11-26
Fix ens_names labels error

antonio-mendes:
This PR fixes an alert our labels manager was firing - https://dune.height.app/T-5002

The problem was basically that there were some `names` that were too long. This meant Postgres was failing on insert as this variable is used on one of the indexes and it was too big. To fix this I just limited the size of the `names`.

You can see it wasn't super common and so 99.9% of the data will be there:
```
 name_length | count 
-------------+-------
       38894 |     1
       10000 |     2
        4754 |     1
        1186 |     1
        1185 |     1
         670 |     1
         268 |     2
         256 |     1
         252 |     1
         224 |     1
         182 |     1
         171 |     1
         132 |     1
         127 |     1
         105 |     1
(15 rows)

Time: 111.169 ms
```

### DIFF
--- a/labels/ethereum/ens_names.sql
+++ b/labels/ethereum/ens_names.sql
@@ -28,5 +28,6 @@ SELECT
 FROM
     ethereumnameservice."ETHRegistrarController_3_evt_NameRegistered"
 WHERE
-    evt_block_time >= '{{timestamp}}';
-
+    evt_block_time >= '{{timestamp}}'
+    AND
+    LENGTH(name) < 10000;


masquot:
Thank you @antonio-mendes. Great catch!

You only modified the clause for the `ethereumnameservice."ETHRegistrarController_3_evt_NameRegistered` contract it seems. Is it not needed to also modify the clause for the other contracts?

antonio-mendes:
Hey @masquot ,
In this case it's ok like this, since that's the only controller getting new data. V1 and V2 are not getting new data as far as I can tell and they didn't cause issues while they were 👍

masquot (approved):

masquot:
Great, thank you @antonio-mendes, this has been merged.

PR #565, opened 2021-11-26
feat: add Poolsuite Executive Member to tokens.sql

stets:
The purpose of this change is to add the Poolsuite token and NFT to `tokens.sql`

I've checked that:

First PR here, hopefully it looks okay!


masquot (approved):
Thank you @stets for your first contribution!

PR is great. Just made one small edit. For the import to work correctly, we need the different columns to be separated by tabs instead of columns.

Poolsuite Executive Member has been added to the token list :rocket:

PR #568, opened 2021-11-28
Create OVM1 ERC20 Balances Table for OVM 2

MSilb7:
Creating a table to pass ERC20 token balances from the end of OVM1 to OVM2. This table will be used as the starting point from which we can calculate ERC20 balances by OVM2 transactions.

The 'ovm1' schema should be used for additional OVM1 data being passed to OVM2 (i.e. contracts created, Uniswap pool events, etc).

This is my first time creating an abstraction table, so please double-check it!

I've checked that:


masquot:
Great to see you create your first abstraction @MSilb7!

Your ideas make perfect sense. The code just needs a few tweaks:
- If your goal was to setup a composite primary key, this can be done by adding an extra row in the `CREATE TABLE` statement.
- Defining the `value` field as `INT` will not work as some values are too big for the integer type.
- The `contract_address` needs a double backslash in the `COPY` statement

I created a file with the proposed fixes here. Feel free to download and use in your next commit if this is what you are trying to achieve:
https://raw.githubusercontent.com/duneanalytics/abstractions/patch-2/optimism2/ovm1/erc20_balances.sql

MSilb7:
> Great to see you create your first abstraction @MSilb7!
> 
> Your ideas make perfect sense. The code just needs a few tweaks:
> 
> * If your goal was to setup a composite primary key, this can be done by adding an extra row in the `CREATE TABLE` statement.
> * Defining the `value` field as `INT` will not work as some values are too big for the integer type.
> * The `contract_address` needs a double backslash in the `COPY` statement
> 
> I created a file with the proposed fixes here. Feel free to download and use in your next commit if this is what you are trying to achieve: https://raw.githubusercontent.com/duneanalytics/abstractions/patch-2/optimism2/ovm1/erc20_balances.sql

Hi, thanks so much, super helpful! You got what I was trying to do with the primary keys, good to know going forward. Just pushed a new commit with those updates.

masquot (approved):

PR #569, opened 2021-11-29
GPv2 - View Trades

bh2smith:
Adding a view for all "valued" trades on Gnosis Protocol v2. This view is generated as an aggregation of 

- `gnosis_protocol_v2."GPv2Settlement_evt_Trade"`
- `prices.usd` and
- `erc20.tokens`

and an example of this results from this view can be found [here](https://dune.xyz/queries/37241/73825)

I've checked that:

* [x] the query produces the intended results - see above link.

### DIFF
--- a/ethereum/gnosis_protocol_v2/view_trades.sql
+++ b/ethereum/gnosis_protocol_v2/view_trades.sql
@@ -35,20 +35,23 @@ WITH trades_with_prices AS (
                 tx_hash,
                 order_uid,
                 owner,
-                sell_token                             as sell_token_address,
+                sell_token                    as sell_token_address,
                 (CASE
                      WHEN s.symbol IS NULL THEN TEXT(sell_token)
                      ELSE s.symbol
-                    END)                               as sell_token,
-                buy_token                              as buy_token_address,
+                    END)                      as sell_token,
+                buy_token                     as buy_token_address,
                 (CASE
                      WHEN b.symbol IS NULL THEN TEXT(buy_token)
                      ELSE b.symbol
-                    END)                               as buy_token,
-                sell_amount / 10 ^ s.decimals          as units_sold,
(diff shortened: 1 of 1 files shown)


masquot:
Thank you @bh2smith. Looks great. :+1: 

Unfortunately, we don't allow querying `dune_user_generated` from within user created abstractions. Might it work for you to add the extra tokens from the extended view to `erc20.tokens` directly? Or is there a specific reason why this would not make sense?

fyi the schema in your PR is `gnosis_protocol_v2` and this namespace exists as this is the place where the contracts are decoded.

bh2smith:
Hey @masquot - I had a feeling you wouldn't like the idea of the user generated view for extended token list. The reason for this is that the list keeps growing and I have developed a semi-automatic way of keeping it updated here in this view...

Of course I could append these tokens to the list, but it will inevitably become out of sync and then this view will be built on data that requires manual PRs to populate completely.

I have spoken today with @bernatfp demoing a possible future where we can automatically keep the `erc20.tokens` updated.

To summarize:

- Missing Tokens Query: https://dune.xyz/queries/236085
- Fetch and format token data by script: https://github.com/gnosis/gp-v2-dune
- Build/Update/Refresh Extended Token List: as user generated view **This could be directly replaced with erc20.tokens**
- Using The Extended Token List: As demonstrated here.


In order to get this view merged, I could replace the user generated view here with `erc20.tokens` and circle back to work towards helping automate updates the updates.

bh2smith:
I have changed the token source to `erc20.tokens` rather than the user generated view (as requested). Please let me know if there is anything else that looks suspicious (for example the large block of commented code that is essentially an example use case of the view introduced here).

masquot (approved):
Thanks so much @bh2smith!

Looks perfectly fine.

PR #573, opened 2021-11-30
update erc20 token list

bh2smith:
I tried to sorting this list using `sed -n '/\\\\/ p' tokens.sql | sort -k2 > sorted.sql` but macOS sort is different than on linux and the diff was too big to be worth it.

related to #569


masquot (approved):
Thank you for trying to sort @bh2smith!

Files sorted and PR merged :rocket:

PR #577, opened 2021-12-01
Create user_addresses.sql

MSilb7:
Saving a list of all addresses that sent a transaction in OVM1. This is used for tracking new vs returning addresses.

Other note, these lists are generated by exporting csv files from queries on the ovm1 database. Is there a way to do this by querying directly?

I've checked that:


MSilb7:
Making updates to fields included + fixing the bad copy line. Will re-submit

MSilb7:
Ah, the updates came through here, re-opening.

masquot (approved):
Thank you @MSilb7. PR is perfect :rocket: 

Unfortunately there is no shortcut right now for the current flow as these are two separate databases.

PR #579, opened 2021-12-02
Create ovm1 user_addresses_by_tx_type.sql

MSilb7:
Creating user_addresses_by_tx_type.sql . This is similar to user_address.sql , except that this is broken out by a user's first transaction of a specific type, or using a specific app. We can use this to accurately measure new/returning users and retention metrics in the ovm2.0 db.

I've checked that:


masquot (approved):
Looks great! Thank you @MSilb7

PR #584, opened 2021-12-06
Changed the source of maker liquidation events

dsalv:
* Changed the source of maker liquidation events from Clipper and Flipper to Cat and Dog contracts to capture all liquidation auctions.

Documentation: https://docs.makerdao.com/smart-contract-modules/dog-and-clipper-detailed-documentation
Test query: [Query](https://dune.xyz/queries/263871)


masquot (approved):
Thank you for this update @dsalv. Ran some test and looks fine!

Issue #593, opened 2021-12-06
MStable -> Dune Analytics


NatalieTop:
add trades from DEX MStable to Ethereum dex.trades table on Dune Analytics 

see an example: https://github.com/duneanalytics/abstractions/blob/f935c24521ae260f3ada576daf5d66fc87e5bb2f/ethereum/dex/trades/insert_1inch_lp.sql 

https://docs.dune.xyz/

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 264 years, 11 months from now.
Please review their action plans below:


**1) [echolon166](https://gitcoin.co/echolon166) has been approved to start work.**

Hello, already submitted PR's for 3 other bounties and would like to take on this one as well.

I will add trades from MStable to the dex.trades table on Dune Analytics.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/duneanalytics/abstractions/593/100027266).

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work for 100.0018 1INCH (254.0 USD @ $2.54/1INCH) has been submitted by__:

<hr>

* Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/duneanalytics/abstractions/593/100027266)
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/duneanalytics/abstractions/593/100027266).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
* $6,078,788.06 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. Started 3. **Submitted** 4. Done 

<hr>

__The funding of 100.0018 1INCH (259.0 USD @ $2.59/1INCH)  attached to this issue has been approved & issued to @echolon166.__  

 
* Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/duneanalytics/abstractions/593/100027266)
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
[...]

PR #624, opened 2021-12-09
dex.trades: add mStable

Echolon166:
Added mStable to dex.trades table as requested in #593.
Results of the mStable query can be checked at the dune_user_generated.mstable table.

I've checked that:


masquot (approved):
Thank you @Echolon166 for the great work.

Checks:
- [ ] token_a is the token bought 
- [x] basic check vs https://mstable.app/#/musd/stats
- [x] USD amounts also for `ETH` -> not applicable here

Looks fine to me.

PR #626, opened 2021-12-10
Update Gnosis Protocol Trade View

bh2smith:
This [transaction](https://etherscan.io/tx/0x8b79d40b068f1ad481910c623e028a4c85ebeffb14a7a277ddf670315c65a729) shows a trade where 25K USDC was bought, but the current version of the trades table values this trade at .20 cents because the prices.usd table has a strange price for the sell token. Instead of giving sell token price the priority when valuing trades we take the max when both token prices are available.

Here is the POC query: https://dune.xyz/queries/288467 (search for \x8b79d40b068f1ad481910c623e028a4c85ebeffb14a7a277ddf670315c65a729 in query results)

Here is the current view results:

Screenshot 2021-12-10 at 10 30 44 AM

### DIFF
--- a/ethereum/gnosis_protocol_v2/view_trades.sql
+++ b/ethereum/gnosis_protocol_v2/view_trades.sql
@@ -75,15 +75,25 @@ WITH trades_with_prices AS (
                 units_bought,
                 atoms_bought,
                 -- We use sell value when possible and buy value when not
+                -- We use sell value when possible and buy value when not
                 (CASE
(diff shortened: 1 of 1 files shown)


masquot (approved):
Thank you for the update @bh2smith! View updated :rocket:

bh2smith:
Thanks! Does the view get rebuilt on merge or does it still follow the cronjob schedule?

masquot:
Rebuild is done on merge and should be completed. :)

PR #632, opened 2021-12-11
GPv2 - Adding Registered Solvers View

bh2smith:
The primary purpose for introducing this is the `view_batches` table (which take include solver names from here) coming in a follow up PR. I thought it would be appropriate to have this view in the DB before making use of it. If not, I can merge the follow up PR directly into here.

* [x] the query produces the intended results: POC query [here](https://dune.xyz/queries/290419)

### DIFF
--- a/ethereum/gnosis_protocol_v2/view_solvers.sql
+++ b/ethereum/gnosis_protocol_v2/view_solvers.sql
@@ -1,0 +1,43 @@
+BEGIN;
+DROP MATERIALIZED VIEW IF EXISTS gnosis_protocol_v2.view_solvers;
+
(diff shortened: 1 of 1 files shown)


masquot (approved):
Thank you @bh2smith.

Noticed you did not include the closing `COMMIT` statement. So added this, and also an index and cron job to refresh the materialized view. Feel free to create a new commit if I did not get the intention of your code correctly.

bh2smith:
Thanks for fixing that up. I think this particular table only really needs to be refreshed once every day or week, so I'll change the cronjob schedule here. If it's not an urgent matter on your side, I'd just wait to include the change until the next PR touching this directory.

Issue #586, opened 2021-12-06
Convergence X -> Dune Analytics


NatalieTop:
add trades from DEX Convergence X to Ethereum dex.trades table on Dune Analytics 

see an example: https://github.com/duneanalytics/abstractions/blob/f935c24521ae260f3ada576daf5d66fc87e5bb2f/ethereum/dex/trades/insert_1inch_lp.sql 

https://docs.dune.xyz/

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 100.0018 1INCH (248.0 USD @ $2.48/1INCH)  attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/duneanalytics/abstractions/586/100027259).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/duneanalytics/abstractions/586/100027259).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://discord.gg/gitcoin/'>Gitcoin's Discord</a>
* $5,677,211.14 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work has been started__.


These users each claimed they can complete the work by 264 years, 11 months from now.
Please review their action plans below:


**1) [gall0x](https://gitcoin.co/gall0x) has been approved to start work.**

Hello, I'd like to work on this bounty, I'm new to pgsql but I've experience in PL/SQL so I think I can do it. Also I just submitted the Convergence Factory contract for decoding (0x4eef5746ED22A2fD368629C1852365bf5dcb79f1)

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/duneanalytics/abstractions/586/100027259).

gitcoinbot:
Issue Status: 1. Open 2. **Started** 3. Submitted 4. Done 

<hr>

__Work for 100.0018 1INCH (256.0 USD @ $2.56/1INCH) has been submitted by__:

<hr>

* Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/duneanalytics/abstractions/586/100027259)
[...]

PR #635, opened 2021-12-11
Add convergence to ethereum dex.trades

gall0x:
Hi this adds convergence swaps to Ethereum dex.trades table.
Issue #586 
You can see the intended results in dune_user_generated."convergence_trades"

I've checked that:

### DIFF
--- a/ethereum/dex/trades/insert_convergence.sql
+++ b/ethereum/dex/trades/insert_convergence.sql
@@ -1,0 +1,131 @@
+CREATE OR REPLACE FUNCTION dex.insert_convergence(start_ts timestamptz, end_ts timestamptz=now(), start_block numeric=0, end_block numeric=9e18) RETURNS integer
+LANGUAGE plpgsql AS $function$
+DECLARE r integer;
+BEGIN
+WITH rows AS (
+    INSERT INTO dex.trades (
+        block_time,
+        token_a_symbol,
+        token_b_symbol,
+        token_a_amount,
+        token_b_amount,
+        project,
+        version,
+        category,
+        trader_a,
+        trader_b,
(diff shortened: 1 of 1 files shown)


dsalv:
Thank you for your contribution @gall0x 

I executed the query and loaded the results in a test query here: https://dune.xyz/queries/300695

1. I have compared the last 24H volume from the query results and convergence metrics https://convx.conv.finance/explore - seems like the results do not match?

2. Based on the design of the table token_a refers to the token bought. Please, make sure that the query follows the same pattern

gall0x:
Hi @dsalv ,

1) The Convergence website is not open source so i don't know the way they calculate that value, but I changed the timeframe parameters:
```
 > (SELECT NOW() - INTERVAL '1 DAY')
 < (SELECT now())
```
And it seams like the 24 trading volume is ok (some times there is a difference because of the delay in processing the last logs in the different platforms, Dune, Thegraph, etc.),but most of the time it's pretty the same. Here is the query https://dune.xyz/queries/317030

2) Regarding the token_a issue, not sure about what should I change, in the query the token_a is the token going out of the pool so it’s the token bought, also that’s the way it’s made in all the adapters made for Uniswap forks (please see [insert_uniswap_v2.sql](https://github.com/duneanalytics/abstractions/blob/master/ethereum/dex/trades/insert_uniswap_v2.sql), [insert_sushi.sql](https://github.com/duneanalytics/abstractions/blob/master/ethereum/dex/trades/insert_sushi.sql), etc).

masquot (approved):
Thank you @gall0x. Double-checked and the results looks fine to me.

Sorry for the delay in getting your PR merged.

PR #636, opened 2021-12-11
dex.trades: add Smoothy Finance

Echolon166:
Added Smoothy Finance to dex.trades table as requested in #600 .
Results of the Smoothy Finance query can be checked at the dune_user_generated.smoothy_finance table.

Tokens traded were stored as id's and there doesn't seem to be a way to get token addresses directly from their contracts, hence the CASE statements linking the id's to their respective addresses.

I've checked that:


dsalv:
Thank you for your contribution @Echolon166 

I executed the query and loaded the results in a test query here: https://dune.xyz/queries/303665

 I have compared the volume from the query results and the metrics on the following page https://dappradar.com/ethereum/defi/smoothy-finance - there is a deviation of 19% on the 17 of December. Please, make sure that all the swap contracts are covered by your query.

 Based on the design of the table token_a refers to the token bought. Please, make sure that the query follows the same pattern

masquot (approved):
Looks fine to me.

PR #642, opened 2021-12-14
[WIP] GPv2 - Increase refresh frequency of trades view

bh2smith:
This PR aims to improve performance of the trades view query and changes the cron schedule for `gnosis_protocol_v2.view_trades` from 30 minutes to 5.  

The changes to the query are as follows:

- Create a temp table for the existing records of the view (will be appended to and "reinserted" later)
- Drop the existing view
- Execute the raw query to populate trade records (but only over the **past day** - if this query winds up being reliable we could even reduce this to 1 hour or less)
- Assemble the past day trades in a table according to the schema of the existing trades
- Take the union of the existing trades with the new trades (with special consideration for potential - and likely - duplicates)

At this point we have a complete account of the entire trade history. However because the erc20.tokens table is updated infrequently (and manually), it can occur that trades older than the '1 day' specified as our cutoff would fall behind our update window and be forever without token details. 

In the second step of this PR we query these complete "preliminary results" for rows with unknown token data,
namely having NULL values for `units_bought` OR `units_sold` (because these can't be populated without knowing token decimals).

We then go through a similar construction as usual (to populate the relevant fields) with any new token information that may have become available in the erc20.tokens table.

Note that the second phase of this construction will only ever yield different results than the preliminary whenever the erc20.tokens table is updated (not very often).

I would like to tag a couple folks from my team who may find this change interesting and/or useful for other parts of our backend infrastructure (cc @josojo and/or @fleupold)

One thing I was unable to test in the POC query (linked below) is the syntactic correctness of this TEMP table, so I will highlight a parts of the code that may require special attention apart from us seeing expected results. 

## Test Plan
When I run the [PoC](https://dune.xyz/queries/309267) I see "Last run took 28 seconds".

## Potential Risks

These changes work only under the assumption that the trades view already exists. If we in into an issue that cron job responsible for updating this view fails for longer than 1 day, then we would have to repopulate the table.

### DIFF
--- a/ethereum/gnosis_protocol_v2/view_trades.sql
+++ b/ethereum/gnosis_protocol_v2/view_trades.sql
@@ -31,7 +31,7 @@ WITH trades_with_prices AS (
                                          THEN '\xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'
                                      ELSE trades."buyToken"
                                      END)
-                                 AND s.minute > NOW() - interval '1 day'
+                                 AND b.minute > NOW() - interval '1 day'
                                  AND b.minute = date_trunc('minute', evt_block_time)
     WHERE evt_block_time > NOW() - interval '1 day'
 ),
--- a/ethereum/gnosis_protocol_v2/view_trades.sql
+++ b/ethereum/gnosis_protocol_v2/view_trades.sql
@@ -157,6 +157,6 @@ CREATE INDEX view_trades_idx_6 ON gnosis_protocol_v2.view_trades (tx_hash);
 
 
 INSERT INTO cron.job (schedule, command)
-VALUES ('*/30 * * * *', 'REFRESH MATERIALIZED VIEW CONCURRENTLY gnosis_protocol_v2.view_trades')
+VALUES ('*/10 * * * *', 'REFRESH MATERIALIZED VIEW CONCURRENTLY gnosis_protocol_v2.view_trades')
 ON CONFLICT (command) DO UPDATE SET schedule=EXCLUDED.schedule;
 COMMIT;
--- a/ethereum/gnosis_protocol_v2/view_trades.sql
+++ b/ethereum/gnosis_protocol_v2/view_trades.sql
@@ -1,7 +1,12 @@
 BEGIN;
+CREATE TEMP TABLE existing_trades as (
+    SELECT *
+    FROM gnosis_protocol_v2.view_trades
+);
 DROP MATERIALIZED VIEW IF EXISTS gnosis_protocol_v2.view_trades;
 
 CREATE MATERIALIZED VIEW gnosis_protocol_v2.view_trades AS
+-- 1. Fetch the trades over the past day and populate with most recent prices, token details and unit amounts
 WITH trades_with_prices AS (
     SELECT evt_block_time               as block_time,
            evt_tx_hash                  as tx_hash,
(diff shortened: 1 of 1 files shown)


masquot:
Hey @bh2smith.
Average running time of the cron job over the last few days is around 5 min.

bh2smith:
I am wondering if the longer times are a result of there being a read lock on the prices table while it is being updated. Would it be worth shifting this job by a minute or two so that updates don't overlap with others?

masquot:
That makes sense. We can definitely give this a try.

bh2smith:
I also have this fun new idea that should dramatically improve this query once and for all. 

Given that historical information (i.e. everything before today) is static, what if - instead of rebuilding the entire view every time - we only query now minus X minutes and take the union of the existing view with the recent results? I will start working on this ASAP. I think it will also save your backend a ton of resources.

masquot:
Thank you for this great write-up. It makes me wonder if it makes sense to replace the Postgres view with a Postgres table?

bh2smith:
That could also work, but it doesn't really flow with the framework of most things defined in that repo, if we do that it would be nice to have an outline of how you might want me to declare the table here, unless I just create table like usual. Also, to allow us some time to migrate all our dashboard that depend on this, it might be good to keep both for a bit while I make the necessary changes elsewhere.

masquot:
You can just create a table as usual in a separate file with indices. I should be able to replace the view by the table in a single transaction. I will create a test version of the table so you can review before I do that.

If the column names and types remain the same, I don't expect there to be be additional work on your side.

bh2smith:
Okie, this sounds like a great idea. Could we perhaps schedule a short call sometime early next week to just discuss some of the expected syntax. I imagine I could probably do it on my own, just wanted to double check a couple of things. 

I would more or less wrap the content of [this query](https://dune.xyz/queries/288467) in to an UPDATE TABLE statement (assuming the table already exists)

Note that this query, as it is now, fetches the trades for the past two hours and attempts to assign some values to it some of the records would be entirely new entries to the table and some of them updates to other records.

One other thing that is not captured by this logic is the "unknown" tokens. When a token isn't contained in erc20.tokens, we use the token address instead of the symbol and wind up with null values for the units bought and sold fields. This would also mean that the query would also need to "remember to check for entires having unknown tokens" and try to update those as well. However the erc20 token list doesn't get updated every two hours so this would be a different sort of maintenance task.

ethereum/gnosis_protocol_v2/view_trades.sql
```diff
@@ -1,7 +1,12 @@
 BEGIN;
+CREATE TEMP TABLE existing_trades as (
+    SELECT *
+    FROM gnosis_protocol_v2.view_trades
+);
 DROP MATERIALIZED VIEW IF EXISTS gnosis_protocol_v2.view_trades;
```
bh2smith, line 6: This is the part that the POC query can't really test and should be perhaps tested before merge.

masquot: Your idea is great, but Postgres doesn't allow it:
> `psql:test_new_view_trades.sql:337: ERROR:  materialized views must not use temporary tables or views`

It works if your write the results of the view to a regular table:
```
CREATE TABLE gnosis_protocol_v2.test_new_existing_trades as (
    SELECT *
    FROM gnosis_protocol_v2.view_trades
);
```
But you can never `DROP` this table as the MATERIALIZED VIEW` depends on it:
```
psql:test_new_view_trades.sql:347: ERROR:  cannot drop table gnosis_protocol_v2.test_new_existing_trades because other objects depend on it
DETAIL:  materialized view gnosis_protocol_v2.test_new_view_trades depends on table gnosis_protocol_v2.test_new_existing_trades
```

My hunch is that your idea might work if you create a table like this and have it updated in its own cron job a few times a day. The creation of the `MATERIALIZED VIEW` using your new approach feels much faster.

bh2smith:
Closing in favour of new approach #853

PR #643, opened 2021-12-14
Add Unifi to dex.trades on Ethereum

lolboysg:
Added Unifi DEX to dex.trades table on Ethereum as requested in #603.
Results of the Unifi query can be checked at the dune_user_generated.unifi table.

I've checked that:


dsalv:
Thank you for your contribution @lolboysg

I executed the query and loaded the results in a test query here: https://dune.xyz/queries/303778

I have compared the volume from the query results and the metrics on the following page https://unifi.report/ - seems like the results do not match? Please, make sure that all the swap contracts are covered by your query. It is possible that some contracts are not decoded on Dune yet.

lolboysg:
@dsalv true I've forgot to add v1 contracts doing it rn

lolboysg:
@dsalv I don't really get it how the result doesn't match because so far it does (don't forget it's only ethereum)

lolboysg:
In fact the only thing missing are the V1 contracts but they don't have any volumes for a few months

dsalv:
Yes, thanks for clarifying, the volume on ethereum is indeed ~720 on that day, since the most of the volume comes from other networks. Please, make sure that your submission follow 1inch pattern where token_a corresponds to token bought.

dsalv:
The interface looks similar to Uniswap V2. Please, make sure that the following check `CASE WHEN "amount0In" = 0` covers all edge cases. You can see in this [PR](https://github.com/duneanalytics/abstractions/pull/687) that this check doesn't work if  `amount0In` was used in a different context. Additional check might be required.

lolboysg:
You're true it's a fork from Uniswap v2. However I don't think there might me additional checks required so far as I've searched in the router contract code. Anyway thanks a lot for the feedback

dsalv:
Thank you @lolboysg

Checks:


- [ ] token_a is the token bought

- [ ] basic check vs https://unifi.report/ 

- [ ] USD amounts also for ETH -> not applicable here

Looks fine to me.

lolboysg:
@masquot could you review this ?

masquot:
Sorry for the delay @lolboysg. Approved and merged :rocket:

masquot (approved):

PR #644, opened 2021-12-15
adds numeric2bytea and bytea2numeric functions to polygon dataset

dparker2:
Hello, currently these functions [do not work on polygon datasets](https://dune.xyz/queries/296788), I am assuming this is because they do not exist within the `polygon/public` folder, like they do for both `ethereum/public` and `xdai/public`. This copies the `xdai` implementations into the `polygon` folder.

I've checked that:

* [X] the query produces the intended results **(same implementation works on xdai)**
* [ ] views are prefixed with `view_`, functions with `fn_`. **(is this applicable here?)**


masquot (approved):
Thank you @dparker2 for adding these UDFs to Polygon.

PR #646, opened 2021-12-16
GPv2: fix trade values in dex.trades

bh2smith:
It has come to our attention that the trade usd amounts were in disagreement between our `trades_view` and `dex.trades`. This is due to recent developments in one place and not the other: 

[Here](https://dune.xyz/queries/299492) is a query showing the differences 

[Here](https://dune.xyz/queries/299557) is a query showing that the proposed change to dex.trades is working as "expected"

Essentially, we join the trade events with the trade view table so to fetch the usd valuation from there (instead of evaluating in both places). This would mean that continuous development of the trades view should result in changes being reflected in both places.

Note that I have also taken this opportunity to replace trade valuations of `-0.01` for `NULL` whenever trade value is "unknown"

I've checked that:


masquot (approved):
Looks great! Thank you @bh2smith

PR #650, opened 2021-12-16
Limit size of nft users collections label

antonio-mendes:
Still related to https://dune.height.app/T-5414 - limiting the size of name so this works with our current implementation of labels


masquot (approved):
Thank you @antonio-mendes

PR #655, opened 2021-12-17
Temp HotFix: refresh baches only every hour

bh2smith:
We have discovered that [this view](https://dune.xyz/queries/38617/76300) has not been refreshed in over a day and I suspect it is because it takes more than 30 minutes to run the query but the query is being rerun every 30 minutes. This is only a temporary solution to a problem we have been discussing in more detail [here](https://github.com/duneanalytics/abstractions/pull/642#issuecomment-994505104)


masquot (approved):

bh2smith:
Just want to point out that the actual cause of this issue is the following 

https://github.com/duneanalytics/abstractions/issues/656

PR #667, opened 2021-12-21
GPv2: Correct dex swaps for aggregator solvers

bh2smith:
This POC query demonstrates expected results: https://dune.xyz/queries/290498

The purpose is to correctly count the number of dex swaps when using aggregator solutions by counting the dex category entries in dex.trades for the corresponding transaction.

### Another independent change (HOT FIX)

Also, as a hot fix, we remove the unique index in the trade view because we now support partially fillable orders.


masquot (approved):
Changes merged @bh2smith :rocket:

PR #684, opened 2021-12-27
Fix cron in insert_l1_gas_price_oracle_updates - Optimism2

MSilb7:
I've checked that:


masquot (approved):
This update has been implemented as well.

MSilb7:
Hm, still not seeing the table update beyond the initial 1000, any ideas? Maybe force a few more blocks in and see if it starts functioning?

masquot:
The code works fine when the `start_block` input equals 1:
`SELECT ovm2.insert_l1_gas_price_oracle_updates(1,10000);`

But it fails when this input is bigger than 1:
`SELECT ovm2.insert_l1_gas_price_oracle_updates(100,10000);`
Error message:
```
ERROR:  null value in column "l1_gas_price" of relation "l1_gas_price_oracle_updates" violates not-null constraint
DETAIL:  Failing row contains (100, null, null).
```

MSilb7:
> The code works fine when the `start_block` input equals 1: `SELECT ovm2.insert_l1_gas_price_oracle_updates(1,10000);`
> 
> But it fails when this input is bigger than 1: `SELECT ovm2.insert_l1_gas_price_oracle_updates(100,10000);` Error message:
> 
> ```
> ERROR:  null value in column "l1_gas_price" of relation "l1_gas_price_oracle_updates" violates not-null constraint
> DETAIL:  Failing row contains (100, null, null).
> ```

Ah yeah I think that's a weird edge case since I had the second insert start exactly at 100 rather than give a buffer. For future-proofing, I should add something where it defaults to pulling the most recent L1 Gas Price if there are no updates between the start block and first new update. On it.

MSilb7:
Figured out this issue, PRd here: https://github.com/duneanalytics/abstractions/pull/695

PR #687, opened 2021-12-28
Add extra check to uniswapV2 dex trades to clear duplicate names

dsalv:
**Problem**
Currently, `dex.trades` abstraction contains trades with similar names in `token_a` and `token_b` fields. The problem is caused by flexible usage of swap function in Uniswap V2. The original logic contains `amount0In` type checks to identify bought and sold tokens to populate amount and contract address fields. However, the function might be used in different contexts such as this https://docs.uniswap.org/protocol/guides/flash-integrations/flash-callback#initiating-a-swap where original fields are used in different context. As a result, originally implemented checks do not cover all edge cases.

**Solution proposed**
This check `"amount0In" = 0' `decides if `token_a` or `token_b` should be used as sold token. To cover additional cases where `"amount0In"' might be non-zero we can add this check `"amount1Out" = 0`.

**Documentation**
I generated two different user_generated_views. The first one is the query from the original abstraction and the second one contains the change. The number of records with duplicated names was reduced from ~3.7m to ~1.1k at the time of writing this comment.

**query1**
https://dune.xyz/queries/318654

**query2**
https://dune.xyz/queries/318210


masquot (approved):
Great work @dsalv.

PR #695, opened 2021-12-30
Fix insert l1 gas prices in op2

MSilb7:
Per this issue: https://github.com/duneanalytics/abstractions/pull/684#issuecomment-1002837392

Fixed an error where the first block would always return null, invalidating the constraint.

Tested here:https://dune.xyz/queries/315523

I've checked that:


masquot (approved):
Thank you @MSilb7. Function updated. Ran successfully :rocket:

PR #696, opened 2021-12-30
Backfill OP2 Prices + Add Logic for Gaps

MSilb7:
This creates a one-off backfill job "backfill_gaps_insert_approx_prices_from_dex_data.sql" add adds logic to pull the most recent price for each token at the start of a new price update in "insert_approx_prices_from_dex_data.sql" to *hopefully* fix the error where tokens would show a null price if they did not have any trades during the first hour of the time range.

I've checked that:


masquot (approved):
Thanks, changes applied.

PR #702, opened 2021-12-30
Update L1 gas Price Update Cron

MSilb7:
Missing a closing parentheses on the cron. Hopefully it updates now.

I've checked that:

### DIFF
--- a/optimism2/ovm2/insert_l1_gas_price_oracle_updates.sql
+++ b/optimism2/ovm2/insert_l1_gas_price_oracle_updates.sql
@@ -91,9 +91,9 @@ WHERE NOT EXISTS (
 
 INSERT INTO cron.job (schedule, command)
 VALUES ('5,15,25,35,45,55 * * * *', $$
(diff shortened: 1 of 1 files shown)


masquot (approved):
Those are dangerous to overlook.

MSilb7:
> Those are dangerous to overlook.

It's working!

MSilb7:
> > Those are dangerous to overlook.
> 
> It's working!

Actually, this is having the same 7 hour delay issue that the prices table had. Maybe it's using timestamptz vs timestamp? Usually I get an error if I don't do timestamptz.

masquot:
Running the code within cron job manually fails with:
> ERROR:  null value in column "l1_gas_price" of relation "l1_gas_price_oracle_updates" violates not-null constraint
DETAIL:  Failing row contains (1753538, null, null).

MSilb7:
> 1753538

So weird, can't recreate that. There should never be any nulls. Just added another Pr to try to fix it: https://github.com/duneanalytics/abstractions/pull/708

PR #709, opened 2022-01-04
Update backfill_gaps_insert_approx_prices_from_dex_data.sql

MSilb7:
testing if making these timestamps instead of timestamptz helps

I've checked that:


masquot:
@MSilb7 Can you please point me to a query or dashboard that shows the delay so I can run some tests on our side.

MSilb7:
> @MSilb7 Can you please point me to a query or dashboard that shows the delay so I can run some tests on our side.

Hi, yeah it looks like the multi-hour delay is resolved, but this shows how there are many null values for prices: https://dune.xyz/queries/349970

This could be due to backfill not running, or something I still need to figure out in the main prices insert.

Also, any recommendations on how to backfill "new" tokens (i.e. we add a token to the token list today, but it would've had a price for the past few weeks).

PR #714, opened 2022-01-05
Price backfill for eth pairs

dsalv:
**Problem:**
The prices for new altcoins are not available on the platform right after their creation. This [function](https://github.com/duneanalytics/abstractions/blob/master/ethereum/dex/backfill_usd_prices.sql) is used to backfill the prices when new coins are added to the platform. However it overwrites the usd_values for the swaps where the price was taken from the ETH value and the price for the swapped altcoin is still not available.

**Solution proposed:**
Add `prices.layer1_usd` for ETH pairs, to use ETH values during backfill.

**Documentation:**
I created two different queries. The first one only uses `prices.usd` and the second one uses `prices.layer1_usd` in addition to that.

**[query1](https://dune.xyz/queries/333086)**

**[query2](https://dune.xyz/queries/330647)**

### DIFF
--- a/ethereum/dex/backfill_usd_prices.sql
+++ b/ethereum/dex/backfill_usd_prices.sql
@@ -18,13 +18,13 @@ FROM
 			d.evt_index,
 			d.trade_id,
 			d.usd_amount as original,
-			d.token_a_amount_raw / 10 ^ pa.decimals * pa.price as pa_price,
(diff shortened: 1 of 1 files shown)


masquot (approved):
Thank you @dsalv. Elegant solution.

Backfill is running. Will merge PR when completed.

PR #724, opened 2022-01-09
Update $KAWA contract, add $xKAWA

letsbecomehuman:
- The KAWA contract was migrated to V2, PR reflects that update
- xKAWA V2 added, as well

I've checked that:


masquot (approved):
Thank you for adding the updated versions. I also kept the old version in there.

PR #736, opened 2022-01-12
Update l1 gas price oracle cron in op2

MSilb7:
It works! But adding a 5000 block buffer to the end of the cron so that inner joins with this table work when blocks are produced in between cron runs. Let me know if you have other potential solution ideas!

I've checked that:


masquot:
Might another potential solution be to have this run every minute?

MSilb7:
> Might another potential solution be to have this run every minute?

Yeah that would. I was unsure if that was too much for the cron scheduler, but every minute would be perfect. I can edit it.

masquot (approved):
Great!

PR #744, opened 2022-01-13
fix balancer trades

mendesfabio:
Balancer recently launched a Boosted Pool which broke a bit the `dex.trades` logic. 

1. Swaps where `tokenIn` or `tokenOut` is the pool token are actually joins/exits in that kind of pool. I added a `WHERE` clause at lines 102-103 to exclude them.

2. Trading against the Boosted Pool will lead to swaps between BPTs (Balancer Pool Tokens). Since we don't have them in the `prices.usd` table, the pool volume is currently zero. I created a `VIEW` to estimate the BPT prices and added to `dex.trades`. 

I've checked that:


masquot (approved):
Looks great. Fixed some typos. Thank you @mendesfabio

PR #749, opened 2022-01-16
Update to pull actual block_time

MSilb7:
Currently, this insert pulls the block time of when the gas price oracle was last updated. This change pulls the block_time from the actual block. This should help join performance, since we can join on block time and block number, rather than only block_number.

Would we be able to drop a regenerate the `ovm2.l1_gas_price_oracle_updates` so this change is reflected throughout?

I've checked that:


masquot (approved):
Update completed and table regenerated.

PR #755, opened 2022-01-18
GPv2: Mainnet Daily average Price injection to xdai trades

bh2smith:
For our upcoming snapshot purposes, we would like to (at least temporarily) include daily average prices injected from mainnet. Because there is no cross-chain query capability, we manually insert the entries resulting from this mainnet query:

https://dune.xyz/queries/358563 

To limit the number of manual entries inserted, we have taking the average daily prices for only three tokens (GNO, STAKE and WETH).

The POC query is here - https://dune.xyz/queries/358616 (although it appears to take a while to run). Hopefully it can complete.


masquot (approved):
Sorry for the delay @bh2smith. This has been merged.

PR #758, opened 2022-01-21
Add GAMMA token

tuckertranslator:
The token for Gamma Strategies, GAMMA, is not listed. ERC20 contract address: 0x6bea7cfef803d1e3d5f7c0103f7ded065644e197

I added it to the list. There is another GAMMA listed but they have different contract addresses of course. This will make some query scripts easier.

I've checked that:


masquot (approved):
Thank you for adding GAMMA @tuckertranslator.

PR #761, opened 2022-01-22
Update insert_approx_prices_from_dex_data.sql

MSilb7:
Remove restriction to Uniswap v3. Add 1inch to the list of DEXs (For now, I think it makes sense to manually add DEXs, in case one that's way off from everything else gets added).

I've checked that:


masquot (approved):
Thank you @MSilb7. Updated the function.

PR #770, opened 2022-01-26
add 0x tables and dex.trades insertion on OP

sui414:
Notes:
- tested queries
- not sure if the cron job schedule makes sense, plz feel free to tune to most reasonable params
- the 3 sql files have dependency sequence, so it has to be created in order of: 
 `zeroex.view_affiliate_data.sql` >> `zeroex.view_0x_api_fills` >>  dex.trades insert_zeroex

------
I've checked that:


sui414:
cc @MSilb7  for review

masquot (approved):
Thank you @sui414!

The cron schedules make sense, and very useful that you shared the dependency sequence!

PR #792, opened 2022-02-03
Update insert_approx_prices_from_dex_data.sql

MSilb7:
Adding in a check to make sure we don't pull in DEX calculated prices for the ETH address which represents L1 deposits.

Also adding in a NULL check for the final insert.

Will add in a follow-up script to delete the old ETH token prices and rewrite as a copy of WETH prices.

I've checked that:

### DIFF
--- a/optimism2/prices/insert_approx_prices_from_dex_data.sql
+++ b/optimism2/prices/insert_approx_prices_from_dex_data.sql
@@ -481,11 +481,11 @@ $$)
 ON CONFLICT (command) DO UPDATE SET schedule=EXCLUDED.schedule;
 
 --once per day run of the last 30 days to handle for new tokens
-INSERT INTO cron.job (schedule, command)
-VALUES ('1 0 * * *', $$
-    SELECT prices.insert_approx_prices_from_dex_data(
-        (SELECT MAX(hour) - interval '30 days' FROM prices.approx_prices_from_dex_data),
-        (SELECT DATE_TRUNC('hour', now()) + interval '1 hour')
-    );
-$$)
-ON CONFLICT (command) DO UPDATE SET schedule=EXCLUDED.schedule;
+-- INSERT INTO cron.job (schedule, command)
+-- VALUES ('1 0 * * *', $$
+--     SELECT prices.insert_approx_prices_from_dex_data(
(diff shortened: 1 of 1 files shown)


MSilb7:
Update: Seems like it's all tokens that are carrying forward old tokens rather than re-calcuating the latest prices. Trying to debug, but also wondering if anyone else notices what the bug could be? (Starting 1/24 ETH price only peridically updated. It's been stuck on 2489 for the past week

MSilb7:
Tried testing in my own test queries. The issue seems to be that the prices query runs before the dex insert completes. Is there a way to force waiting for the dex insert to complete?

masquot:
> Tried testing in my own test queries. The issue seems to be that the prices query runs before the dex insert completes. Is there a way to force waiting for the dex insert to complete?

I had to pause the cron job for `insert_approx_prices_from_dex_data` unfortunately as it no longer completes.

Suggested work-around: include both scripts in the same cron job and wrap them in a single transaction:
```
BEGIN;
SELECT [function1];
SELECT [function2];
COMMIT;
```

MSilb7:
> > Tried testing in my own test queries. The issue seems to be that the prices query runs before the dex insert completes. Is there a way to force waiting for the dex insert to complete?
> 
> I had to pause the cron job for `insert_approx_prices_from_dex_data` unfortunately as it no longer completes.
> 
> Suggested work-around: include both scripts in the same cron job and wrap them in a single transaction:
> 
> ```
> BEGIN;
> SELECT [function1];
> SELECT [function2];
> COMMIT;
> ```

Makes sense. Just removed the 30 day cron job update. Left a comment that I think I can create a version that mimics this: https://github.com/duneanalytics/abstractions/blob/master/ethereum/prices/insert_prices_from_dex_data.sql (pull from dex.trades) which should be way simpler and run better. I'll make a PR to combine the functions like you mentioned.

Other maybe better workaround: If I provided a token address mapping table, would we be able to pull prices for the "main" tokens from the same API that prices.usd on Eth populates from? Or is there some functionality we (Optimism) need to work with CoinGecko/CoinPaprika on?

masquot:
Fixed in #865

PR #794, opened 2022-02-04
Update ens_name_reverse.sql

zxsasha:
Added all possible Unicode Whitespace characters(https://en.wikipedia.org/wiki/Whitespace_character) to prevent malicious issues in the future.

I've checked that:


antonio-mendes (approved):
Looking good, thanks for this update @zxsasha !

zxsasha:
@antonio-mendes at the moment every time a new ens name is set to an address, it is added to the label table. But a 'ens name reverse' type must be in one instance per address. Can I ask you to use upsert instead of insert in order to have only one latest 'ens namer reverse' record per address? 

the example below returns two records
`select labels.get('\x69c488Bcda156379B6661f08A35dB627E5D467Dd', 'ens name reverse')`

zxsasha:
@antonio-mendes ok i ended up making the separate view https://dune.xyz/queries/458531 . going to use this view. please discuss with a team should we really store ens reverse records in the "labels" table.

PR #796, opened 2022-02-05
Add nft trades index

th3o3ry:
I've checked that:

this adds 2 indexes to nft.trades for querying (contract_address, token_id) and (tx_hash). i added IF NOT EXISTS and CONCURRENTLY to existing table + index creation. i'm happy to unwind that if you like. the actual index addition is the 2nd and 3rd commits of this PR.

### DIFF
--- a/ethereum/nft/trades/trades.sql
+++ b/ethereum/nft/trades/trades.sql
@@ -1,4 +1,4 @@
-CREATE TABLE nft.trades (
+CREATE TABLE IF NOT EXISTS nft.trades (
     block_time timestamptz NOT NULL,
     nft_project_name text,
     nft_token_id text,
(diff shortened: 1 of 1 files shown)


masquot:
Thank you @th3o3ry for proposing these updates to the `nft.trades` table and indexes.

Could you please provide more context on why you propose to have `CONCURRENTLY` added on the already existing indexes? There are some [caveats](https://www.postgresql.org/docs/13/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) related to using concurrent indexes, so I have a preference for keeping the existing indexes as they are.

th3o3ry:
> Thank you @th3o3ry for proposing these updates to the `nft.trades` table and indexes.
> 
> Could you please provide more context on why you propose to have `CONCURRENTLY` added on the already existing indexes? There are some [caveats](https://www.postgresql.org/docs/13/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY) related to using concurrent indexes, so I have a preference for keeping the existing indexes as they are.

I don't have a good reason. I'll revert that commit and leave the 2 adds. Is it helpful to leave in the added IF NOT EXISTS on the table create so the sql can be run in place? Do you want the new indexes to be created with CONCURRENTLY?

masquot (approved):
Thank you @th3o3ry. Your latest commit is perfect! Adding the new indexes right now. Will take some time.

masquot:
completed :rocket:

PR #822, opened 2022-02-11
Update insert_daily_token_balances.sql

MSilb7:
Hi, i found a very very dumb mistake where I wasn't filtering for the date range on WETH wraps/unwraps.

I did some testing and I think this is the last missing piece for data to be correct. But it requires re-running the whole table again :(

I've checked that:


masquot (approved):
Thank you @MSilb7. Refill completed.

PR #841, opened 2022-02-16
update chron

3StepCapital:
updated cron

I've checked that:

* [yes] the query produces the intended results
* [yes ] the folder name matches the schema name
* [yes] the schema name exists in Dune
* [yes] views are prefixed with `view_`, functions with `fn_`.
* [yes] the filename matches the defined view, table or function and ends with .sql
* [yes] each file has only one view, table or function defined  
* [yes] column names are `lowercase_snake_cased`

### DIFF
--- a/polygon/QiDao/view_balances_by_day.sql
+++ b/polygon/QiDao/view_balances_by_day.sql
@@ -76,6 +76,6 @@ order by ft."date",ft."collateral_token_symbol"
 );
 
 INSERT INTO cron.job(schedule, command)
-VALUES ('3 * * * *', $$REFRESH MATERIALIZED VIEW qidao.view_balances_by_day$$)
+VALUES ('3 * * * *', $$REFRESH MATERIALIZED VIEW qidao.view_balances_by_day;$$)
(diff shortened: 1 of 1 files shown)


ThreeStepCapital:
@masquot I can't test locally for chrons but I anticipate these will work fine

ThreeStepCapital:
Hey @masquot I made some changes and the query is much faster, when I tested it ran in about 1 minute...but the changes I made don't logically explain why the query is now faster. 

Change 1:
Add a limit to (SELECT * FROM qidao."view_contract_token_label" LIMIT 5000)
but this table only has 33 rows in it total, but adding a limit here sped up the query a lot, which doesn't make any sense

Change 2:
remove WHERE ft."collateral_token_symbol" is not NULL

This clause should make the table smaller, and therefore faster, but this where clause makes the query so slow it doesn't work.

masquot:
Great job @ThreeStepCapital. Updated the script, reset the cron jobs and they seem to be running fine.

masquot (approved):

PR #845, opened 2022-02-18
Gnosis Chain Extended Token List

### LINKED ISSUE #844: Gnosis Chain: erc20.MissingTokens
This query here shows ~140 active erc20 tokens which are missing form the token list. 

https://dune.xyz/queries/432914

I have not been able to determine how this list is populated or maintained from within this repo so posting as an issue. In the mean time, I am using this token list extension user generated view (here is where the data for the tokens can be found).

https://dune.xyz/queries/432843

bh2smith:
I was unable to tell how erc20.tokens is populated in this project so I created a table `erc20.extended_tokenlist` which takes all the tokens from erc20 and extends it by a bunch of missing tokens.

PoC Query showing desired results (as a user generated view). 
https://dune.xyz/queries/432843

Would be important to double check that the table creation works as expected.

Note that we also add native xDAI to stablecoins (since it was not there already). 

Closes #844


masquot (approved):
Thank you @bh2smith. Great solution.

PR #853, opened 2022-02-21
GPv2: Archived Trade History

bh2smith:
I've checked that:

* [ ] the query produces the intended results (Would like a second pair of eyes on this).


ethereum/gnosis_protocol_v2/trades.sql
```diff
@@ -0,0 +1,33 @@
+CREATE TABLE gnosis_protocol_v2.trades
+(
+    app_data           text,
+    atoms_bought       integer     NOT NULL,
+    atoms_sold         integer     NOT NULL,
```
masquot, line 5: Did a testrun with the `insert` function and got the following error:
>  ERROR:  integer out of range

Unfortunately, it is not clear from the error which of the `integer` fields in the table has the `out of range` problem.

Background info: [html](https://www.postgresql.org/docs/13/datatype-numeric.html)

ethereum/gnosis_protocol_v2/insert_trades.sql
```diff
@@ -0,0 +1,196 @@
...
+                   "buyToken"                   as buy_token,
+                   ("sellAmount" - "feeAmount") as sell_amount,
+                   "buyAmount"                  as buy_amount,
+                   "feeAmount"                  as fee_amount,
+                   s.price                      as sell_price,
+                   b.price                      as buy_price
+            FROM gnosis_protocol_v2."GPv2Settlement_evt_Trade" trades
+                     LEFT OUTER JOIN prices.usd as s
+                                     ON trades."sellToken" = s.contract_address
+                                         AND s.minute > TO_DATE('2021/03/03', 'YYYY/MM/DD') --! Deployment Date
```
masquot, line 23: Do you have a specific reason for not using `start_ts` in the `ON` clause here (and also in similar statement below)?

ethereum/gnosis_protocol_v2/insert_trades.sql
```diff
@@ -0,0 +1,196 @@
...
+                       sell_token,
+                       sell_value_usd,
+                       trader,
+                       trade_value_usd,
+                       tx_hash,
+                       units_bought,
+                       units_sold
+                FROM valued_trades
+                ORDER BY block_time DESC
+                ON CONFLICT DO NOTHING -- TODO should probably use DO UPDATE and put those with trade_value in
```
masquot, line 168: Is this the syntax you are looking for:
```
ON CONFLICT (order_uid, tx_hash) DO UPDATE SET trade_value_usd = EXCLUDED.trade_value_usd, field_2 = EXCLUDED.field_2, field_3 = EXCLUDED.field_3 [...]
```

masquot (changes requested):

masquot:
Thank so much @bh2smith for clearly documenting the intentions of the code!

You mentioned:
> In reality we will find that sometimes prices and tokens are missing... we do need some way of going further back than 1 day to back fill tokens and prices (as they affect many other values in the table).

Maybe one solution for the above is to setup multiple cron jobs:
1. Current data cron job running frequently during the day
2. One daily cron job going back 3 days in time and updating the rows of that day
3. One daily cron job going back 30 days in time. E.g. today, Feb 23 this cron job updates the rows for 24 Jan
-> The number and frequency of cron jobs can be adapted to your needs.

ethereum/gnosis_protocol_v2/insert_trades.sql
bh2smith: No specific reason, I could probably use the start ts here as well!

ethereum/gnosis_protocol_v2/trades.sql
bh2smith: I think in need to use numeric for these large wei values.

ethereum/gnosis_protocol_v2/insert_trades.sql
bh2smith: Actually, perhaps this inequality is not needed at all since in the following line we have an equality condition on the very same field. I had introduced this a long time ago thinking that the performance might be better if we restrict first the date to be in a certain range and followed by an equality. Perhaps this was a silly thought and is completely unnecessary. WDYT?

masquot: Makes sense to remove.

masquot (approved):
Thank you @bh2smith. Changes look great to me.

Trying to write the table. Will keep you posted here.

ethereum/gnosis_protocol_v2/insert_trades.sql
bh2smith: Done!

PR #857, opened 2022-02-22
Hot Fix: Batch View Query to include batches without interactions

bh2smith:
It was noticed that our Batch view query is not returning anything when batches have no interactions (this is because the query was faulty and explicitly excluded them - by accident of course). This update changes that.

[PoC Query](https://dune.xyz/queries/443833) - search for hash `\x0382115863e8200b1d1fe74baf49501c2e3b2901e41777a5e8276b8d5b3a7922` as an example of the bug fix (which was not in the table before).


masquot (approved):
Thank you @bh2smith

PR #862, opened 2022-02-24
[GPv2] Refactor Solver View

bh2smith:
Following up on #842

Previously we had manually updated the `active` column when including solver metadata - however this value can be computed entirely from event history now that the `AllowListAuthenticator` contract event have been indexed.

PoC Query Here: https://dune.xyz/queries/450092

This gives us a complete account of all solvers and whether or not they are currently registered. The only manual maintenance here then is to give names and environments when new solvers are added.

Test Plan:

Notice, if you fork the PoC Query and remove any of the of the manually inserted rows from metadata - the solver will still appear in the list with `environment = 'new'` and `name = 'Uncatalogued'`

cc @josojo

### DIFF
--- a/ethereum/gnosis_protocol_v2/view_solvers.sql
+++ b/ethereum/gnosis_protocol_v2/view_solvers.sql
@@ -60,7 +60,7 @@ known_solver_metadata (address, environment, name) as (
                  (decode('56d4ed5e49539ebb1366c7d6b8f2530f1e4fe753', 'hex'), 'test', 'Test Solver 2')
          ) as _
 )
--- Combining the metadata with activation status for final table
+-- Combining the metadata with current activation status for final table
 select address,
        case when environment is not null then environment else 'new' end as environment,
        case when name is not null then name else 'Uncatalogued' end      as name,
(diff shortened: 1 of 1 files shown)


ethereum/gnosis_protocol_v2/view_solvers.sql
```diff
@@ -1,64 +1,77 @@
...
-SELECT *
-FROM (
-         select *
-         from known_solvers
-         union
-         select *
-         from unknown_solvers
-     ) as _;
+)
+-- Combining the metadata with activation status for final table
```
josojo, line 63: A little bit above, you are introducing  block numbers, when the solvers get's activated and deactived.
```
    select solver, evt_block_number, evt_index, True as activated
    from gnosis_protocol_v2."GPv2AllowListAuthentication_evt_SolverAdded"
    union
    select solver, evt_block_number, evt_index, False as activated
    from gnosis_protocol_v2."GPv2AllowListAuthentication_evt_SolverRemoved"
    ```
 But here you just store the final end result. Hence, a deactivated solver will have two rows: activated and deactivated, without the block number info. Is this useful in some sense? Or should the block number also be in this table.

Comment (bh2smith) [reply ethereum/gnosis_protocol_v2/view_solvers.sql:63]: I think you might be missing the next sub query which selects the most recent event related to a distinct solver where we choose the relevant boolean value to determine if the solver is active or not active.

Comment (bh2smith) [reply ethereum/gnosis_protocol_v2/view_solvers.sql:63]:
```suggestion
-- Combining the metadata with current activation status for final table
```

josojo (approved):

masquot (approved):
Thank you @bh2smith and @josojo

PR #864, opened 2022-02-26
Mirror writers 022422

andrewhong5297:
I've checked that:

same data format as before, just fixed one data mistake I had


masquot (approved):
Thank you @andrewhong5297. Tables have been updated.

PR #868, opened 2022-02-28
[Trivial] Typo in Solver name

bh2smith:
Looks like our string comparison is case sensitive and the Paraswap dex-swaps counter was not returning expected results because of the typo `ParaSwap != Paraswap`


masquot (approved):
Thank you for fixing this typo @bh2smith.

PR #874, opened 2022-03-03
OP2 prices - fix bridge token

MSilb7:
Realized an issue where bridge tokens (calculated by the ratio to the underlying) were not updating. This is because we were always picking the "carry forward from previous prices" price rather than the actual updated price.

Updates here should fix this. We want to change the logic so we always pick the updated bridge token price (should never have a DEX trade).

I've checked that:


MSilb7:
Prices table issue should be resolved now.

masquot (approved):
Thank you @MSilb7

PR #897, opened 2022-03-13
OP2 - Insert Curve, Zipswap to dex.trades

MSilb7:
Adding ZIpswap and Curve to OP2 dex.trades

- Zipswap: Should be fairly straightforward, it's events are a Uni v2 fork.
- Curve: More complicated. Have logic up top to map tokens to their pool arg id. Then the metapools (1 token + base pool, i.e. sUSD - 3crv) have weird decimals behavior, where the extra token's decimals are always used if it was the one that was bought (not sure if anyone knows more about this, but I found the pattern that worked?)

There may be some collision with the Clipper PR (https://github.com/duneanalytics/abstractions/pull/888) - we should make sure that Clipper, Zipswap, and Curve all make it in to the dex insert.

We can add Zipswap to the dex prices insert, as some tokens are only traded there (i.e. ZIP, gOHM). Since curve is weird, and we already have reliable prices for the tokens it has (stables), I'll leave it out until we're 100% confident that its robust enough.

I've checked that:


masquot (approved):
Looks great @MSilb7!

PR #906, opened 2022-03-15
op2 refactor Chainlink Price Feed Inserts

MSilb7:
~The underlying token is a fraction of total market cap, was incorrectly assigned.~

~Is it possible to re-run the chainlink price inserts over ~32 days so we can overwrite this underlying token error?~

Recognized that I needed to refactor the Chainlink price feed insert for a few reasons:
1) We can use one price feed for multiple tokens (i.e. native AAVE, and Synthetix's sAAVE)
2) Tokens like TCAP use a chainlink price feed (i.e. total market cap), then adjust to a lower decimal unit ($1.39T to $139)

This removes underlying token for the chainlink feeds table, but adds a new feeds to token mapping table. Then, respective updates to chainlink's insert prices and the larger dex and prices inserts follow.

Since the tables change, we'd need to delete and re-create the `chainlink. oracle_addresses` and `chainlink.view_price_feeds` tables. All prices except for TCAP are unaffected. So, we'd also need to run the prices backfill `prices_backfill_02_01_22`  since TCAP prices started pulling in in Feb (it was never traded, so dex.trades is unaffected).

I've checked that:


masquot:
Thank you @MSilb7 

Just caught my attention that you use `NULL` in a `COPY` statement here which does not result in a `NULL` value in Postgres. Suggest to replace all `NULL`s with no value at all if you want the field to be NULL. If you decide to do so, you need to keep the tab as a delimiter for correct entry.

masquot:
Thank you @MSilb7 

Just caught my attention that you use `NULL` in a `COPY` statement here which does not result in a `NULL` value in Postgres. Suggest to replace all `NULL`s with no value at all if you want the field to be NULL. If you decide to do so, you need to keep the tab as a delimiter for correct entry.

MSilb7:
> Thank you @MSilb7
> 
> Just caught my attention that you use `NULL` in a `COPY` statement here which does not result in a `NULL` value in Postgres. Suggest to replace all `NULL`s with no value at all if you want the field to be NULL. If you decide to do so, you need to keep the tab as a delimiter for correct entry.

Good point. As part of this, I also recognized that I needed to refactor the Chainlink price feed structure. I adjusted the PR description. Should be all good to go now, but we should delete and restart the 2 chainlink tables mentioned above (still building as we fly a bit haha).

masquot (changes requested):

optimism2/chainlink/oracle_token_mapping.sql
```diff
@@ -0,0 +1,41 @@
...
+	underlying_token_address bytea,
+	extra_decimals numeric,
+		PRIMARY KEY (proxy,underlying_token_address, extra_decimals)
+);
+
+BEGIN;
+DELETE FROM chainlink.oracle_token_mapping *;
+
+COPY chainlink.oracle_token_mapping (proxy, underlying_token_address, extra_decimals) FROM stdin;
+AAVE / USD	\\x338ed6787f463394D24813b297401B9F05a8C9d1	\\x76FB31fb4af56892A25e32cFC43De717950c9278	0	--AAVE
```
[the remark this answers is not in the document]
MSilb7: Got it, is there a better way you can think of to track what's what without adding an extra column? I'll save in a local text file for now.

MSilb7:
Made fixes + modified chainlink logic that was causing synth price feeds to stall after mid-Feb.

Order should be:
- Chainlink insert updates
- Backfill query last

masquot:
Wrrote the new oracle files to the DB:
- [x] oracle_token_mapping.sql
- [x] oracle_addresses.sql
- [x] view_price_feeds.sql

But the `insert_price_feeds` file does not write any results to DB. It just returns 0 rows. Any ideas why?

Paused cron job for now.

MSilb7:
> Wrrote the new oracle files to the DB:
> 
> * [x]  oracle_token_mapping.sql
> * [x]  oracle_addresses.sql
> * [x]  view_price_feeds.sql
> 
> But the `insert_price_feeds` file does not write any results to DB. It just returns 0 rows. Any ideas why?
> 
> Paused cron job for now.

Hi, I got it to run on user generated.

I may have missed to add, since I added new columns to the chainlink, the base view_price_feeds table should be re-created. I'll check if there are any other discrepancies.

MSilb7:
Re-pasted the query in. Not sure why it's not working.

To validate, you're looking at the cron in prices and trades inserts correct? https://github.com/duneanalytics/abstractions/blob/master/optimism2/dex/prices_and_trades_inserts.sql

optimism2/chainlink/insert_price_feeds.sql
```diff
@@ -18,66 +19,76 @@ WITH gs AS (
...
+FROM (
 	SELECT
 	DATE_TRUNC('hour',block_time) AS dt
 	, feed_name
 	, AVG(bytea2numeric(topic2)::decimal/(10^decimals)::decimal) AS price
-	,"proxy", "address", underlying_token_address
+	,"proxy", "address"
 	FROM optimism.logs l
-	INNER JOIN chainlink.oracle_addresses cfa
+	INNER JOIN dune_user_generated.chainlink_feed_addresses cfa
```
masquot, line 37: Trying to run this latest version. There is still have a reference to `dune_user_generated` here.

MSilb7: Good catch, updated.

masquot:
Thank you @MSilb7. Pushed a commit that changed the placement of the `rows` INSERT statement as it makes testing and reasoning about the code easier.

Testing `chainlink.insert_price_feeds` in the UI (without the `rows` INSERT statement) yields no results:
https://dune.xyz/queries/540164

MSilb7:
> Thank you @MSilb7. Pushed a commit that changed the placement of the `rows` INSERT statement as it makes testing and reasoning about the code easier.
> 
> Testing `chainlink.insert_price_feeds` in the UI (without the `rows` INSERT statement) yields no results: https://dune.xyz/queries/540164

Ah thanks this was super helpful. Found the issue. My generated series up top was still pulling a field from the old table, so prices were always null (and therefore being dropped).

I updated it to pull the proxy and address along with underlying token. But also realized that I should add proxy and address to the base table. So, what I think steps from here are:

- Delete and re-create `chainlink/oracle_addresses.sql`
- Delete and re-create `chainlink/view_price_feeds.sql`
- Run `chainlink/insert_price_feeds.sql` including the original fill
* chainlink crons should be all good here *
- Update `prices/insert_approx_prices_from_dex_data.sql`
- Run backfill `dex/adhoc_backfills/prices_backfill_02_01_22.sql`

masquot:
> > Thank you @MSilb7. Pushed a commit that changed the placement of the `rows` INSERT statement as it makes testing and reasoning about the code easier.
> > Testing `chainlink.insert_price_feeds` in the UI (without the `rows` INSERT statement) yields no results: https://dune.xyz/queries/540164
> 
> Ah thanks this was super helpful. Found the issue. My generated series up top was still pulling a field from the old table, so prices were always null (and therefore being dropped).
> 
> I updated it to pull the proxy and address along with underlying token. But also realized that I should add proxy and address to the base table. So, what I think steps from here are:
> 
> * Delete and re-create `chainlink/oracle_addresses.sql`
> * Delete and re-create `chainlink/view_price_feeds.sql`
> * Run `chainlink/insert_price_feeds.sql` including the original fill
> 
> * chainlink crons should be all good here *
> 
> * Update `prices/insert_approx_prices_from_dex_data.sql`
> * Run backfill `dex/adhoc_backfills/prices_backfill_02_01_22.sql`

Great @MSilb7! Completed all steps. Cron job restarted.

masquot (approved):

PR #907, opened 2022-03-16
Create view_rari_fuse_ftokens.sql

scottincrypto:
Creates view with all Rari Capital Fuse Pools, the ftokens and underlying tokens with associated metadata in each pool.

This data is really difficult to get in Dune due to multiple methods of creating ftokens.  Often the data required is hidden in binary blobs in function calls.  This view was created from an eth-brownie script which queries the Rari Fuse contracts directly to get the required data.  This will need to be updated from time to time to keep it current.

This query has been tested in a user generated table at dune_user_generated.rari_capital_fuse_ftokens and produces the desired results.

I've checked that:


masquot (approved):
Thanks so much @scottincrypto for creating this view with Rari Capital Fuse Pools

PR #926, opened 2022-03-23
add backfill UPDATE logic and fix a placeholder address

sui414:
this PR is initially because our [table](https://dune.xyz/0x/0x-API-and-Matcha-on-Optimism) was missing token price for a period. & @MSilb7 had a improvement recently.

- updated the constants (the OP 0x Exchange Proxy) address we use for taker addr placeholder. (a side small issue i found recently)
- added the backfill ON CONFLICT DO UPDATE SET logic in `zeroex.insert_0x_api_fills()` function per @masquot 's suggestion. (thank you!!)

would love a complete backfill for the OP table `zeroex.view_0x_api_fills` 🙏 first tx for reference: https://optimistic.etherscan.io/tx/1691336

* might need to be careful about coordinating the task's schedule dependency with `dex.insert_zeroex()` for `dex.trades` table to make sure that one won't miss out records during this backfill.

### DIFF
--- a/optimism2/zeroex/view_0x_api_fills.sql
+++ b/optimism2/zeroex/view_0x_api_fills.sql
@@ -39,7 +39,7 @@ WITH zeroex_tx AS (
             logs.contract_address,
             block_time AS block_time,
             substring(DATA,13,20) AS maker,
-            '\xdef1c0ded9bec7f1a1670819833240f027b25eff'::bytea AS taker,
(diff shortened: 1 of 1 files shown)


masquot (approved):
Thanks so much @sui414 for this PR and the clear instructions. :+1: 

Did a complete backfill for `zeroex.view_0x_api_fills` and for `dex.trades`.

PR #931, opened 2022-03-26
[Easy] Add Entry to gpv2.view_solvers

bh2smith:
PoC Query: https://dune.xyz/queries/538257

Adding withdraw service to solver view. 

Note that we also change a column selection here because it was choosing the null version of the address resulting from the outer join.


masquot (approved):
Thanks so much @bh2smith

PR #962, opened 2022-04-12
OP2 1Inch - Handle for Limit Order

MSilb7:
We've found that 1Inch on Optimism has many trades that do not route through DEXs. One case is Limit Orders. These trades should not be counted as an aggregator, but rather as a dex trade via 1Inch. We'll need to delete and re-fill the 1inch trades for this to be fixed.

Steps:
- Delete 1Inch trades from dex.trades
- Re-insert all 1inch trades to dex.trades
- Continue with the cron as normal.

I've checked that:

### DIFF
--- a/optimism2/dex/insert_oneinch.sql
+++ b/optimism2/dex/insert_oneinch.sql
@@ -29,6 +29,13 @@ WITH rows AS (
         evt_index,
         trade_id
     )
+	
+,limit_orders AS ( --limit order trades use 1inch, but do not route to a DEX
+	SELECT evt_tx_hash FROM oneinch."LimitOrderProtocol_evt_OrderFilled"
+	WHERE evt_block_time BETWEEN start_ts AND end_ts
+	)
+	
+
     SELECT
         dexs.block_time,
         erc20a.symbol AS token_a_symbol,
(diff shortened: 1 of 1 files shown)


MSilb7:
Found this issue via chatting with @hildobby (tag isn't working, so let's see if Github figures it out)

hildobby (approved):
PR LGTM!

Before the PR, those txs are only included in the 'Aggregator' type of Optimism's dex.trades: https://dune.xyz/queries/594285

So this method removes 82 txs that are only in the Aggregator category: https://dune.xyz/queries/601164

So now it seems to be mostly synthetic swaps left according to that last query, figuring out what to do with those is step 2, maybe a change for another PR though 🤔

jeff-dude:
@MSilb7 thanks for the submission & details. i noticed when you declare the new 'limit_orders' CTE, it is formatted incorrectly. since the 'rows' CTE contains the entire insert/select statement(s), you would need to reuse 'WITH' as a new nested CTE to establish it. below is an example (not perfect, but gives the idea):
https://stackoverflow.com/questions/11741267/postgresql-nested-cte-and-union

i'll close this PR but feel free to edit and push new commit & reopen it when ready. i have everything prepped and ready to reload once fixed.

MSilb7:
Hi, wasn't able to re-open this, but opened a new PR here: https://github.com/duneanalytics/abstractions/pull/970

PR #964, opened 2022-04-12
[Ethereum] Set Protocol component pricefeed

jackiep00:
I've checked that:

This PR is for a daily price feed for the underlying components of Sets created using Set Protocol. These prices are important for estimating TVL and whatnot, and is intended to work across the entire ecosystem.

I've validated the insert function and backfilling function in this query here:
https://dune.xyz/queries/594861
This doesn't run in one go because the backfilling operation takes longer than 30 minutes, but you can execute each insert statement individually.

Most of the operative logic is clarified in the README. Architecturally it's modeled off of insert_prices_from_dex_data, and uses the dex pricing as a key input.


jeff-dude:
@jackiep00 we should not rely on data within the 'dune_user_generated' schema, it is meant to be temporary. please submit new PR which moves data in the mapping table within this schema to the 'setprotocol_v2' schema.

PR #967, opened 2022-04-13
Set protocol component prices

jackiep00:
I've checked that:

This PR is for a daily price feed for the underlying components of Sets created using Set Protocol. These prices are important for estimating TVL and whatnot, and is intended to work across the entire ecosystem.

I've validated the insert function and backfilling function in this query here:
https://dune.xyz/queries/594861
This doesn't run in one go because the backfilling operation takes longer than 30 minutes, but you can execute each insert statement individually.

Most of the operative logic is clarified in the README. Architecturally it's modeled off of insert_prices_from_dex_data, and uses the dex pricing as a key input.

Update: Addressed feedback on putting the token mapping into the PR instead of relying on `dune_user_generated`

### DIFF
--- a/ethereum/setprotocol_v2/set_component_token_mapping.sql
+++ b/ethereum/setprotocol_v2/set_component_token_mapping.sql
@@ -1,0 +1,33 @@
+-- https://dune.xyz/queries/576574
+-- There are certain wrapped components of Set tokens that map closely 1:1 with their corresponding token
+-- but are not actively traded, since they're just individual tokens that represent claims on other tokens
+
+
+create table if not exists setprotocol_v2.set_component_token_mappings
+(
+  symbol varchar,
+  component_address bytea,
+  token_type varchar,
+  mapped_symbol varchar,
+  mapped_component_address bytea         
+);
+
+create index if not exists on setprotocol_v2.set_component_token_mappings(component_address, mapped_component_address);
+create index if not exists on setprotocol_v2.set_component_token_mappings(component_address);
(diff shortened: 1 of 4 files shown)


jackiep00:
@jeff-dude Saw your feedback and resubmitted here!

ethereum/setprotocol_v2/set_component_token_mapping.sql
```diff
@@ -0,0 +1,33 @@
...
+create table if not exists setprotocol_v2.set_component_token_mappings
+(
+  symbol varchar,
+  component_address bytea,
+  token_type varchar,
+  mapped_symbol varchar,
+  mapped_component_address bytea         
+);
+
+create index if not exists on setprotocol_v2.set_component_token_mappings(component_address, mapped_component_address);
```
jeff-dude, line 15: @jackiep00 i've created this table & inserted the rows, however, the indices were not created due to syntax error. refer to your other table script, where there is a name to be applied before the 'on' command. please fix that syntax and push a commit to this PR, then i'll apply them. i'll continue on without the index here

jeff-dude:
@jackiep00 here is where we stand:

- [ ] set_component_token_mappings table created and populated
- [ ] set_component_token_mappings indices created
- [ ] daily_component_prices table & indices created
- [ ] daily_component_prices history loaded
- [ ] insert function created
- [ ] cron job inserted and active

once complete, we can merge & close

jackiep00:
@jeff-dude alright, I added the name and fixed the syntax error. Sorry about that!

jeff-dude (approved):

PR #968, opened 2022-04-14
[Polygon] Add tokens to erc20.tokens

jackiep00:
These are some components used in Set that haven't yet been added to erc20.tokens yet.

One thing I want to call attention to is that MATIC hasn't been added to erc20.tokens yet, so I added it here. Not sure if that's deliberate or not.

I've checked that:


jeff-dude (approved):
@jackiep00 thanks for the additions. i committed one minor fix but otherwise it's complete

PR #985, opened 2022-04-21
Add 37 ERC20s to erc20.tokens

bh2smith:
Based on the missing token query here: https://dune.com/queries/236085


jeff-dude (approved):
thank you @bh2smith the tokens were added

PR #988, opened 2022-04-25
[GPv2] Uniform Solver Names

bh2smith:
We had started making queries that required special cases for non-uniform solver names...

```
    where 
        name = 'DexCowAgg' or
        name = 'CowDexAg' or 
        name = 'Quasimodo' or
        name = 'QuasiModo'
```


jeff-dude:
@bh2smith fyi i'm looking at this, but there are downstream dependencies (view_batches) which need dropped to allow this recreation. both may be down for a bit to get it back up and running.

bh2smith:
That's ok but I'm not sure why that's the case. View batches just reads from this and will be corrected on its next run... which is every thirty minutes. right?

jeff-dude:
> That's ok but I'm not sure why that's the case. View batches just reads from this and will be corrected on its next run... which is every thirty minutes. right?

i was having issues attempting to drop view_batches, i will work with the team on this. in the meantime, i resumed the cron jobs until we make these changes.

bh2smith:
I am still not sure I understand the problem you are facing here: A previous PR updating this view was entirely fine.

https://github.com/duneanalytics/abstractions/pull/931

Notice that the view is dropped and recreated on every run of this cron job.

The `view_batches` simply uses the values here to include solver name. Could it be that the view batches cron job was running at the time this was being updated and the DB had a lock on it? Perhaps its worth checking how long the batches view regeneration is taking to populate these days? I hope its not longer than 30 minutes.

jeff-dude:
> I am still not sure I understand the problem you are facing here: A previous PR updating this view was entirely fine.
> 
> #931
> 
> Notice that the view is dropped and recreated on every run of this cron job.
> 
> The `view_batches` simply uses the values here to include solver name. Could it be that the view batches cron job was running at the time this was being updated and the DB had a lock on it? Perhaps its worth checking how long the batches view regeneration is taking to populate these days? I hope its not longer than 30 minutes.

apologies for the confusion. there aren't any issues, just an order of operations that needs to happen to process this change.

the full script here isn't run on each cron job execution, just the refresh statement. in order to change a materialized view, i have to handle any downstream dependencies (i.e. view_batches in this instance) before i can continue on the view here. the command to drop view_batches was just taking longer than expected. once it's dropped, i can recreate both, then reenable the cron jobs to refresh. there are no changes to view_batches, just dropping until view_solvers is updated. this is in process fyi.

jeff-dude (approved):
@bh2smith the rebuilds are complete and cron jobs reenabled. fyi -- the view_batches cron job refresh takes ~20-25 minutes each run, so only a few minutes before the next refresh at half hour mark.

PR #1002, opened 2022-05-01
OP2 Eth Transfers - add deposit/withdrawal handling

MSilb7:
Hi, making an update to eth_transfers to include deposit and withdrawal transactions. They use a mock ERC20 transfer, rather than traces.

Since these transactions would be missing from the earlier runs, would we be able to re-run the earlier inserts (without the where clauses) to backfill these transactions?

I've checked that:

### DIFF
--- a/optimism2/eth/insert_eth_transfers.sql
+++ b/optimism2/eth/insert_eth_transfers.sql
@@ -19,7 +19,7 @@ WITH rows AS (
     SELECT 
     r."from",
     r."to",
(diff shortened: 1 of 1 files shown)


jeff-dude:
@MSilb7 i will take a look into this & try to rerun the history

jeff-dude (approved):
@MSilb7 after adding the union, postgres complained about the hardcoded string into bytea column. i pushed an update here to fix that. history should be loaded for new erc20 event transfers & enabled moving forward. let me know if anything seems off, but should be all set.

PR #1006, opened 2022-05-02
[Ethereum] [Set Protocol] add anchor prices for daily component prices

jackiep00:
I noticed a bug where between insertion operations, the query would "forget" the last known price for imputing the component prices. I introduced this fix so that each insertion operation isn't missing a baseline price for all of the components.

Can we truncate the table and rerun the backfill? Or should I include the table truncation operation right before the backfill commands? (`insert_prices_from_dex_data.sql`, which i'm using as a template, doesn't include any pre-backfill truncate commands) 

I've checked that:

* [X] the query produces the intended results
* [X] the folder name matches the schema name
* [X] the schema name exists in Dune
* [X] views are prefixed with `view_`, functions with `fn_`.
* [X] the filename matches the defined view, table or function and ends with .sql
* [X] each file has only one view, table or function defined  
* [X] column names are `lowercase_snake_cased`

PR #1007, opened 2022-05-03
[Set Protocol] component views for TVL

jackiep00:
There's a series of views that are used as part of the TVL queries at Set, and even though they can be handled with dune user generated queries, it's better aesthetically to have them as part of the Set schema

I've checked that:


jeff-dude:
we can focus on this PR after #1006 is resolved

jeff-dude (approved):
@jackiep00 added commit with minor fixes, otherwise all objects should be built and/or edited where necessary

PR #1015, opened 2022-05-04
GPv2: Batches Table (an alternative to view_batches)

bh2smith:
Similar to the previous refactor of `gnosis_protocol_v2.view_trades` into `gnosis_protocol_v2.trades` (here - #853) we do the same for `view_batches` which has become very large and whose update schedule is now about as long as the refresh rate. This PR creates a batches table which is being appended to instead of completely recreated as a view every 5 minutes. 

As a second step we would eliminate the view_batches and free up some resources on Dune servers. 

I've checked that:

* [ ] the query produces the intended results - This I can't actually check, but the query is quite similar to the `view_batches` which does produce intended results. 
* [x] the schema name exists in Dune (schema name will exist after the table is created - here)

Additionally, with this merged, we should be more easily able to update `view_solvers` without conflict (cc #1013)

### DIFF
--- a/ethereum/gnosis_protocol_v2/view_batches.sql
+++ b/ethereum/gnosis_protocol_v2/view_batches.sql
@@ -24,7 +24,7 @@ WITH batch_counts AS (
                 sum(trade_value_usd) as batch_value,
                 sum(fee_usd)         as fee_value,
                 price                as eth_price
-         from gnosis_protocol_v2."view_trades"
+         from gnosis_protocol_v2.trades
                   JOIN prices.usd as p
                        ON p.contract_address = '\xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'
                            AND p.minute = date_trunc('minute', block_time)
--- a/ethereum/gnosis_protocol_v2/batches.sql
+++ b/ethereum/gnosis_protocol_v2/batches.sql
@@ -1,0 +1,23 @@
+CREATE TABLE gnosis_protocol_v2.batches
+(
+    block_time      timestamptz NOT NULL,
+    num_trades      int8        NOT NULL,
+    dex_swaps       int8        NOT NULL,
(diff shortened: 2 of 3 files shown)


ethereum/gnosis_protocol_v2/view_batches.sql
```diff
@@ -24,7 +24,7 @@ WITH batch_counts AS (
                 sum(trade_value_usd) as batch_value,
                 sum(fee_usd)         as fee_value,
                 price                as eth_price
-         from gnosis_protocol_v2."view_trades"
+         from gnosis_protocol_v2.trades
```
bh2smith, line 27: I suspect this may have also been an issue with the view taking so long to generate, but this is just a hunch.

jeff-dude: okay, i will rebuild this batches view (until you later delete)

jeff-dude:
batches table was created, backfilled and cron jobs active

jeff-dude:
> This is very exciting news, but I think there may be something wrong
> 
> ```
> select count(*) new_count from gnosis_protocol_v2."batches"
> ```
> 
> is about 114K
> 
> while
> 
> ```sql
> select count(*) new_count from gnosis_protocol_v2."batches"
> ```
> 
> is about 200K
> 
> Maybe its because the job backing filling the last six months doesn't happen till midnight...

looks like you accidentally used the same query twice there, but i think i see your point (comparing to view_batches). you are right though, i ran your command to update through 1/1/2022 but it stopped there. my assumption was you were waiting for the cron job of six months back to run to pick up 2022 data.

jeff-dude (approved):
okay, batches table is built and should be refreshed on it's schedule for last six months (all of 2022). view_batches is rebuilt and cron job enabled again. keep an eye on things and let me know any feedback

PR #1016, opened 2022-05-04
CoW Protocol: Remove all occurences of view_trades

bh2smith:
It has been several months since #853 was merged and enough time has passed to remove our dependence on this trades view. As the second part of the refactor described in #853, we eliminate the `view_trades` table and update schedule from your database. 

This may require a deletion on your side something like 
```
DROP MATERIALIZED VIEW IF EXISTS gnosis_protocol_v2.view_trades;
```


jeff-dude:
i'll start this cleanup soon, fyi

jeff-dude (approved):
@bh2smith files updated, view file removed, view_trades dropped in db, cron job deactivated. should be all set to focus on trades table only now

PR #1019, opened 2022-05-04
Add solana prices

aalan3:
Initial prices for Solana using coinpaprika. Keeping the same format as the rest for now.


jeff-dude (approved):
@aalan3 fixed an 'id' typo, but otherwise should be set

PR #1020, opened 2022-05-04
Polygon token update

jackiep00:
These are some of the tokens that have price feeds available in prices.usd but aren't in erc20.tokens.

I've checked that:


jeff-dude (approved):
@jackiep00 thank you, tokens added

PR #1024, opened 2022-05-05
[Ethereum] [Set Protocol] Price feed bug fix

jackiep00:
The cron job command doesn't update the current day. It turns out that this is because the insertion script is noninclusive of the end_date, so the cron job should pad an additional day to include the current date in the update.

I've checked that:


jeff-dude (approved):
@jackiep00 cron job command updated

PR #1033, opened 2022-05-10
Solana price update

aalan3:
Added some more tokens

### DIFF
--- a/prices/solana/coinpaprika.yaml
+++ b/prices/solana/coinpaprika.yaml
@@ -47,3 +47,8 @@
   symbol: USDT
   address: Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB
   decimals: 6
+- name: shdw-genesysgo-shadow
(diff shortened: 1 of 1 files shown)


prices/solana/coinpaprika.yaml
```diff
@@ -47,3 +47,18 @@
   symbol: USDT
   address: Es9vMFrzaCERmJfrF4H2FYD4KCoNkY11McCe8BenwNYB
   decimals: 6
+- name: shdw-genesysgo-shadow
```
jeff-dude, line 50: this comment would apply to the name attribute on all entries, but i'm curious if we should give the name the cleaner look, similar to how it is on solscan?
https://solscan.io/token/SHDWyBxihqiCj6YekG2GUr7wqKLeLAMK1gHZck9pL6y
for this one, it'd be 'Shadow Token'

this is minor feedback, but curious your thoughts?

jeff-dude: good to know 👍

PR #1034, opened 2022-05-10
view_pools.sql

yulesa:
Creates a table consolidating all Curve Pools with their underlying coins and gauge.

An equal user_generated view can be found here: https://dune.com/queries/450397

Regular pools had to be manually inputted.

I've checked that:


jeff-dude (approved):
@yulesa i committed one minor syntax fix, otherwise you should be all set 👍

PR #1037, opened 2022-05-10
[Ethereum] [Set protocol] Add cETH and cWBTC to component price feed

jackiep00:
Added two extra steps to the insertion process. cETH and cWBTC aren't often traded, so their prices from dex data are either inaccurate or missing. To get their prices we pull directly from the Compound exchange rate events.

After making this change, we should truncate the table and re-run the backfill again.

I've checked that:

### DIFF
--- a/ethereum/setprotocol_v2/insert_daily_component_prices.sql
+++ b/ethereum/setprotocol_v2/insert_daily_component_prices.sql
@@ -256,7 +256,6 @@ with ceth_eth_exchange_rate as (
 , day_series as (
   select generate_series(start_time::date - interval '1 day' -- include an extra day for the anchor price
                         , end_time::date, '1 day') as day
-  from ceth_eth_exchange_rate
 )
 , anchor_prices as (
     select dcp.date as day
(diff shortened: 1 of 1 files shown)


jeff-dude:
working through backfill with up-to-date function now 👍

ethereum/setprotocol_v2/insert_daily_component_prices.sql
```diff
@@ -240,6 +240,214 @@ with initial_components as (
   
 )
 SELECT count(*) + r INTO r from rows;
+-------------------------------------------------------------------------------------------------------------------------
+-- Step 3: Update prices for cETH using Compound-specific query
```
jeff-dude, line 244: it appears step 3 has duplicates on the values you implied in the 'ON CONFLICT' statement during inserts. please refer to this query to identify duplicates:
https://dune.com/queries/699027

i'm unable to load due to this failure

jeff-dude (changes requested):

ethereum/setprotocol_v2/insert_daily_component_prices.sql
jackiep00: Alright fixed, sorry about that, I was keeping two copies of the code, one using `dune_user_generated` and the other using `setprotocol_v2` and i didn't propagate all the changes over

jeff-dude (approved):
@jackiep00 backfill complete & cron job enabled again 🙌

PR #1038, opened 2022-05-11
[Polygon] [Set Protocol] KPI Views

jackiep00:
These are a series of views to measure Set KPIs on Polygon.

I've checked that:


jeff-dude (approved):
@jackiep00 i pushed a few commits with minor assumptions. the files here reflect what lives in the database now. if you need to tweak something, please open new PR on the merged files.

PR #1045, opened 2022-05-13
Update insert_curve.sql

yulesa:
Updating the dex table with all the current decoded curve pools.

View table: (need to restrict the interval)
https://dune.com/queries/462468

Query comparing with token terminal data:
https://dune.com/queries/736759

I've checked that:

### DIFF
--- a/ethereum/dex/trades/insert_curve.sql
+++ b/ethereum/dex/trades/insert_curve.sql
@@ -27,21 +27,1320 @@ WITH rows AS (
         evt_index,
         trade_id
     )
+
+    WITH exchange_n_exchangUnderlying AS (
+
+        SELECT evt_block_time, buyer, tokens_bought, tokens_sold, bought_id, sold_id, contract_address, evt_tx_hash, evt_index
+            FROM curvefi."17PctCypt_evt_TokenExchange"
+            UNION ALL
+
+        SELECT evt_block_time, buyer, tokens_bought, tokens_sold, bought_id, sold_id, contract_address, evt_tx_hash, evt_index
+            FROM curvefi."2CRV_evt_TokenExchange"
+            UNION ALL
+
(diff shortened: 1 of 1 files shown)


yulesa:
I have a question in the category column for each trade.
I've hardcoded everything from the curve protocol evts and calls as 'DEX'. Transactions coming from dex aggregators are included in the middle of those evts. Should I have separated them?

jeff-dude:
> I have a question in the category column for each trade. I've hardcoded everything from the curve protocol evts and calls as 'DEX'. Transactions coming from dex aggregators are included in the middle of those evts. Should I have separated them?

i can take a closer look at this question beginning of next week before finalizing the change.

also, would this change require a truncate & historical backfill? or since it's a new set of decoded tables, history can be run without a truncate prior?

yulesa:
I have made a PR for a modification in the [view_pools.sql](https://github.com/duneanalytics/abstractions/blob/master/ethereum/curvefi/view_pools.sql) that will require some modifications here as well.

jeff-dude (approved):
@yulesa function updated, history refreshed & schedule active 👍  should be set

markusbkoch:
@yulesa @jeff-dude This seems to be the most recent change to dex.trades involving Curve, and now I don't see Curve data in the table anymore. See eg [dex-metrics](https://dune.com/hagaetc/dex-metrics)
image

Also this
image


Could something be wrong with the PR, or maybe just missing a backfill+cron, or am I missing something?

jeff-dude:
> @yulesa @jeff-dude This seems to be the most recent change to dex.trades involving Curve, and now I don't see Curve data in the table anymore. See eg [dex-metrics](https://dune.com/hagaetc/dex-metrics) image
> 
> Also this image
> 
> Could something be wrong with the PR, or maybe just missing a backfill+cron, or am I missing something?

@markusbkoch thanks for asking. there is an open PR #1068 which is fixing an issue @yulesa found after this PR. at this point, curve data was deleted, but i may load majority curve data today while filtering out the known issues. then we can reload that data post-fix. hope this helps.

PR #1050, opened 2022-05-15
Add products to Pika protocol, namely APE, AXS, and UNI.

RplusT:
This PR adds $APE, $AXS, and $UNI as new underlying assets and markets under Pika Protocol. These new additions are specified in the `insert_pika_v1.sql` and `insert_pika_v2.sql` files.

I've checked that:


jeff-dude:
@RplusT is it fair to assume when you make these types of changes (including any future ones) that a full historical run needs to be done again -- using the dates provided in function calls of the script here? and since you have an update statement on the 'on conflict', we don't need to truncate and reload?

RplusT:
@jeff-dude yes, that is correct

PR #1055, opened 2022-05-17
add view vebal votes

mendesfabio:
I've checked that:

### DIFF
--- a/ethereum/balancer/view_vebal_votes.sql
+++ b/ethereum/balancer/view_vebal_votes.sql
@@ -1,0 +1,89 @@
+DROP VIEW IF EXISTS balancer.view_vebal_votes;
+
+CREATE VIEW balancer.view_vebal_votes AS
+
+WITH calendar AS (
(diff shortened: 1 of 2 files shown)


ethereum/balancer/view_vebal_votes.sql
```diff
@@ -0,0 +1,92 @@
...
+        AND v.weight > 0
+    )
+
+SELECT
+    round_id,
+    start_date,
+    end_date,
+    gauge,
+    provider,
+    ((bias - slope * (end_timestamp - block_timestamp)) * weight) AS vote
```
jeff-dude, line 88: the values in this line need some sort of aggregation function applied since the first five fields are grouped. please take the full view query here into the dune app, apply agg functions here & ensure the query runs as expected, then apply changes here with a commit

jeff-dude (changes requested):

ethereum/balancer/view_vebal_votes.sql
mendesfabio: done! thanks @jeff-dude 🙏

PR #1057, opened 2022-05-18
[GPv2] Fix Batch Indices

bh2smith:
The indices were copy-pasted from the view. Added `IF NOT EXISTS` to `CREATE TABLE` to avoid any issue.


jeff-dude (approved):
@bh2smith good catch, updated

PR #1059, opened 2022-05-19
Create LPtokens_based_on_liquidity.sql

yulesa:
I've checked that:

* [ x] the filename matches the defined view, table or function and ends with .sql


jeff-dude:
@yulesa please take a close look at my commit and let me know your thoughts 👍

yulesa (commented):
In accordance.

yulesa:
One doubt that I have is. The dex.liquidity table is daily based, so making this table on hourly might be difficult for all DEX LP tokens. Of course, having the price for every hour would be better than on daily, but maybe the burden is too big.

PR #1063, opened 2022-05-20
Remove Curve from OP2 dex.trades

MSilb7:
Something is off with the Curve insert & I think this is what's slowing down the table. It's inserting way more trades than actually exist.

Commenting out for now until i can resolve it.

I've checked that:


jeff-dude (approved):
@MSilb7 cron job updated. we did notice this job would continuously run (sometimes for days in one run) recently, did you confirm it's just because of curve? otherwise, we were going to explore next week.

jeff-dude:
> > @MSilb7 cron job updated. we did notice this job would continuously run (sometimes for days in one run) recently, did you confirm it's just because of curve? otherwise, we were going to explore next week.
> 
> Hi, in the dex trades table I noticed Curve had over 1k times as many rows as it should have - there was a ton of duplication. I found what was causing redundant rows, so now only 17k trades come in: #1069
> 
> ^ if we merge this PR, then wipe and reinsert the Curve trades, we should be good to add it back to the cron.

got it!

PR #1067, opened 2022-05-20
Remove all cronjobs for GPv1

bh2smith:
This PR should free up some resources on your server and this protocol has been deprecated. Along with this code change, I suppose we may also want to execute the following deletion:

```sql
DELETE FROM cron.job WHERE command ILIKE 'gnosis_protocol.%';
```
BUT ONLY UNDER THE `ethereum` namespace!

Leaving in Draft State and awaiting approval from another team member (cc @fleupold and/or @anxolin)

### DIFF
--- a/ethereum/gnosis_protocol/view_balances.sql
+++ b/ethereum/gnosis_protocol/view_balances.sql
@@ -33,7 +33,7 @@ JOIN gnosis_protocol.view_movement movement
 CREATE UNIQUE INDEX IF NOT EXISTS view_balances_id ON gnosis_protocol.view_balances (trader, token) ;
 CREATE INDEX view_balances_1 ON gnosis_protocol.view_balances (token);
 
-INSERT INTO cron.job (schedule, command)
-VALUES ('*/5 * * * *', 'REFRESH MATERIALIZED VIEW CONCURRENTLY gnosis_protocol.view_balances')
-ON CONFLICT (command) DO UPDATE SET schedule=EXCLUDED.schedule;
-COMMIT;
+-- INSERT INTO cron.job (schedule, command)
+-- VALUES ('*/5 * * * *', 'REFRESH MATERIALIZED VIEW CONCURRENTLY gnosis_protocol.view_balances')
+-- ON CONFLICT (command) DO UPDATE SET schedule=EXCLUDED.schedule;
+-- COMMIT;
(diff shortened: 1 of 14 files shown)


anxolin (commented):
Good call @bh2smith!

I'm 100% fine, GPv1 is deprecated. Afait only dxDAO was using it after we migrated to v2, and maybe they don't use it any more.

anxolin (approved):

bh2smith:
Thanks for your reply @anxolin - 

It appears (according to this that the protocol is still be operated on xdai: https://dune.com/queries/805833?d=6). We could keep the xdai cron job running. I will change that here.

> removing the views all-together
I don't think deleting the views is necessary, it would break a lot of queries and dashboards and there isn't really any harm in leaving them there.

jeff-dude (approved):
@bh2smith all of these should be deactivated now. note that view_trade_stats didn't seem to exist as a cron job on either ethereum or xdai.

rzeydelis:
> Good call @bh2smith!
> 
> I'm 100% fine, GPv1 is deprecated. Afait only dxDAO was using it after we migrated to v2, and maybe they don't use it any more.
> 
> If you want you can confirm with them if they want to maintain any of it, but probably they would want to use v2. I'd be even fine removing the views all-together, although not sure if someone would be upset of that breaking old dashboards

Do you know where I can find V2? I actually depended on the view_price_batch.sql data.

jeff-dude:
> > Good call @bh2smith!
> > I'm 100% fine, GPv1 is deprecated. Afait only dxDAO was using it after we migrated to v2, and maybe they don't use it any more.
> > If you want you can confirm with them if they want to maintain any of it, but probably they would want to use v2. I'd be even fine removing the views all-together, although not sure if someone would be upset of that breaking old dashboards
> 
> Do you know where I can find V2? I actually depended on the view_price_batch.sql data.

v2 objects here: https://github.com/duneanalytics/abstractions/tree/master/ethereum/gnosis_protocol_v2

bh2smith:
@rzeydelis - Are you asking about gnosis_protocol_v2? If so, there is no `view_price_batch` table for the more recent version of the exchange. Could you elaborate a bit on how you depend on this data?

rzeydelis:
@bh2smith I use it here in this dashboard for StakeWise to calculate TVL, Rewards, etc.: https://dune.com/beeezeee/SWISE-BZ

bh2smith:
OK I see here that you are using this table to get GNO token prices. I would highly recommend using `prices.usd` in place of this and furthermore do you joins on `contract_address` instead of `symbol`. However, for any queries on Gnosis Chain there is no `prices.usd` table.

Here, I have extracted a GNO price feed from the GPv2 trades view table, forked your query to demonstrate how you can get fairly reliable GNO price feed:

https://dune.com/queries/895620

rzeydelis:
Oh, thank you! This is super helpful.

I did some further digging and noticed some nulls (which is fine because I am looking for the daily avg anyways), but just wanted to leave my findings in here for future reference. My guess is that it might be staking rewards.

https://dune.com/queries/899220

bh2smith:
I think that link you shared is either a private query or hasn't been saved yet (so I can't view it).

PR #1069, opened 2022-05-20
OP2 Curve - fix base pool uniqueness

MSilb7:
Duplicate Curve entries in dex.trades were caused by an error in the abse pools CTE where we weren't forcing uniqueness, and trades were being duplicated 100s of times. This fixes that error.

To rectify impacted data, can we also delete all Curve trades from dex.trades and reinsert from the beginning?

I've checked that:


jeff-dude (approved):
@MSilb7 curve function should be updated and cron job enabled again

PR #1071, opened 2022-05-21
erc20.tokens: add 10 missing tokens

bh2smith:
Same as usual list taken from: https://dune.com/queries/236085


jeff-dude (approved):
@bh2smith added 👍

PR #1074, opened 2022-05-23
ovm2 l1 gas price - add unique indices

MSilb7:
Adding unique indices to try to make this table faster to query.

Let me know if there are other obvious-seeming ideas! Definitely not a pro at indices.

In short: Block Number is always unique, block time can be duplicated, but this table is often filtered by block time / joined on block number.

I've checked that:


jeff-dude (approved):
@MSilb7 indices are added

PR #1087, opened 2022-05-27
T 12720 add project and version to opensea trades

soispoke:
I've checked that:

If we want to have nft trades tables w/ all platforms and their versions, we need to add specific columns (project, version) to our opensea abstraction.

I added project and version columns to opensea.trades


soispoke (commented):
I also fixed some bugs w/ this (I kind of got confused between branches, should be fine after this PR)

PR #1089, opened 2022-05-27
Update ovm contract project names

MSilb7:
Small updates to handle for name override edge cases, where we define a name in creator address mapping that we later override in project name mapping.

I've checked that:


jeff-dude (approved):
@MSilb7 updated 👍

PR #1092, opened 2022-05-29
DRAFT: OP2 Update insert_project_name_mapping.sql

MSilb7:
DRAFT: Updating additional mappings

I've checked that:


jeff-dude:
similar comment here from #1094

jeff-dude (approved):
@MSilb7 pushed a few minor fixes, otherwise should be all set 👍

PR #1096, opened 2022-05-30
T 12432 performance testing optimize and partition by on dbt models new

soispoke:
This PR results from a discussion we had w/ @aalan3, with the aim of optimizing/reducing build time for incremental tables.
This resulted in 2 changes:
- Running OPTIMIZE on incremental tables on-run-end
- Replacing the incremental filter by a static filter,  namely  `NOW() - interval 2 days`

I've checked that:

### DIFF
--- a/spellbook/dbt_project.yml
+++ b/spellbook/dbt_project.yml
@@ -129,8 +129,8 @@ on-run-end:
   - "{{ alter_tblproperties_opensea_txns_day() }}"
   - "{{ alter_tblproperties_opensea_active_traders_day() }}"
 
-  - "OPTIMIZE transfers_ethereum_erc20_agg_hour"
-  - "OPTIMIZE transfers_ethereum_erc20_agg_day"
-  - "OPTIMIZE opensea_ethereum_trades"
-  - "OPTIMIZE opensea_solana_trades"
-  - "OPTIMIZE opensea_magiceden_trades"
+  - "OPTIMIZE transfers_ethereum.erc20_agg_hour"
+  - "OPTIMIZE transfers_ethereum.erc20_agg_day"
+  - "OPTIMIZE opensea_ethereum.trades"
+  - "OPTIMIZE opensea_solana.trades"
+  - "OPTIMIZE opensea_magiceden.trades"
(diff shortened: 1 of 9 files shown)


spellbook/models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -29,5 +29,5 @@ AND ARRAY_CONTAINS(log_messages, 'Program log: Instruction: ExecuteSale')
 AND block_time > '2021-09-01'
 {% if is_incremental() %}
 -- this filter will only be applied on an incremental run
-AND block_time > (select max(block_time) from {{ this }})
+AND block_time > now() - interval 2 days
```
aalan3, line 32: This should be `block_date` right? Or does block_time also work well?

Comment applies for the occurrences below as well

spellbook/dbt_project.yml
```diff
@@ -127,4 +127,10 @@ on-run-end:
   - "{{ alter_tblproperties_opensea_trades() }}"
   - "{{ alter_tblproperties_opensea_volume_day() }}"
   - "{{ alter_tblproperties_opensea_txns_day() }}"
-  - "{{ alter_tblproperties_opensea_active_traders_day() }}"
\ No newline at end of file
+  - "{{ alter_tblproperties_opensea_active_traders_day() }}"
+
+  - "OPTIMIZE transfers_ethereum.erc20_agg_hour"
```
aalan3, line 132: These commands are gonna run on create right? But will they also run on incremental runs? Ideally these should periodically. For context, optimize for prices run hourly (but inserts happen every 5 min or so there).

aalan3 (commented):
Very nice! I had 2 questions.

spellbook/models/magiceden/solana/magiceden_solana_trades.sql
soispoke: block_time works as well (I tested it and it also takes the exact same time to run) 👍

spellbook/dbt_project.yml
soispoke: These commands are actually ran every time a incremental run updates as well (so quarter hourly for now)

soispoke (commented):
thansk for the review and the help on this @aalan3

aalan3 (approved):

PR #1099, opened 2022-05-31
nft tokens

dot2dotseurat:
This PR adds a static table from legacy caladan and was copied over using abstraction_migration_tools. 

A few open Qs: 
- I added an accepted values check to confirm standards and one contract has no standard or symbol. '0x8c9f364bf7a56ed058fc63ef81c6cf09c833e656' 
This seems to be the SuperRare Auction House https://etherscan.io/address/0x8c9f364bf7a56ed058fc63ef81c6cf09c833e656
Should this address be handled elsewhere?
If it should be here, I don't want to add empty standard as an acceptable value. What could I put instead? 

- Some of the tokens are ERC20. Does that make sense for an NFT token? 

I've checked that:


soispoke (changes requested):
Based on what I read it's fine to have SuperRare as a token standard https://help.superrare.com/en/articles/5790642-series-contracts-101, so we should replace empty standards by "superrare" when it's related to the SuperRare Auction House

I also think it's ok to have erc20 tokens in there!

Is the dbt cloud test failing because of the empty standards ?

soispoke (approved):

soispoke (commented):
Replaced empty standard by superrare and added superrare to the test, that is now passing

soispoke (commented):
@dot2dotseurat The nft_tokens should also be included in the table_properties macro so it's exposed in the Dune app

soispoke (approved):
Just thought it made more sense to move everything to the tokens/ethereum folder instead of the nft one, but we can revert this in the future if we change our mind

PR #1100, opened 2022-06-01
add erc721 balances on v2

hildobby:
I've checked that:

* [X] the directory tree matches the pattern /sector/blockchain/ e.g. /tokens/ethereum

### DIFF
--- a/spellbook/models/balances/ethereum/balances_ethereum_erc721_day.sql
+++ b/spellbook/models/balances/ethereum/balances_ethereum_erc721_day.sql
@@ -1,0 +1,35 @@
+{{ config(
+        alias='erc721_day'
+        )
+}}
+
+with
+    days as (
+        select
+            explode(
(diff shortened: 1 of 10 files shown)


dot2dotseurat:
@hildobby I pushed a commit with some changes. The project wasn't compiling and I made some small edits to the schema file to fix them. Then I ran the project and there were some SQL errors around renamed columns and trailing commas. 

I also reorganized a bit by moving the models in the respective erc20 and erc721 folders.

soispoke:
The hourly test is taking too long to compute, I think it would make sense to add a WHERE statement (for the last day for example) so both they daily/hourly tests run smoothly

dot2dotseurat (approved):

kophyo1234 (commented):
Duplicate of #

kophyo1234 (commented):
#1100

PR #1104, opened 2022-06-01
clean up description in schema files for documentation

soispoke:
Small PR to clean up description in schema files for documentation

I've checked that:

### DIFF
--- a/spellbook/models/opensea/solana/opensea_solana_schema.yml
+++ b/spellbook/models/opensea/solana/opensea_solana_schema.yml
@@ -24,12 +24,14 @@ models:
         description: "UTC event block time"
       - name: token_symbol
         description: "Token Symbol"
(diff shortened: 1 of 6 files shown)


spellbook/models/opensea/solana/opensea_solana_schema.yml
```diff
@@ -24,12 +24,14 @@ models:
...
+      - name: token_address
+        description: "Token Address"
       - name: amount
         description: "Value of the trade at time of execution in the original currency"
       - name: amount_usd
         description:  "USD value of the trade at time of execution"
-      - name: trader
-        description:  "Trader"
+      - name: traders
+        description:  "Traders"
```
dot2dotseurat, line 34: if the description is just the same as the name of the column, I think we can just skip adding a description.

PR #1106, opened 2022-06-01
T 12902 optimize macro instead of on run end

soispoke:
Instead of having multiple OPTIMIZE commands running on-run-end, I just implemented a macro that I attach to the weekly job, as well as the newly created "Manually Deploy New Version" job :)
Capture d’écran 2022-06-01 à 13 13 59

I've checked that:

### DIFF
--- a/spellbook/macros/optimize_tables.sql
+++ b/spellbook/macros/optimize_tables.sql
@@ -1,0 +1,32 @@
+{% macro optimize_tables() %}
+{%- if flags.full_refresh -%}
+{% set sql_1 %}
+OPTIMIZE dbt_thomas_transfers_ethereum.erc20_agg_hour;
+{% endset %}
+
+{% set sql_2 %}
+OPTIMIZE transfers_ethereum.erc20_agg_day;
+{% endset %}
+
+{% set sql_3 %}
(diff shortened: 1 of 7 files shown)


spellbook/macros/optimize_tables.sql
```diff
@@ -0,0 +1,32 @@
...
+
+{% set sql_4 %}
+OPTIMIZE opensea_solana.trades;
+{% endset %}
+
+{% set sql_5 %}
+OPTIMIZE magiceden_solana.trades;
+{% endset %}
+
+{% do run_query(sql_1) %}
```
dot2dotseurat, line 23: slight preference for the set sql to be named after the optimize. 
so for instance `sql_4` -> `optimize_opensea_solana`

dot2dotseurat (approved):
lgtm

spellbook/macros/optimize_tables.sql
soispoke: Agreed, done !

soispoke (commented):
And yeah run-operation is pretty cool, I also discovered the "flags" feature this morning (https://docs.getdbt.com/reference/dbt-jinja-functions/flags), but it's only for dbt core

Thanks for the reviews !

PR #1108, opened 2022-06-01
Clarify dbt setup steps

dot2dotseurat:
Steps below are not relevant, this is a setup update. 

I've checked that:

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,24 @@
 Welcome to your [Spellbook](https://youtu.be/o7p0BNt7NHs). Cast a magical incantation to tame the blockchain.
 
 ### Dune contributors using the project
-Follow instructions from [Databricks](https://docs.databricks.com/dev-tools/dbt.html) on how to set up dbt core until you reach the dbt init step. 
+Use the pipfile to create a pipenv.
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,9 +4,24 @@
...
+```
 
-Enter `.` or other placeholders after running `dbt init`. This will not connect to the database but you have access to some dbt actions. When you are prompted to choose a target, please enter `wizard` so we know you are an external contributor.
+If the env is created successfully, enter the env shell. (If you run into Python version errors, try installing the matching version or editing the python version in the Pipfile but don't commit your change).
+```
+pipenv shell
+```
+
+Set up the dbt spellbook project. (You must run this step from the spellbook directory)
+Select Databricks and enter `.` or other placeholders after running `dbt init` for the host, HTTP path, and token. This will not connect to the database but you have access to some dbt actions.
```
soispoke, line 19: nit: Maybe rephrase this the the dbt init comes first in the sentence

soispoke (approved):

README.md
dot2dotseurat: yes, thats a great point. I'll update.

PR #1109, opened 2022-06-02
Tokemak

mfuribondo:
I've checked that:


mfuribondo:
please take a look at the new cron schedules.  looks like the curve pool stats one was either timing out or wasnt added to the cron schedule because i noticed it wasnt updating during my debugging.  thanks!

ethereum/tokemak/view_tokemak_PCAs_daily.sql
```diff
@@ -0,0 +1,111 @@
...
+        , tl.is_dollar_stable  
+            FROM (
+            SELECT  b."date", b.wallet_address, b.source_name, b.symbol, tl.address, b.display_name, b.tokemak_qty 
+                FROM tokemak."view_tokemak_wallet_balances_daily" b INNER JOIN 
+                tokemak.view_tokemak_lookup_tokens tl on tl.address = b.token_address AND tl.is_pool = false AND tl.symbol <>'TOKE'
+                LEFT JOIN tokemak."view_tokemak_prices_usd_eth_daily" tp on tp."contract_address" = b."token_address" and tp."date" = b."date"
+                WHERE tl.is_pool = false and tl.symbol <>'TOKE' 
+                ORDER BY source_name, symbol, wallet_address
+            ) b
+            INNER JOIN tokemak."view_tokemak_lookup_tokens" tl on tl.address = b.token_address and is_pool = false and is_liability=false ) as t GROUP BY 1,2,3
```
jeff-dude, line 28: b.token_address doesn't exist in the select within subquery, therefore the join is failing to recognize the column

jeff-dude (changes requested):

jeff-dude (approved):
@mfuribondo latest changes should all be applied 👍

PR #1112, opened 2022-06-02
Update .gitignore

soispoke:
Remove target folder from .gitignore

Checks below don"t apply

I've checked that:

### DIFF
--- a/spellbook/.gitignore
+++ b/spellbook/.gitignore
@@ -1,9 +1,3 @@
 
-target/
 dbt_packages/
 logs/
-
-# exclusions
-!target/manifest.json
-!target/run_results.json
-!target/sources.json


philippWassibauer:
please make sure this makes sense for your setup. checking in rendered files in git can have some consequences, eg: more work merging on your end. 

that being said, the aim of this would be that it is easier to pick up the `target` folder with github actions and deploy it to S3 where it would be served as a static website. the main issue here being, if I understand it correctly, that to render the folders one needs to have a connection to the prod database and I would prefer not having these types of secrets within Github Actions. 

If there is an easier way, without DB connection, to render the `target` folder from within Github Actions then let's reconsider this.

kophyo1234 (approved):

philippWassibauer:
let's wait with merging this until we have all the information.

soispoke (commented):
@philippWassibauer Yes let's double check everything before we settle on this.

The folders are rendered by the connection set up in dbt Cloud, and I don't think it depends on secrets within Github actions (@dot2dotseurat can you confirm ?).

I think the only connection is between dbt Cloud and the prod database, and then dbt Cloud has write access to the spellbook repo. 

The files generated in the target folder for the documentation come from the dbt docs generate command, attached to a job triggered in dbt Cloud. 

I think the only people who can generate new files in the target folder are ones that have access to dbt Cloud (only team members). But it would be good to hear opinions/thoughts on this from others :)

philippWassibauer:
The Github secrets only come into play if we render the target folder within Github actions. I checked with Florent and he would also prefer (but not definite no) not to have prod database credentials in the Github Secrets.

dot2dotseurat:
The target folder is generated by dbt core. We should keep it in the gitignore because it generates with everyones personal schemas. So for example, it would generate the queries with `dbt_meghan` or `dbt_thomas` prepended. So whenever we committed after running dbt compile we would be adding files with our personal schemas and I don't think that makes sense for the repo. 

wizards would have their own target folders because they will be running dbt compile. 

I see this is already closed so maybe there was a discussion elsewhere but we should leave it in the gitignore.

soispoke:
@dot2dotseurat I just chose to close it because it was open for a while and it seemed like we moved on to other options :) 
But very good to know I thought it would just generate when running it from dbt cloud (not core) !

PR #1114, opened 2022-06-02
Version 2 of Clipper contracts for ETH Mainnet and Polygon for `dex.trades`

aothman:
Clipper recently replaced its contracts on Mainnet and Polygon. This PR updates the `dex.trades` table on each network with the new contracts, whose trades will be as version "2" of the project "Clipper". The specific function implementation is called `insert_clipper_v2` on both networks.

I've checked that:

* [ X ] I tested the relevant subquery on dune.com
* [ N/A ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ X ] the directory tree matches the pattern /sector/blockchain/ e.g. /tokens/ethereum
* [ N/A ] if adding a new model, I added a test
* [ X ] the filename is unique and ends with .sql
* [ X ] each file has only one view, table or function defined  
* [ X ] column names are `lowercase_snake_cased`


ethereum/dex/trades/insert_clipper_v2.sql
```diff
@@ -0,0 +1,126 @@
...
+        tx_hash,
+        tx_from,
+        tx_to,
+        trace_address,
+        evt_index,
+        trade_id
+    )
+    SELECT
+        dexs.block_time,
+        pa.symbol AS token_a_symbol,
```
jeff-dude, line 32: for lines 32-35, you pull this data from prices, but in v1 script, it pulls from erc20 tokens table. do you want to keep consistency? i think it'll be more reliable too, as prices table can fall behind as it depends on data from outside API.

ethereum/dex/trades/insert_clipper_v2.sql
```diff
@@ -0,0 +1,126 @@
...
+    FROM (
+        -- Clipper v2 "Caravel" on ETH Mainnet
+        SELECT
+            t.evt_block_time AS block_time,
+            'Clipper' AS project,
+            '2' AS version,
+            'DEX' AS category,
+            t."recipient" AS trader_a,
+            NULL::bytea AS trader_b,
+            "inAmount" AS token_b_amount_raw,
```
jeff-dude, line 66: in v1 script, token_a = "inAmount" while this v2 script has token_b = "inAmount" -- do they need flipped here?

jeff-dude (changes requested):

ethereum/dex/trades/insert_clipper_v2.sql
aothman: Updated to pull symbol and decimals data from the erc20 tokens table.

ethereum/dex/trades/insert_clipper_v2.sql
jeff-dude: a change to v1 would require a reload, yes. that is up to you. are there downstream queries/dashboards which have a manual workaround on the flipped values, that would also need updated if we change?

ethereum/dex/trades/insert_clipper_v2.sql
```diff
@@ -0,0 +1,128 @@
...
+    AND project = 'Clipper'
+    AND version = '2'
+);
+
+INSERT INTO cron.job (schedule, command)
+VALUES ('*/10 * * * *', $$
+    SELECT dex.insert_clipper_v2(
+        (SELECT max(block_time) - interval '1 days' FROM dex.trades WHERE project='Clipper' AND version='2'),
+        (SELECT now() - interval '20 minutes'),
+        (SELECT max(number) FROM ethereum.blocks WHERE time < (SELECT max(block_time) - interval '1 days' FROM dex.trades WHERE project='Clipper')),
```
jeff-dude, line 125: i don't have access, but i pushed a change locally and added the version filter from line 123 to this spot as well. please add as a commit prior to merge

polygon/dex/trades/insert_clipper_v2.sql
```diff
@@ -0,0 +1,126 @@
...
+        tx_hash,
+        tx_from,
+        tx_to,
+        trace_address,
+        evt_index,
+        trade_id
+    )
+    SELECT
+        dexs.block_time,
+        pa.symbol AS token_a_symbol,
```
jeff-dude, line 32: did you want to change these similar to the ethereum script? i can change locally and run, but then would need your access to push commit here

ethereum/dex/trades/insert_clipper_v2.sql
aothman: Let's leave it alone then. Not hurting anything and it's no longer active.

jeff-dude (approved):
@aothman thank you! should be all set

PR #1126, opened 2022-06-06
reorganizing folders

dot2dotseurat:
I think our file structure is drifting away from a structure. We currently have NFT and DEX projects at the top level but then nft and dex (balances) directories also at the top. 

I think the project directories should be nested under their "sector". So Opensea and MagicEden should be under nft and Uniswap should be under dex. 

Then we can try to follow `sector/protocol/chain` or `sector/protocol/concept/chain` depending on complexity. 
`dex/erc20/balances/ethereum` 
`dex/uniswap/ethereum`
`nft/opensea/solana`

below what is currently in master
Screen Shot 2022-06-06 at 12 02 35 PM

N/A below
I've checked that:


dot2dotseurat:
ok, you're right we should match whats in the data explorer.

PR #1127, opened 2022-06-06
setup video

dot2dotseurat:
Screenshot video of setting up the project per wizards requests. I tried a couple things to reformat this and upload as a gif but kept running into issues. I think this is ok but a bit larger than I'd like. 

N/A
I've checked that:


soispoke (commented):
Nice, this looks good. The only thing I was thinking is maybe add a dbt debug command run before the first dbt compile since it's recommended by dbt ?

dot2dotseurat:
@soispoke the debug step sadly will fail because it's checking for the database connection

soispoke:
Oh I see, LGTM then !

soispoke (approved):

PR #1128, opened 2022-06-06
Add Uniswap & SushiSwap to zerion.trades

hildobby:
I've checked that:


jeff-dude (approved):
@hildobby i pushed a commit and then ran on the db, you should be all set 👍  let me know if anything needs adjusted

PR #1130, opened 2022-06-06
[dbt test] update nft tokens to 2022-06-04

dot2dotseurat:
@jeff-dude  I copied over @0xbitfly's work here so the CI tests would run. For context, we have an issue where commits from forked repos don't get picked up by our CI tool so I just made a new branch off of your branch. 

That test caught a missing comma in the v2 version of the tokens table. I pushed a small commit that fixes that issue. 

If you want to test your V2 contributions, when you run `dbt compile` (after installing it) and it creates a new folder called target with your compiled SQL.  You can copy the SQL there and run it directly on dune.com to check it. 

I've checked that:

* [n/a] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [n/a] if adding a new model, I added a test

### DIFF
--- a/ethereum/nft/tokens.sql
+++ b/ethereum/nft/tokens.sql
@@ -26526,6 +26526,2416 @@ COPY nft.tokens (contract_address, name, symbol, standard) FROM stdin;
 \\xfeee2b2cbdf19ec13313bab02b775c5d278fbae1	Mutant Yeti Dao	MYD	erc721
 \\xffac1e0a91ffea4746359e222f5f3fcba7b4aebe	U R	WORLD	erc721
 \\xffba892d276e4687e44edfa9de529851559f3634	Elongevity Enigma	ELONENIGMA	erc721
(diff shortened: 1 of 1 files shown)


0xbitfly:
Nice job! So for the v2 version PR, I should check the dbt compile result sql on dune?

dot2dotseurat:
> 

yep, let me know if you have questions or run into issues with that. what should happen is your query is compiled and you can run it directly on dune.com to test.

side note: 
it's a little more complicated when you are referencing another view/table that you are creating in the dbt project because your personal schema gets prepended to the view/table name so nft.trades would become wizard_nft.trades etc but we are working on a fix for that. the prepended schema thing is a feature in dbt so colleagues can work on the same table without collisions but for us since everyone it working on dune.com it's a little wonky. 

but in general yes. compile and run on dune.com. if you get an error about a table not existing, check to see if it's because the wizard schema is prepended to a table name.

jeff-dude:
fyi the v1 engine tokens table has been refreshed with the updated file here

jeff-dude (approved):
@dot2dotseurat apologies if you were waiting on my approval, but all set from my side, if you're good 👍

PR #1143, opened 2022-06-08
Created View for Opensea Seaport

sohwak:
I've checked that:


jeff-dude (approved):
@sohwak the view has been created 👍

PR #1150, opened 2022-06-10
fix balancer views

mendesfabio:
I've checked that:

### DIFF
--- a/ethereum/balancer/view_vebal_votes.sql
+++ b/ethereum/balancer/view_vebal_votes.sql
@@ -76,7 +76,6 @@ WITH calendar AS (
         ON v.round_id <= r.round_id
         AND r.round_id < v.next_round
         AND r.end_timestamp <= v.unlocked_at
-        AND v.weight > 0
     )
 
 SELECT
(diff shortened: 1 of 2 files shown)


ethereum/balancer/view_vebal_votes.sql
```diff
@@ -76,7 +76,6 @@ WITH calendar AS (
         ON v.round_id <= r.round_id
         AND r.round_id < v.next_round
         AND r.end_timestamp <= v.unlocked_at
-        AND v.weight > 0
```
jeff-dude, line 79: i'm unable to drop this view as 'dune_user_generated.vw_tianqi_gauge_reward' depends on it -- is this your user generated object? would you like to back it up, then drop, then allow this rebuild?

ethereum/balancer_v1/view_liquidity.sql
```diff
@@ -58,7 +58,7 @@ CREATE MATERIALIZED VIEW balancer_v1.view_liquidity AS (
             1,
             2
         HAVING
-            SUM(sample_size) > 3
+            SUM(sample_size) > 5
```
jeff-dude, line 61: same comment here on this file with same user generated table restricting the rebuild

ethereum/balancer/view_vebal_votes.sql
mendesfabio: it's not mine but I know the user who maintains it. I don't think we need to back it up - I'm just fixing some issues in both views here and didn't change anything else.

jeff-dude (approved):
@mendesfabio updated

PR #1141, opened 2022-06-08
Change materialization for nft/uniswap trades

soispoke:
This PR improves both create and query speed by changing the materialization strategy for nft and dex trades:
Here is an example for uniswap, but the same logic applies to nft trades (e.g., opensea_ethereum_trades and opensea_solana_trades UNIONed ALL to give opensea_trades)

Before:
uniswap_v2_trades                 uniswap_v3_trades
     Incr. Table                                   Incr. Table
                  
                          uniswap_trades
                                 View

Now:
uniswap_v2_trades                 uniswap_v3_trades
           View                                          View
                  
                          uniswap_trades
                                 Incr. Table

I've checked that:

* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] the directory tree matches the pattern /sector/blockchain/ e.g. /tokens/ethereum
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each file has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`

PR #1155, opened 2022-06-13
Revert "Change materialization for nft/uniswap trades (#1141)"

soispoke:
Ok so I thought the change of materialization, but for some (very obscure) reasons the magiceden trades take forever to create a table so it's probably best to revert this so we're back to normal (and running), and then investigate further why magiceden won't work properly

This reverts commit 0356bd532cb353af15c429af1e2355d1bb5dfa8b.

I've checked that:


aalan3 (approved):

soispoke:
Thanks @aalan3 !

PR #1158, opened 2022-06-13
Update insert_1inch.sql

yulesa:
Making 1inch on polygon similar to ethereum, including the Limit Order Protocol in the same insert.

A dune_user_generated.oneinch_dex_trades table was created to visualize all data although some double inclusions have been made to this table because I needed to break the insertions into many jobs.

Queries using this table, even though data is duplicated:
https://dune.com/queries/893408 : query that summarizes some of the most important stats on 1inch.
https://dune.com/queries/906627 : query that summarizes some of the most important stats on 1inch Limit order protocol.

View table simulating the insertion, but for limited time frames:
https://dune.com/queries/896051

I've checked that:

### DIFF
--- a/polygon/dex/trades/insert_1inch.sql
+++ b/polygon/dex/trades/insert_1inch.sql
@@ -27,12 +27,12 @@ WITH rows AS (
         evt_index,
         trade_id
     )
-    SELECT
+        SELECT
         dexs.block_time,
-        pa.symbol AS token_a_symbol,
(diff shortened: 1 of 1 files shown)


jeff-dude:
reload is in process. i made a few minor tweaks for performance, will push when the code completes.

jeff-dude (approved):
@yulesa this should be rebuilt and automation enabled

PR #1159, opened 2022-06-14
add new pool trade events for hashflow on all chains & init op tables

pani-poori:
for the new trade events changes on ETH, Polygon, BSC -- here are the partial test of the new subqueries:

[Ethereum PR Test](https://dune.com/queries/907494)
[Polygon PR Test](https://dune.com/queries/907550)
[BSC PR Test](https://dune.com/queries/907573)

After this PR is merged, hashflow table will need a backfill from 2022-04-09.
There is no code change needed for inserting into dex.trades -- as it's a simple select from hashflow tables -- once the table's backfill is done, they will need a backfill from same dates (all chains start from 2022-04-09); and after that all data will flow automatically.

For the Optimism new hashflow tables, i tested the whole functions with dune_user_generated view:

[table function test](https://dune.com/queries/907557)
[function test result](https://dune.com/queries/907605) looks good.

They will need a backfill since 2022-05-07. Also added the insert_hashflow for dex.trades table on Optimism, which also need a backfill from same date.

cc our CTO @gxmxni

----
I've checked that:


jeff-dude (approved):
@pani-poori these should all be built/rebuilt 👍

PR #1161, opened 2022-06-14
Update erc20.tokens: add 19 missing tokens including HOP

bh2smith:
Fetched from Missing Tokens Query: https://dune.com/queries/236085


jeff-dude (approved):
@bh2smith added 👍

PR #1164, opened 2022-06-14
Remove duplicates from `transfers_ethereum.erc20` originating in `erc20_ethereum.evt_transfer`

dot2dotseurat:
This PR addresses an issue raised by @MSilb7. https://github.com/duneanalytics/abstractions/issues/1120 

`erc20_ethereum.evt_transfer` has duplicated values keyed on `evt_tx_hash` and `evt_index`.  I will open an internal issue to identify the root cause. 

Screen Shot 2022-06-14 at 4 38 53 PM

Screen Shot 2022-06-14 at 4 39 00 PM

Screen Shot 2022-06-14 at 4 40 51 PM

In the meantime, `transfers_ethereum.erc20` is a slightly munged version of this table that we use to create token balances. I'm proposing using this view as a substitute until we fix the root cause. 

I've added a test on the a unique key coalesced from:
-  'send' || '-' || evt_tx_hash || '-' || evt_index || '-' || `to`  for sent transfers 
-  'receive' || '-' || evt_tx_hash || '-' || evt_index || '-' || `from` for received transfers 
-  'withdrawn' || '-' || evt_tx_hash || '-' || evt_index || '-' || src for withdrawn WETH
-  'deposit' || '-' || evt_tx_hash || '-' || evt_index || '-' || dst for deposited WETH

I added the tx type (send , receive etc) because I found there were transactions where the `to` and `from` addresses were the same address so the original unique key was insufficient. 
example: https://etherscan.io/tx/0xa3a30dc42d0276aca0ccc36aa3fb6644bee589afed121bae1c36c163e855d875

This PR also adds the WETH deposits and withdrawals into the transfers tables. Previously this was removed because those tables were not ready yet. 

I've checked that:

### DIFF
--- a/spellbook/models/transfers/ethereum/erc20/transfers_ethereum_erc20.sql
+++ b/spellbook/models/transfers/ethereum/erc20/transfers_ethereum_erc20.sql
@@ -3,8 +3,8 @@
 with
     sent_transfers as (
         select
-            evt_tx_hash || '-' || evt_index || '-' || to as unique_tx_id,
-            to as wallet_address,
+            'send' || '-' || evt_tx_hash || '-' || evt_index || '-' || `to` as unique_tx_id,
+            `to` as wallet_address,
             contract_address as token_address,
             evt_block_time,
             value as amount_raw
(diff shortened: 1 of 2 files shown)


soispoke (commented):
Ok, wow I wasn't expecting exact duplicates when you told me about this issue I thought it came from the abstractions we created, not the decoded data. We should let the DP team know about this, I'll open a Height task so we can figure this out.

Nice work fixing it in the meantime !

soispoke (approved):

spellbook/models/transfers/ethereum/erc20/transfers_ethereum_erc20.sql
```diff
@@ -3,8 +3,8 @@
 with
     sent_transfers as (
         select
-            evt_tx_hash || '-' || evt_index || '-' || to as unique_tx_id,
-            to as wallet_address,
+            'send' || '-' || evt_tx_hash || '-' || evt_index || '-' || `to` as unique_tx_id,
```
soispoke, line 6: unique_transfer_id might be better than unique_tx_id ?

kophyo1234 (approved):

PR #1169, opened 2022-06-15
standardize unique_trade_id

soispoke:
Standardize descriptions and unique_trade_id columns + its associated tests across existing abstractions

I've checked that:

### DIFF
--- a/spellbook/models/magiceden/solana/magiceden_solana_schema.yml
+++ b/spellbook/models/magiceden/solana/magiceden_solana_schema.yml
@@ -11,8 +11,8 @@ models:
     description: >
         MagicEden trades on Solana
     columns:
-      - name: unique_id
-        description:  "Unique id, combination of signatures[0] and id"
+      - name: unique_trade_id
+        description:  "Unique trade id, combination of signatures[0] and id"
         tests:
           - unique
       - &blockchain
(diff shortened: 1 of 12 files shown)


soispoke (commented):
dbt cloud is passing but was taking way too long + too much ressources

spellbook/models/magiceden/solana/magiceden_solana_schema.yml
```diff
@@ -11,10 +11,8 @@ models:
     description: >
         MagicEden trades on Solana
     columns:
-      - name: unique_id
-        description:  "Unique id, combination of signatures[0] and id"
-        tests:
```
dot2dotseurat, line 16: are we taking this out because it's taking too long to run this test?

soispoke (commented):
Yes @dot2dotseurat I've added this in the previous comment but those tests have been running and passing dozens of time and take a loooot of compute/run time

dot2dotseurat:
> Yes @dot2dotseurat I've added this in the previous comment but those tests have been running and passing dozens of time and take a loooot of compute/run time

alright, I think it's too complicated for right now but it would be cool if we could find a way to run these tests once a week. I think dbt core gives us enough tools to configure including and excluding tests but let's figure that out later. 

Like maybe we can use tags tests that run only for certain jobs.

dot2dotseurat (approved):

soispoke:
Yeah I like the idea of running these tests once a week, I'll take a shot at it tomorrow :)

soispoke (commented):
@dot2dotseurat Just so you know, I reverted the changes I made on tests here and will open a new PR to tackle this particular issue

PR #1171, opened 2022-06-15
Update DEPLOYMENT.md

dot2dotseurat:
I pushed a change for an incremental table that changed a column name. This worked on the PR job but resulted in errors in the production job because the new column name wasn't present in the original incremental table.

This is just a small note in the deployment.md to for future deployers to avoid my mistake. 

N/A below
I've checked that:


jeff-dude (approved):
this file change makes sense. are you able to link the change you're referring to?

PR #1177, opened 2022-06-16
Drop dbt cloud pr schemas macro

soispoke:
Current attempt at creating a macro to periodically delete dbt_cloud_pr schemas

I've checked that:


soispoke (commented):
Not sure why this isn't working tbh, maybe we just can't delete schemas from dbt directly ?

soispoke:
Closing this now since I found an alternative solution implementing a weekly job directly on Databricks :)

PR #1178, opened 2022-06-16
add categories to erc721 balances

soispoke:
After having integrated the category column (PR by Hildobby), I think it makes sense to include it in erc721 balances 

I've checked that:


jeff-dude (approved):
@soispoke i like it 👍

soispoke (commented):
thanks @jeff-dude for reviewing it !

PR #1188, opened 2022-06-17
add transfers and balances for erc1155 tokens on ethereum

soispoke:
Also added a test against erc1155 transfers on v1 (no erc1555 balances on v1 afaik) + checked individual tx_hash on etherscan

I've checked that:


spellbook/models/transfers/ethereum/erc1155/transfers_ethereum_erc1155_rolling_day.sql
```diff
@@ -0,0 +1,18 @@
+{{ config(
+        alias ='erc1155_rolling_day',
+        materialized ='view'
```
dot2dotseurat, line 3: is this redundant? is view the default materialization for this directory in the dbt project?

spellbook/tests/transfers_ethereum/transfers_ethereum_erc1155_assert_samples_single.sql
```diff
@@ -0,0 +1,13 @@
+-- Bootstrapped correctness test against legacy Postgres values.
+-- Also manually check etherscan info for the first 5 rows
```
dot2dotseurat, line 2: we could make that manual check into a actual test but we should have enough coverage using the legacy vals.

spellbook/models/transfers/ethereum/erc1155/transfers_ethereum_erc1155_rolling_day.sql
soispoke: Yes that's right ! Just modified this

dot2dotseurat (approved):
lgtm! thanks for adding the test. tedious but useful!

soispoke (commented):
thanks !

PR #1191, opened 2022-06-17
fix incremental strategy for erc721 and erc20 tables

soispoke:
The fact that we found out very late about duplicates in erc20/721 transfers + transfers tables took quite a long time to build got me investigating the incremental strategy we used.

Turns out we didn't include the unique_key parameter while using the merge strategy, which basically means the table was rebuilt everytime, without checking for uniqueness.

This PR adresses this issue, and improves both build time and compute costs

I've checked that:


dot2dotseurat (approved):

soispoke:
thanks !

PR #1193, opened 2022-06-20
Fix incremental run time issues

soispoke:
In a previous PR, the aim was to standardize the incremental strategy across spells: https://github.com/duneanalytics/abstractions/commit/9ae14c5289a85605314bcd6fd49efcb05567eecc 

 We were using the « merge » strategy, which allows to update fields if the underlying data changed for the whole table, and append new data. While I remain convinced the merge strategy is the best one to ensure data quality, this unfortunately came at a cost: Quarter hourly jobs could not run under 15 minutes, which caused jobs to queue up.   To mitigate this, I propose to use the « default » incremental strategy, that just appends new data incrementally (without updating the whole dataset if needed). The big advantage is that tables are updated around twice as fast as the merge strategy. To mitigate the fact that existing data can’t be updated, I added uniqueness tests to all incremental tables so we can regularly make sure there are no duplicates/data quality is good.

I've checked that:


masquot (approved):
thank you for the walk-through @soispoke. lgtm.

PR #1202, opened 2022-06-21
Add Clipper Coves and Relevant ERC20 tokens to Polygon

aothman:
Clipper Coves extend Clipper to trade arbitrary ERC20 tokens. This PR adds Clipper Coves into the dex.trades table and adds a handful of the most popular ERC20 tokens trading on Clipper Coves that are currently undefined.

I've checked that:

* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory) - **tested inner sub-query**
* [ NA ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ NA ] if adding a new model, I added a test


polygon/dex/trades/insert_clipper_coves.sql
```diff
@@ -0,0 +1,132 @@
...
+        FROM
+            clipper."ClipperCove_evt_CoveSwapped" t
+    ) dexs
+    INNER JOIN polygon.transactions tx
+        ON dexs.tx_hash = tx.hash
+        AND tx.block_time >= start_ts
+        AND tx.block_time < end_ts
+        AND tx.block_number >= start_block
+        AND tx.block_number < end_block
+    LEFT JOIN erc20.tokens erc20a ON erc20a.contract_address = dexs.token_a_address
```
jeff-dude, line 88: i added the tokens from your edited file, however, i believe there may still be some missing from this dex. with missing tokens in erc20.tokens, there can be missing data in amount fields and symbols. you may look to add more tokens to the file using this as reference:
https://dune.com/queries/956351?d=7

in addition to the tokens table entries, may want to look at adding all the tokens to prices file as well, if missing:
https://github.com/duneanalytics/abstractions/blob/master/prices/polygon/coingecko.yaml

we should look to get the best data quality we can prior to initial history backfill. let me know if this all makes sense

jeff-dude (changes requested):

polygon/dex/trades/insert_clipper_coves.sql
aothman: Thanks. I added a couple more notable tokens and added Coingecko feeds for all of the new entries. The biggest remaining ones are wormhole tokens - making a decision about whether wormhole USDC is "real" USDC is beyond the scope of this PR.

polygon/dex/trades/insert_clipper_coves.sql
```diff
@@ -0,0 +1,132 @@
...
+        -- ClipperCove Polygon
+        SELECT
+            t.evt_block_time AS block_time,
+            'Clipper' AS project,
+            'coves1' AS version,
+            'DEX' AS category,
+            t."recipient" AS trader_a,
+            NULL::bytea AS trader_b,
+            "inAmount" AS token_b_amount_raw,
+            "outAmount" AS token_a_amount_raw,
```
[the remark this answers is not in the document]
aothman: Thanks, I've made this change and also added a default value of 0 if we don't know anything about either token price!

jeff-dude:
@aothman apologies for inconvenience, but can you remove the prices file from this PR and create a separate one with just that file? i'll need to merge that first, then get prices populated, then come back here and populate dex trades

polygon/dex/trades/insert_clipper_coves.sql
aothman: I'll add up to the ones with 20 txs. Since Clipper Coves can trade any ERC20 token we won't be able to cover _everything_. The value calculation will just default to 0 - a failsafe - if it's swapping tokens we don't know anything about.

jeff-dude: yep, fully understand. just double checking. all good!

jeff-dude (approved):
@aothman thank you for the quick commits. it's been backfilled and automation enabled 🙌

PR #1203, opened 2022-06-22
feat: add aBNBc on BNB Chain

ittikorns:
I've checked that:

* [X] the directory tree matches the pattern /sector/blockchain/ e.g. /tokens/ethereum


jeff-dude (approved):
thank you @ittikorns

PR #1207, opened 2022-06-22
import new price

umbnat92:
New price for token:
- ETH
- soETH
- wstETH
- BTC
- DFL
- 1SOL


jeff-dude (approved):
@umbnat92 we can merge these into the repo, however, you may not see the data until next week. @aalan3 can update here when we have the data flowing through

PR #1211, opened 2022-06-24
feat: add insert_paraswap for dex.trades on BSC

harshagrawal13:
I've checked that:

### DIFF
--- a/binance_smart_chain/dex/trades/insert_paraswap.sql
+++ b/binance_smart_chain/dex/trades/insert_paraswap.sql
@@ -1,0 +1,156 @@
+CREATE OR REPLACE FUNCTION dex.insert_paraswap(start_ts timestamptz, end_ts timestamptz=now(), start_block numeric=0, end_block numeric=9e18) RETURNS integer
+LANGUAGE plpgsql AS $function$
+DECLARE r integer;
+BEGIN
+WITH rows AS (
+    INSERT INTO dex.trades (
+        block_time,
+        token_a_symbol,
+        token_b_symbol,
+        token_a_amount,
+        token_b_amount,
+        project,
+        version,
+        category,
+        trader_a,
+        trader_b,
+        token_a_amount_raw,
+        token_b_amount_raw,
+        usd_amount,
+        token_a_address,
+        token_b_address,
+        exchange_contract_address,
+        tx_hash,
+        tx_from,
+        tx_to,
+        trace_address,
+        evt_index,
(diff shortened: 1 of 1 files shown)


jeff-dude:
@harshagrawal13 thanks for the PR!

there appears to be a list of tokens missing from the prices.usd feed, query for reference:
https://dune.com/queries/972603?start_ts_d83555=2022-06-01+00%3A00%3A00
are these needed for downstream consumption? the key here is that it leads to null values for USD amount which tends to be used in report queries often. if the volumes are small and/or tokens aren't wanted, let me know.

the prices file can be edited here:
https://github.com/duneanalytics/abstractions/blob/master/prices/bsc/coinpaprika.yaml

if adding prices, please do so in smaller batches (15-20 tokens per PR)

jeff-dude:
similar idea with bep20 tokens file, with maybe less importance but still worth a look, query for reference:
https://dune.com/queries/972977
when null, symbols and raw amount fields can be null. this may impact downstream report queries.

bep 20 tokens can be added here:
https://github.com/duneanalytics/abstractions/blob/master/binance_smart_chain/bep20/tokens.sql
not much concern on number of tokens per PR here

jeff-dude:
> > @harshagrawal13 thanks for the PR!
> > there appears to be a list of tokens missing from the prices.usd feed, query for reference: https://dune.com/queries/972603?start_ts_d83555=2022-06-01+00%3A00%3A00 are these needed for downstream consumption? the key here is that it leads to null values for USD amount which tends to be used in report queries often. if the volumes are small and/or tokens aren't wanted, let me know.
> > the prices file can be edited here: https://github.com/duneanalytics/abstractions/blob/master/prices/bsc/coinpaprika.yaml
> > if adding prices, please do so in smaller batches (15-20 tokens per PR)
> 
> Hey! So the idea is that as more txs occur, newer tokens will be used and their prices would thus be needed too. I can create a list of PRs to add the existing tokens that are missing but is there a way to automatically keep on adding more tokens as they come up in transactions?

at this time, we don't have automation of the tokens. it will be explored in the new v2 engine and spellbook repo over time.

jeff-dude:
hi @harshagrawal13 -- do you still want to add a few more tokens to bsc file & polygon file (for #1212) or do you want me to proceed as-is?

jeff-dude (approved):
@harshagrawal13 as discussed on TG, i've gone ahead and built this and it's now live on bsc db

PR #1212, opened 2022-06-24
feat: add insert_paraswap to dex."trades" on Polygon

harshagrawal13:
I've checked that:


jeff-dude:
similar comments as #1211 -- both example queries from there apply here. for reference, here is prices:
https://dune.com/queries/973177

i didn't write out query for token symbols / amounts from join on erc20 tokens, but you can apply similar logic from other PR.

jeff-dude (approved):
@harshagrawal13 as agreed upon in TG, i've proceeded to build this and it's now live on polygon db

PR #1214, opened 2022-06-24
Import new token price solana

umbnat92:
stSOL price added


jeff-dude (approved):
thank you @umbnat92, fyi @aalan3

PR #1220, opened 2022-06-28
[CoW] Solverlist Cleanup

bh2smith:
This is just a cosmetic change to the solverlist values making it more readable (for those who are redirected here as a source of truth). 

Note that this is based on #1219, but they can be merged together at once if its cleaner (just wanted to keep the changes independent).


jeff-dude (approved):
@bh2smith should be all updated on both PRs

bh2smith:
So does that mean I should close this? Or will it be merged?

PR #1221, opened 2022-06-28
replace unique_tx_id by unique_transfer_id to match schema

soispoke:
This was causing the erc20 transfers uniqueness test to fail

*For Dune Engine V2*
I've checked that:


jeff-dude:
to clarify: the job was failing due to column name mismatch to the table schema, not unique tests on actual data? the change appears to be only column name, not column values. just double checking as i was a bit confused on the wording in the description

soispoke:
Yeah thanks for asking @jeff-dude it was a name mismatch to the table schema 👍

jeff-dude (approved):
sounds good!

soispoke:
Thanks !

PR #1224, opened 2022-06-29
python version

dot2dotseurat:
Python do be pythoning. Switching versions. 

Adding a commit to spellbook to trigger the action to test it and then I'll drop the commit if it works. 

*For Dune Engine V2*
I've checked that:


soispoke (commented):
LGTM ! Let's see if it works now

soispoke (approved):

dot2dotseurat:
This is failing due to the creds in dbt but not the python version. I'm going to squash are merge but the job will continue to fail until the creds are fixed.

PR #1226, opened 2022-06-29
Update Oracles to OP

MSilb7:

*For Dune Engine V2*
I've checked that:


jeff-dude (approved):
@MSilb7 updated, thank you

PR #1229, opened 2022-06-30
insert 1inch into binance smart chain dex trades

yulesa:
Here are the tests conducted:

dune_user_generated function insert_oneinch: https://dune.com/queries/976454
Inserted data: https://dune.com/queries/900194
Statistics on the dune_user_generated table:  https://dune.com/queries/977445 and https://dune.com/queries/981257

I found issue with these tokens: ('\xa4b6E76bba7413B9B4bD83f4e3AA63cc181D869F', '\xe76804b43f17fc41f226d63fd2a676df409d4678', '\x587c16b84c64751760f6e3e7e32f896634704352'). Be in mind that these tokens are filtered on the statistic queries.
I think they are scam tokens, copying tokens with real value and shouldn't be on the prices.usd table. This needs to be fixed before approving the PR.

*For Dune Engine V2*
I've checked that:


jeff-dude (approved):
@yulesa this should be good to go

PR #1230, opened 2022-06-30
edit bsc prices file

jeff-dude:
remove four tokens which lead to irregular usd amounts in dex trades
issue found when building 1inch script to populate dex trades
#1229


dot2dotseurat (approved):
lgtm

PR #1231, opened 2022-06-30
erc721 category causing dupes

dot2dotseurat:
Multiple categories per collection is causing duplicate rows which our unit tests alerted us to. 

I'm dropping this column until we have a fix that will convert the category column into categories. 

*For Dune Engine V2*
I've checked that:

### DIFF
--- a/spellbook/models/balances/ethereum/erc721/balances_ethereum_erc721_hour.sql
+++ b/spellbook/models/balances/ethereum/erc721/balances_ethereum_erc721_hour.sql
@@ -28,8 +28,7 @@ SELECT
     b.wallet_address,
     b.token_address,
     b.tokenId,
-    nft_tokens.name as collection,
-    nft_tokens.category as category
+    nft_tokens.name as collection
 FROM daily_balances b
 INNER JOIN hours d ON b.hour <= d.hour AND d.hour < b.next_hour
 LEFT JOIN {{ ref('tokens_ethereum_nft') }} nft_tokens ON nft_tokens.contract_address = b.token_address
--- a/spellbook/models/tokens/ethereum/tokens_ethereum_schema.yml
+++ b/spellbook/models/tokens/ethereum/tokens_ethereum_schema.yml
@@ -41,7 +41,7 @@ models:
       sector: nft
       contributors: dot2dotseurat, hildobby
     config:
-      tags: ['tokens', 'ethereum', 'nft']
+      tags: ['tokens', 'ethereum', 'nft', 'erc721']
     description: >
         Selection of NFT token addresses.
     columns:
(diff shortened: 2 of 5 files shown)


dot2dotseurat:
@hildobby for transparency

spellbook/models/tokens/ethereum/tokens_ethereum_schema.yml
```diff
@@ -54,7 +54,7 @@ models:
       - name: standard
         tests:
           - accepted_values:
-              values: ['erc721', 'erc1155' , 'erc20', 'cryptopunks','superrare']
+              values: ['erc721', 'erc1155' , 'erc20', 'cryptopunks','superrare', 'unknown']
```
soispoke, line 57: Do we know what 'unknown' tokens are / where they come from ?

spellbook/models/balances/ethereum/erc721/balances_ethereum_erc721_hour.sql
```diff
@@ -22,14 +22,13 @@ with
...
 
-SELECT
+SELECT distinct
     'ethereum' as blockchain,
     d.hour,
     b.wallet_address,
     b.token_address,
     b.tokenId,
-    nft_tokens.name as collection,
-    nft_tokens.category as category
```
soispoke, line 32: Is deleting the whole category column the only/best way ? It seems like it's an easy/temporary fix but it would still be cool to include categories in erc721 balances. Wouldn't distinct fix it ?

dot2dotseurat: no, just a temp fix. we should try to do a group by list aggregation but I need to figure out how to do it in sparksql.

spellbook/models/tokens/ethereum/tokens_ethereum_schema.yml
dot2dotseurat: Screen Shot 2022-07-01 at 10 19 10 AM
dencentaland wearables?

soispoke: wow ok, i guess that is unknown

spellbook/models/balances/ethereum/erc721/balances_ethereum_erc721_hour.sql
soispoke: Ok, if you think it'll take a long time to find the right fix we can push this but it made me a bit hesitant at first because it kind of feels like it wouldn't be too much work to get it fixed without excluding categories

soispoke: I'll still approve I don't want it to block you or anything

soispoke (approved):

PR #1236, opened 2022-07-01
[V2] Update ERC20 Tokens

bh2smith:
As requested in https://github.com/duneanalytics/abstractions/pull/1228 here is an update for the erc20 tokens in V2 Engine.

Based on this PoC Query: https://dune.com/queries/857522

We add the 129 missing token details detected.

Additional Notes:
1. This PR consists of two commits (the first adding the tokens) and the second fixing the awkward file format that was here before. For tokens added see this commit: https://github.com/duneanalytics/abstractions/pull/1236/commits/03b5dd8555f52576d74231aeb932d3d389b06151
2. It appears you are using lower case strings, so I made sure that they were all lower case. Please let me know if you prefer a different format.
3. Added ETH under the contract address `0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee` (as is common notation in the ecosystem).
4. The script used to generate this content is adapted from Dune V1 script here: https://github.com/cowprotocol/data-misc/pull/7


dot2dotseurat (approved):
lgtm and sorry I didn't see this earlier.

bh2smith:
No worries. I am working on a setup that may actually be used to keep both systems updated simultaneously (from the same source). Will experiment a bit and get back to you.

dot2dotseurat:
> No worries. I am working on a setup that may actually be used to keep both systems updated simultaneously (from the same source). Will experiment a bit and get back to you.

hell yeah. lfg.

PR #1239, opened 2022-07-02
Update curve pools & dex liquidity

### LINKED ISSUE #1216: dex.liquidity version for Curve is wrong
All the version in dex.liquidity for curve is `'1'` cause that's what was inserted in the [query](https://github.com/duneanalytics/abstractions/blob/9a7e4d9412bcb251a156f58f5e2a0cdb483df843/ethereum/dex/liquidity/insert_curve.sql)

A look at `distinct version` on tables `dex.trades` and `curvefi."view_pools"` would result in
1. Factory V1 Meta
2. Factory V1 Plain
3. Factory V2
4. Regular

I have 2 suggested solution
1. Use the same version as `dex.trades` and `curvefi.view_pools` by simply updating version in [liquidity query](https://github.com/duneanalytics/abstractions/blob/9a7e4d9412bcb251a156f58f5e2a0cdb483df843/ethereum/dex/liquidity/insert_curve.sql) starting from [`curve_pool_tokens`](https://github.com/duneanalytics/abstractions/blob/9a7e4d9412bcb251a156f58f5e2a0cdb483df843/ethereum/dex/liquidity/insert_curve.sql#L8-L18) downwards.
2. Take it a step further and edit [curvefi.view_pools](https://github.com/duneanalytics/abstractions/blob/7777cf26da41799117057543c3403adfac00fcd5/ethereum/curvefi/view_pools.sql) by updating `Regular` to either `Regular V1` or `Regular V2` so that there's clearer knowledge of V1 vs V2.

Challenge with step 2 is that it might affect older queries and Messari projects. But it's probably worth it knowing the difference between Regular V1 and Regular V2. Therefore I'm seeking approval or suggestion from @yulesa too.

bizzyvinci:
This is to resolve #1216 

- [x] Split Regular version to V1 and V2
- [x] Add 3Crv price
- [x] Add version to liquidity
- [x] Change ETH to WETH to keep consistency with dex.trades and make price calculation possible (This implementation might be wrong, would check it later) 

I don't know why some pools are missing in liquidity ([sample query](https://dune.com/queries/957508)). Is it because of when dex.liquidity was created? @duneteam

Requesting review from @yulesa too


jeff-dude:
> This is to resolve #1216
> 
> * [x]  Split Regular version to V1 and V2
> * [x]  Add 3Crv price
> * [x]  Add version to liquidity
> * [x]  Change ETH to WETH to keep consistency with dex.trades and make price calculation possible (This implementation might be wrong, would check it later)
> 
> I don't know why some pools are missing in liquidity ([sample query](https://dune.com/queries/957508)). Is it because of when dex.liquidity was created? @duneteam
> 
> Requesting review from @yulesa too

thank you @bizzyvinci for the PR.

first step -- can you move the prices file update to its own PR, that way i can push through and backfill in the meantime?

as for why some pools may be missing, i think @yulesa mentioned in the linked issue that dex trades had been updated recently, but not liquidity. looks like a large number of decoded tables may have been missing at the time (or didn't exist).

i'll wait a bit to see if @yulesa has any input to add on this PR.

yulesa:
The values don't seem correct. I've run your query here only for this year: https://dune.com/queries/1013121
Values are at least double what they should.

Another issue is there are 3 metapools, you've added 3CRV, but there are ren and sBTC pools. My query calculates the price of those tokens based on the liquidity of those pools.

A good practice is to provide a query, showing your results in the PR.

This is your query:
image
This is mine: 
image
This is etherscan:
image

bizzyvinci:
This should still be on hold cause there are some things I want to check and more commits.

bizzyvinci:
The reason for the double to triple value is incorrect join statement. The values in the [query](https://dune.com/queries/1022197) seems close now except for `stETH Concentrated`. This is because `erc20.token_balances` return values that are about 7 times the real value as show in this [query](https://dune.com/queries/1022356). Also, I'm thinking ren and sBTC price should be left out for simplicity and uniformity with other liquidity queries.

jeff-dude:
closing due to inactivity

PR #1241, opened 2022-07-04
T 13866 looksrare transactions on v2

soispoke:

This PR adds the LooksRare platform (Ethereum) on Dune v2.
It includes LooksRare transactions, mints, burns, and fees tables
Using the merge into strategy, I made rows are unique. I also added a test against Dune v1 nft.trades_v2_beta table
I made sure the new tables have the appropriate metadata and can be exposed on the Dune App + are integrated to get included in nft.transactions/trades/mints/burns.

*For Dune Engine V2*
I've checked that:


soispoke:
Said it failed w/ dbt cloud but this is because of: accepted_values_tokens_ethereum_nft_standard__erc721__erc1155__erc20__cryptopunks__superrare (not related)
Everything else passed 👍

bhndt (approved):

spellbook/models/looksrare/ethereum/looksrare_ethereum_events.sql
```diff
@@ -0,0 +1,157 @@
...
+        ask.contract_address AS contract_address,
+        ask.evt_tx_hash AS tx_hash,
+        ask.evt_block_number AS block_number,
+        ask.evt_index AS evt_index,
+        roy.evt_index as roy_event_index,
+        CASE -- CATEGORIZE Collection Wide Offers Accepted 
+            WHEN strategy = '0x86f909f70813cdb1bc733f4d97dc6b03b8e7e8f3' THEN 'Collection Offer Accepted'
+            ELSE 'Offer Accepted' 
+            END AS category    
+    FROM {{ source('looksrare_ethereum','looksrareexchange_evt_takerask') }} ask
```
jeff-dude, line 36: in the config, it's set to incremental & the sources file for looksrare seems to call out freshness on block time column, but how does dbt know to read only incremental data rather than the entire source dataset? is that automatically applied, or should it be explicitly written in this code?

jeff-dude: if it needs to be written out, i think all table reads and joins to larger tables (i.e. ethereum transactions outside this CTE) should include filters on dates, if possible?

spellbook/macros/alter_table_properties.sql
```diff
@@ -1,5 +1,5 @@
 {% macro alter_table_properties() %}
-{%- if target.name == 'prod'-%}
+{%- if target.name == 'dbt_thomas'-%}
```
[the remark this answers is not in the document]
soispoke: Yes ! Thanks a lot for spotting this, this was changed back to prod

spellbook/models/looksrare/ethereum/looksrare_ethereum_events.sql
soispoke: - You're right it should be explicitly specified (otherwise it does recompute everything every time it runs!), I added this to the code:
{% if is_incremental() %}
WHERE looks_rare.block_time > now() - interval 2 days
{% endif %} 

- Good thinking, I added a time filter on both erc721/1155 tables + ethereum.transactions, starting when the LooksRare platform launched (which I set to january 1st 2022)

jeff-dude: re: time filter on both erc721/1155 tables + ethereum.transactions -- i think we should restrict the joins to same filter as looks_rare subquery (i.e. > now() - interval 2 days). any reason the filters can't match between the two? 

over time, > '2022-01-01' can still take more time than necessary to process rows we don't need in current execution

spellbook/models/looksrare/ethereum/looksrare_ethereum_events.sql
```diff
@@ -154,4 +156,9 @@ LEFT JOIN  {{ ref('nft_ethereum_aggregators') }} agg ON agg.contract_address = t
 LEFT JOIN {{ source('prices', 'usd') }} p ON p.minute = date_trunc('minute', looks_rare.block_time)
     AND p.contract_address = currency_contract
     AND p.blockchain ='ethereum'
-LEFT JOIN {{ ref('tokens_ethereum_erc20') }} erc20 ON erc20.contract_address = currency_contract
\ No newline at end of file
+LEFT JOIN {{ ref('tokens_ethereum_erc20') }} erc20 ON erc20.contract_address = currency_contract
+WHERE tx.block_time > '2022-01-01'
```
jeff-dude, line 160: this may be more efficient as AND tx.block_time > '2022-01-01' under line 152 inner join constraints (but also same comment as above on date filter variable values vs. hardcoded)

spellbook/models/looksrare/ethereum/looksrare_ethereum_events.sql
```diff
@@ -154,4 +156,9 @@ LEFT JOIN  {{ ref('nft_ethereum_aggregators') }} agg ON agg.contract_address = t
...
+{% if is_incremental() %}
+-- this filter will only be applied on an incremental run
+AND looks_rare.block_time > now() - interval 2 days
```
jeff-dude, line 163: the if is incremental is helpful, thank you 👍 

one other point, it may be more efficient to move this if block into the looks_rare CTE itself (after line 38 and after line 64 on both tables) that way the CTE doesn't do full scan, then outer query filters, we can get ahead of the filtering

soispoke (commented):
- Added AND tx.block_time > '2022-01-01' under line 152 inner join constraints
- We can't have the same filters on erc721/1155 tables, since the > now() - interval 2 days is part of the merge incremental strategy and should only be applied on time for the whole table (that's why it's embedded in the jinja {% if is_incremental() %}  macro)
For the rest I'd rather play it safe and have time filters set to when the platform launch

spellbook/models/looksrare/ethereum/looksrare_ethereum_events.sql
```diff
@@ -0,0 +1,164 @@
...
+INNER JOIN {{ source('ethereum','transactions') }} tx ON tx_hash = tx.hash
+AND tx.block_time > '2022-01-01'
+LEFT JOIN erc_transfers erc ON erc.evt_tx_hash = tx_hash AND erc.token_id_erc = token_id
+LEFT JOIN {{ ref('tokens_ethereum_nft') }} tokens ON tokens.contract_address =  nft_contract_address
+LEFT JOIN  {{ ref('nft_ethereum_aggregators') }} agg ON agg.contract_address = tx.to
+LEFT JOIN {{ source('prices', 'usd') }} p ON p.minute = date_trunc('minute', looks_rare.block_time)
+    AND p.contract_address = currency_contract
+    AND p.blockchain ='ethereum'
+LEFT JOIN {{ ref('tokens_ethereum_erc20') }} erc20 ON erc20.contract_address = currency_contract
+{% if is_incremental() %}
```
jeff-dude, line 161: my last point is on this incremental logic -- can we move this jinja macro from here to within the looks_rare CTE? it would result in lines 161-164 being copied to after line 38 & line 64 (with edit to table/column name for block time), then removed from 161-164. this would allow the CTE to do the filtering if incremental enabled, then outer query will already have the designated subset you are ultimately filtering to, which simplifies join to eth tx's and prices

jeff-dude (changes requested):

soispoke (commented):
@jeff-dude I tried you approach but it failed and gave this error, I think we should keep it as is since `this` is supposed to refer to the whole table

soispoke (commented):
@jeff-dude I tried you approach but it failed and gave this error, I think we should keep it as is since `this` is supposed to refer to the whole table

jeff-dude:
> 

why do we need to read the full decoded tables each time, if it's set to incremental? and even if not incremental based on dbt command, WHERE clause is ignored regardless of location.

the error is in the screenshot, decoded tables have different naming standard for block time. it should be (or close to):
for ask alias table:
{% if is_incremental() %}
WHERE ask.evt_block_time >= (select max(block_time) from {{ this }})
{% endif %} 

for bid alias table:
{% if is_incremental() %}
WHERE bid.evt_block_time >= (select max(block_time) from {{ this }})
{% endif %} 


i'm not sure what {{ this }} refers to, but i assume target table merging into? if so, then 'block_time' should be right column name in the condition of where clause.

soispoke (commented):
@jeff-dude What I'm trying to say is that it won't work even with the right alias, since {{ this }} has to refer to the entire model and the merge strategy needs a unique_id column to be applied on since we do need to do a quick full run each time to check if fields were modified since last time we incrementally updated the table, here are some more docs: https://docs.getdbt.com/docs/building-a-dbt-project/building-models/configuring-incremental-models

jeff-dude:
> @jeff-dude What I'm trying to say is that it won't work even with the right alias, since {{ this }} has to refer to the entire model and the merge strategy needs a unique_id column to be applied on since we do need to do a quick full run each time to check if fields were modified since last time we incrementally updated the table, here are some more docs: https://docs.getdbt.com/docs/building-a-dbt-project/building-models/configuring-incremental-models

sorry to hammer at this point, but i think it may still be valid 🙂 there is no difference in row output when filtering on incremental how you have it now vs. moving the filter into the CTE. either way, same results, just improved performance when filtering within the CTE.

the link you sent actually calls that out too:
_Optimizing your incremental model
For more complex incremental models that make use of Common Table Expressions (CTEs), you should consider the impact of the position of the is_incremental() macro on query performance. In some warehouses, filtering your records early can vastly improve the run time of your query!_

soispoke (commented):
@jeff-dude Ok no thanks for pushing you are actually right ! Moving it inside the CTEs does work, and does speed up the build time. SOrry for not seeing this earlier !

spellbook/models/looksrare/ethereum/looksrare_ethereum_events.sql
```diff
@@ -0,0 +1,170 @@
...
+        bid.evt_tx_hash AS tx_hash,
+        bid.evt_block_number AS block_number,
+        bid.evt_index AS evt_index,
+        roy.evt_index as roy_event_index,
+        'Buy' as category
+    FROM {{ source('looksrare_ethereum','looksrareexchange_evt_takerbid') }} bid
+    LEFT JOIN {{ source('looksrare_ethereum','looksrareexchange_evt_royaltypayment') }} roy ON roy.evt_tx_hash = bid.evt_tx_hash
+    AND roy.evt_index = bid.evt_index - 4
+     {% if is_incremental() %} -- this filter will only be applied on an incremental run 
+     AND bid.evt_block_time >= (select max(block_time) from {{ this }})
```
jeff-dude, line 69: same as other comment, but for bid alias here using WHERE instead

jeff-dude (changes requested):
final thing before approval 👍

soispoke (commented):
Done :)

soispoke (commented):
Thank you @jeff-dude !

PR #1244, opened 2022-07-05
Update OP Kwenta DEX Insert

MSilb7:
Updating the Optimism DEX Trades insert for Kwenta to better pick the tokens traded.

Would we be able to wipe and re-insert the Kwenta trades? Some more recent trades are showing sETH for sETH because of multi-hops causing an issue with the prior construction.

Eventually we'd need to figure out some verification that we're picking the synth token contract addresses and not another token with the same symbol (maybe add a 'project name' field to erc20 tokens?)

*For Dune Engine V2*
I've checked that:


jeff-dude:
fyi -- the history was reloaded and we removed 753,159 rows & reloaded 73,663. i'm assuming the large decrease in rows is due to you removing all the multi-hops mentioned above?

if this feels off, feel free to submit another PR with any updates. it was relatively quick to rebuild.

jeff-dude (approved):

PR #1245, opened 2022-07-05
ENS Spells Migration

antonio-mendes:
Height Task: https://dune.height.app/T-13877

*For Dune Engine V2*
I've checked that:


spellbook/models/ens/ens_view_registries.sql
```diff
@@ -0,0 +1,14 @@
+{{config(alias='view_registries')}}
+SELECT 
+  node,
+  label,
+  min(evt_block_time) AS min_evt_block_time, 
+  max(evt_block_time) AS max_evt_block_time, 
+  count(*)
```
soispoke, line 7: This one is a bit weird, but basically it fails because count(*) needs an alias (e.g., count(*) as count_ens_registries)

soispoke (commented):
Everything should work once these changes are made 👍

soispoke (commented):
Finished reviewed + commited the appropriate changes so you can have an example @antonio-mendes

soispoke (commented):
LGTM, but I think we might need to materialize those (querying them takes forever)

spellbook/dbt_project.yml
```diff
@@ -128,5 +131,26 @@ seeds:
             block_time: timestamp
             tx_hash: string
             amount: string
-
+    ens:
+      ethereum:
```
soispoke, line 135: Here, if you had ethereum then you need to make sure the files are under ens/ethereum/filename.csv (they are currently under ens/filename.csv)

soispoke: So you either need to remove ethereum from the dbt_project.yml file or to add a folder in which you place the files

soispoke (approved):

PR #1247, opened 2022-07-05
Uniswap unique trade id + version

dot2dotseurat:
Uniswap trades table was failing on the following error:

`Cannot perform Merge as multiple source rows matched and attempted to modify the same\ntarget row in the Delta table in possibly conflicting ways`

Likely that means the unique_trade_id wasn't completely unique. I've appended the version to the start of each ID to unique-ify it. 

*For Dune Engine V2*
I've checked that:

### DIFF
--- a/spellbook/models/uniswap/ethereum/uniswap_ethereum_schema.yml
+++ b/spellbook/models/uniswap/ethereum/uniswap_ethereum_schema.yml
@@ -72,7 +72,7 @@ models:
       project: uniswap_v2
       contributors: masquot, soispoke
     config:
-      tags: ['ethereum','uniswap_v2','trades']
+      tags: ['ethereum','uniswap_v2','trades', 'uniswap']
     description: >
         Uniswap V2 trades on Ethereum
     columns:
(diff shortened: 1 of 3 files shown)


spellbook/models/uniswap/ethereum/uniswap_ethereum_schema.yml
```diff
@@ -101,7 +101,7 @@ models:
       project: uniswap_v3
       contributors: masquot, soispoke
     config:
-      tags: ['ethereum','uniswap_v3','trades']
+      tags: ['ethereum','uniswap_v3', 'trades', 'uniswap']
```
soispoke, line 104: This just makes me think of something I've found useful for tags, is separating 'uniswap' and 'v3' (but adding uniswap_v3 is fine too)

spellbook/models/uniswap/ethereum/uniswap_v3_ethereum_trades.sql
```diff
@@ -3,7 +3,7 @@
 }}
         
 SELECT
-    tx_hash || '-' || evt_index::string as unique_trade_id,
+    'v3' || '-' || tx_hash || '-' || evt_index::string as unique_trade_id,
```
soispoke, line 6: Feels weird to me that it was the overlap between v2 and v3 that caused duplicates on unique_trade_id

soispoke (commented):
If this really fixes it, LGTM :)

soispoke (approved):

soispoke (commented):
Looks like it didn't fix it, opening a new PR for this

PR #1252, opened 2022-07-06
update 3Crv in erc20 and add its price

bizzyvinci:
Backstory: #1239#issuecomment-1175253542


jeff-dude (approved):
@bizzyvinci thank you for moving these to separate PR

PR #1256, opened 2022-07-06
Update APE Price Feed OP2

MSilb7:

*For Dune Engine V2*
I've checked that:


jeff-dude:
@MSilb7 that should all be fine. i'm going to pause the cron jobs now & get to fixing the above

jeff-dude (approved):
@MSilb7 the oracle_addresses file was updated, then backfill on view_price_feeds was run. i noticed there was no data for ape / usd feed_name. i then noticed only 23/46 of the oracle addresses are represented in the table.

```
--106,717
select count(1)
from chainlink."view_price_feeds"
;

--23 of 46 total feeds from chainlink.oracle_addresses
select feed_name
    , count(1) as total_count
from chainlink."view_price_feeds"
-- where feed_name like '%APE%'
group by feed_name
order by 2 desc
;
```

i'll close this PR as the change is reflected, but if you think based on the above there is still an issue, please feel free to submit new PR on function to load the feed table to fix the gaps.

PR #1263, opened 2022-07-08
[CoW Protocol] Use CSV for named solvers

bh2smith:
In an effort to make the raw solver data more legible/accessible we store the solver name data in CSV which is populated via `COPY <TABLE> FROM`.  The `view_solvers` is then constructed as a join of this `named_solvers` table. 

A couple of open concerns I have about this setup:

1. Not sure if the file path for the `COPY FROM` is correct (would require dune team)
2. In named solvers the address table is string but should probably be bytea. This would amount to using `\\x` or possibly just `\x` prefix instead of `0x` prefix.
3. I am still wondering what will happen on a follow up PR when solvers are updated... Seems like there would be conflict. 
One option is to TRUNCATE the table immediately before the copy, but I suspect this would conflict with the solver view that depends on it... Another option would be something similar to `ON CONFLICT do nothing`, but then I guess any conflict would result in no update. Is there some equivalent command like `ON CONFLICT insert those which do not conflict?`

If this technique does actually work out well here, then I would also propose we do something similar for `erc20.tokens` that can be used to keep both the legacy platform and v2 engine both in sync simultaneously. That is keep the token list in a single CSV that is used to generate the tokens table in both places.

cc @jeff-dude and @dot2dotseurat (pertaining to https://github.com/duneanalytics/abstractions/pull/1236#issuecomment-1178034245)


jeff-dude:
thanks for proving this out in a PR, i'll take a closer look soon. i'm fairly certain that file path will need altered, but i can try to see what would work in it's place

jeff-dude:
here are my thoughts after a closer look:
1. the csv file in it's own PR, done first, which allows merge into repo
2. once merged, can access raw csv file ideally
3. switch COPY command to \copy -- the former requires certain access rights, the latter seems to be a good alternative (https://stackoverflow.com/questions/29004860/how-to-allow-permission-to-access-csv-file-using-postgres-in-ubuntu)
4. when csv is in master, should be able to flip the URL to refer to 'master' rather than 'a68ea6d2c3045d02c29c974556d0d230533a1263' (i pushed a commit to show what copy may look like)

in it's current state, psql command line isn't recognizing that URL path. i'd like to try it post-merge to see if we can continue down this path. otherwise, we'd need some sort of way to download the csv either locally or on the postgres server and run that way, which isn't as efficient.

bh2smith:
Great suggestions @jeff-dude! I've updated this branch to only include the CSV and #1284 is the follow up PR.

jeff-dude (approved):
let's see if this works 🤞

PR #1265, opened 2022-07-08
Add base58decode

aalan3:
This adds a function that makes it possible to decode base58 strings which is very useful for Solana data.

I'm not sure if this should just be a macro or if a function will work.


theflyhigh:
it will be very useful if we can decode base58 data of solana,I prefer it to be a function

soispoke (commented):
@aalan3 Thanks for the submission 👍

There was one minor issue: You wrote the macro as a SQL function, so I changed it so it fits dbt/jinja.
I also created the UDF SQL function in the back end so users can either use it as a macro in the following way : 

select
  {{ base58_decode('column_name') }}
  ...
from a_given_table

OR 

Use the function this way:

select
  base58_decode(column_name)
  ...
from a_given_table

Both will work :)

soispoke (approved):

PR #1266, opened 2022-07-08
add wstETH to prices dataset

markusbkoch:
add wstETH to prices dataset


jeff-dude (approved):
@markusbkoch thank you for the add

PR #1276, opened 2022-07-09
Trigger spellbook deploys on spellbook updates in the abstraction repo

philippWassibauer:
this will trigger the workflow in the spellbook-deploy repo every time the spellbook repo is changed, therefore implementing a continuous deploy.


soispoke (commented):
LGTM

soispoke (approved):

PR #1278, opened 2022-07-11
T 13807 fix tests by only querying relevant time range

soispoke:
This PR implements changes aim to improve run time of spell tests. 
=> it sucesfully reduces runtime from 24 minutes to 10 minutes without affecting tests quality

*For Dune Engine V2*
I've checked that:


spellbook/models/opensea/ethereum/opensea_ethereum_schema.yml
```diff
@@ -153,11 +153,8 @@ models:
       - &royalty_fee_currency_symbol
         name: royalty_fee_currency_symbol
         description: "Symbol of the token in which fees are paid out"
-      - &unique_trade_id
-        name: unique_trade_id
+      - name: unique_trade_id
```
jeff-dude, line 156: any reason this one can't also follow the '&' pattern, instead of using it downstream in line 203 below?

jeff-dude (approved):
other than the one comment, lgtm

spellbook/models/opensea/ethereum/opensea_ethereum_schema.yml
soispoke: Thanks for pointing this out, the only reason here is I don't want to test uniqueness on this particular table, since  it's a view and will take forever. I'd rather test uniqueness in the opensea_ethereum.transactions table, that's why the &unique_trade_id was only implemented later on line 203 :)

soispoke (commented):
Thanks for taking the time to review this :)

PR #1279, opened 2022-07-11
nft transactions to nft events

soispoke:
After giving it some more thought and implementing the first version of it, I actually think nft.transactions shoul be changed to nft.events.

Events correspond better to the tables built on top of it (nft.mints/burns/trades), and transactions have a very precise meaning for EVM blockchains. We could actually do an nft.transactions abstraction that would only give info about transaction value/gas fees at the transaction level (not the log/transfer level) later on.

*For Dune Engine V2*
I've checked that:


jeff-dude (approved):
i don't have strong conviction either way on transactions vs. events, but the code changes look straightforward to me

soispoke (commented):
Thanks 👍 !

PR #1280, opened 2022-07-11
Refactor OVM Get Contracts

MSilb7:

Dune v1 - Refactored the Optimism contract inserts so it runs faster and requires less updating. I believe the cron from this was paused, can we re-run it?

Once this gets through, I can create an update script to overwrite some contract with missing created times that the prior version overwrote. An alternative is to delete and re-write this table.

*For Dune Engine V2*
I've checked that:


jeff-dude:
@MSilb7 i've rebuilt all objects as intended here & paused both cron jobs. how do you want me to proceed prior to resuming cron jobs? update statement as mentioned or drop all delete and restart (if so, which table(s))?

MSilb7:
> @MSilb7 i've rebuilt all objects as intended here & paused both cron jobs. how do you want me to proceed prior to resuming cron jobs? update statement as mentioned or drop all delete and restart (if so, which table(s))?

Hi awesome! Dropping works, I actually want to add one additional field if we're doing that. So I'll make that change, then get on the order of what to update.

jeff-dude:
> > @MSilb7 i've rebuilt all objects as intended here & paused both cron jobs. how do you want me to proceed prior to resuming cron jobs? update statement as mentioned or drop all delete and restart (if so, which table(s))?
> 
> Hi awesome! Dropping works, I actually want to add one additional field if we're doing that. So I'll make that change, then get on the order of what to update.

sounds good, just tag me here when ready

MSilb7:
> > > @MSilb7 i've rebuilt all objects as intended here & paused both cron jobs. how do you want me to proceed prior to resuming cron jobs? update statement as mentioned or drop all delete and restart (if so, which table(s))?
> > 
> > 
> > Hi awesome! Dropping works, I actually want to add one additional field if we're doing that. So I'll make that change, then get on the order of what to update.
> 
> sounds good, just tag me here when ready

Made an update to add in the creation tx hash as a field.

Order:
1. Drop `ovm2.get_contracts`
2. Re-Create `ovm2.get_contracts`
3. Run insert - should take ~2 mins
```SELECT ovm2.insert_get_contracts('01-01-2021'::timestamptz,NOW())
WHERE NOT EXISTS (
SELECT *
FROM ovm2.get_contracts
);
```
4. Re-enable crons in `insert_get_contracts.sql` and `insert_updated_contract_info.sql`

optimism2/ovm2/get_contracts/insert_get_contracts.sql
```diff
@@ -174,7 +174,7 @@ GROUP BY 1,2,3,4,5,6
                   fifth_level -- check for contract factories 4 layers down --as of 3/12, this ran through all contracts
               ) con
               LEFT JOIN ovm2."contract_creator_address_list" cc ON con.creator_address = cc.creator_address
-              LEFT JOIN ovm2."contract_creator_address_list" ccf ON con.creator_address = ccf.contract_creator_if_factory
+              LEFT JOIN ovm2."contract_creator_address_list" ccf ON con.creator_address = ccf.creator_address
```
MSilb7, line 177: Ah had a typo here, will fix. Shouldn't create any major impact for now, but I'll check. Can be fixed with an update if needed.

PR #1281, opened 2022-07-12
Update nft_ethereum_aggregators.sql

soispoke:

Adding aggregators to the nft_ethereum_aggregators table

*For Dune Engine V2*
I've checked that:

### DIFF
--- a/spellbook/models/nft/ethereum/nft_ethereum_schema.yml
+++ b/spellbook/models/nft/ethereum/nft_ethereum_schema.yml
@@ -1,0 +1,17 @@
+version: 2
+
+models:
+  - name: nft_ethereum_aggregators
+    meta:
+      blockchain: ethereum
+      sector: nft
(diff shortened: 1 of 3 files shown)


spellbook/models/nft/ethereum/nft_ethereum_schema.yml
```diff
@@ -0,0 +1,17 @@
+version: 2
```
jeff-dude, line 1: maybe a misunderstanding of how dbt is setup, but why wouldn't this fit into the other schema file?
https://github.com/duneanalytics/abstractions/blob/master/spellbook/models/nft/nft_schema.yml

soispoke (commented):
@jeff-dude For now, we have used blockchain specific schema files. This way, we can have an nft_aggregators.sql file in the future and define its schema in nft.yml

jeff-dude (approved):

PR #1284, opened 2022-07-13
[CoW Protocol] Known Solvers Part 2

bh2smith:
~~This is based on #1263 addressing the suggestion https://github.com/duneanalytics/abstractions/pull/1263#issuecomment-1182347900 to first get the CSV data into the project and then use the permanent link for the copy statement here.~~

Reverts #1263 (see discussion below).


ethereum/gnosis_protocol_v2/solver_names.sql
```diff
@@ -0,0 +1,10 @@
+CREATE TABLE IF NOT EXISTS gnosis_protocol_v2.solver_names
+(
+    address     bytea,
+    environment text,
+    name        text
+);
+
+\copy gnosis_protocol_v2.solver_names FROM 'https://raw.githubusercontent.com/duneanalytics/abstractions/a68ea6d2c3045d02c29c974556d0d230533a1263/ethereum/gnosis_protocol_v2/solvers.csv' with (format csv,header true, delimiter ',');
```
bh2smith, line 8: TODO - Once #1263 is merged, use a permanent link to the CSV file with the solver names.

jeff-dude: from what i'm seeing, it's possible a URL won't work. the copy command appears to expect the file to be stored locally. while i can use a file path on my local git, i'd prefer not to put that path into our master branch. if another teammate looks to run, the path would need edited each time

jeff-dude: ignore that -- thanks to the legend @antonio-mendes it appears we can read from the URL as intended

jeff-dude (approved):
this should be all set, take a minute to review, then feel free to remove draft status to merge

ethereum/gnosis_protocol_v2/solver_names.sql
bh2smith: Party time!

ethereum/gnosis_protocol_v2/view_solvers.sql
```diff
@@ -20,75 +20,15 @@ registered_solvers as (
...
 -- Combining the metadata with current activation status for final table
 select solver as address,
        case when environment is not null then environment else 'new' end as environment,
        case when name is not null then name else 'Uncatalogued' end      as name,
        active
 from registered_solvers
-         left outer join known_solver_metadata
-                         on solver = replace(address, '0x', '\x')::bytea;
+         left outer join gnosis_protocol_v2.solver_names
+                         on solver = replace(address::text, '0x', '\x')::bytea;
```
bh2smith, line 31: I had actually forgotten the original reason I had this. Actually the CSV itself would be more readable if it had 0x addresses. Perhaps I should revert that and make the named solver table with strings. Although I think the effort here is probably more than anyone wants to deal with.

jeff-dude: for updating solver list: yes, i believe an initial PR on just that file to update, merge, then refer in next PR may be best option. the alternative would require us to alter the URL reference each time. in our scenario, the URL can remain static, as we merge into master first.

feel free to proceed as you see fit on the second point about readability.

bh2smith: Not so sure I'm keen on having to make two PRs, just to update the solver names... maybe we should forget this whole idea. It was supposed to make things cleaner.

jeff-dude: yeah, the other option would be to include csv file in same PR, then a dune teammate would need to determine the branch used and alter the URL link prior to running the command. for example, this URL would be:
https://raw.githubusercontent.com/duneanalytics/abstractions/d16117c907fe67e57c0db1eddc3efe554aa74461/ethereum/gnosis_protocol_v2/solver_names.csv

'd16117c907fe67e57c0db1eddc3efe554aa74461' would be specific to this PR.

if you want to abandon the idea, feel free to close this, then we can revisit again if ideas surface

bh2smith: Yes I think I may abandon this. Sorry to waste your time here. Will close it tomorrow. Or turn it into a revert of the previous one.

bhndt: and I apologize in advance for the stupidity

ethereum/gnosis_protocol_v2/solver_names.csv
```diff
@@ -0,0 +1,56 @@
+address,environment,name
```
bhndt, line 1: 0xCa0b051804f57412600C2597d62E688d813694Ba, Ethereum, holytrade

bh2smith: What is this all about?

bh2smith:
Good day folks. I have turned this PR into a revert of #1263 since it no longer seems like a feasible option.

jeff-dude (approved):

PR #1288, opened 2022-07-14
Fix erc721 balances

0xBoxer:
Fixing erc721 balances, the prior version had problems and the test was was incorrectly configured.

Proof that current version is broken with correctly defined test:
https://dune.com/queries/1039163

*For Dune Engine V2*
I've checked that:

balances latest in one query
https://dune.com/queries/1038527

balances daily in one query
https://dune.com/queries/1035454

testing the test
https://dune.com/queries/1039115


spellbook/models/balances/ethereum/erc721/balances_ethereum_erc721_day.sql
```diff
@@ -13,22 +13,18 @@ with
...
     b.tokenId,
-    nft_tokens.name as collection
-FROM daily_balances b
-INNER JOIN days d ON b.day <= d.day AND d.day < b.next_day
-LEFT JOIN {{ ref('tokens_ethereum_nft') }} nft_tokens ON nft_tokens.contract_address = b.token_address
+    b.num_tokens,
+    nft_tokens.name as collection,
+    nft_tokens.category
+FROM days d
+INNER JOIN {{ ref('transfers_ethereum_erc721_rolling') }} b ON (b.evt_block_time <= d.day AND d.day < b.next_evt)
```
jeff-dude, line 26: minor feedback: for readability, even though with inner join the results are the same, can we switch the FROM to be on {{ ref('transfers_ethereum_erc721_rolling') }} and the JOIN on days? as an outsider looking in for first time, seems that would flow better

spellbook/models/balances/ethereum/erc721/balances_ethereum_erc721_hour.sql
```diff
@@ -13,22 +13,18 @@ with
...
     b.tokenId,
-    nft_tokens.name as collection
-FROM daily_balances b
-INNER JOIN hours d ON b.hour <= d.hour AND d.hour < b.next_hour
-LEFT JOIN {{ ref('tokens_ethereum_nft') }} nft_tokens ON nft_tokens.contract_address = b.token_address
+    b.num_tokens,
+    nft_tokens.name as collection,
+    nft_tokens.category
+FROM hours d
+INNER JOIN {{ ref('transfers_ethereum_erc721_rolling') }} b ON (b.evt_block_time <= d.hour AND d.hour < b.next_evt)
```
jeff-dude, line 26: same as other comment

minor feedback: for readability, even though with inner join the results are the same, can we switch the FROM to be on {{ ref('transfers_ethereum_erc721_rolling') }} and the JOIN on hours? as an outsider looking in for first time, seems that would flow better

spellbook/models/balances/ethereum/erc721/balances_ethereum_erc721_latest.sql
```diff
@@ -2,11 +2,30 @@
...
+    b.tokenId,
+    b.num_tokens,
+    b.updated_at,
     nft_tokens.name as collection,
-    updated_at
-FROM {{ ref('transfers_ethereum_erc721_rolling_hour') }}
-LEFT JOIN {{ ref('tokens_ethereum_nft') }} nft_tokens ON nft_tokens.contract_address = token_address
+    nft_tokens.category
+FROM minutes d
+INNER JOIN {{ ref('transfers_ethereum_erc721_rolling') }} b ON (b.evt_block_time <= d.minute AND d.minute < b.next_evt)
```
jeff-dude, line 26: same as other comment

minor feedback: for readability, even though with inner join the results are the same, can we switch the FROM to be on {{ ref('transfers_ethereum_erc721_rolling') }} and the JOIN on minutes? as an outsider looking in for first time, seems that would flow better

jeff-dude (changes requested):
minor readability requests, open to feedback of leaving as-is vs. changing

jeff-dude:
overall, these changes make sense to me. i am a bit confused though, when i search the dune engine v2 explorer for 'transfers', i don't see the erc721 tables? when i search 'balances' though, i do see the erc721 tables. curious why that may be. otherwise, i can approve, but maybe we look for @soispoke and @dot2dotseurat feedback as well before merge

jeff-dude (approved):

spellbook/models/balances/ethereum/erc721/balances_ethereum_erc721_day.sql
0xBoxer: we are projecting state on days here, makes more sense to me to do it this way around 🤷 

I do not object to changing this around, your call.

jeff-dude: i certainly don't feel strongly about a change, so i'm good with as-is 👍

0xBoxer:
https://dune.com/queries/1071709

Inner join with nft.tokens leaves 108 contracts which are affected.

These ones are a headscratcher, currently still trying to find out what contributes to them not properly being tracked.
I suspect it's a missing erc721 transfer event log despite moving tokens, but still doing investigations.

https://dune.com/queries/1079330

all erc721 tokens leaves me with 60k affected contracts

This is a lot of scam tokens which are improperly implemented erc721 tokens, often hidden behind a non verified proxy. Don't worry about these too much.


Now just manually excluding these kinda nullifies the test, I would like to find a different way of filtering these out, but that proves to be complciated.

0xBoxer:
there was some miscommunication here, the table I reference in the queries above is not the updated version.

Now I am totally lost on which tests ran against which versions, could you clarify that here @soispoke?

I suspect my version will still have some troubles as there is some mishapps in some smart contracts, but no nearly as much as the queries above would indicate.

Would be awesome if you can drop me a table with the updated scripts in the prod db somewhere so I can debug without 10 minute query times.

0xBoxer:
One thing I have found for example is this contract: [Tripsters](https://etherscan.io/address/0xc7457f480416e145093edb355f626faa0fc7f90e) which literally had the issue of minting token_id's twice 💀 
Look [here](https://etherscan.io/token/0xc7457f480416e145093edb355f626faa0fc7f90e?a=2116)

I suspect there is a few others which also have this error and potentially some other weird happenings.

dot2dotseurat:
> there was some miscommunication here, the table I reference in the queries above is not the updated version.
> 
> Now I am totally lost on which tests ran against which versions, could you clarify that here @soispoke?
> 
> I suspect my version will still have some troubles as there is some mishapps in some smart contracts, but no nearly as much as the queries above would indicate.
> 
> Would be awesome if you can drop me a table with the updated scripts in the prod db somewhere so I can debug without 10 minute query times.

it's really late where @soispoke is so I'll try to jump in here. catching up on what exactly is happening though. I'll dm you.

0xBoxer:
I think there is a whole block full of decoded erc721 transactions missing 😅 

https://dune.com/queries/1082463

context and everything is in the query

dot2dotseurat:
> I think there is a whole block full of decoded erc721 transactions missing 😅
> 
> https://dune.com/queries/1082463
> 
> context and everything is in the query

alright, we can't fix that from this spell so let's discuss in a height task.

PR #1296, opened 2022-07-18
[ERC20] Update Tokens

bh2smith:
Adding 21 tokens to erc20 table on both legacy and v2 platforms.


jeff-dude (approved):
thanks @bh2smith v1 engine is updated, v2 looks good too. @dot2dotseurat any feedback or good to merge?

PR #1297, opened 2022-07-18
fix synthetix.trades

0xBoxer:
We changed the naming convention of the SNX tables apparently, this ensures that the dex.trades insert will work correctly again

cc: @antonio-mendes


0xBoxer:
you might have to backfill some data @jeff-dude, I didn't look to closely at the logic here, but from my understanding it should be fine due to the way the cron job is structured

PR #1302, opened 2022-07-19
delete gamedao

0xBoxer:
token only exists on bsc, is in here by accident

### DIFF
--- a/prices/ethereum/coinpaprika.yaml
+++ b/prices/ethereum/coinpaprika.yaml
@@ -1765,11 +1765,6 @@
   symbol: XYZ
   address: 0x618679df9efcd19694bb1daa8d00718eacfa2883
   decimals: 18
-- name: dao_gamedao
-  id: dao-gamedao
-  symbol: DAO
-  address: 0x0f51bb10119727a7e5ea3538074fb341f56b09ad
-  decimals: 18
 - name: tcr_tracer_dao
   id: tcr-tracer-dao
   symbol: TCR


bh2smith (approved):

jeff-dude (approved):

bh2smith:
Will this token's price history also be removed from the `prices.usd` table?

bh2smith:
Ah that was just a raw query on the prices.usd table to demonstrate the whacky prices. If anything, any query that depends on this price feed is entirely inaccurate and it should just be removed completely.

bh2smith:
I know that our trades table is using prices from here though, so it might require dropping some values from there as well. Or repopulating the table from scratch.

jeff-dude:
> I know that our trades table is using prices from here though, so it might require dropping some values from there as well. Or repopulating the table from scratch. Will have to check with the team about how to proceed here.
> 
> This also mean dex.trades likely contains valuations based on this.... uh oh.

i don't think we'd need to retroactively delete rows. when it comes to dex trades (or even protocol-specific abstractions prior to loading dex trades), the prices table is a left join, and if matched, populates usd amount in most cases. i think we could just update the table(s) to null for usd price where applicable, as that is what would happen if the left join doesn't find the contract address in prices like it won't moving forward.

if this specific token isn't on coinpaprika, but is being traded on dex's, would we want to look to add to coinpaprika?

0xBoxer:
but wait this token only exists on bsc and we only join on contract_address, this should have not been picked up by dex.trades or similar?

Just dropping it from the prices.usd table should be good.

There is the very fringe case where people used the wrong symbol to join this pricefeed into their queries/views, but that seems rather outlandish.

jeff-dude:
i should be able to write a query to validate exactly what situation we're in

bh2smith:
It's definitely appearing by address on mainnet. I will share an example with us tomorrow.

jeff-dude:
it's on etherscan:
https://etherscan.io/token/0x0f51bb10119727a7e5ea3538074fb341f56b09ad

also api returns results, but is_active = false, which is an issue:
https://api.coinpaprika.com/v1/coins/dao-gamedao

we'd need to reactive it with them, maybe here:
https://coinpaprika.com/add/

0xBoxer:
my bad, the game-dao token only exists on bsc, this token indeed exists on mainnet.

So this token now has the price of game-dao in dex.trades and maybe other tables, that indeed is a problem.

jeff-dude:
i'm going to outline where my head is at, see how we'd like to proceed (i know you guys mentioned some of this, just summarizing to be sure):

1. an incorrect 'id' value was applied to the coinpaprika file in the ethereum directory (i.e. this PR to remove it)
2. the coinpaprika API still found results, but found this: https://bscscan.com/token/0x21a807b748423e5c37a914fb4fb40850037dafdd -- which pulled a different token price from bsc rather than ethereum, hence the odd looking prices
3. gnosis v2 trades table loaded data which joined to prices on this address (address in this PR was right, just not the id to coinpaprika), now usd values in that table likely off
4. dex trades reads in data from table in step 3, so values there would be off as well

how should we proceed?
will @bh2smith look to add the right token (https://etherscan.io/token/0x0f51bb10119727a7e5ea3538074fb341f56b09ad) to coinpaprika (i did quick search, didn't see it, but i may not be looking in the right spot)?

i could look to delete historical prices data for this incorrect entry, then history should refill if added later. we could then write queries to update the usd value as needed on affected rows.

i only see ~200 tx's which hit this address:
```
    select
      *
    from
      gnosis_protocol_v2.trades
    where
      (
        buy_token_address = '\x0f51bb10119727a7e5ea3538074fb341f56b09ad'
        or sell_token_address = '\x0f51bb10119727a7e5ea3538074fb341f56b09ad'
      )
```

bh2smith:
Maybe its not such a huge problem to just leave it. We could always drop recreate the entire trades table using the insert_trades function, but then it would also require recreating the `dex.trades` table as well. It is also unclear if other exchange projects in dex.trades has been using this value as well (which I suspect they are). Sounds like a lot of work on your side. I could help assist if needed, but maybe its just not worth it.

PR #1315, opened 2022-07-25
Add hex2numeric function to Postgres

antonio-mendes:
Adding the `hex2numeric` function to Caladan. This function was actually already deployed on the Ethereum db and being used but was not documented here.


jeff-dude (approved):
@antonio-mendes approving but wanted to validate: this version of the function lives in the 'dune_user_generated' schema -- is that expected? the function with same name in 'public' schema looks a bit different

antonio-mendes:
Yeah @jeff-dude I saw it was deployed from `user_generated` but the function is equivalent to what I'm sharing here.

I think it's ok we get this on the repo just to have the documentation that it exists but we don't actually need to redeploy the function in `public` since everyone can access it anyway. What do you think?

If you'd rather have things more tidy, feel free to drop the user generated one and deploy this instead 👍 . Both are fine with me

PR #1318, opened 2022-07-25
Add cex addresses

hildobby:
Added a list of known CEX addresses I could identify

*For Dune Engine V2*
I've checked that:


jeff-dude:
proposed some updates via a commit. curious to hear input @dot2dotseurat @hildobby if you think this is good, or we can easily revert as needed 👍

hildobby:
> proposed some updates via a commit. curious to hear input @dot2dotseurat @hildobby if you think this is good, or we can easily revert as needed 👍

I thought about this and while yes I did check solely on ethereum, those addresses are often use across chains so I kept it without ethereum mentioned but I'm not sure what the best route is here 🤔 

Also, I see you re;oved the schema='cex' but I discussed this with @dot2dotseurat and we said that keeping the schema='cex' would be best so we have cex.addresses (but the list remains in the addresses folder). This is because in the future I plan on (and im sure others also will) adding stuff like bridges.adddresses for example and it is likely that other abstractions will appear under the bridges schema.

jeff-dude:
> > proposed some updates via a commit. curious to hear input @dot2dotseurat @hildobby if you think this is good, or we can easily revert as needed 👍
> 
> I thought about this and while yes I did check solely on ethereum, those addresses are often use across chains so I kept it without ethereum mentioned but I'm not sure what the best route is here 🤔
> 
> Also, I see you re;oved the schema='cex' but I discussed this with @dot2dotseurat and we said that keeping the schema='cex' would be best so we have cex.addresses (but the list remains in the addresses folder). This is because in the future I plan on (and im sure others also will) adding stuff like bridges.adddresses for example and it is likely that other abstractions will appear under the bridges schema.

thank you for the context. no problem, happy to help revert. i figured the best way to ask questions was to look at it this way.

dot2dotseurat:
@jeff-dude @hildobby just triggered the tests and things look good. Let's revert the schema back to `cex` so the table name can be `cex.addresses` and then this is good to go.

jeff-dude:
pushed a commit to revert back to original code and naming standard, feel free to merge if looks good

jeff-dude (approved):

PR #1319, opened 2022-07-26
[Aztec Connect] ETH price feed and balance bug fixes

jackiep00:
1. ETH price duplication error from some ETH being transacted via dex data
2. ETH transfer duplication error due to some bad joins

I've added in a truncate and re-insert statement for the insert function redefinition script - should truncate everything and re-insert the tables from there

*For Dune Engine V2*
I've checked that:


jeff-dude (approved):
@jackiep00 function updated & history reloaded, automation back on schedule, we can monitor to see how it goes

PR #1321, opened 2022-07-26
Add prices.usd_latest

hildobby:
prices.usd is very often used to get the latest price of an/multiple assets, this table should help easily get just that.

*For Dune Engine V2*
I've checked that:


dot2dotseurat:
I ran these tests locally when I made this commit and they are passing.

dot2dotseurat (approved):
lgtm

PR #1328, opened 2022-07-27
[CoW Protocol] Update Gnosis solvers names

harisang:
This PR renames some of the solvers owned by Gnosis that are currently participating in the CoW Protocol solver competition, since there was some ambiguity with their original names. In particular, they were originally named after the API they were using, while now we have introduced the prefix `Gnosis_` to explicitly show that these are owned by Gnosis.

E.g., the `1inch` solver, which is a solver built by Gnosis and is using the 1inch API, is now named `Gnosis_1inch`


bh2smith (approved):
Hey! This looks great, thanks for making this! Could you also apply the same changes to the view for xdai solvers aswell?

https://github.com/duneanalytics/abstractions/blob/master/xdai/gnosis_protocol_v2/view_solvers.sql

harisang:
> Hey! This looks great, thanks for making this! Could you also apply the same changes to the view for xdai solvers aswell?
> 
> https://github.com/duneanalytics/abstractions/blob/master/xdai/gnosis_protocol_v2/view_solvers.sql

Right, I forgot about it, sorry! Just did it.

jeff-dude (approved):
thank you both, updated

PR #1334, opened 2022-07-29
urgent patch : avoiding division_by_zero

sohwak:

Some of the cases transacted with seaport contracts without any NFT. 

https://etherscan.io/tx/0xda5ea90ec1cad79968ef47c3f38e5c2b785e76a5e5bd05b538d437d7a364e059
https://etherscan.io/tx/0xb5892d5bdec28c091efa683ae4d178251db587a2295d49e7ae77c0adc235c43d

This patch is to avoid those kinds of cases.

*For Dune Engine V2*
I've checked that:


sohwak:
I will re-submit this PR

PR #1333, opened 2022-07-29
Add sETH2 coinpaprika.yaml

rzeydelis:
Add StakeWise Staked Eth token.

Brief comments on the purpose of your changes:

*For Dune Engine V2*
I've checked that:

* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory) **-- not writing a sql query**
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables *-- not writing a sql query**
* [ ] if adding a new model, I added a test **-- no new model**
* [ ] the filename is unique and ends with .sql *-- not writing a sql query**
* [ ] each sql file is a select statement and has only one view, table or function defined   *-- not writing a sql query**
* [ ] column names are `lowercase_snake_cased` *-- not writing a sql query**

When you are ready for a review, tag duneanalytics/data-experience. We will re-open your forked pull request as an internal pull request. Then your spells will run in dbt and the logs will be avaiable in Github Actions DBT Slim CI. This job will only run the models and tests changed by your PR compared to the production project.

PR #1338, opened 2022-07-31
Update coinpaprika.yaml - add SWISE token

rzeydelis:

*For Dune Engine V2*
I've checked that:

* [n/a ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [n/a ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [n/a] if adding a new model, I added a test
* [n/a ] the filename is unique and ends with .sql
* [n/a ] each sql file is a select statement and has only one view, table or function defined  
* [n/a ] column names are `lowercase_snake_cased`


masquot:
Integrated #1333

PR #1342, opened 2022-08-01
[Erc20.tokens] add 13 missing tokens

bh2smith:
Tokens taken from This query: https://dune.com/queries/236085


jeff-dude (approved):
added another recent token from #1339

PR #1349, opened 2022-08-04
OP2 - update insert lookbacks

MSilb7:

Extend some insert time lookbacks since the db fell behind by > 7 days (if the db gets to within 7 days before this PR, we should look to try to backfill the ~11-7 day period that may have been missed.

*For Dune Engine V2*
I've checked that:


jeff-dude (approved):
updated 👍 let me know if there are gaps in the data at all after a few runs

MSilb7:
> updated 👍 let me know if there are gaps in the data at all after a few runs

We got this through just in time haha so hopefully we'll be good to go.

PR #1352, opened 2022-08-04
Safe

sche:

Here I create the `safe_ethereum.safes` and `safe_ethereum.eth_transfers` tables that will contain the full list of created safes and eth transfers for that safes.

*For Dune Engine V2*
I've checked that:

@duneanalytics/data-experience

### DIFF
--- a/spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
+++ b/spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
@@ -19,6 +19,7 @@ select
     array_join(et.trace_address, ',') as trace_address
 from {{ source('ethereum', 'traces') }} et
 join {{ ref('safe_ethereum_safes') }} s on et.from = s.address
+    and et.from != et.to -- exclude calls to self to guarantee unique key property
     and success = true
     and (lower(call_type) not in ('delegatecall', 'callcode', 'staticcall') or call_type is null)
     and try_cast(date_trunc('day', et.block_time) as date) = s.block_date
@@ -39,6 +40,7 @@ select
     array_join(et.trace_address, ',') as trace_address
 from {{ source('ethereum', 'traces') }} et
 join {{ ref('safe_ethereum_safes') }} s on et.to = s.address
+    and et.from != et.to -- exclude calls to self to guarantee unique key property
     and success = true
     and (lower(call_type) not in ('delegatecall', 'callcode', 'staticcall') or call_type is null)
     and try_cast(date_trunc('day', et.block_time) as date) = s.block_date
--- a/spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
+++ b/spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
@@ -21,8 +21,7 @@ from {{ source('ethereum', 'traces') }} et
 join {{ ref('safe_ethereum_safes') }} s on et.from = s.address
     and et.from != et.to -- exclude calls to self to guarantee unique key property
     and success = true
-    and (lower(call_type) not in ('delegatecall', 'callcode', 'staticcall') or call_type is null)
-    and try_cast(date_trunc('day', et.block_time) as date) = s.block_date
+    and (lower(call_type) not in ('delegatecall', 'callcode', 'staticcall') or call_type is null)    
 where et.block_time > '2018-11-24' -- for initial query optimisation
 {% if is_incremental() %}
 -- to prevent potential counterfactual safe deployment issues we take a bigger interval
@@ -43,7 +42,6 @@ join {{ ref('safe_ethereum_safes') }} s on et.to = s.address
     and et.from != et.to -- exclude calls to self to guarantee unique key property
     and success = true
     and (lower(call_type) not in ('delegatecall', 'callcode', 'staticcall') or call_type is null)
-    and try_cast(date_trunc('day', et.block_time) as date) = s.block_date
 where et.block_time > '2018-11-24' -- for initial query optimisation
 {% if is_incremental() %}
 -- to prevent potential counterfactual safe deployment issues we take a bigger interval
--- a/spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
+++ b/spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
@@ -2,7 +2,8 @@
     config(
         materialized='incremental',
         alias='eth_transfers',
-        unique_key = 'tx_hash',
+        partition_by = ['block_date'],
+        unique_key = ['block_date', 'tx_hash'],
         on_schema_change='fail',
         file_format ='delta',
         incremental_strategy='merge'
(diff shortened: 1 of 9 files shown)


sche:
@tschubotz fyi

jeff-dude:
thank you for the PR -- we'll look to get this built asap 👍

spellbook/models/safe/ethereum/safe_ethereum_safes.sql
```diff
@@ -0,0 +1,54 @@
...
+    and et.block_time > (select max(creation_time) from {{ this }}) - interval '1 day'
+    {% endif %}
+        
+union all
+    
+select contract_address as address, 
+    '1.3.0' as creation_version, 
+    evt_block_time as creation_time, 
+    evt_tx_hash as tx_hash
+from gnosis_safe_ethereum.GnosisSafev1_3_0_evt_SafeSetup
```
jeff-dude, line 51: the goal is to avoid hardcoding table names. can you build a new sources YAML file in this directory for 'gnosis_safe' specific decoded tables? you can mimic an existing sources file:
https://github.com/duneanalytics/abstractions/blob/master/spellbook/models/transfers/ethereum/transfers_ethereum_sources.yml

you can leave out column descriptions if they are obvious based on column name. that's more a nice-to-have and helps with auto-document generation for other users.

with that said, would it make more sense to rename this whole directory to 'gnosis_safe' rather than 'safe'? i ask due to the schema in the db being 'gnosis_safe_ethereum' (side note: can have 'ethereum' directory nested in too, allows for more blockchains later on).

spellbook/models/safe/ethereum/safe_ethereum_schema.yml
```diff
@@ -0,0 +1,43 @@
...
+      project: safe
+      contibutors: scherbovich, tschubotz
+    freshness:
+      warn_after: { count: 12, period: hour }
+      error_after: { count: 24, period: hour }
+    config:
+      tags: ['safe', 'ethereum']
+    description: "Safe addresses"
+    columns:
+      - name: address
```
jeff-dude, line 16: minor feedback, but good to know for now & future spell builds:
you can enable "reuse" of columns downstream in the same file, so if two models have the same column, you can reuse it without defining twice. an example in this file is 'address' column in both. could look like:
      - &address
        name: address
        description:  "Safe contract address"

spellbook/models/safe/ethereum/safe_ethereum_schema.yml
```diff
@@ -0,0 +1,43 @@
...
+      project: safe
+      contibutors: scherbovich, tschubotz
+    freshness:
+      warn_after: { count: 12, period: hour }
+      error_after: { count: 24, period: hour }
+    config:
+      tags: ['safe', 'transfers', 'ethereum']
+    description: "Eth transfers for safes"
+    columns:
+      - name: address
```
jeff-dude, line 37: then simply:
- *address

which picks up the prior record of '&' prefix above and reuses across models. this is more useful when there's lots of overlap, but good to have for consistency

spellbook/models/safe/ethereum/safe_ethereum_schema.yml
jeff-dude: comment continued on line 37

spellbook/models/safe/ethereum/safe_ethereum_safes.sql
jeff-dude: once built, this hardcoded table reference can use the '{{ source ( ) }}' syntax, similar to {{ref( )}} i've seen used in this PR

jeff-dude (changes requested):
@sche thank you for the PR. overall, great job on your first spell! you've covered most of what needs done. i have left a few relatively minor pieces of feedback. happy to help if needed.

spellbook/models/safe/ethereum/safe_ethereum_safes.sql
sche: Hi @jeff-dude!

Thank you for the feedback. I am back from a vacation and will adjust the PR today.
We rebranded, and now the project name is `Safe`. It would be good to deprecate all the `gnosis_safe_` names with time and change them to `safe_`. I don't know how to make it smooth, so I will use already existing decoded tables for now. Let's discuss in Discord how we could migrate properly!

>  can have 'ethereum' directory nested in too

I already have an 'ethereum' directory nested in the 'safe' directory, or did you mean some other place?

sche: I made adjustments @jeff-dude

jeff-dude: thank you!

spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
```diff
@@ -0,0 +1,40 @@
...
+    et.block_time,
+    -value as amount_raw
+from {{ source('ethereum', 'traces') }} et
+join {{ ref('safe_ethereum_safes') }} s on et.from = s.address
+    and et.block_time > '2018-11-24' -- for initial query optimisation
+    and success = true
+    and (lower(call_type) not in ('delegatecall', 'callcode', 'staticcall') or call_type is null)
+    {% if is_incremental() %}
+    -- to prevent potential counterfactual safe deployment issues we take a bigger interval
+    and et.block_time > (select max(block_time) from {{ this }}) - interval '10 days'
```
jeff-dude, line 23: just an fyi, i pushed a commit to update the filter to block_time as it's the valid field (same applies to line 39)

spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
```diff
@@ -0,0 +1,40 @@
+{{ 
+    config(
+        materialized='incremental',
+        alias='eth_transfers',
+        unique_key = 'address',
```
jeff-dude, line 5: this table is using merge, however, it appears address is not unique. i ran this the first time, loaded as expected, then the second time it failed on unique constraints. you may need to either add multiple fields and/or add a column which concats a few for unique value

jeff-dude (changes requested):

spellbook/tests/safe/eth_transfers_check_sum.sql
```diff
@@ -0,0 +1,11 @@
...
+
+with unit_tests as (
+    select round(sum(amount_raw) / 1e18, 0) as total
+    from {{ ref('safe_ethereum_eth_transfers') }}
+    where block_time between '2022-01-01' and '2022-01-03'
+)
+
+select case when total = 2996 then true else false end as test
+from unit_tests
+where test = false
```
jeff-dude, line 11: you can't refer back to 'test' in the same query here, you'll either need to reuse the case logic fully in the where clause or move this column up into the CTE above, then refer here

spellbook/tests/safe/safes_check_count.sql
```diff
@@ -0,0 +1,11 @@
...
+
+with unit_tests as (
+    select count(*) as total
+    from {{ ref('safe_ethereum_safes') }}
+    where creation_time between '2022-01-01' and '2022-08-03'
+)
+
+select case when total = 43704 then true else false end as test
+from unit_tests
+where test = false
```
jeff-dude, line 11: same idea here

jeff-dude (changes requested):

spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
sche: thanks!

spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
sche: True, my bad. I added `tx_hash` field that should be always unique.

spellbook/tests/safe/eth_transfers_check_sum.sql
sche: Thanks! Changed

sche:
@jeff-dude I made all requested changes. Could you please check it? 🙏

jeff-dude (approved):
thanks again for all the help getting this into a good spot 👍 this will be merged into the repo & will be bundled as part of a deployment likely early next week

jeff-dude:
pushed one last commit which adds the safe tables to our table properties file & optimize script

PR #1353, opened 2022-08-04
introduce dex trades to spellbook using uniswap v1/v2 contract data

jeff-dude:
build out the infrastructure for dex trades using uniswap v1/v2 data as a starting point. ultimately, five database objects will be built:

1. uniswap_v1_ethereum_trades
2. uniswap_v2_ethereum_trades
3. uniswap_ethereum_trades (future iterations will add v3)
4. uniswap_trades (to allow for cross-chain integration, if applicable)
5. dex_trades

this is considered a first iteration and looking for plenty of feedback from teammates, users, etc.

note that the column names have been reworked from v1 engine to make it more readable. also open to feedback on more columns, less columns & any other ideas to improve dex trades. with this laying the foundation on the v2 engine, we'd like to get a good starting point to iterate upon with other protocol data.

*For Dune Engine V2*
I've checked that:


jeff-dude:
@duneanalytics/team-data-experience fyi for first draft on dex trades 👍

spellbook/macros/optimize_tables.sql
```diff
@@ -56,10 +56,22 @@ OPTIMIZE nft.burns;
...
+{% set uniswap_ethereum_trades %}
+OPTIMIZE uniswap_ethereum.trades;
+{% endset %}
+
 {% set uniswap_trades %}
 OPTIMIZE uniswap.trades;
 {% endset %}
 
+{% set dex_trades %}
+OPTIMIZE dex.trades;
```
dot2dotseurat, line 72: are we sure this should be a table? should it be a view of unioned tables instead?

spellbook/models/dex/dex_schema.yml
```diff
@@ -0,0 +1,82 @@
...
+      sector: dex
+      contributors: jeff-dude
+    config:
+      tags: ['ethereum','uniswap','trades', 'dex','cross-chain']
+    description: >
+        DEX trades on all chains across all contracts and versions
+    columns:
+      - &blockchain
+        name: blockchain
+        description: "Blockchain which the DEX is deployed"
```
dot2dotseurat, line 16: Setting a very high column description standard :) 

I do feel like maybe this is a bit obvious and although its a significant improvement on
description: "Blockchain" 
maybe we don't need a description at all. 

It may make writing schemas less intimidating if we reserve descriptions for ambiguous columns. 

But that does mean we have to guess which columns aren't clear to users.

spellbook/models/dex/dex_schema.yml
```diff
@@ -0,0 +1,82 @@
...
+        name: token_bought_amount
+        description: "Value of the token bought at time of execution in the original currency"
+      - &token_sold_amount
+        name: token_sold_amount
+        description: "Value of the token sold at time of execution in the original currency"
+      - &token_bought_amount_raw
+        name: token_bought_amount_raw
+        description: "Raw value of the token bought at time of execution in the original currency"
+      - &token_sold_amount_raw
+        name: token_sold_amount_raw
```
dot2dotseurat, line 45: this description I think is quite meaningful fwiw. this clarifies the field significantly.

spellbook/models/uniswap/ethereum/uniswap_v1_ethereum_trades.sql
```diff
@@ -0,0 +1,105 @@
...
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = 'unique_trade_id'
+    )
+}}
+WITH dexs AS
+(
+        -- Uniswap v1 TokenPurchase
+    SELECT
+        t.evt_block_time AS block_time
```
dot2dotseurat, line 14: Lets add a block_date field and then add that as a partition key in the config

spellbook/macros/optimize_tables.sql
jeff-dude: i'm following the standard set by nft trades, which is also an incremental table and included in the optimize statements. maybe an assumption that it was discussed and agreed upon. we can explore unions and default view, would be curious to see performance difference. either way, i assume we should match between dex and nft

spellbook/models/dex/dex_schema.yml
jeff-dude: agree on all points here. i wanted to put it out there for users to know it's an option. happy to hear from potentially @dsalv @aalan3 or @0xBoxer on this. 

keep all column descriptions, then clean up wording on each where applicable?
remove descriptions from the obvious columns (help determine what's obvious)?

spellbook/models/dex/dex_schema.yml
jeff-dude: could be a good field to leave a description on 👍

spellbook/models/uniswap/ethereum/uniswap_v1_ethereum_trades.sql
jeff-dude: definitely interested in adding and seeing this in action. will push a commit and resolve this

spellbook/models/dex/dex_trades.sql
```diff
@@ -0,0 +1,18 @@
+{{ config(
+        alias ='trades',
+        materialized ='incremental',
+        file_format ='delta',
+        incremental_strategy='merge',
+        unique_key='unique_trade_id'
+        )
+}}
+SELECT *
```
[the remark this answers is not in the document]
jeff-dude: i left the select * on outer query, but replaced inner query with specific columns in all levels of the process to load dex trades

spellbook/macros/optimize_tables.sql
jeff-dude: set back to views for everything outside of the first layer reading from decoded tables, which are left incremental

dot2dotseurat (approved):

PR #1358, opened 2022-08-07
opensea-solana transfers added

sohwak:

I think I will make it to replace the current `opensea_solana.events`, but I am going to make it as a temporary version, not a fixed version so I just name this as `opensea_solana.transfers`.

Because I'm not sure if it's perfect. and in order to test and confirm, it can't execute with single query because Solana data is too heavy to run with one query.

I'm going to make it `opensea_solana.events` after testing and modifying if needed.

The point is, I want to keep the table in operation until the table in development is complete. If you let me know if there's any other way, I'll apply like that.

*For Dune Engine V2*
I've checked that:

### DIFF
--- a/polygon/opensea/backfill_trades.sql
+++ b/polygon/opensea/backfill_trades.sql
@@ -1,0 +1,36 @@
+CREATE OR REPLACE FUNCTION opensea.backfill_trades() RETURNS boolean
+LANGUAGE plpgsql AS $function$
+BEGIN
+
+-- backfill usd_price, fee_usd_price, royalty_usd_amount
+update opensea.trades
+   set usd_amount = new.new_usd_amount
+      ,fee_usd_amount = new.new_fee_usd_amount
+      ,royalty_usd_amount = new.new_royalty_usd_amount
+  from (select a.tx_hash
+              ,a.trade_id
+              ,a.original_amount_raw / 10^p.decimals * p.price as new_usd_amount
+              ,a.fee_amount_raw / 10^p.decimals * p.price as new_fee_usd_amount
+              ,a.royalty_amount_raw / 10^p.decimals * p.price as new_royalty_usd_amount
+          from opensea.trades a
+               inner join prices.usd p on p.contract_address = a.currency_contract
+                            		   and p.minute = date_trunc('minute', a.block_time)
+         where a.usd_amount is null
+        ) as new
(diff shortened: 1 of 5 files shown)


sohwak:
also added opensea trades abstraction on Polygon (Dune v1)

sohwak:
example: 
- opensea-solana query : https://dune.com/queries/1141087
- opensea-polygon query : https://dune.com/queries/1146806

sohwak:
forgot to mention one thing: for Solana query, I execute it for just "1 day" duration, and it took around 2 min to get result.

jeff-dude:
> also added opensea trades abstraction on Polygon (Dune v1)

i'd suggest pulling the polygon portion out of this PR and submitting a separate PR with just that code, i'll be able to merge it in quicker, then we can focus on spellbook items here

jeff-dude:
in regards to your comment on putting in a 'dummy' name for the object until it's built and tested, i think we can explore that as a last resort, but we should try to test via the dune web app query as much as possible and feel confident in full historical. for instance, you have the query set to 1 day, maybe you can expand it to larger timeframe (a week, a month) and test those results and get to level of comfort you are looking for.

worst case scenario, if tested a full month (and maybe even month timeframes multiple times for different months) and still find bugs, we can always rebuild later.

jeff-dude:
> > also added opensea trades abstraction on Polygon (Dune v1)
> 
> i'd suggest pulling the polygon portion out of this PR and submitting a separate PR with just that code, i'll be able to merge it in quicker, then we can focus on spellbook items here

fyi, i'm running the polygon script on v1 behind the scenes, but i still suggest to pull the code out of this branch/PR and put in it's own branch/PR, that way i can close it out. i pushed one change to this branch, so be sure to refresh locally, then create new

sohwak:
Thank you for letting me know.

I will separate PR for Dune v1 and Spellbook. I was going to do that, but I was not used to it and I thought it would be no problem once.

As for the test, I will try to test it if possible. but for solana, the query does not run in more than 5 days. I think it's probably processing too much data. I'm doing my best for the query when I wrote but I'm worried if I get it wrong. In any case, the current `opensea_solana.events` is the table already in use, so I am afraid that it will affect other dashboards when it has wrong data. I will happy if you give me more note about this part.

I will close this PR and open separated PRs

PR #1363, opened 2022-08-09
nft trades seaport contract -- switch tx_hash and block_number to load columns correctly

jeff-dude:
bug fix to rearrange columns to ensure loaded as expected.

current state:
tx_hash values live in block_number field
block_number values live in tx_hash field

future state post-fix:
values in correct columns

after code merge, this will require a new dbt deployment, deletion of historical data, reload full history


jeff-dude:

aalan3 (approved):

jeff-dude:
@sohwak fyi -- there's a bug in v2 engine nft trades for seaport data, this will be merged and data rebuilt soon

sohwak:
@jeff-dude 
Can you give me a moment to check? There was a miscalculation in the price part, so I was going to request it after revising it, but I think it would be good to correct it here.

jeff-dude:
> @jeff-dude Can you give me a moment to check? There was a miscalculation in the price part, so I was going to request it after revising it, but I think it would be good to correct it here.

sure, i will need to do a manual cleanup (delete from nft trades where project is seaport) and then deploy with latest changes to reload history. if you plan to change seaport data as well, please do so in another PR and i'll pair them up together and deploy after

sohwak:
ok. This is from wyvern on Dune v2, right? I was confused with the seaport. just proceed with this commit.

jeff-dude:
> ok. This is from wyvern on Dune v2, right? I was confused with the seaport. just proceed with this commit.

no, it's actually seaport data on v2 engine.

change is to the query reading from this file: opensea_v3_ethereum_events

which reads from: seaport_ethereum_transfers

if you're making changes to seaport, i'll wait to bundle

sohwak:
@jeff-dude I think it is better to do this merge first. Modifying seaport took more time than I thought. I will submit PR mine seperately.

jeff-dude:
Close T-15199

PR #1364, opened 2022-08-09
Sudoswap AMM to nft trades

andrewhong5297:

*For Dune Engine V2*
I've checked that:

---

Compiled SQL model has been tested here (with some where constraints to get it to run quicker lol): https://dune.com/queries/1153457?d=1

Not sure of the easiest way to test, seems other tests just check the count of trades in a time period. A good test would probably be just to check that number of rows added for a given period is the same as the count of rows across `LSSVMPair_general_call_swapTokenForSpecificNFTs`, `LSSVMPair_general_call_swapTokenForAnyNFTs`, and `LSSVMPair_general_call_swapNFTsForToken` for the same period. But open to suggestions.


jeff-dude:
> Brief comments on the purpose of your changes:
> 
> _For Dune Engine V2_ I've checked that:
> 
> * [x ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
> * [x ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables
> * [ ]  if adding a new model, I added a test
> * [x]  the filename is unique and ends with .sql
> * [x]  each sql file is a select statement and has only one view, table or function defined
> * [x]  column names are `lowercase_snake_cased`
> 
> When you are ready for a review, tag duneanalytics/data-experience. We will re-open your forked pull request as an internal pull request. Then your spells will run in dbt and the logs will be avaiable in Github Actions DBT Slim CI. This job will only run the models and tests changed by your PR compared to the production project.
> 
> Compiled SQL model has been tested here (with some where constraints to get it to run quicker lol): https://dune.com/queries/1153457?d=1
> 
> Not sure of the easiest way to test, seems other tests just check the count of trades in a time period. A good test would probably be just to check that number of rows added for a given period is the same as the count of rows across `LSSVMPair_general_call_swapTokenForSpecificNFTs`, `LSSVMPair_general_call_swapTokenForAnyNFTs`, and `LSSVMPair_general_call_swapNFTsForToken` for the same period. But open to suggestions.

thank you for the new spell PR 🙌 

for tests, since we're still in the early stages, we're looking for any creative approaches and also to keep them lightweight and easy to implement/follow/process.

an example that i'm looking to build out right now for initial dex trades build is to hand-build a csv file (could be google sheet exported as csv, then imported to project) with a handful of tx's that fit certain criteria, i.e. token bought from the model is truly the token bought. maybe as simple as 3-5 tx's per project/contract.

i also think aggregate tests are a good approach as you mentioned. you could write a query which aggregates the decoded tables over a time period, say a month or few months, then do the same with the built model. join them together and ensure a match. another aggregate approach would be to follow a similar pattern, but sum() the amount and compare to a value and ensure it matches up to a certain % threshold.

hope that helps, look forward to seeing a test or two applied

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+                , call_tx_hash
+                , call_trace_address
+                , call_block_time
+                , call_block_number
+                , call_success
+                , nftRecipient as call_from
+                , 'Buy' as trade_category 
+            FROM {{ source('sudo_amm_ethereum','LSSVMPair_general_call_swapTokenForSpecificNFTs') }}
+        ) s
+        {% if is_incremental() %}
```
jeff-dude, line 49: we've found it beneficial performance wise to move these incremental filters to within the CTEs -- for example, you'd apply this if block and filter to all three queries in the swaps CTE

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+            , output_pair as pair_address
+            , call_block_time as block_time
+        FROM {{ source('sudo_amm_ethereum','LSSVMPairFactory_call_createPairETH') }}
+        WHERE call_success
+    ),
+
+    owner_fee_update as (
+        SELECT 
+            *
+        FROM {{ source('sudo_amm_ethereum','LSSVMPair_general_evt_FeeUpdate') }}
```
jeff-dude, line 70: i'd apply the incremental filter here, since you use the date in join downstream

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+    ),
+
+    protocol_fee_update as (
+        SELECT
+            *
+        FROM {{ source('sudo_amm_ethereum','LSSVMPairFactory_evt_ProtocolFeeMultiplierUpdate') }}
```
jeff-dude, line 76: i'd apply the incremental filter here, since you use the date in join downstream

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+    ),
+
+    eth_traces as (
+        SELECT 
+            * 
+        FROM {{ source('ethereum', 'traces') }}
```
jeff-dude, line 82: spark will not like these CTEs on the base tables (traces, transactions, etc). any select * on a table of this volume will be inefficient. i would remove the CTEs on these and look to join directly downstream and apply as many filters as possible to enhance performance without losing records

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+    ),
+
+    eth_transactions as (
+        SELECT 
+            *
+        FROM {{ source('ethereum','transactions') }}
```
jeff-dude, line 88: same comment as traces

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+    nft_ethereum_aggregators as (
+        SELECT 
+            *
+        FROM {{ ref('nft_ethereum_aggregators') }}
+    ),
+
+    prices_usd_eth as (
+        SELECT 
+            *
+        FROM {{ source('prices','usd') }}
```
jeff-dude, line 106: i would either apply the incremental filter here or remove the CTE and join directly downstream, with the incremental filter as part of the join

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+                    , COALESCE(fu.newfee, pc.initialfee)/1e18 as ownerfee --most recent ownerfee, depends on bonding curve to implement it correctly. See explanation in fee table schema.
+                    , COALESCE(fu.evt_block_time, pc.block_time) as fee_update_time
+                    , pfu.newMultiplier/1e18 as protocolfee --most recent protocolfee, depends on bonding curve to implement it correctly. See explanation in fee table schema.
+                    , pfu.evt_block_time as protocolfee_updatetime
+                    , pc.nftcontractaddress
+                    , pc.asset_recip
+                FROM swaps
+                JOIN pairs_created pc ON pc.pair_address = contract_address --remember swaps from other NFT addresses won't appear!
+                LEFT JOIN owner_fee_update fu ON swaps.call_block_time >= fu.evt_block_time AND swaps.contract_address = fu.contract_address
+                LEFT JOIN protocol_fee_update pfu ON swaps.call_block_time >= pfu.evt_block_time
```
jeff-dude, line 140: without looking closer, is block time the only applicable join condition? should there be more?

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+            , ARRAY_AGG(distinct CASE WHEN substring(input,1,10)='0x42842e0e' THEN bytea2numeric_v2(substring(input,139,64))::int ELSE null END) 
+                as token_id
+            , sb.call_tx_hash
+            , sb.call_from
+            , sb.pair_address
+            , sb.nftcontractaddress
+            , sb.ownerfee
+            , sb.protocolfee
+        FROM swaps_w_fees sb 
+        LEFT JOIN eth_traces tr
```
jeff-dude, line 165: once changes are made for removing CTE, i'd ensure that block date/time is in the join conditions as either a filter on 'tr' alias directly on the is_incremental logic or matching the field to the from table

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,240 @@
...
+            , sc.royalty_fee_amount*pu.price as royalty_fee_amount_usd
+            , sc.platform_fee_amount*pu.price as platform_fee_amount_usd
+            , tx.from 
+            , tx.to 
+            , 'sudoswap-' || sc.tx_hash || '-' || sc.nft_contract_address || sc.token_id::string || '-' || sc.seller || '-' || sc.amount_original::string || 'Trade' AS unique_trade_id
+        FROM swaps_cleaned sc
+        LEFT JOIN prices_usd_eth pu ON pu.blockchain='ethereum'
+            AND date_trunc('minute', pu.minute)=date_trunc('minute', sc.block_time)
+            --add in `pu.contract_address = sc.currency_address` in the future when ERC20 pairs are added in.
+        LEFT JOIN eth_transactions tx ON tx.hash=sc.tx_hash
```
jeff-dude, line 234: after the CTE cleanup, i'd ensure this is filtered as much as possible. we're looking into more performance improvements related to base table joins on tx_hash

jeff-dude (changes requested):

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
andrewhong5297: nope, the protocol fee is updated across all pair contracts (checks against the factory)

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
andrewhong5297: can't filter this one since this is just checking for most recent fee update - so if the last updated swap was after the fee update then we'd lose data here.

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
andrewhong5297: same comment as above

andrewhong5297:
addressed comments and added a raw versus abstractions aggregated count test (not sure if final select is right for dbt tests). 

We can't use etherscan because all the data is missing from events/calls lol, so we'd have to maybe manually calculate results for some time period and save that as a seed to reference. Lmk if you think we should do that, we can use a subset of results from [this compiled query](https://dune.com/queries/1153457) after I've manually checked it against etherscan for errors.

andrewhong5297:
found an edge case for `0x9122f4c63076d7d1d68a82f8cae149fa7dbd6e7dee3ba43eba59b541deb80a57` that I need to debug

andrewhong5297:
okay, I've found a temporary fix for private transactions (since it seems all of them only make one swap, so we can aggregate trade data across all the traces without a duplication issue). Monitoring this [here](https://dune.com/queries/1158049), long term we need a fix for decoding private txs in call tables. 

would be great to add `dbt_utils` to packages so we can add some `not null` tests in the schemas, Not sure if it's compatible with databricks though.

spellbook/dbt_project.yml
```diff
@@ -103,6 +103,13 @@ models:
       ethereum:
         +schema: transfers_ethereum
         +materialized: view
+    
+    sudoswap:
```
jeff-dude, line 107: fyi, moved this to the models section, rather than seeds section below

spellbook/tests/sudoswap/ethereum/sudoswap_ethereum_assert_swaps.sql
```diff
@@ -0,0 +1,90 @@
...
+                    SELECT 
+                        contract_address
+                        , call_tx_hash
+                        , call_trace_address
+                        , call_block_time
+                        , call_block_number
+                        , call_success
+                        , tokenRecipient as call_from
+                        , 'Sell' as trade_category 
+                    FROM sudo_amm_ethereum.LSSVMPair_general_call_swapNFTsForToken
```
jeff-dude, line 32: any reason your recent commit switched back from {{ source( ) }} syntax to hardcode? looks like line 14 still has it, i think you're good to switch back

jeff-dude: i can continue to dig deeper into the models for now, tests will come last, but i'll look closer at this test soon

jeff-dude:
> Thanks Jeff. I've addressed comments and added a raw versus abstractions aggregated count test (not sure if final select is right for dbt tests).
> 
> We can't use etherscan because all the data is missing from events/calls lol, so we'd have to maybe manually calculate results for some time period and save that as a seed to reference. Lmk if you think we should do that, we can use a subset of results from [this compiled query](https://dune.com/queries/1153457) after I've manually checked it against etherscan for errors.

we're keeping tests relatively light for now while we work through performance on tx_hash and such. i'd say seed file isn't critical at this time. it can be considered in future enhancement

jeff-dude:
> okay, I've found a temporary fix for private transactions (since it seems all of them only make one swap, so we can aggregate trade data across all the traces without a duplication issue). Monitoring this [here](https://dune.com/queries/1158049), long term we need a fix for decoding private txs in call tables.
> 
> would be great to add `dbt_utils` to packages so we can add some `not null` tests in the schemas, Not sure if it's compatible with databricks though.

you can apply not null checks to columns in the schema files. there are four built-in dbt tests you can apply as-is. i've used it like this:
          - &evt_tx_hash
            name: evt_tx_hash
            description: "Primary key of the transaction"
            tests:
              - not_null

spellbook/tests/sudoswap/ethereum/sudoswap_ethereum_assert_swaps.sql
andrewhong5297: Ah I think I was copying and pasting from compiled to dune and back to make sure the query ran correctly. I'll correct it tmr

andrewhong5297:
test fixes made!

jeff-dude:
thank you!

i'm looking closer at the models now. i may push a few changes related to our pipeline improvements discovered the last few days

andrewhong5297:
Sounds good! Thanks Jeff!

jeff-dude:
pushed a few changes, hope they make sense. mostly for enhancements for our pipeline.

looks like the model is building well. you can take a peak yourself:
```
select count(1), min(block_date), max(block_date)
from dbt_jeff_sudoswap_ethereum.events
;
```

spellbook/tests/sudoswap/ethereum/sudoswap_ethereum_assert_swaps.sql
```diff
@@ -0,0 +1,90 @@
...
+            , COUNT(distinct call_tx_hash) as num_txs
+        FROM fil_swaps
+    ),
+
+    abstractions_swaps as
+    (
+        SELECT
+            COUNT(*) as num_trades
+            COUNT(distinct tx_hash) as num_txs
+        FROM {{ ref('nft_trades') }} nft
```
jeff-dude, line 79: i think ultimately this test will be fine, but as of now, sudoswap isn't added to nft trades, so i can't run this as of now

jeff-dude: here is the compiled test if you want to mess with it to help:

```
-- test to see if # trades in raw = # trades in processed.
WITH 
    raw_swaps as (
        WITH
            pairs_created as (
                SELECT 
                    _nft as nftcontractaddress
                    , _initialNFTIDs as nft_ids
                    , _fee as initialfee
                    , _assetRecipient as asset_recip
                    , output_pair as pair_address
                    , call_block_time as block_time
                FROM sudo_amm_ethereum.LSSVMPairFactory_call_createPairETH
                WHERE call_success
                AND _nft = '0xef1a89cbfabe59397ffda11fc5df293e9bc5db90'
            ),

            fil_swaps as (
                SELECT 
                    * 
                FROM (
                    SELECT 
                        contract_address
                        , call_tx_hash
                        , call_trace_address
                        , call_block_time
                        , call_block_number
                        , call_success
                        , tokenRecipient as call_from
                        , 'Sell' as trade_category 
                    FROM sudo_amm_ethereum.LSSVMPair_general_call_swapNFTsForToken
                    join pairs_created pc ON contract_address = pc.pair_address
                    where call_block_time >= '2022-07-10' AND call_block_time <= '2022-08-10'
                    AND call_success = true

                    UNION ALL
                    SELECT 
                        contract_address
                        , call_tx_hash
                        , call_trace_address
                        , call_block_time
                        , call_block_number
                        , call_success
                        , nftRecipient as call_from
                        , 'Buy' as trade_category 
                    FROM sudo_amm_ethereum.LSSVMPair_general_call_swapTokenForAnyNFTs
                    join pairs_created pc ON contract_address = pc.pair_address
                    where call_block_time >= '2022-07-10' AND call_block_time <= '2022-08-10'
                    AND call_success = true

                    UNION ALL
                    SELECT
                        contract_address
                        , call_tx_hash
                        , call_trace_address
                        , call_block_time
                        , call_block_number
                        , call_success
                        , nftRecipient as call_from
                        , 'Buy' as trade_category 
                    FROM sudo_amm_ethereum.LSSVMPair_general_call_swapTokenForSpecificNFTs
                    join pairs_created pc ON contract_address = pc.pair_address
                    where call_block_time >= '2022-07-10' AND call_block_time <= '2022-08-10'
                    AND call_success = true
                ) s
            )

        SELECT 
            COUNT(*) as num_trades
            , COUNT(distinct call_tx_hash) as num_txs
        FROM fil_swaps
    ),

    abstractions_swaps as
    (
        SELECT
            COUNT(*) as num_trades
            COUNT(distinct tx_hash) as num_txs
        FROM dbt_jeff_nft.trades nft
        WHERE blockchain = 'ethereum' AND project = 'sudoswap' and version = 'v1'
        AND block_time >= '2022-07-10' AND block_time <= '2022-08-10'
        AND nft_contract_address = '0xef1a89cbfabe59397ffda11fc5df293e9bc5db90'
    )

SELECT 
    (SELECT num_trades FROM abstractions_swaps) - (SELECT num_trades FROM raw_swaps) as trades_mismatch
    , (SELECT num_txs FROM abstractions_swaps) - (SELECT num_txs FROM raw_swaps) as txs_mismatch
FROM 1
WHERE (SELECT num_trades FROM abstractions_swaps) - (SELECT num_trades FROM raw_swaps) > 0
    OR (SELECT num_txs FROM abstractions_swaps) - (SELECT num_txs FROM raw_swaps) > 0
```

andrewhong5297:
Ah im not by my computer, but there's a join on traces that's `call_trace_address[0]=trace_address[0]` which should be `call_trace_address=trace_address[0:cardinality(call_trace_address)]` because some swap calls are deeper than index 0 depending on routers.

Will make the fix when I'm home but feel free to fix that if you're trying to build the model today

jeff-dude:
i can rebuild the model in the dbt_jeff_* schema for you to test again, but what do you want to do about your automated test?

as mentioned in previous comment, that test won't show anything until nft trades loads sudoswap data -- is that the next logical step after this PR? to script out load to nft trades?

jeff-dude:
> i can rebuild the model in the dbt_jeff_* schema for you to test again, but what do you want to do about your automated test?
> 
> as mentioned in previous comment, that test won't show anything until nft trades loads sudoswap data -- is that the next logical step after this PR? to script out load to nft trades?

you could write a more simplified one for this PR, then use that in nft trades PR?

andrewhong5297:
The test should be fine right? Not sure what a simplified version would look like, either way we would be testing it against nft trades and not the temp generated table?

andrewhong5297:
> The test should be fine right? Not sure what a simplified version would look like, either way we would be testing it against nft trades and not the temp generated table?

compiled abstraction results: https://dune.com/queries/1174032?d=1
raw test results: https://dune.com/queries/1174027?d=1

The test is currently passing. isn't it possible to build the model directly once with `dbt run +sudoswap_ethereum_trades`, then maybe we can change the test to run against that table instead of `nft.trades`?

jeff-dude:
> > The test should be fine right? Not sure what a simplified version would look like, either way we would be testing it against nft trades and not the temp generated table?
> 
> compiled abstraction results: https://dune.com/queries/1174032?d=1 raw test results: https://dune.com/queries/1174027?d=1
> 
> The test is currently passing. isn't it possible to build the model directly once with `dbt run +sudoswap_ethereum_trades`, then maybe we can change the test to run against that table instead of `nft.trades`?

yes, my confusion was around the usage of nft trades in the test. sudoswap data doesn't exist in nft trades and won't with this PR deployment either.

we can either add the logic to push this data into nft trades, then run this test, or alter the test to point to this final object rather than nft trades

andrewhong5297:
> yes, my confusion was around the usage of nft trades in the test. sudoswap data doesn't exist in nft trades and won't with this PR deployment either.
> 
> we can either add the logic to push this data into nft trades, then run this test, or alter the test to point to this final object rather than nft trades

Ohh I see what you mean. Can we add this into `nft.trades`? that was the initial goal. I can take a stab at updating schema/refs but not sure if there's anything else that needs to be updated too

jeff-dude:
> > yes, my confusion was around the usage of nft trades in the test. sudoswap data doesn't exist in nft trades and won't with this PR deployment either.
> > we can either add the logic to push this data into nft trades, then run this test, or alter the test to point to this final object rather than nft trades
> 
> Ohh I see what you mean. Can we add this into `nft.trades`? that was the initial goal. I can take a stab at updating schema/refs but not sure if there's anything else that needs to be updated too

take a peek here: https://github.com/duneanalytics/abstractions/pull/1298/files

that is a good example of adding through to nft trades

andrewhong5297:
kk just tried adding it, lmk if I missed a step somewhere

jeff-dude:
> kk just tried adding it, lmk if I missed a step somewhere

yep, looked good. pushed a commit with a few more tweaks to match new design patterns at nft object level. i'll test it locally again now

spellbook/models/nft/nft_events.sql
```diff
@@ -1,23 +1,207 @@
...
+                token_standard,
+                trade_type,
+                number_of_items,
+                trade_category,
+                evt_type,
+                seller,
+                buyer,
+                amount_original,
+                amount_raw,
+                currency_symbol,
```
jeff-dude, line 185: this looks like the one column expected in nft objects that wasn't built out in sudoswap events model -- do you want to force it blank ('') or want to push a commit with that column derived in the sudoswap model?

jeff-dude:
okay -- i'll stop pushing commits now 😂

all objects are built with 'dbt_jeff_' preface (sudoswap schema, nft schema)

i updated the test case and it's running as expected now 👍 

the only outstanding item is if you want to add the 'currency symbol' column in the base sudoswap events model, then have it propagate all the way through to other sudoswap views and nft schema views. right now, i'm forcing '' as you can see in some of the views.

andrewhong5297:
added in 'ETH' for now! I'll do another PR later once ERC20 pairs are live.

andrewhong5297:
It seems the distinct array_agg created text objects instead of keeping int token ids, it ran fine when i pasted it in the editor. Not sure if this is a dbt specific thing? Maybe need to cast token ID as int before array_agg?

spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -0,0 +1,256 @@
...
+        -- we traces to get NFT and ETH transfer data because sudoswap doesn't emit any data in events for swaps, so we have to piece it together manually based on trace_address.
+        SELECT
+            sb.call_block_time
+            , sb.call_block_number
+            , sb.trade_category
+            , CASE WHEN sb.trade_category = 'Buy' THEN SUM(value)/(1+sb.ownerfee+sb.protocolfee)
+                ELSE SUM(value)/(1-sb.ownerfee-sb.protocolfee)
+                END as base_price
+            , SUM(value) as trade_price_eth --should give total value of the trade (buy or sell)
+            , ARRAY_AGG(distinct CASE WHEN substring(input,1,10)='0x42842e0e' THEN bytea2numeric_v2(substring(input,139,64))::int ELSE '' END)
```
jeff-dude, line 149: i wonder if the "ELSE '' END" is what is forcing a string within the array. maybe instead of '' it should be NULL?

andrewhong5297: Ah yeah that's probably it. Lemme fix

jeff-dude: okay, the db says array(int) now -- want to confirm?

jeff-dude (approved):

0xRobin:
@andrewhong5297 Great work on this, I wish I'd seen it earlier so I could have helped!

Going to add some of my thoughts here so they don't get lost. 
I'll have time to follow up on this next week and propose some changes.

1.  I don't think it's a good idea to populate the existing `royalty fee` fields with something that isn't royalty related. 
**Lets move these LP fees to their own (new) fields**, they are also pretty misleading as they aren't really materialized in the protocol (they just result in higher floor/offer spread). That aside, having some stats on these *virtual* fees accrued is fun.

2. The trade price calculation seems to be off, this is due to the way we're using the subtrace values here. 
`BUY`: we are **including** the protocol fee into the trade price (cause the buyer pays full amount to pool, pool relays the protocol fee).
`SELL`: we're **not including** the protocol fee into the trade price (seller receives net amount, pool pays for protocol fees)
I believe it's convention to **include** the protocol fees (and royalty fees, which are 0 here) into the trade price for the other nft.trades abstractions. **We should aim to follow the same conventions.**  
You could still keep your existing subtrace matching logic but change
https://github.com/duneanalytics/abstractions/blob/eb2e5ebd443547d072db7309a6ea5d74adc757c1/spellbook/models/sudoswap/ethereum/sudoswap_ethereum_events.sql#L145-L148
To something like
```sql
, CASE WHEN sb.trade_category = 'Buy' THEN SUM(value)/(1+sb.ownerfee+sb.protocolfee)
                ELSE SUM(value)*(1+sb.protocolfee)/(1-sb.ownerfee-sb.protocolfee)
                END as base_price
, CASE WHEN sb.trade_category = 'Buy' THEN SUM(value)
                ELSE SUM(value)*(1+sb.protocolfee)
                END as trade_price
```

andrewhong5297:
haha next time!

1. moved!
2. Sure, including fees makes sense. I'm fairly sure that we can use `base_price` the way I've calculated it, not sure why for "Sell" you're multiplying by protocol fee and then dividing it again?

PR #1368, opened 2022-08-11
Replace trader_a values for 1inch in dex.trades

fatayri:
Currently in dex.trades trader_a and tx_from are exactly the same for 1inch transactions. With this PR we replace the trader_a with the "from" value from the trace, not the transaction.

*For Dune Engine V2*
I've checked that:

* [] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)

### DIFF
--- a/ethereum/dex/trades/insert_1inch.sql
+++ b/ethereum/dex/trades/insert_1inch.sql
@@ -60,7 +60,7 @@ WITH rows AS (
             '1inch' AS project,
             version,
             'Aggregator' AS category,
-            tx."from" AS trader_a,
+            min_trace.trader_a AS trader_a,
             NULL::bytea AS trader_b,
             to_amount AS token_a_amount_raw,
             from_amount AS token_b_amount_raw,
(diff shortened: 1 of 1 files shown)


jeff-dude:
thank you @fatayri for the PR. i assume this means you'd like to truncate dex trades for 1inch (attached script) and then reload fresh with the new field?

i can look to do that tomorrow, it'll take a good amount of time i'd imagine to reprocess, so the data will be gone for a bit.

fatayri:
Hey @jeff-dude! Thanks for the prompt reply.

Do we actually have any other options other than truncating and replacing? If not, it should be fine to do it so.

jeff-dude:
> Hey @jeff-dude! Thanks for the prompt reply.
> 
> Do we actually have any other options other than truncating and replacing? If not, it should be fine to do it so.

the other option is that you can supply an update statement here in the comments, but that may end up a bit complex with all the different project names, versions and categories contained within. it also may take even more time than a delete/reload.

let me know your thoughts between the two and i'll look to perform on monday morning

fatayri:
Hey, sorry for the super long reply. We considered our options and decided that it might make more sense to truncate the values and fully rewrite it. But it's better to let it run over the weekend. 

If it's not too late for today and you have some spare time - it would be perfect. If not, we can run it next Friday.

jeff-dude:
> Hey, sorry for the super long reply. We considered our options and decided that it might make more sense to truncate the values and fully rewrite it. But it's better to let it run over the weekend.
> 
> If it's not too late for today and you have some spare time - it would be perfect. If not, we can run it next Friday.

no problem, i'll look to do it over the weekend at some point

jeff-dude:
this is still running and backfilling history, if anyone is noticing the data missing

jeff-dude (approved):
@fatayri this fix should be applied and backfilled again 👍

PR #1370, opened 2022-08-11
fix tx_hash / block_number column switch on nft events

jeff-dude:
nft events table has a bug with the two columns switched


jeff-dude:
@duneanalytics/team-data-experience if anyone is online, would appreciate an approval 👍

aalan3 (approved):

PR #1376, opened 2022-08-14
Update OP v1 tables

MSilb7:
Add tokens, contract creators, name mapping.

---

*For Dune Engine V2*
I've checked that:


optimism2/ovm2/get_contracts/insert_project_name_mapping.sql
```diff
@@ -50,6 +50,11 @@ Layerzero	Layer Zero
 Xtoken	xToken
 Instadapp	InstaDapp
 Lifi	LiFi
+Optimistic Exporer	Optimistic Explorer - Get Started NFT
+ironbank	Iron Bank
+Ironbank	Iron Bank
```
jeff-dude, line 55: postgres seems to treat line 54 & 55 as the same -- likely not case sensitive, leading to duplicates and failed load. i had to remove line 55 to run. will this work?

optimism2/ovm2/get_contracts/insert_project_name_mapping.sql
```diff
@@ -50,6 +50,11 @@ Layerzero	Layer Zero
...
+iron_bank	Iron Bank
+Iron_bank	Iron Bank
```
jeff-dude, line 57: same comment

optimism2/ovm2/get_contracts/insert_project_name_mapping.sql
jeff-dude: i ran if you wanted to check output

jeff-dude (approved):
closing this with the outstanding comments. please open new PR if the changes didn't cover all that was needed

PR #1380, opened 2022-08-15
Migrate Optimism `uniswap_v3.view_pools` to V2

chuxinh:
Migrating this abstraction from V1 to V2: https://github.com/duneanalytics/abstractions/tree/master/optimism2/uniswap_v3
cc: @MSilb7 

*For Dune Engine V2*
I've checked that: 

* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory): https://dune.com/queries/1170735?d=1


chuxinh:
new one here: https://github.com/duneanalytics/abstractions/pull/1389

PR #1383, opened 2022-08-16
Gnosis oneinch dex trades

jacubcub:
This is a V1 abstraction table so the below checklist doesn't really apply...

Adding 1inch trades to gnosis chain dex.trades table. Gnosis chain only has 1inch V4 on it. It was added early 2022 so only need to insert from 2022 forward.

I created this function manually and inserted into dune_user_generated.trades on the gnosis db to make sure it works and the data looks good.

*For Dune Engine V2*
I've checked that:


jeff-dude (approved):
thank you @jacubcub this is now loaded. i pushed one commit for an fyi.

PR #1384, opened 2022-08-16
chore(prices): add SPS and DEC token

yashtbc:
add tokens
- SPS
- DEC


jeff-dude (approved):
thank you @yashtbc for the add

PR #1382, opened 2022-08-15
remove duplicates from ethereum erc20 tokens view

jeff-dude:
remove duplicates from ethereum erc20 view

PR #1385, opened 2022-08-16
Prices for arbitrum, avalanche_c, gnosis and optimism

aalan3:
This PR adds USDT, USDC, WBTC and DAI for the newer chains that don't have prices yet. This also adds prices for OP.


jeff-dude (approved):
we can approve and push this, but we may want to consider writing a test which checks for duplicates in this file as it's manually maintained and can easily contain errors. this is happening in erc20 tokens (#1382)

PR #1387, opened 2022-08-16
Adds ClipperVerifiedCaravel to dex.trades

aothman:
Clipper redeployed on August 5th with a new contract that implements the same ABI as the existing `v2` contract. The old contract is no longer in use. This code updates `dex.trades` to read and update swaps from the new contract calling it version 3.


jeff-dude (approved):
@aothman updated and backfilled 👍

PR #1391, opened 2022-08-17
modified Seaport to add a zone address belonging to OpenSea

sohwak:

(Dune Legacy)
- Added a zone address belonging to OpenSea
- execute `backfill_by_zone_address.sql` once to update previous rows

*For Dune Engine V2*
I've checked that:


jeff-dude:
do you also need to add an update statement for nft trades, since you applied the same change? should be similar format, just a few more filters

jeff-dude (approved):
@sohwak this is done. if you want to update nft trades, feel free to submit another PR with similar script to update

sohwak:
I thought about `nft.trades` (Dune v1) too, but non-Opensea cases are already excluded so it is hard to update but insert needed.
On the other hand, I don't think there would be a big difference overall (hundreds rows) so I didn't add it.

I will apply when modifying `nft.trades` on Dune v2.

PR #1400, opened 2022-08-19
Fix address format and lowercase for Uniswap

aalan3:
I was using this query to test performance and noticed some small errors. The addresses were still in v1 format and one of the addresses was not in lowercase.


jeff-dude (approved):
@aalan3 thank you! 

maybe we should add a macro which is applied to any hardcoded hash value to lower and replace '\' with '0' to be safe?

PR #1411, opened 2022-08-22
patch division-by-zero on seaport abstraction

sohwak:

modified : some "division-by-zero" cases which were not covered in the previous patch.

affected tables:
- `nft.trades`, `platform = 'OpenSea'` : since 20 Aug 2022
- `seaport.transfers`  : since 20 Aug 2022

please backfill affected tables too.

Thank you.

*For Dune Engine V2*
I've checked that:

### DIFF
--- a/ethereum/nft/trades/insert_seaport.sql
+++ b/ethereum/nft/trades/insert_seaport.sql
@@ -93,6 +93,7 @@ with p1_call as (
                           ,a.*
                       from p1_evt a
                     ) a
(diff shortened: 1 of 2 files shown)


jeff-dude (approved):
@sohwak the functions are updated and the cron commands should pick up where the data left off, we can monitor

sohwak:
Really appreciate your prompt support. I just confirmed that the data is well loaded into both tables.

PR #1419, opened 2022-08-23
fix duplicated unique key

sche:
Fix the unique key constraint of the `safe_ethereum.eth_transfers` table.

*For Dune Engine V2*
I've checked that:

### DIFF
--- a/spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
+++ b/spellbook/models/safe/ethereum/safe_ethereum_eth_transfers.sql
@@ -3,7 +3,7 @@
         materialized='incremental',
         alias='eth_transfers',
         partition_by = ['block_date'],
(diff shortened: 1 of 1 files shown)


sche:
@jeff-dude now it should work 🤞
But the initial table should be recreated.

jeff-dude:
closing due to #1428 

@sche we did a reorg on our public repo today. to make it easier for you, since i have my env setup, i've mimicked the changes here. in the meantime, i'd recommend refreshing your local env to fit the new repo structure.

change summary:
1. changed repo name from 'abstractions' to 'spellbook'
2. renamed default branch from 'master' to 'main'
3. moved all v1 engine abstractions into 'deprecated-dune-v1-abstractions'
4. removed 'spellbook' directory at root and moved all objects within (i.e. models directory you use here) to the root level

it might be easiest to start fresh with a fork of the repo, config local env again.

PR #1437, opened 2022-08-25
sudoswap improved price calculations

0xRobin:
following the discussion we had after the merge of: https://github.com/duneanalytics/spellbook/pull/1397
This PR includes:
- improved ETH tracing to determine the correct trade amount
- aggregator fields population
- test with example trades

*For Dune Engine V2*
I've checked that:


0xRobin:
@andrewhong5297 give this a look if you can.
I haven't finished checking everything but query seems to run.

andrewhong5297:
Awesome!!! I'll take a look tonight, agree on having tests for different trade types. 

Probably need to have pool type checks with trade types, and add in the total protocol fee check as a sanity check.

Likely also need an aggregator check (I saw Vasa tweet that gem was integrating next week).

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -66,42 +69,60 @@ WITH
...
             WHERE call_success = true
             {% if is_incremental() %}
             AND call_block_time >= (select max(block_time) from {{ this }})
             {% endif %}
         ) s
-    ),
+    )
+
+    -- this join should be removed in the future when more call trace info is added to the _call_ tables, we need the call_from field to track down the eth traces.
+    , swaps_with_calldata as (
```
andrewhong5297, line 84: why are we pulling this data here instead of later on in swaps_with_traces? seems expensive to join twice on ethereum.traces

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -111,70 +132,88 @@ WITH
...
-            , SUM(value) as trade_price --should give total value of the trade (buy or sell)
+            , SUM(
+                CASE WHEN sb.trade_category = 'Buy' -- caller buys, AMM sells
+                THEN (
+                    CASE WHEN tr.from = sb.call_from THEN value -- amount of ETH payed
+                    WHEN (tr.to = sb.call_from AND sb.call_from != sb.asset_recip) THEN -value --refunds unless the caller is also the asset recipient, no way to discriminate there.
+                    ELSE 0 END)
+                ELSE ( -- caller sells, AMM buys
+                    CASE WHEN tr.from = sb.pair_address THEN value -- all ETH leaving the pool, nothing should be coming in on a sell.
+                    ELSE 0 END)
```
andrewhong5297, line 182: for sells, where are we adding back in protocol fee and owner fee to get to actual price? since that's subtracted before ETH is transferred out of pool

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -111,70 +132,88 @@ WITH
...
-            , CASE WHEN sb.trade_category = 'Buy' THEN SUM(value)/(1+sb.ownerfee+sb.protocolfee)
-                ELSE SUM(value)/(1-sb.ownerfee-sb.protocolfee)
-                END as base_price
-            , SUM(value) as trade_price --should give total value of the trade (buy or sell)
+            , SUM(
+                CASE WHEN sb.trade_category = 'Buy' -- caller buys, AMM sells
+                THEN (
+                    CASE WHEN tr.from = sb.call_from THEN value -- amount of ETH payed
+                    WHEN (tr.to = sb.call_from AND sb.call_from != sb.asset_recip) THEN -value --refunds unless the caller is also the asset recipient, no way to discriminate there.
+                    ELSE 0 END)
```
andrewhong5297, line 179: I still don't fully understand refunds, mind adding in the comments here an example tx or flow of ETH? is it price paid then partial value is sent back or something?

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
0xRobin: We are joining on different conditions here so I don't think there's a way around it.
Here we need the tracedata of the trace that corresponds to the call, later we need all the traces from the subtraces that came from the call. 
We should push for more standard trace data to be included into the function call tables by default (like `from` and `to` fields). That would make this join not needed anymore (as stated in the inline comment)

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
0xRobin: protocol fee is included (cause that also gets transferred out of the pool)
owner fee should not be included cause it's not actually payed.
I'll write my thought framework out in the general comments here. 
Might even be worth including it in a readme in the repo.

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
0xRobin: Yes so since these function calls are just 'payable' one can send however much ETH you'd like when interacting with a pool directly. The ETH that was not used for the transfer is sent back to the caller at the end. 
Nice example:
https://etherscan.io/tx/0x1c74e5f2717bc4f9d791ad89183eb351919fdfb7569968a65cddc9a1a7b1b5eb
- 1ETH call `swapTokenForAnyNFTs`
- 0.5010 ETH to the pool recipient
- 0.0025 ETH to the protocol
- 0.4964 ETH refund

0xRobin:
## Main `nft.trades` framework
#### (this is not sudo specific)
When reporting trades and determine the actual value exchanged I like to think of states and state transformations.
A basic NFT trade can be distilled to these parties:

B) Buyer, S) Seller F) Facilitators (marketplaces/protocols/aggregators/...)
```mermaid
flowchart LR;
    B(Buyer) -->|ETH| S(Seller);
    S(Seller) -->|NFT| B(Buyer);
    B(Buyer) -.->|ETH| F(Facilitators);
```
These are the state transformations for each:
```mermaid
stateDiagram-v2
    b0 : B0
    b1 : B1
    s0 : S0
    s1 : S1
    f0 : F0
    f1 : F1
    b0 --> b1 : +1 NFT <br> - ΔB ETH
    s0 --> s1 : -1 NFT <br> + ΔS ETH
    f0 --> f1 : + ΔF ETH
```

Where in general it must hold that:
$\Delta_B = \Delta_S + \sum_{i}{\Delta_{F_i}}$

In short:
* The Buyer pays what the Seller and all intermediate Facilitators receive (fees)
* **The trade value is what the Buyer pays**


----
### sudoswap
If we apply this to sudo, it becomes clear that we can correctly determine:
-  `the trade value` by tracing down what the buyer paid.
- `the fee value` by tracing down what the protocol received

Which is what we are doing here:
https://github.com/duneanalytics/spellbook/blob/de1d02844ea17c2de06b5e070bfcb7b4893f23c6/models/sudoswap/ethereum/sudoswap_ethereum_events.sql#L174-L187

What we are not seeing here is the so-called `swap fee` set by a trade pool owner (owner fee).
I would say the swap fee has no influence in the value exchanged, it only influences the spread when setting both ask and bid prices on the same moving price curve. When you are listing NFTs at 10% above the current floor price, you are not *collecting* a 10% fee when that ask gets filled. 

You should look at it as a spread setting instead of a fee setting where it results in profit if (and only if) spread from both the ask and bid sides are captured. 
Now that I have thought about it more, I believe we should remove the owner fee fields from this base table as it will probably lead to a lot of misinformation about pool profits. 
It's still interesting to calculate the 'virtual swap fees' that were used in all the pool trades, but more thought is needed to present it correctly to the public then just a flat fee collected on each swap. 

I hope this gives a good/better understanding at the thought framework I'm currently employing. :smile:

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
andrewhong5297: ah I see, didn't notice that you changed `call_from` to `trade_recipient` in the earlier CTE. makes sense, yeah would be nice to have `call_from`, `call_to`, and `value` in the decoded tables.

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
andrewhong5297: Hmm, if you look at the bonding curves you'll see that owner fee is added/subtracted from the total price. 

For example here https://github.com/sudoswap/lssvm/blob/main/src/bonding-curves/LinearCurve.sol#L166 

You can `feeMultiplier` of the trade pool is subtracted from the value of ETH actually sent to the seller.

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
0xRobin: yeah that field name was causing me lots of confusions. :laughing:

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
0xRobin: The `feeMultiplier` is subtracted from the baseprice to get to the bid price
It's added to the baseprice to get to the ask price.

It's a spread setting instead of fee setting.

I like to think in extremes to clear things up,
If you have a pool with 10 ETH internal spot price, a feemultiplier of 90% and only 1.5 ETH liquidity.
That bid price will be ~1ETH, ask price will be ~19ETH
Someone can come in and sell an NFT to the pool for 1ETH.

The pool gains 1 NFT and pays 1 ETH (I'm ignoring protocol fees for simplicity)
Does it make sense to say the pool **payed and received** 9 ETH in swap fees in this trade?

andrewhong5297: Yes because we want totals? You could make same argument for if protocol fee was 90%, I don't follow

andrewhong5297:
wait we still need owner fees in there! I thought we agreed to keep all fees in prices and those columns too. I can't accept deleting those columns completely.

andrewhong5297:
if you take those out there's no way for us to back into real spot price either if we're doing slippage analysis later

0xRobin:
> if you take those out there's no way for us to back into real spot price either if we're doing slippage analysis later

You can still calculate internal spot price as `protocol_fee_amount/protocol_fee_percentage`

andrewhong5297:
> > if you take those out there's no way for us to back into real spot price either if we're doing slippage analysis later
> 
> You can still calculate internal spot price as `protocol_fee_amount/protocol_fee_percentage`

lol fair, but I am still against taking it out since I think it's useful data to have.

0xRobin:
> I think it's useful data to have.

will revert then, no harm in including it. :+1:

0xRobin:
@jeff-dude this is now ready for review

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -111,128 +135,150 @@ WITH
...
+            , sb.pool_fee
             , sb.protocolfee
-        FROM swaps_w_fees sb
-        LEFT JOIN {{ source('ethereum', 'traces') }} tr
+            , project_contract_address
+            -- these 2 are used for matching the aggregator address, dropped later
+            , router_caller
+            , call_from
+        FROM swaps_w_fees sb 
+        LEFT JOIN {{ source('ethereum', 'traces') }} tr
```
jeff-dude, line 205: can this be an inner join?
please see alan latest PR on performance enhancement changes: https://github.com/duneanalytics/spellbook/pull/1465/files#diff-1ed6be0ef7436ccdcb014cb5e12ea867d6dd7d14ec655a15c522ae2186f29833

jeff-dude: @aalan3 fyi, this PR changes sudoswap model which you look to change in linked PR

seeds/sudoswap/sudoswap_ethereum_example_trades.csv
```diff
@@ -0,0 +1,24 @@
+block_number,tx_hash,trade_category,nft_contract_address,token_id,amount_original,platform_fee_amount,pool_fee_amount
```
jeff-dude, line 1: when adding a new seed file, the dbt project yaml file should be edited to include the new directory and file for seeds. please see line 148 forward in here: https://github.com/duneanalytics/spellbook/blob/main/dbt_project.yml

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -9,32 +9,35 @@
 }}
 
 --base table CTEs
-WITH
+WITH
```
aalan3, line 12: nitpick: if you turn on whitespace removal on save in your editor these will go away and the diff will be easier to review

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
```diff
@@ -247,25 +293,26 @@ WITH
             AND pu.minute >= '2022-4-1'
             {% endif %}
             --add in `pu.contract_address = sc.currency_address` in the future when ERC20 pairs are added in.
-        LEFT JOIN {{ source('ethereum', 'transactions') }} tx
-            ON tx.hash=sc.tx_hash
+        LEFT JOIN {{ source('ethereum', 'transactions') }} tx
```
jeff-dude, line 296: similar idea here with inner join. you could either wait for other PR to merge and then merge main into your branch, or match changes

0xRobin: merged in main :heavy_check_mark:

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
0xRobin: done :heavy_check_mark:

seeds/sudoswap/sudoswap_ethereum_example_trades.csv
0xRobin: :heavy_check_mark:

models/sudoswap/ethereum/sudoswap_ethereum_events.sql
0xRobin: merged in main with the performance changes :heavy_check_mark:

jeff-dude:
@0xRobin i pushed a few minor commits. the model runs as expected. i did compile the code and grab the test query on new seed file, it appears rows are returned:
https://dune.com/queries/1221898

tests/sudoswap/ethereum/sudoswap_ethereum_assert_samples.sql
```diff
@@ -18,10 +18,10 @@ WITH trades as (
     , case when (t.platform_fee_amount = ex.platform_fee_amount) then true else false end as correct_platform_fee
     , case when (t.pool_fee_amount = ex.pool_fee_amount) then true else false end as correct_pool_fee
     from trades t
-    outer join examples ex
+    join examples ex
```
0xRobin: We actually want to do outer join here because we want the the full union of the two sets.

0xRobin:
> @0xRobin i pushed a few minor commits. the model runs as expected. i did compile the code and grab the test query on new seed file, it appears rows are returned:

@jeff-dude thanks for this!
I manually checked the data and the issue was actually with the seed file where there were often not enough decimals reported. Should be resolved now, can you revalidate the test?

tests/sudoswap/ethereum/sudoswap_ethereum_assert_samples.sql
jeff-dude: got it, i think syntax needs to be 'full outer join' or it fails, i can push a commit if so

jeff-dude:
> > @0xRobin i pushed a few minor commits. the model runs as expected. i did compile the code and grab the test query on new seed file, it appears rows are returned:
> 
> @jeff-dude thanks for this! I manually checked the data and the issue was actually with the seed file where there were often not enough decimals reported. Should be resolved now, can you revalidate the test?

rebuilding now

jeff-dude (approved):
thank you @0xRobin and @andrewhong5297 

fyi @soispoke this model will need a full refresh in prod

PR #1438, opened 2022-08-25
add metadata tag for dbt run commands

jeff-dude:
on hardcoded views which add metadata to other objects, include tag for 'metadata' to allow ease of orchestration in dbt


jeff-dude:

jeff-dude:
closing this for now but i do think tags will be the key to successful job orchestration as we grow. just an fyi @soispoke @dsalv

PR #1444, opened 2022-08-26
Nft trades performance test

aalan3:
This PR tests forcing a one week window on incremental updates for the two slowest nft trades tables.

For some reason there is a bunch of whitespace in these files so my editor removed it. To review the actual code changes look at these two commits:
https://github.com/duneanalytics/spellbook/commit/96f81d1db2622c9fb31e21a81332d61a25281a1c
https://github.com/duneanalytics/spellbook/commit/452c2d12ee6d9ccf233e0d1cc4b2c29c27776313

If this makes performance better we can do it for the rest of the nft trades tables as well.

### DIFF
--- a/models/opensea/ethereum/opensea_v1_ethereum_events.sql
+++ b/models/opensea/ethereum/opensea_v1_ethereum_events.sql
@@ -10,15 +10,15 @@
 }}
 
 WITH wyvern_call_data as (
-SELECT 
+SELECT
   call_tx_hash,
   call_block_time,
   CASE WHEN contains('0x68f0bcaa', substring(calldataBuy,1,4)) THEN 'Bundle Trade'
         ELSE 'Single Item Trade'
   END AS trade_type,
-  CASE WHEN contains('0xfb16a595', substring(calldataBuy,1,4)) THEN 'erc721'  
-        WHEN contains('0x23b872dd', substring(calldataBuy,1,4)) THEN 'erc721' 
-        WHEN contains('0x96809f90', substring(calldataBuy,1,4)) THEN 'erc1155' 
+  CASE WHEN contains('0xfb16a595', substring(calldataBuy,1,4)) THEN 'erc721'
+        WHEN contains('0x23b872dd', substring(calldataBuy,1,4)) THEN 'erc721'
+        WHEN contains('0x96809f90', substring(calldataBuy,1,4)) THEN 'erc1155'
         WHEN contains('0xf242432a', substring(calldataBuy,1,4)) THEN 'erc1155'
   END AS token_standard,
   addrs [0] as project_contract_address,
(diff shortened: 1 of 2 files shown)


dot2dotseurat (approved):

0xRobin:
Any reason to go for a week long window?
This will always recalculate the full last 7 days (even if last incremental run was 1 hour ago)
It will also skip events should there ever be a window > 7 days where no run was done.

Should we not use something like:
```sql
{% if is_incremental() %}
AND tx.block_time >= (select max(block_time) from {{ this }})
{% endif %}
```

(I'm unsure how big the impact of the select-max statement  is on performance though)

0xRobin:
ah sure, makes sense!
I was not thinking about out-of-order events. :+1:

jeff-dude:
> @0xRobin
> 
> Using a week long window is just to be safe, there should be very few cases where new data come in after that and since we use the `tx.block_time >= (select max(block_time) from {{ this }})` filter it will also skip these old transactions anyway. If new data doesn't come in for 7 days, we likely have a bigger issue and would likely need to full refresh. But it is tradeoff.
> 
> The reason why using a week is faster is because it's a static value, where max block_time is a calculated value derived. The former allows the query planner to push it down as a filter to scan less partitions which should give significant performance gains vs now where it's doing a full scan.

can't we test this on local dbt setup first? i'm curious to see equality vs. range filter performance difference on left joins to larger objects like ethereum transactions.

while this approach cuts down the range to 7 days, i.e. 7 partitions on block_date, it's few files to scan through.

with the equality approach on the left join, if the incoming source data from wyvern decoded date is already filtered on incremental (i.e. an hourly run), then the block_date would be ~one day of data and then spark would only need to search that one partition on ethereum transactions.

i'd like to see what the query plan says though, if it actually does that

jeff-dude:
the other performance issue with this script is it does a select all on wyvern decoded data with no incremental date filter, so it pulls all history every time and joins to ethereum transactions on the last 7 days worth, but since it's a left join, all of wyvern history comes through.

i think first step is to filter incoming decoded source data

aalan3:
You bring up a good point about filtering decoded data, so I'll try to explain this a bit better.

Looking at the query profile for these types of queries the slowest part of the query is the transactions table, both fetching and processing. So by filtering one date, which the table is partitioned by, we can skip a bunch of data and get good performance.

We do a lot of fetches from decoded data but these are not slow and for much less data. Even if we did optimize this it would not have a big effect. This is clear from looking the query profiler, the major thing is transactions (then prices).

Another point here is that the underlying data for decoded tables are not partitioned by date but by namespace and contract name. So filtering by date will not give the same performance gain. We could still do it but for this specific test we are looking to see if we can make a dent on the transactions.

I also actually think that we need to modify the merge into step to really speed up inserts, I'm not sure how much this will help.

PR #1445, opened 2022-08-26
prevent negative veBAL balances

markusbkoch:
Fixes a bug in the veBAL balances view that allowed for negative values
cc @mendesfabio


jeff-dude (approved):
@markusbkoch should be updated 👍

PR #1448, opened 2022-08-28
Opensea fees fix

aalan3:
I discovered that the opensea fees table references to `dbt_thomas...` so I updated it and the other references to use the refs macro.


soispoke (approved):
LGTM, thanks !

PR #1449, opened 2022-08-28
Nft performance 2

aalan3:
This PR adds more time filters on incremental updates on the nft trades tables, for the ethereum.transactions and prices.usd tables. It's a follow up of https://github.com/duneanalytics/spellbook/pull/1444

A lot of the changed rows are just whitespace, these files have previously been saved without stripping whitespace.

### DIFF
--- a/models/looksrare/ethereum/looksrare_ethereum_events.sql
+++ b/models/looksrare/ethereum/looksrare_ethereum_events.sql
@@ -163,7 +163,7 @@ INNER JOIN {{ source('ethereum','transactions') }} tx ON tx_hash = tx.hash
     AND tx.block_time > '2022-01-01'
     {% endif %}
     {% if is_incremental() %}
-    AND TRY_CAST(date_trunc('DAY', tx.block_time) AS date) = TRY_CAST(date_trunc('DAY', looks_rare.block_time) AS date)
+    AND tx.block_time >= date_trunc("day", now() - interval '1 week')
     {% endif %}
 LEFT JOIN erc_transfers erc ON erc.evt_tx_hash = tx_hash AND erc.token_id_erc = token_id
 LEFT JOIN {{ ref('tokens_ethereum_nft') }} tokens ON tokens.contract_address =  nft_contract_address
@@ -171,6 +171,9 @@ LEFT JOIN  {{ ref('nft_ethereum_aggregators') }} agg ON agg.contract_address = t
 LEFT JOIN {{ source('prices', 'usd') }} p ON p.minute = date_trunc('minute', looks_rare.block_time)
     AND p.contract_address = currency_contract
     AND p.blockchain ='ethereum'
+    {% if is_incremental() %}
+    AND p.minute >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
 LEFT JOIN {{ ref('tokens_ethereum_erc20') }} erc20 ON erc20.contract_address = currency_contract
 WHERE number_of_items >= 1
 {% if is_incremental() %}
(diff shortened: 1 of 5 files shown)


soispoke:
The fact that most changes are fixing white space makes it a bit hard to review and to spot the "important" changes you made, is there a way to deal with the white space issue separately ?

models/looksrare/ethereum/looksrare_ethereum_events.sql
```diff
@@ -163,17 +163,20 @@ INNER JOIN {{ source('ethereum','transactions') }} tx ON tx_hash = tx.hash
     AND tx.block_time > '2022-01-01'
     {% endif %}
     {% if is_incremental() %}
-    AND TRY_CAST(date_trunc('DAY', tx.block_time) AS date) = TRY_CAST(date_trunc('DAY', looks_rare.block_time) AS date)
+    AND tx.block_time >= date_trunc("day", now() - interval '1 week')
```
soispoke, line 166: I've noticed you often use the `tx.block_time >= date_trunc("day", now() - interval '1 week')` filter

I was wondering why go back one week when we could probably just use something like this instead (which is supposed to take the last timestamp for the previous materialization and not the whole week before today:
`WHERE tx.block_time >= (select max(block_time) from {{ this }})`

models/magiceden/solana/magiceden_solana_events.sql
```diff
@@ -84,18 +84,22 @@ SELECT
...
   log_messages
 FROM {{ source('solana','transactions') }}
-LEFT JOIN prices.usd p 
+LEFT JOIN prices.usd p
   ON p.minute = date_trunc('minute', block_time)
-  AND p.symbol = 'SOL'        
+  AND p.blockchain is NULL
+  AND p.symbol = 'SOL'
+  {% if is_incremental() %}
+  AND p.minute >= date_trunc("day", now() - interval '1 week')
```
soispoke, line 101: Same as previous comment here (not going to comment the same thing everywhere but same thought process)

soispoke (commented):
As a more general comment, which also applies to your other PR (I'll write it there as well):
I think it might help to add performance improvements on total build/update time when using the dbt cloud cluster (or one that has the same parameters) to evaluate how much these changes improve performance and avoid back and forth changes
Just posting the build time "before" and "after" changes would be super helpful imo

models/looksrare/ethereum/looksrare_ethereum_events.sql
aalan3: This needs to be a static value for the filter to be pushed down when fetching the files otherwise it does the filtering after doing a full scan. We could potentially make this shorter as well.

aalan3:
>The fact that most changes are fixing white space makes it a bit hard to review and to spot the "important" changes you made, is there a way to deal with the white space issue separately ?

Yeah I don't know why these files were like this. But basically search for `now() - interval '1 week')` and you should find all the changes.

aalan3:
@soispoke rebased this now so should be easier to review without all the whitespace changes.

models/magiceden/solana/magiceden_solana_events.sql
```diff
@@ -95,7 +95,11 @@ SELECT
 FROM {{ source('solana','transactions') }}
 LEFT JOIN prices.usd p
   ON p.minute = date_trunc('minute', block_time)
+  AND p.blockchain is NULL
```
aalan3, line 98: If we don't have this we can get multiple rows per line if another `SOL` is added in the future.

soispoke: Thanks ! I'll review this and run it on my end as well tmw 👍

PR #1451, opened 2022-08-28
Opensea incremental update performance fix

aalan3:
This PR speeds up the opensea spell on incremental updates. It does so by pusing down a `block_time` filter on `ethereum.traces`.  

Since the `ethereum.transactions` table is now filtered on this query, the slowest part is a full scan on `ethereum.traces`. This happens because to get the fees (via the opensea fees view), we do a query on `ethereum.traces` by `tx_hash` and `trace_address` which results in a full scan.

By `block_time` is added the fees table, we can then use a filter on we join on fees in the main query which prevents a full scan on incremental updates

### DIFF
--- a/models/opensea/ethereum/opensea_ethereum_schema.yml
+++ b/models/opensea/ethereum/opensea_ethereum_schema.yml
@@ -22,6 +22,8 @@ models:
       tags: ['ethereum','opensea','v1','fees']
     description: OpenSea v1 fees on Ethereum
     columns:
+      - name: block_time
+        description: "Block time of transaction"
       - name: trace_address
         description: "Trace address for fees"
       - name: tx_hash
--- a/models/opensea/ethereum/opensea_v1_ethereum_events.sql
+++ b/models/opensea/ethereum/opensea_v1_ethereum_events.sql
@@ -83,7 +83,13 @@ SELECT
   fees.fee_currency_symbol,
   call_trace_address
   FROM wyvern_call_data wc
-  LEFT JOIN {{ ref('opensea_v1_ethereum_fees') }} fees ON fees.tx_hash = wc.call_tx_hash AND fees.trace_address = wc.call_trace_address),
+  LEFT JOIN {{ ref('opensea_v1_ethereum_fees') }} fees ON
+    fees.tx_hash = wc.call_tx_hash
+    AND fees.trace_address = wc.call_trace_address
+    {% if is_incremental() %}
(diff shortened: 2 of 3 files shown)


soispoke:
I really like the idea, but as I mentioned in the other PR it would be nice to add metrics on performance improvements using total build/update time when using the dbt cloud cluster (or one that has the same parameters) to evaluate how much these changes improve performance and avoid back and forth changes
Just posting the build time "before" and "after" changes would be super helpful imo

soispoke (commented):
If it turns out to be increase the incremental build time drastically, we should look to apply these changes to every incremental spell

Also, I'm not 100% convinced by the now() - interval trick anymore, getting the last timestamp is more efficient imo

aalan3:
I've tested this manually in databricks, the query takes much less time and was about 3x faster. It's a bit difficult to get a exact/objective value but I also verified in the query planner that this is in fact fetching less data.

Here is some data, `execution_time_ms` is faster and `read_files_count/bytes` is much lower. 

Here's with the filter:
```
{'compilation_time_ms': 9681,
 'execution_time_ms': 154949,
 'network_sent_bytes': 54481926805,
 'photon_total_time_ms': 3605369,
 'pruned_bytes': 3637449923767,
 'pruned_files_count': 906920,
 'read_bytes': 55558331564,
 'read_cache_bytes': 10288873627,
 'read_files_count': 1337,
 'read_partitions_count': 24,
 'read_remote_bytes': 45269457937,
 'result_from_cache': False,
 'rows_produced_count': 19856187,
 'rows_read_count': 532685588,
 'spill_to_disk_bytes': 0,
 'task_total_time_ms': 4565052,
 'write_remote_bytes': 0}
 ```

Here's without
```
{'compilation_time_ms': 10064,
 'execution_time_ms': 599445,
 'network_sent_bytes': 60067074375,
 'photon_total_time_ms': 13210993,
 'pruned_bytes': 2818219947821,
 'pruned_files_count': 897467,
 'read_bytes': 489181671320,
 'read_cache_bytes': 58862024495,
 'read_files_count': 6155,
 'read_partitions_count': 2603,
 'read_remote_bytes': 430319646825,
 'result_from_cache': False,
 'rows_produced_count': 19856187,
 'rows_read_count': 7111808534,
 'spill_to_disk_bytes': 0,
 'task_total_time_ms': 14933759,
 'write_remote_bytes': 0}
 ```

dot2dotseurat (approved):

PR #1452, opened 2022-08-29
add safemoon token to prices

RockandGrohl:

*For Dune Engine V2*
I've checked that:

* [X ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ X] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ X] if adding a new model, I added a test
* [ X] the filename is unique and ends with .sql
* [X ] each sql file is a select statement and has only one view, table or function defined  
* [X ] column names are `lowercase_snake_cased`

### DIFF
--- a/packages.yml
+++ b/packages.yml
@@ -1,0 +1,5 @@
+- name: sfm-safemoon
+  id: sfm-safemoon    # the id from above
+  symbol: SFM 				      # the asset ticker
+  address: 0x42981d0bfbaf196529376ee702f2a9eb9092fcb5       # the contract address
+  decimals: 9              # the number of decimals the token contract uses


jeff-dude:
@RockandGrohl thank you for your first PR!

to add a new price, you want to edit this file (can delete the new one you created): https://github.com/duneanalytics/spellbook/blob/main/models/prices/prices_tokens.sql

you can see the API result here which the prices process will grab info from: https://api.coinpaprika.com/v1/coins/sfm-safemoon

jeff-dude:
everything within looks correct, you could rename the 'name' value to be 'SafeMoon' to match here: https://bscscan.com/token/0x42981d0bfbaf196529376ee702f2a9eb9092fcb5

that's more a nice-to-have, but may be worth it

RockandGrohl:
Hi J,

I'm such a novice, I can't tell if there's more I need to do, or if it's done, or even how to check if it's done.

I think it's really hard as a complete newb to GitHub to be expected to do this. Perhaps a form submission on Dune.com would be simpler?

When you say things like Fork the repo or delete your fork from the spellbook or send a push request, it just means nothing to me as you may as well be speaking another language.

Sorry to be a pest on this!
________________________________

PR #1453, opened 2022-08-29
Modify sources.yml files to avoid source freshness errors

soispoke:

*For Dune Engine V2*
I've checked that:

### DIFF
--- a/models/opensea/ethereum/opensea_ethereum_sources.yml
+++ b/models/opensea/ethereum/opensea_ethereum_sources.yml
@@ -2,9 +2,6 @@ version: 2
 
 sources: 
   - name: opensea_ethereum
-    freshness:
-      warn_after: { count: 12, period: hour }
-      error_after: { count: 24, period: hour }
     tables:
       - name: wyvernexchange_call_atomicmatch_
         loaded_at_field: call_block_time
--- a/models/nomad/ethereum/nomad_ethereum_sources.yml
+++ b/models/nomad/ethereum/nomad_ethereum_sources.yml
@@ -2,9 +2,6 @@ version: 2
 
 sources: 
   - name: nomad_ethereum
-    freshness:
-      warn_after: { count: 12, period: hour }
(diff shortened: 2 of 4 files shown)


soispoke:
@duneanalytics/team-data-experience let me know if anyone has time to review/approve this one :)

models/opensea/ethereum/opensea_ethereum_sources.yml
```diff
@@ -2,9 +2,6 @@ version: 2
 
 sources: 
   - name: opensea_ethereum
-    freshness:
```
aalan3, line 5: Can you explain why this is dropped? I'm assuming you want to look at the actual tables which are below, but why was this there in the first place in that case?

dot2dotseurat: Are we just removing this because it runs too slow? I would say we just don't run the check source freshness test in dbt cloud rather than remove it from the project. that makes it easier to turn the tests back on when we resolve the performance issues. n

dot2dotseurat: Oh I just saw the height task. So these are just tables that aren't supposed to be refreshed daily?

soispoke: Should've been more specific: I'm dropping this (which runs the same freshness test for all sources) by the same freshness test in the same file but only applied when it makes sense for specific tables. Not removing this because it's too slow, removing it on ALL sources tables and applying it on SPECIFIC tables because for example new uniswap v1 pools are juste not created every 12/24 hours so it makes sense to exclude it from freshness tests.

The opensea one was added in the first place because back then the wyvern contract was active and refreshing all the time (but is now replaced by their new seaport contract)

@aalan3 @dot2dotseurat

soispoke: 👍 exactly @dot2dotseurat

aalan3 (approved):

PR #1454, opened 2022-08-29
Dex trades performance

aalan3:
This is a follow up on https://github.com/duneanalytics/spellbook/pull/1444 and https://github.com/duneanalytics/spellbook/pull/1449.

This PR adds the date filters to the uniswap queries. 

It also adds the static dates for the first uniswap transactions in each version, for when we are not doing an incremental update. Before this was using `min(block_time)` from the `dexs` query. It's a value that doesn't change and hardcoding it will give some performance gain when redoing the whole table.


soispoke (dismissed):
Tagging @jeff-dude as well for visibility

jeff-dude (approved):
makes sense 👍

PR #1456, opened 2022-08-29
feat(prices): add tbc, sps, and dec token on bnb chain

yashtbc:
adding tbc, sps and dec token to prices file


models/prices/prices_tokens.sql
```diff
@@ -167,6 +167,10 @@ VALUES
     ("znn-zenon", "bnb", "Znn", "0x84b174628911896a3b87fa6980d05dbc2ee74836", 8),
     ("azy-amazy", "bnb", "AZY", "0x7b665B2F633d9363b89A98b094B1F9E732Bd8F86", 18),
     ("amt-amazy", "bnb", "AMT", "0xf625069dce62df95b4910f83446954b871f0fc4f", 18),
+    ("tbc-terablock", "bnb", "TBC", "0x9798dF2f5d213a872c787bD03b2b91F54D0D04A1", 18),
```
aalan3, line 170: Can you please lowercase the address here and below.

aalan3 (changes requested):

models/prices/prices_tokens.sql
yashtbc: done, thanks!

aalan3 (approved):

PR #1460, opened 2022-08-30
Tornado Cash: Fix sources, lowercase and improve materialization strategy

soispoke:

*For Dune Engine V2*
I've checked that:


models/tornado_cash/tornado_cash_deposits.sql
```diff
@@ -35,13 +35,13 @@ FROM
                 AND et.block_time >= (select min(evt_block_time) from {{ source('tornado_cash_ethereum','eth_evt_Deposit') }})
                 {% endif %}
                 {% if is_incremental() %}
-                AND et.block_time >= (select max(block_time) from {{ this }})
+                AND et.block_time >= date_trunc("day", now() - interval '1 week')
```
aalan3, line 38: Shouldn't the `LEFT JOIN` above be an `INNER JOIN` instead? And then you could drop the where filter below.

models/tornado_cash/tornado_cash_deposits.sql
```diff
@@ -35,13 +35,13 @@ FROM
                 AND et.block_time >= (select min(evt_block_time) from {{ source('tornado_cash_ethereum','eth_evt_Deposit') }})
```
aalan3, line 35: Can't comment on the actual line but line 32  `source('ethereum','transactions_0006'`) should be ` source('ethereum','transactions')`.

jeff-dude: i believe that source was used specifically for performance reasons, but agreed to revert back if we're comfortable with new performance enhancement changes (could then also remove from source file)

models/tornado_cash/tornado_cash_deposits.sql
jeff-dude: added this common check to review doc, but many existing queries from v1 will have left joins to look out for

aalan3: There are cases where you'd want to do a left join as I discovered [here](https://github.com/duneanalytics/spellbook/pull/1465/files#diff-5bf74bdf14e57e88e2a0385fed61e5f0a5ba0d554f05f56bf255e413ca9eec7aR86). But generally it seems like it should be inner.

jeff-dude:
lgtm after changes applied from alan feedback

hildobby:
Discussed this in DMs with @jeff-dude, imo naming should be consistent across the board and since `avalanche_c` is avalanche's default blockchain name, all abstractions should also stick with it for consistency, don't you guys think?

PS: Apologies for not changing the names myself when doing the first tornado cash PR, I will make sure to be careful on names being lowercase going forward!

soispoke (commented):
mades some small changes regarding lowercasing blockchains and moving from avalanche to avalanche_c for consistency

macros/alter_table_properties.sql
```diff
@@ -560,17 +560,17 @@ ALTER TABLE airdrop_optimism.addresses SET TBLPROPERTIES ('dune.public'='true',
 {% endset %}
 
 {% set tornado_cash_deposits %}
-ALTER TABLE tornado_cash.deposits SET TBLPROPERTIES ('dune.public'='true',
-                                                'dune.data_explorer.blockchains'='["ethereum", "bnb", "avalanche", "gnosis", "optimism", "arbitrum"]',
+ALTER VIEW tornado_cash.deposits SET TBLPROPERTIES ('dune.public'='true',
```
jeff-dude, line 563: shouldn't these both be 'table'?

models/tornado_cash/tornado_cash_deposits.sql
jeff-dude: @soispoke can you update this?

jeff-dude (changes requested):

jeff-dude:
to @aalan3 point, should we use this PR to update all joins to base tables as inner join?

models/tornado_cash/tornado_cash_deposits.sql
soispoke: @jeff-dude Done

macros/alter_table_properties.sql
soispoke: You're right they should, pushed a fix

soispoke (commented):
@jeff-dude Done, updated all joins with base tables

aalan3 (approved):

soispoke:
Thanks !

macros/alter_table_properties.sql
```diff
@@ -579,16 +579,16 @@ ALTER TABLE airdrop_optimism.addresses SET TBLPROPERTIES ('dune.public'='true',
...
                                                 'dune.data_explorer.category'='abstraction',
                                                 'dune.data_explorer.abstraction.type'='project',
                                                 'dune.data_explorer.abstraction.name'='tornado_cash',
                                                 'dune.data_explorer.contributors'='["hildobby", "dot2dotseurat"]');
 {% endset %}
 
 {% set tornado_cash_withdrawals %}
-ALTER TABLE tornado_cash.withdrawals SET TBLPROPERTIES ('dune.public'='true',
-                                                'dune.data_explorer.blockchains'='["ethereum", "bnb", "avalanche", "gnosis", "optimism", "arbitrum"]',
+ALTER VIEW tornado_cash.withdrawals SET TBLPROPERTIES ('dune.public'='true',
```
jeff-dude, line 590: @soispoke looks like this one didn't get updated back to 'table'

aalan3: Was fixed here https://github.com/duneanalytics/spellbook/pull/1475/files#diff-89d8392c4db5b996c368bef69b027f7ccd442b9135924630007065e3a1d5569cR590

PR #1470, opened 2022-08-31
modify seaport column to avoid integer overflow (Dune v1)

sohwak:

(Dune v1) Currently, an integer overflow is occurring in the seaport table on Polygon so altering columns datatype is needed.

*For Dune Engine V2*
I've checked that:


sohwak:
example txn:
https://polygonscan.com/tx/0x62d208e2c1c3419cc0d4e31147610efda2a2301283ce0d4545c69d12fc8d8f60

jeff-dude (approved):
this should be updated. i can try to monitor today.

sohwak:
Thank you for your prompt support. Everything seems to work well now.

PR #1483, opened 2022-09-02
T 15666 fix dbt source snapshot freshness warn instead of error

soispoke:
This PR aims to use the source freshness utility better. Since the DX team doesn't ultimately errors/delays on base and decoded tables, I propose to have the source freshness command attached to our daily job to check if everything looks good, but instead of throwing errors (that ultimately end up in failed jobs), throwing warnings so we can alert the data platform team without having to worry about failed jobs on our end

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:

### DIFF
--- a/models/tornado_cash/tornado_cash_sources.yml
+++ b/models/tornado_cash/tornado_cash_sources.yml
@@ -3,7 +3,7 @@ version: 2
 sources:
   - name: tornado_cash_bnb
     freshness:
-      warn_after: { count: 12, period: hour }
+      warn_after: { count: 7, period: day }
     tables:
       - name: TornadoCashBNB_evt_Deposit
         loaded_at_field: evt_block_time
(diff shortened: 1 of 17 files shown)


models/tornado_cash/tornado_cash_sources.yml
```diff
@@ -3,7 +3,7 @@ version: 2
 sources:
   - name: tornado_cash_bnb
     freshness:
-      warn_after: { count: 12, period: hour }
+      warn_after: { count: 7, period: day }
```
aalan3, line 6: Why 7 days here and 12h below? I think these should also be 12h if it's not too much

aalan3: oh lol okay then it's fine I guess

aalan3 (approved):

soispoke:
thanks @aalan3

PR #1486, opened 2022-09-02
wizard test schema and tests

dot2dotseurat:
This purpose of this PR is to publish a github workflow that runs on workflow dispatch and upon merge that will run the modfied dbt models and tests compared to the production run. 

This workflow is protected by a codeowners file and concurrent runs per branch are not allowed. the workflow times out at 60 minutes. 

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


.github/CODEOWNERS
```diff
@@ -0,0 +1,6 @@
+# Each line is a file pattern followed by one or more owners.
```
a-monteiro, line 1: We don't need this file unless we want to get tagged if someone tries to change this, it won't actually prevent changes.

.github/protected-workflows.yml
```diff
@@ -0,0 +1,30 @@
+# "events" is a map between Github events and rules.
```
a-monteiro, line 1: This file isn't required

.github/workflows/dbt_slim_ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    timeout-minutes: 60
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+
+      - name: Add git_sha to schema
+        env:
+          GIT_SHA: ${{ github.ref }}
+        run: sed -ie 's!^\(\s*schema:\)[^.]*!\1 git_$GIT_SHA!' profiles.yml
```
a-monteiro, line 26:
```suggestion
        run: sed -ie 's!^\(\s*schema:\)[^.]*!\1 git_$GIT_SHA!' profiles.yml
```
        run: "/runner/change_schema.sh git_$GIT_SHA"

.github/workflows/dbt_slim_ci.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  dbt-test:
+    runs-on: [ self-hosted, linux, spellbook ]
+    timeout-minutes: 60
+
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v2
+
```
a-monteiro, line 22: - name: Set environment variables
```suggestion
      - name: Set environment variables
        run: |
          echo "GIT_SHA=${GITHUB_SHA::7}" >> $GITHUB_ENV
```
I'm not sure if this works 😅

dsalv (approved):

PR #1503, opened 2022-09-05
Transfers optimize

aalan3:
This PR should bring transfers time filtering to the same speed as the rest of the tables.

The reason why it doesn't work atm is because there's a `date_trunc` in the `where`. We don't need to truncate the date there, we can just pass the value.

Thanks to @jkylling for pointing this out and solving this!


soispoke (commented):
Oh interesting, thanks @aalan3 and @jkylling !

soispoke (approved):

soispoke:
Can this be applied to erc721 and erc1155 transfers as well ?

aalan3:
>Can this be applied to erc721 and erc1155 transfers as well ?

It can but those tables are not yet z-ordered but we can still change the queries. I can follow up with PR after this one.

PR #1508, opened 2022-09-05
Correctly trunc date for transfers

aalan3:
I may have made a mistake in the previous PR. Without truncating the date, we risk losing some events that happened in between the truncated date and when the query is running. This is also a problem because the `merge into` updates records, so currently rows from a week ago might be updated when they shouldn't. 

We should do either do a full refresh of transfers after this is deployed or delete data from today to 7 days ago and refill.


soispoke:
lgtm 👍

soispoke (approved):

PR #1509, opened 2022-09-05
Enable `ovm1` data UI

chuxinh:
Add `ovm1` to explorer UI

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


chuxinh:
Also would like to know if `airdrop_optimism.address` can also be updated? I think it's referencing the csv file name or instead i can update what the alter table marco is calling? 
We'd like to have it changed to `op_airdrop_1`

soispoke (commented):
Hi @chuxinh, thanks for your submission. It seems like you're are trying to expose tables that don't exist on Dune v2 ?

(if if does exist and I missed it, could you send me the links of where the table are in the spellbook github repo please ? https://github.com/duneanalytics/spellbook)

chuxinh:
> Hi @chuxinh, thanks for your submission. It seems like you're are trying to expose tables that don't exist on Dune v2 ?
> 
> (if if does exist and I missed it, could you send me the links of where the table are in the spellbook github repo please ? https://github.com/duneanalytics/spellbook)

Thanks @soispoke! 
So those `ovm1` data have been loaded as csv as i'm aware of and I don't have any link to them. So it's very similar to what the team has done in this airdrop address list that you did: https://github.com/duneanalytics/spellbook/pull/1420

soispoke:
Oh alright, let me get back to you quickly on these ones then, I think we need to find a better way to have these tables deployed/exposed somehow.

soispoke:
@chuxinh I think one good way to do this would be to turn these .csv files into static tables that can be seen, modified and exposed easily on spellbook. 

A good example of this format is tokens here: https://github.com/duneanalytics/spellbook/blob/main/models/tokens/ethereum/tokens_ethereum_erc20.sql

What do you think ?

chuxinh:
> @chuxinh I think one good way to do this would be to turn these .csv files into static tables that can be seen, modified and exposed easily on spellbook.
> 
> A good example of this format is tokens here: https://github.com/duneanalytics/spellbook/blob/main/models/tokens/ethereum/tokens_ethereum_erc20.sql
> 
> What do you think ?

Thanks for getting back on this! Yea originally Jeff and Antonio helped on this and uploaded as CSVs. 
Is it possible that we can get some help on your side to reorganise the ovm1 existing schema and those tables?

soispoke:
Yes ofc, what would be super useful is to transform those csv files into static SQL tables as I mentioned (in this PR for example), and then I can polish everything up, make sure schemas are right and tables are exposed properly. But since I don't have access to the original files rn, I still need you to reformat/resubmit those files as tables if that's ok?

chuxinh:
> Yes ofc, what would be super useful is to transform those csv files into static SQL tables as I mentioned (in this PR for example), and then I can polish everything up, make sure schemas are right and tables are exposed properly. But since I don't have access to the original files rn, I still need you to reformat/resubmit those files as tables if that's ok?

Thank you!! 
So all the original files are resting here: https://github.com/duneanalytics/spellbook/tree/main/deprecated-dune-v1-abstractions/optimism2/ovm1 
So i guess it's better to have them loaded as `.sql` files as the example you showed. However, i think i also remember Meghan has mentioned about the limitations on github storage (or something similar that i'm not using the right word for). Is that still a concern or I can proceed

soispoke:
Oh thanks for pointing out where the original files are. Ok yeah I remember something like this. Meghan is OOO but how about we try to go through the process for 1 file so you're not spending too much time on this (maybe with the biggest file ?), and if everything goes well we just do the same for all the other ones.

chuxinh:
Right, I will leave this one open for now and create a separate one and see how it goes
Will do that my time tomorrow :D

PR #1512, opened 2022-09-06
Add static cex labels 

soispoke:
This PR is meant to give an example of static labels added to the spellbook repo for CEX on ethereum.

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -102,6 +102,10 @@ models:
       ethereum:
         +schema: balances_ethereum
         +materialized: view
+  
+    labels:
+      +schema: labels
+      +materialized: view
 
     tokens:
       +schema: tokens
--- a/models/labels/cex/labels_cex_ethereum.sql
+++ b/models/labels/cex/labels_cex_ethereum.sql
@@ -1,0 +1,496 @@
+{{config(alias='cex_ethereum')}}
+
+SELECT blockchain, address, name, category, contributor, source, created_at, updated_at
+FROM (VALUES
+    -- Binance, Source: https://etherscan.io/accounts/label/binance
+    (array('ethereum'),'0x3f5ce5fbfe3e9af3971dd833d26ba9b5c936f0be', 'Binance 1', 'cex', 'hildobby', 'static', timestamp('2022-08-28'), now())
+    , (array('ethereum'), '0xd551234ae421e3bcba99a0da6d736074f22192ff', 'Binance 2', 'cex', 'hildobby', 'static', timestamp('2022-08-28'), now())
+    , (array('ethereum'), '0x564286362092d8e7936f0549571a803b203aaced', 'Binance 3', 'cex', 'hildobby', 'static', timestamp('2022-08-28'), now())
+    , (array('ethereum'), '0x0681d8db095565fe8a346fa0277bffde9c0edbbf', 'Binance 4', 'cex', 'hildobby', 'static', timestamp('2022-08-28'), now())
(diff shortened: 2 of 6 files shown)


models/labels/cex/labels_cex_ethereum.sql
```diff
@@ -0,0 +1,496 @@
+{{config(alias='cex_ethereum')}}
+
+SELECT blockchain, address, name, category, contributor, source, created_at, updated_at
+FROM (VALUES
+    -- Binance, Source: https://etherscan.io/accounts/label/binance
+    (array('ethereum'),'0x3f5ce5fbfe3e9af3971dd833d26ba9b5c936f0be', 'Binance 1', 'cex', 'hildobby', 'static', timestamp('2022-08-28'), now())
```
aalan3, line 6: shouldn't `updated_at` be the same as `created_at`

dbt_project.yml
```diff
@@ -102,6 +102,10 @@ models:
       ethereum:
         +schema: balances_ethereum
         +materialized: view
+  
+    labels:
+      +schema: labels
+      +materialized: view
```
jeff-dude, line 108: if you leave out the subdirectories (i.e. ethereum) here, will all schemas in the 'labels' directory default to 'labels'?

models/labels/cex/labels_cex_ethereum.sql
jeff-dude: i could see it either way, but if they're the same, then we should just have ```updated_at``` only and leave out ```created_at```

jeff-dude, line 6: if this model is specific to ethereum, do we need to allow an array value for ```blockchain``` since it'll always be ethereum to match the model?

models/labels/labels_all.sql
```diff
@@ -0,0 +1,8 @@
+{{ config(
+    alias = 'all',
+    materialized = 'table',
+    file_format = 'delta')
+}}
+
+-- Static Labels
+SELECT * FROM {{ ref('labels_cex') }}
```
jeff-dude, line 8: what are the future use cases of adding labels here, will they have the same columns and be able to just ```select *``` union?

models/labels/cex/labels_cex_ethereum.sql
jeff-dude: just noticed #1513 which i think clarifies my question here. while this PR may not use it, others will in the other PR, then later combined into one object

models/labels/labels_all.sql
jeff-dude: i think #1513 may answer this as well, looks like you'll always conform to that standard

models/labels/cex/labels_cex_ethereum.sql
soispoke: The idea here is to have created_at reflect the moment at which contributors create the query/static labels for the first time, while updated_at reflects the last date at which the table was updated/rebuilt

soispoke: Yes exactly, for some more examples you can look at # 1474 as well :)

models/labels/labels_all.sql
soispoke: yes, that's also right 👍

soispoke (commented):
Yes that's right 👍

soispoke (commented):
thanks @aalan3 and @jeff-dude for taking the time to review this

jeff-dude (approved):
i think i'm good with this 👍

soispoke:
thanks @jeff-dude !

PR #1516, opened 2022-09-06
Test add optimism airdrop addresses 

soispoke:

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -28,6 +28,12 @@ models:
     ens:
       +schema: ens
       +materialized: view
+    airdrop:
+      +schema: airdrop
+      +materialized: view
+      optimism:
+        +schema: airdrop_optimism
+        +materialized: view
     dex:
       +schema: dex
       +materialized: view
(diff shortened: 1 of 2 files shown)


dbt_project.yml
```diff
@@ -28,6 +28,12 @@ models:
     ens:
       +schema: ens
       +materialized: view
+    airdrop:
+      +schema: airdrop
+      +materialized: view
```
aalan3, line 33: This and below should be `table`?

aalan3 (approved):
One question otherwise lgtm

dbt_project.yml
soispoke: @aalan3 oh right, I just enforced it directly in the model but you're right, it's safer to have it by default here too

soispoke: all good now

soispoke:
@chuxinh @MSilb7 In this PR, I experimented with the airdrop optimism addresses table to improve submissions of big/static tables into V2. It turns out materializing it as tables and following the format I used here works well. @chuxinh could you use the same format for the other tables you wanted to add here #1509 ? @MSilb7 Can you check and lmk if I can delete the old airdrop table and keep this one instead ?

PR #1517, opened 2022-09-06
remove uniqueness test on hash for optimism transactions

soispoke:

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:

### DIFF
--- a/models/base_sources/optimism_base_sources.yml
+++ b/models/base_sources/optimism_base_sources.yml
@@ -10,7 +10,7 @@ sources:
       - name: transactions
         tests:
           - unique:
-              column_name: "(tx_hash || '-' || hash || '-' || block_number)"
+              column_name: "(hash || '-' || block_number)"
         loaded_at_field: block_time
         description: "An Optimism transaction refers to an action initiated by an externally-owned account (i.e., an account managed by a human, not a contract)."
         columns:
--- a/models/base_sources/optimism_base_sources.yml
+++ b/models/base_sources/optimism_base_sources.yml
@@ -8,6 +8,9 @@ sources:
       warn_after: { count: 12, period: hour }
     tables:
       - name: transactions
+        tests:
+          - unique:
+              column_name: "(tx_hash || '-' || hash || '-' || block_number)"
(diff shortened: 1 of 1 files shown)


chuxinh:
Thanks for this! 
I think it'd be better instead of dropping unique_ids unique test, we should instead update the unique id to `tx_hash` and `block number` instead. That also means the table needs a rebuild if necessary?

soispoke (commented):
@chuxin we wouldn't need to rebuild the table, we could change the test to something similar to the commit I last pushed. However, the trace_address doesn't seem to be a column in the transactions table ? (so we can't use it as a unique id)

chuxinh:
> @chuxin we wouldn't need to rebuild the table, we could change the test to something similar to the commit I last pushed. However, the trace_address doesn't seem to be a column in the transactions table ? (so we can't use it as a unique id)

ohhh i got confused here. thought you were referring to `transfer_optimism_eth`. yea that looks good to me!

chuxinh:
is this going to re-enable of incremental build for `transfers_optimism.eth` after merge?
it's been paused since Aug 27

soispoke:
@chuxinh ok great, will merge today then 👍

And unfortunately I think the pause on optimism is due to other factors not related to this particular issue

chuxinh:
just wonder is there's any error or issue that's blocking the incremental build of `transfers_optimism.eth`?

soispoke:
Opened a PR here #1522 to improve the incremental filters, let's also wait for @aalan3 or @jeff-dude's inputs to make sure there aren't any other issues with that table

jeff-dude:
updated the unique test to remove ```tx_hash``` which doesn't exist on base table, results pass

jeff-dude:
> @chuxinh Opened a PR here #1522 to improve the incremental filters, let's also wait for @aalan3 or @jeff-dude's inputs to make sure there aren't any other issues with that table

this PR looks good (and it's merged already), plus dbt shows that this model has been refreshing today 👍

jeff-dude (approved):

PR #1523, opened 2022-09-07
Add cryptopunks to engine v2 + extend generic test framework

masquot:
### Adding a Generic Test Framework

This code builds upon the work of @jeff-dude who introduced generic tests in the context of spellbook. This PR continues his work by adding a generic test that can be applied in many contexts: `tests/generic/compare_column_values_to_seed_values.sql`. This results in the following process:

#### Steps to add unit tests to a model
1. Create a CSV seed file with sample rows specific to the project you are integrating into spellbook.
E.g. `seeds/cryptopunks/cryptopunks_trades_buys_bids_seed.csv`
Every row in this file should contain expected values that you get from a trusted place such as a project's reporting. In this case, we used the [CryptoPunks project's website detail pages](https://cryptopunks.app/cryptopunks/details/5293). It's important here to **make sure that the column names in the seed file match those in the model**.

2. Add the test in the schema.yml file for the model
E.g. `models/cryptopunks/ethereum/cryptopunks_ethereum_schema.yml`
We added the following lines to test if `seller` values from seed and model match up:

```
- &seller
  name: seller
  description: "Seller wallet address"
  tests:
    - compare_column_values_to_seed_values:
        seed_file_location: ref('cryptopunks_trades_buys_bids_seed')
```

### Why is this PR WIP?

The transformation logic in [`cryptopunks_ethereum_events.sql`](./models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql) needs a little more work review. I does not handle certain edge cases such as sales initiated by aggregators / vaults.

### General checks:

Join logic:

Incremental logic:


models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql
```diff
@@ -0,0 +1,124 @@
...
+    tx.from as tx_from,
+    tx.to as tx_to,
+    "0" AS platform_fee_amount_raw,
+    "0" AS platform_fee_amount,
+    "0" AS platform_fee_amount_usd,
+    "0" AS platform_fee_percentage,
+    "0" AS royalty_fee_amount_raw,
+    "0" AS royalty_fee_amount,
+    "0" AS royalty_fee_amount_usd,
+    "0"  AS royalty_fee_percentage,
```
hildobby, line 110: Why the `"0"`s here and not `0`s? I believe that the other marketplace abstractions use ints not strings

models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql
```diff
@@ -0,0 +1,124 @@
...
+        NULL,
+        NULL AS `from`,
+        fromAddress AS `bid_from`,
+        evt_block_number,
+        evt_index,
+        evt_block_time,
+        evt_tx_hash,
+        contract_address
+      FROM
+        {{ source('cryptopunks_ethereum','CryptoPunksMarket_evt_PunkBidEntered') }}
```
soispoke, line 73: We might want to add incremental filters on decoded tables as well here: `evt_block_time >= date_trunc("day", now() - interval '1 week')`

models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql
```diff
@@ -0,0 +1,124 @@
...
+    "0" AS platform_fee_percentage,
+    "0" AS royalty_fee_amount_raw,
+    "0" AS royalty_fee_amount,
+    "0" AS royalty_fee_amount_usd,
+    "0"  AS royalty_fee_percentage,
+    NULL::string as royalty_fee_receive_address,
+    NULL::string as royalty_fee_currency_symbol,
+    "cryptopunks" || '-' || buys.evt_tx_hash || '-' || buys.punkIndex || '-' ||  buys.from || '-' || buys.evt_index || '-' || "" as unique_trade_id
+FROM buys
+INNER JOIN {{ source('ethereum','transactions') }} tx ON buys.evt_tx_hash = tx.hash
```
soispoke, line 115: We probably want to filter on transactions since it's such a big table, so here we should add: 

`AND tx.block_time >= date_trunc("day", now() - interval '1 week')`

models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql
```diff
@@ -0,0 +1,124 @@
...
+LEFT JOIN {{ ref('nft_ethereum_aggregators') }} agg ON agg.contract_address = tx.to
+LEFT JOIN {{ source('prices', 'usd') }} p ON p.minute = date_trunc('minute', buys.evt_block_time)
+    AND p.contract_address = "0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2"
```
soispoke, line 118: add     `{% if is_incremental() %}
    AND p.minute >= date_trunc("day", now() - interval '1 week')
    {% endif %}`

models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql
```diff
@@ -0,0 +1,124 @@
...
+    buys.punkIndex as token_id,
+    "CryptoPunks" AS `collection`,
+    buys.enhanced_buy_value / power(10,18) * p.price AS amount_usd,
+    "erc20" AS token_standard,
+    NULL::string AS trade_type,
+    NULL::string AS number_of_items,
+    buys.trade_category,
+    buys.from AS seller,
+    buys.`enhanced_to` AS buyer,
+    "Trade" as evt_type,
```
soispoke, line 90: Since all evt_types in this spell are actually trades, there are two options:
Either rename this spell a cryptopunks_ethereum_trades.sql

Or, if we want to keep the possibility of having different types of events for cryptopunks, we should build the cryptopunks_ethereum_trades model anyways, using something like 
here: https://github.com/duneanalytics/spellbook/blob/main/models/looksrare/ethereum/looksrare_ethereum_trades.sql

soispoke (changes requested):
See inline comments

models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql
masquot: The outer function contains the lag() window function which requires all rows to be present to get accurate results. Not sure if other sql coding patterns are possible, but in the current setup incremental updates here seem not possible. Not a big issue here as the number of events is limited.

models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql
masquot: 👍 integrating trades model in next commit

models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql
masquot: ☑️

masquot:
Thank you for your reviews @hildobby @soispoke ! 👍

Ready for merging as far as I am concerned. Unit tests passing locally. Only thing not tested locally is final integration into NFT trades.

hildobby:
Just made a change to make sure the buyer wasn't an aggregator address. Also, there are still a lot of empty buyer, don't have time to look into it today but might be worth fixing before merging 🤔

hildobby:
Oh one final thing: I think that all royalties/marketplace fees should be 0 instead of nulls (0 as an int, not '0'). This would make things clearer for the average `nft.trades` user who could otherwise think that the fees are missing.

What do you think?

masquot:
Thank you for the update to `buyer` @hildobby. Agree on what you say about fees. Updated in new commit.

I know there are users waiting for CryptoPunks to be added to v2. Therefore, I suggest to merge now and make the dataset available. I made a note to look into fine-tuning the code including buyer values after DuneCon.

soispoke (approved):
lgtm guys :)

PR #1525, opened 2022-09-07
WIP T 16269 broken erc721 balances tables

soispoke:
This PR aims to improve/fix erc721 balances on ethereum.
Changes have been added (mainly fix the sign for amounts, and the send/receive fields + some improved incremental filters), but there are still issues with duplicate owners that need to be addressed, see https://dune.com/queries/1246260

The test was also changed (and is currently failing), so that wallet address is not part of it (therefore, it is now actually testing for unique owners)

@hildobby @0xRobin @aalan3 

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


0xRobin:
@soispoke, I fixed some grouping and partitioning issues I found. (see PR mentioned above)
I believe query should now return correct results but I was not able to fully test it.

0xBoxer (approved):
Seems good to me, this should make sure that every NFT (combination of address and id) only exists once in the balances table for any given day or hour.

soispoke:
thanks @0xBoxer

0xBoxer:
@soispoke @0xRobin 

I found a edge case where the script goes wrong:

https://dune.com/queries/1079379


https://etherscan.io/token/0xc7457f480416e145093edb355f626faa0fc7f90e?a=2116


This is a weird contract that minted a token_id twice. 
The currently running script assumes that the first owner is the correct one and actually changes the balance of the first owner on a transfer 20 days later, despite that owner not even having an outgoing transfer for this token.

The second minter is the correct one and sells his NFT a few minutes after minting, this tx is completely missing from our results. Therefore, in the period between 01/05 to 20/05 this token has wrong owner data.

This finding makes me question what usually happens in the script if a token gets traded multiple times within a day, because that seems to be what's going wrong here.

0xRobin:
@0xBoxer Good catch!
This is actually expected behaviour from the current query.
When 2 addresses have positive delta for that day (meaning more incoming then outgoing transfers for that ID) the query itself produces 2 records.
The deduplication happens later on the table merge cause those entries will have identical unique keys. So I guess it's currently quite random which owner will be chosen as we don't have strong ordering guarantees within the source table partitions.

We could update the query to include the deduplication and favour the entry which had the last incoming transfer. Another option might be to  provide ordered output if that guarantees us the deduplication by the merge will choose the correct owner. (Will need to look into spark/databricks docs to confirm)

PR #1533, opened 2022-09-08
Remove freshness block as it's not needed 

agaperste:
Remove freshness block as it's not needed per discussion in https://dune.height.app/T-15666


soispoke (approved):

soispoke (commented):
lgtm, thanks @agaperste !

PR #1534, opened 2022-09-08
add wallet_address to unique_id

soispoke:

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:

### DIFF
--- a/models/transfers/ethereum/erc721/transfers_ethereum_erc721_agg_day.sql
+++ b/models/transfers/ethereum/erc721/transfers_ethereum_erc721_agg_day.sql
@@ -13,7 +13,7 @@ select
     wallet_address,
     token_address,
     tokenId,
-    date_trunc('day', evt_block_time) || '-' || token_address || '-' || tokenId as unique_transfer_id
+    wallet_address || '-' || date_trunc('day', evt_block_time) || '-' || token_address || '-' || tokenId as unique_transfer_id
 from {{ ref('transfers_ethereum_erc721') }}
 {% if is_incremental() %}
 -- this filter will only be applied on an incremental run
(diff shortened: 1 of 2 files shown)


jeff-dude (approved):

soispoke (commented):
thanks @jeff-dude !

0xRobin:
@soispoke @jeff-dude I don't think this is the correct solution to the problem here as this will again introduce multiple reported owners for 1 ID in case of duplicate mints in the same hour/day.
Does the no-duplicate-owners test still pass under this build?
See my comments here: https://github.com/duneanalytics/spellbook/pull/1525#issuecomment-1241290341

In essence the breaking smart contracts (with double mints) are violating ERC721 standards and I believe we should correct their mistake as early as possible in the data flow instead of propagating it.

soispoke:
@0xRobin Yes, tests are passing. This fix was pushed because otherwise the incremental tables (agg_hour and day) couldn't be updated without throwing an error because the unique_id wasn't actually unique.

0xRobin:
> @0xRobin Yes, tests are passing. This fix was pushed because otherwise the incremental tables (agg_hour and day) couldn't be updated without throwing an error because the unique_id wasn't actually unique.

Ok so keys can be duplicates over different incremental runs but aren't automatically deduplicated within 1 query run. Gotcha!
In that case we should deduplicate in the query.

0xRobin:
as you can see here:
https://dune.com/queries/1253709?d=1

this results in 2 entries and we have **no guarantees** which one makes it into the `balances.erc721_day` view

0xBoxer:
I think this particular contract had this issue when 2 token_id's were minted in the same block, we could sort based on evt_index and keep the higher one

PR #1535, opened 2022-09-08
Add gnosis safe airdrop address data

springzh:

Added value based data table for gnosis safe airdrop data. Source file downloaded from https://raw.githubusercontent.com/safe-global/safe-user-allocation-reports/main/safe_user_allocations_reworked.csv 

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


soispoke (changes requested):
Thanks for your submission @springzh ! 

Could you please change reformat this query so the model is addresses_ethereum.safe_airdrop so we can standardize airdrops across different chains ? 
https://github.com/duneanalytics/spellbook/pull/1529?no-redirect=1

macros/alter_table_properties.sql
```diff
@@ -415,6 +415,15 @@ ALTER TABLE airdrop_optimism.addresses_1 SET TBLPROPERTIES('dune.public'='true',
                                                     'dune.data_explorer.contributors'='["soispoke"]');
 {% endset %}
 
+{% set addresses_ethereum_safe_airdrop %}
+ALTER TABLE addresses_ethereum.safe_airdrop SET TBLPROPERTIES('dune.public'='true',
+                                                    'dune.data_explorer.blockchains'='["ethereum"]',
+                                                    'dune.data_explorer.category'='abstraction',
+                                                    'dune.data_explorer.abstraction.type'='sector',
+                                                    'dune.data_explorer.abstraction.name'='safe_airdrop',
```
springzh, line 423: Not sure about this line. What's the usage for the abstraction name? Should it be always set to same as table name?

springzh:
> Thanks for your submission @springzh !
> 
> Could you please change reformat this query so the model is addresses_ethereum.safe_airdrop so we can standardize airdrops across different chains ? #1529

Updated. Please check again. @soispoke

macros/alter_table_properties.sql
soispoke, line 423: Here, I think the type should be `sector`, and the name `addresses`

soispoke (changes requested):

soispoke:
The model file name should be addresses_ethereum_safe_airdrop.sql instead of safe_airdrop.sql

macros/alter_table_properties.sql
springzh: Updated

springzh:
> The model file name should be addresses_ethereum_safe_airdrop.sql instead of safe_airdrop.sql

@soispoke Renamed. Thank you.

soispoke (approved):
lgtm !

PR #1536, opened 2022-09-08
Some fixes for OpenSea, Seaport, LooksRare and X2Y2

hildobby:
Brief comments on the purpose of your changes:

Buyer is sometimes the address of an aggregator, fixed that for OpenSea, Seaport, LooksRare & X2Y2 (This query should ideally always return nothing: https://dune.com/queries/1249102).

Also for LooksRare: fixed sometimes missing `token_standard` and changed NULLs to 0 for royalties/marketplaceplace fees & simplified `number_of_items`.

FYI: Probably needs a full rerun of all those to apply the fixes retroactively :)

*For Dune Engine V2*
I've checked that:
General checks:
* [X] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [X] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [X] if adding a new model, I added a test
* [X] the filename is unique and ends with .sql
* [X] each sql file is a select statement and has only one view, table or function defined  
* [X] column names are `lowercase_snake_cased`
* [X] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [X] if adding a new model, I edited the alter table macro to display new database object (table or view) in UI explorer
* [X] if adding a new materialized table, I edited the optimize table macro

Join logic:
* [X] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

Incremental logic:
* [X] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [X] where block_time >= date_trunc("day", now() - interval '1 week')
* [X] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [X] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #1538, opened 2022-09-09
Remove erc20 join that causes duplicate entries

masquot:
Removed join on erc20 transfers for now as it results in duplication of buy events and duplicate `unique_trade_id` in some cases.

fyi @hildobby an example of a tx that results in duplication is [0x04f019f90a5afea8bdc72dddccdf5f7fb5ec5cdce69877497ad35bdabe07f9c5](https://etherscan.io/tx/0x04f019f90a5afea8bdc72dddccdf5f7fb5ec5cdce69877497ad35bdabe07f9c5#eventlog)

#1536 may very well be affected by this issue also.


soispoke (approved):

soispoke (commented):
lgtm !

PR #1554, opened 2022-09-14
Revert "feat: add ETHW token on trx-tron chain"

aalan3:
Reverts duneanalytics/spellbook#1546

We don't support tron chain and there is no trailing comma on this which is causing problems.


soispoke (approved):

danqing117:
hey, may I know what's wrong with tron chain?

danqing117:
And I would also like to double confirm. Even if ethw is added to prices_tokens.sql file and merged, then if I run `select * from prices.usd where symbol = 'ETHW'`, there will be no results shown?

PR #1563, opened 2022-09-16
Add CoW Protocol Solvers View

gentrexha:
Adding Dune V2 CoW Protocol solvers view

For Dune Engine V2
I've checked that:

- [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
- [x] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables
- [x] if adding a new model, I added a test
- [x] the filename is unique and ends with .sql
- [x] each sql file is a select statement and has only one view, table or function defined
- [x] column names are lowercase_snake_cased


gentrexha:
@bh2smith @soispoke

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -0,0 +1 @@
+# Kept here for future use
```
bh2smith, line 1: I am still new to this, but I guess we don't need to define a schema for views? I will be using this file in a follow up PR for the incremental trades table.

dbt_project.yml
```diff
@@ -191,13 +191,18 @@ models:
...
       +materialized: table
       optimism:
         +schema: omv1_optimism
         +materialized: table
+    cow_protocol:
+      +schema: cow_protocol
+      +materialized: view
+      ethereum:
+        +schema: cow_protocol_ethereum
+        +materialized: view
```
bh2smith, line 205: I am not sure how all the labels work just yet, but I was expecting the solvers to be the materialized view... this makes it look like the entire namespace is materialized view... I see that ovm1 uses `+materialized: table` just above and above that is also `+materialized: view`...

models/cow_protocol/ethereum/cow_protocol_ethereum_sources.yml
```diff
@@ -0,0 +1,38 @@
...
+            name: evt_block_time
+            description: "Timestamp for block event time in UTC"
+          - &evt_index
+            name: evt_index
+            description: "Index value of the transaction"
+          - &evt_tx_hash
+            name: evt_tx_hash
+            description: "Primary key of the transaction"
+            tests:
+              - not_null
```
bh2smith, line 29: Does this mean that "tests do exist"? I would have expected this to be a reference to the test file or something.

bh2smith (approved):
The SQL looks good to me, although you may want to link the [PoC Query](https://dune.com/queries/1276806) also in the PR description. I am a total noob with the spellbook configuration files. Probably best to wait on review by @soispoke.

dbt_project.yml
soispoke: In this case, the +materialized: view means all models in these repositories will be views by default (which is usually a good idea). However, it's very easy to materialize specific models as tables/incremental tables within each model in the config JINJA statement

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
soispoke: We always need a schema, even for views. This is quite important, since it's also used to generate spellbook documentation. Pls make sure to include column names, descriptions, tags, metadata, etc... here

models/cow_protocol/ethereum/cow_protocol_ethereum_sources.yml
soispoke: Here this means that a test has been set up to check if evt_tx_hash column values are not NULL. It'll fail if that's the case

soispoke (changes requested):
thanks for your submission @gentrexha and @bh2smith

gentrexha:
> In this case, the +materialized: view means all models in these repositories will be views by default (which is usually a good idea). However, it's very easy to materialize specific models as tables/incremental tables within each model in the config JINJA statement

Does the one in the model config overwrite the general one?

dbt_project.yml
gentrexha: > In this case, the +materialized: view means all models in these repositories will be views by default (which is usually a good idea). However, it's very easy to materialize specific models as tables/incremental tables within each model in the config JINJA statement

Does the one in the model config overwrite the general one?

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -0,0 +1,25 @@
...
+        description: "Solver's wallet address"
+      - &environment
+        name: environment
+        description: "Solver's environment"
+      - &name
+        name: name
+        description: "Solver's name"
+      - &active
+        name: active
+        description: "Solver's active status"
```
bh2smith, line 25:
```suggestion
        description: "Solver's active status according to the events emitted by GPv2AllowListAuthentication. True if solver has been 'added' and not yet removed, False when the solver has been 'removed'."
```

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -0,0 +1,25 @@
+version: 2
+
+models:
+  - name: cow_protocol_ethereum_events
```
bh2smith, line 4: Not sure, but I think this should be...
```suggestion
  - name: cow_protocol_ethereum_solvers
```

dbt_project.yml
soispoke: Yes, ideally you can remove the `materialized='view'` inside the jinja config in the model file so you avoid defining materialization twice (although it won't break things if you keep things as such)

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -0,0 +1,25 @@
+version: 2
```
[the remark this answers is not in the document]
soispoke: This is just the dbt default config for the schema yml files https://docs.getdbt.com/reference/project-configs/config-version

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
soispoke: Yes, that's right ✔️

gentrexha:
@soispoke I believe I've covered the requested changes. Let me know if there's anything else to do.

soispoke:
@gentrexha If you want to expose this table on the Dune App, the last thing to do is to attribute metadata to your model by adding it to the alter_table_properties.sql macro: https://github.com/duneanalytics/spellbook/blob/main/macros/alter_table_properties.sql

gentrexha:
> @gentrexha If you want to expose this table on the Dune App, the last thing to do is to attribute metadata to your model by adding it to the alter_table_properties.sql macro: https://github.com/duneanalytics/spellbook/blob/main/macros/alter_table_properties.sql

Done :heavy_check_mark: 

P.S. @bh2smith can you please give that latest change a 2nd pair of eyes as well.

macros/alter_table_properties.sql
```diff
@@ -766,6 +766,15 @@ ALTER TABLE ovm1_optimism.user_addresses_weekly_active_cohorts SET TBLPROPERTIES
                                                 'dune.data_explorer.contributors'='["msilb7", "chuxinh"]');
 {% endset %}
 
+{% set cow_protocol_ethereum_solvers %}
+ALTER TABLE cow_protocol_ethereum.solvers SET TBLPROPERTIES ('dune.public'='true',
```
soispoke, line 770: This should be set to `ALTER VIEW` instead 🙏

models/cow_protocol/ethereum/cow_protocol_ethereum_sources.yml
```diff
@@ -0,0 +1,38 @@
...
+    tables:
+      - name: GPv2AllowListAuthentication_evt_SolverAdded
+        description: "GPv2AllowListAuthentication list of events when a Solver was added"
+        columns:
+          - &solver
+            name: solver
+            description: "Solver wallet address"
+          - &contract_address
+            name: contract_address
+            description: "Ethereum address for the liquidity pool used in transaction"
```
bh2smith, line 15: I guess this was copied from elsewhere.
```suggestion
            description: "Ethereum address of contract emitting the event"
```

models/cow_protocol/ethereum/cow_protocol_ethereum_sources.yml
```diff
@@ -0,0 +1,38 @@
...
+          - &evt_block_number
+            name: evt_block_number
+            description: "Block number which processed the unique transaction hash"
+          - &evt_block_time
+            name: evt_block_time
+            description: "Timestamp for block event time in UTC"
+          - &evt_index
+            name: evt_index
+            description: "Index value of the transaction"
```
bh2smith, line 24:
```suggestion
            description: "Index value of the event within transaction"
```

soispoke (approved):

soispoke:
LGTM 🚀

models/cow_protocol/ethereum/cow_protocol_ethereum_sources.yml
bh2smith: I will fix this in the follow up PR...

bh2smith:
When should we expect to see this view in the interface?

soispoke:
@bh2smith right about....now :) 
Capture d’écran 2022-09-19 à 15 23 34

PR #1566, opened 2022-09-17
change deploy condition

dot2dotseurat:
It looks like the spellbook docs at https://spellbook-docs.dune.com are out of date. 

We changed the structure of the directory when we moved v1 into the deprecated folder. I think that after we made that change, this workflow wasn't running because we no longer had a spellbook sub directory. 

@philippWassibauer can you take a look at this change and let me know if you think there is anything else we need to change to keep the docs up to date. 

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:

### DIFF
--- a/.github/workflows/spellbook_deploy_trigger.yml
+++ b/.github/workflows/spellbook_deploy_trigger.yml
@@ -2,8 +2,8 @@ name: Trigger Spellbook Deploy to update the spellbook site
 
 on:
   push:
-    paths:
-      - 'spellbook/**'
+    branches:
+      - 'main'
 
 jobs:
   notify:


philippWassibauer:
Changes look good to me. I had a chat with Bjorn 1-2 months ago and he said he would take over the triggering of the deploy script. So I am not aware if the {{ secrets.PAT_TOKEN }} is set up by now. If not then this will not be enough. I think the action was also disabled at one point. 

@dot2dotseurat do you want me to set up the PAT_TOKEN? 

In any case I will trigger a deploy now so that we are at least at the latest version.

philippWassibauer (approved):

PR #1569, opened 2022-09-17
ENS resolver: fixes

hildobby:

Added a distinct filter to remove duplicates, remove subdomains (broken for now) and add a test.

*For Dune Engine V2*
I've checked that:
General checks:
* [X] if adding a new model, I edited the alter table macro to display new database object (table or view) in UI explorer

Join logic:

Incremental logic:


0xRobin:
Looks like this is mixing both resolver as reverse records..
What are you trying to achieve? 

1) resolver: one entry per existing ENS name 
-> `what address does this ENS resolve to currently?` (multiple records for 1 address possible)
2) reverse: one entry per address 
-> `what ENS should I call this address?` (max 1 distinct record for an address)

soispoke:
tagging @hildobby so we can discuss @0xRobin comments here

hildobby:
yupp I definitely agree here! tbh ENS is very tough to work with and we were constrained by time so we rushed this PR a bit believing that it would also trigger others to come and help improve this abstraction. Our goal here was to surface all ENS names belonging to a single address (both reverse record and resolver). How do you think it should be here @0xRobin?

hildobby:
It would also be cool to have @makoto's opinion here, might be able to help for accuracy. Ideally this abstraction is then used into Dune's labels

0xRobin:
@soispoke @hildobby 
Just based on domain knowledge (haven't looked at contract) I believe we should split ENS into different basis tables
`ens_names`, `ens_records` and `ens_reverse`.
These tables should only contain currently valid entries (no history of what *was* valid)

## Tables

### ens_names
| field | type |
| ------------- | ------------- |
| ens_name | string (readable) **[unique]** |
| token_id | string (only defined for top level ENS) |
| parent | string (ens_name) |
| registrant | string (address, currently only defined for top level ENS) |
| controller | string (address) | 
| expiry_date | datetime |

### ens_records
| field | type |
| ------------- | ------------- |
| token_id | string (hash) |
| ens_name | string (readable) |
| record_type | string ['address' / 'content' / 'text'] |
| record_key | string ['ETH' / 'BTC' / 'com.twitter' / 'avatar' / ...]
| value | string |
| unique_record_id | concat(ens_name, _record_type_) **[unique]**  | 

### ens_reverse
| field | type |
| ------------- | ------------- |
| address | string **[unique]** |
| ens_name | string (readable) | 

Will have to look into the contract to see what happens on transfers/expiry/... to see if we can achieve this from on-chain data. 
I believe retrieval speed here is more important then having the full history parsed in a materialized table, hence my suggestions to only include current valid state and no history. Users that want to query ens record history can still do so with the existing tables. 

## Usage
1. getting all ENS names a wallet controls 
```sql
select 
ens_name
from ens_names
where controller = {{wallet_address}}
```

2. getting all ENS names that resolve to a wallet
```sql
select 
ens_name
from ens_records
where record_type = 'address' and record_key = 'ETH'
and value = {{wallet_address}}
```

3. making a table more readable by using reverse ENS names
```sql
select
coalesce(ens_name, wallet_address)
from some_table t
left join ens_reverse e
on t.wallet_address = e.address
```

models/ens/ens_resolver_addresses.sql
```diff
@@ -115,7 +115,7 @@ WHERE latest_ens = 1
         ) names
 )
 
-SELECT address
+SELECT distinct address
```
makoto, line 118: `distinct` is probably needed because Registrar.setName and Resolver.setName are redundant ( Registrar.setName usually calls Resolver.setName internally like https://github.com/ensdomains/ens-archived-contracts/blob/main/contracts/ens/ReverseRegistrar.sol#L81

models/ens/ens_resolver_addresses.sql
```diff
@@ -128,4 +128,5 @@ FROM
         SELECT name, call_tx_hash AS tx_hash, call_block_time AS block_time, address FROM reverse_records
         ) lastone
     ) fin
-WHERE name_rank = 1
\ No newline at end of file
+WHERE name_rank = 1
+AND !(name regexp '.*\\..+\\..*')
```
makoto, line 132: Can I see the example of subdomains not working? I wonder if I can figure out why it's not working

makoto:
> @soispoke @hildobby Just based on domain knowledge (haven't looked at contract) I believe we should split ENS into different basis tables `ens_names`, `ens_records` and `ens_reverse`. These tables should only contain currently valid entries (no history of what _was_ valid)

FYI, in our [subgraph](https://github.com/ensdomains/ens-subgraph/blob/master/schema.graphql)

we have folloing models (aka `types`)

## domain = similar to `name`

```
type Domain @entity {
  id: ID!                                               # The namehash of the name
  name: String                                          # The human readable name, if known. Unknown portions replaced with hash in square brackets (eg, foo.[1234].eth)
  labelName: String                                     # The human readable label name (imported from CSV), if known
  labelhash: Bytes                                      # keccak256(labelName)
  parent: Domain                                        # The namehash (id) of the parent name
  subdomains: [Domain!]! @derivedFrom(field: "parent")  # Can count domains from length of array
  subdomainCount: Int!                                  # The number of subdomains
  resolvedAddress: Account                              # Address logged from current resolver, if any
  owner: Account!
  resolver: Resolver
  ttl: BigInt
  isMigrated: Boolean!
  createdAt: BigInt!
  events: [DomainEvent!]! @derivedFrom(field: "domain")
}
```

## resolver = similar to `record`

```
type Resolver @entity {
  id: ID!                   # Concatenation of resolver address and namehash
  domain: Domain
  address: Bytes!           # Address of resolver contract

  addr: Account             # Current value of addr record (per events)
  contentHash: Bytes        # Content hash, in binary format.
  texts: [String!]          # Set of observed text record keys
  coinTypes: [BigInt!]      # Set of observed SLIP-44 coin types
  events: [ResolverEvent!]! @derivedFrom(field: "resolver")
}
```

## Account = similar to reverse record but we actually don't track it there

Reverse record is a bit tricky to track as we not only need to check if reverseRecord (address -> name) is set but also forward record (name -> address) is set because anyone can set their eth address to `vitalik.eth` unless we check both directions.

```
type Account @entity {
  id: ID!
  domains: [Domain!]! @derivedFrom(field: "owner")
  registrations: [Registration!] @derivedFrom(field: "registrant")
}
```

## Registrations = owner of .eth names

```
type Registration @entity {
  id: ID!
  domain: Domain
  registrationDate: BigInt!
  expiryDate: BigInt!
  cost: BigInt
  registrant: Account!
  labelName: String
  events: [RegistrationEvent!]! @derivedFrom(field: "registration")
}
```

makoto:
Another FYI.
The current query is inferring that the person who sends the transaction is the address of the primary name/reverse record, but that would change after the new contract release (which emits `ReverseClaimed ` at https://github.com/ensdomains/ens-contracts/blob/8a2423829a28852297ee208357d148987e8dce0f/contracts/registry/ReverseRegistrar.sol#L85)

makoto:
Probably not related to this PR, but why it's doing `UNION ALL`

`reverse_records ` and `register_to_resolve ` at https://github.com/duneanalytics/spellbook/blob/5c75bfdef0ca5bd21143d1c85388d78efe44b958/models/ens/ens_resolver_addresses.sql#L126 to get reverse record?

jeff-dude:
@hildobby how would you like to proceed here? i've applied the WIP tag until we know it's ready for a closer review

jeff-dude:
closing due to above comment. conversation to continue in #1666

Issue #1510, opened 2022-09-06
Incorrect NFT sale prices in nft.trades


catherine-2:
I've noticed there are a lot of incorrect NFT sale prices in the Dune v2 nft.trades table. Seems like a pretty large sized issue -- for Moonbirds Opensea trades over the last 2 months 46% of sales are affected and show the incorrect price. 

Not sure what is causing this as it affects both WETH accepts + ETH buy now listings.

Two examples:
select * 
from nft.trades
where tx_hash in (
    lower('0xf3e876be22e90a8f339c140411e1aa73ca63999ee2c8cd3b0d22054447695b46') -- 11.8 WETH sale showing as 12.685 in amount_original
    , lower ('0x0846007a3049edc73ae36ef069dfba9a11a02ebe8381c34963b715ff66499910') -- 14.99 ETH buy now sale showing as 16.11425 in amount_original
    )

Opensea links:
https://opensea.io/assets/ethereum/0x23581767a106ae21c074b2276d25e5c3e136a68b/5344
https://opensea.io/assets/ethereum/0x23581767a106ae21c074b2276d25e5c3e136a68b/9099

catherine-2:
I figured out how this bug is happening.

Example on 11.8 ETH sale that is showing as 12.685
https://etherscan.io/tx/0xf3e876be22e90a8f339c140411e1aa73ca63999ee2c8cd3b0d22054447695b46

The numbers should be:
Sale price = 11.8
OS fee = 0.295
Royalty = 0.59
Amount to seller = 10.915

Second example on 14.99 that is showing as 16.11425
https://etherscan.io/tx/0x0846007a3049edc73ae36ef069dfba9a11a02ebe8381c34963b715ff66499910

Sale price = 14.99
OS fee = 0.37475
Royalty = 0.7495
Amount to seller = 13.86575


Somehow the amount_original column in the nft.trades table v2 for OpenSea is showing the sale price + OS fee + royalty instead of just the sales price. (e.g. 11.8+0.295+0.59 = 12.685 that is showing up in the data table right now instead of the correct 11.8)

I'm not familiar with the underlying code -- can take a look -- but wanted to flag in the meantime in case someone else is more familiar and knows how this is happening!

catherine-2:
@soispoke saw you were listed as the nft.trades contributor on the docs so wanted to flag for you as well! I searched the underlying code and couldn't tell at first read through how this is happening.

soispoke:
[...]

PR #1572, opened 2022-09-18
quickfix-1510

0xRobin:
Quickfix for https://github.com/duneanalytics/spellbook/issues/1510
resolves double counting all fees in the `p3` subquery for seaport.transfers


soispoke (approved):
merging this because I think it's important not to inflate volume from OpenSea v3. This doesn't mean issues with this query are all resolved

Thanks @0xRobin

sohwak:
Thank you for covering this.

PR #1579, opened 2022-09-19
T 16562 post hook within model config for table properties

soispoke:

It's time (dramatic music plays). The alter_table_properties.sql macro is growing indefinitely, and countless mistakes are made by wizards and us everyday because of it. Here is an example of using a post-hook (credit to @aalan3) within a model to directly define metadata associated with it. While not perfect (e.g., we have to use backslash to skip escape double quotes to define some variables), it's more scalable and will probably solve most of our current problems. I'll open a follow-up PR to apply this change to all our models if you guys @duneanalytics/team-data-experience @augustog think this makes sense 👍 

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


models/nomad/ethereum/nomad_ethereum_view_bridge_transactions.sql
```diff
@@ -1,7 +1,17 @@
 {{
   config(
-        alias='view_bridge_transactions'
-  )
+      alias='view_bridge_transactions',
+      post_hook = "
```
aalan3, line 4: very nice thanks for trying it! I think we should make this block into a macro, this would hopefully prevent typos and other smaller issues. Will also help if we ever need to change a bunch of things.

soispoke: makes sense, I'll give it a try

soispoke: @aalan3 Here's what I got so far, the only thing that's still bothering me is the backslash that's needed to break double quotes but otherwise wdyt ?

aalan3 (approved):

PR #1580, opened 2022-09-19
CoW Protocol Trades

bh2smith:
Migrating the `gnosis_protocol_v2.trades` from [V1 Abstractions](https://github.com/duneanalytics/spellbook/blob/main/deprecated-dune-v1-abstractions/ethereum/gnosis_protocol_v2/insert_trades.sql) table to V2 Spellbook under `cow_protocol_ethereum.trades`.

Note that

- `prices.usd` is joined all the way back to the first emitted event.
- The PoC Query here: https://dune.com/queries/1283229 has been pasted from the results of `dbt compile`, but I have added a block time constraint (so that the results can actually be viewed - it is a very large table)
- I am unable to determine if the dbt project YAML needs to be altered (since this is incremental, but the config file says everything is a view).

cc @gentrexha for internal review.

I've checked that:

**General checks:**
---> This was a bit weird, because I had to use `source('prices', 'usd')`, but used `ref('tokens_erc20')` which seems inconsistent.
---> I was thinking that it would take hours to populate this table, so I suspect it would be more appropriate to add a test once the table has been populated. Is there something better that could be done here @soispoke?
---> Not Applicable according to: https://github.com/duneanalytics/spellbook/pull/1563#discussion_r974171047

**Join logic:**

**Incremental logic:**
--> THIS IS ONLY TRUE FOR INCREMENTAL CHANGES... otherwise the table needs a complete history of prices back to launch date.


macros/alter_table_properties.sql
```diff
@@ -775,6 +775,14 @@ ALTER VIEW cow_protocol_ethereum.solvers SET TBLPROPERTIES ('dune.public'='true'
...
 {% endset %}
 
+{% set cow_protocol_ethereum_trades %}
+ALTER VIEW cow_protocol_ethereum.trades SET TBLPROPERTIES ('dune.public'='true',
+    'dune.data_explorer.blockchains'='["ethereum"]',
+    'dune.data_explorer.category'='abstraction',
+    'dune.data_explorer.abstraction.type'='project',
+    'dune.data_explorer.abstraction.name'='cow_protocol',
+    'dune.data_explorer.contributors'='["bh2smith", "gentrexha"]');
+{% endset %}
```
bh2smith, line 785: It seems to me that this could go into the block above so that all cow_protocol table modifications are made in the same place (instead of defining a new block for each...)

macros/optimize_tables.sql
```diff
@@ -99,6 +99,10 @@ OPTIMIZE archipelago_ethereum.events;
 OPTIMIZE labels.all;
 {% endset %}
 
+{% set cow_protocol_ethereum_table_optimizer %}
+OPTIMIZE cow_protocol_ethereum.trades;
+{% endset %}
+
```
bh2smith, line 105: Similar to the comment above, I have named this in a way that suggests other tables in the same namespace be optimized within this block. Not sure if this considered appropriate by the Dune team.

soispoke (commented):
thanks @bh2smith everything lgtm, the only thing I'm wondering is if we should try to standardize this spell so it can be included in the meta dex.trades spell (see `https://github.com/duneanalytics/spellbook/blob/main/models/dex/dex_trades.sql` for column names)
wdyt ? 
@augustog @dsalv @jeff-dude

bh2smith:
> thanks @bh2smith everything lgtm, the only thing I'm wondering is if we should try to standardize this spell so it can be included in the meta dex.trades spell

While generally I would agree with this, a lot of the dex.trades columns would be superfluous for us and I am essentially trying to stay similar to our existing trades table (so that query and dashboard migration doesn't become too complex). 

When `dex.trades` is ready, I would have no problem filling building the select statement that conforms to their columns... similar to what we have already done here:

https://github.com/duneanalytics/spellbook/blob/main/deprecated-dune-v1-abstractions/ethereum/dex/trades/insert_gnosis_protocol.sql

(note that we insert from the trades abstraction since many of the relevant fields are populated).

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -0,0 +1,159 @@
...
+        from {{ source('gnosis_protocol_v2_ethereum', 'GPv2Settlement_evt_Trade') }}
+                 distribute by evt_tx_hash
+            sort by evt_index
+        ) as _
+    {% if is_incremental() %}
+    WHERE call_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
+    on call_tx_hash = evt_tx_hash
+),
+-- TODO - replace this with abstraction `cow_protocol_ethereum_order_id_app_data_map
```
gentrexha, line 97: Remove todo from here?

soispoke:
@bh2smith this makes sense! 

Do you think we could build the other table suitable for dex.trades right away or in a following PR (but keeping this one as well) then ?

`dex.trades` is ready, we just need to actually add protocols now, so I'm thinking this is a good opportunity to get the ball rolling

gentrexha (approved):
Nice work Ben. Minor nitpicking comments from my side.

bh2smith:
> Do you think we could build the other table suitable for dex.trades right away or in a following PR (but keeping this one as well) then ?

Yes, Absolutely! I would be beyond excited to get us in there ASAP! Would definitely follow up with that immediately (at least once the table is populated).

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -0,0 +1,159 @@
...
+           trader,
+           sell_token                        as sell_token_address,
+           (CASE
+                WHEN ts.symbol IS NULL THEN sell_token
+                ELSE ts.symbol
+               END)                          as sell_token,
+           buy_token                         as buy_token_address,
+           (CASE
+                WHEN tb.symbol IS NULL THEN buy_token
+                WHEN buy_token = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee' THEN 'ETH'
```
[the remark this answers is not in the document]
bh2smith: ETH cannot be sold on the protocol but can be bought (i.e. we unwrap for them).

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -22,4 +22,83 @@ models:
...
+        description: "Owner of the order being traded (aka trader)"
+      - &sell_token_address
+        name: sell_token_address
+        description: "Ethereum address of Sell Token"
+      - &sell_token
+        name: sell_token
+        description: "Symbol of Sell Token"
+      - &buy_token_address
+        name: buy_token_address
+        description: "Ethereum address of Buy Token"
```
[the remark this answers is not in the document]
bh2smith: Will make it consistent.

bh2smith:
Thanks @soispoke - will do - just gonna fix some nits.

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -0,0 +1,159 @@
...
+     join (
+        select evt_tx_hash, evt_index, orderUid
+        from {{ source('gnosis_protocol_v2_ethereum', 'GPv2Settlement_evt_Trade') }}
+                 distribute by evt_tx_hash
+            sort by evt_index
+        ) as _
+    {% if is_incremental() %}
+    WHERE call_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
+    on call_tx_hash = evt_tx_hash
```
soispoke, line 95: there is a small syntax error here

bh2smith: what is the syntax error?

bh2smith: Oh... the JOIN .. ON should come first.

bh2smith:
Ok, I believe this is ready.

soispoke:
@bh2smith The incremental merge into is throwing an error because the  unique_key = ['tx_hash', 'order_uid'] is not unique 

Capture d’écran 2022-09-20 à 10 42 55

bh2smith:
Ok - this is unusual (but known to be possible). I will look into an alternative for unique-id.

soispoke:
I just made this quick query and I'm seeing duplicate rows : https://dune.com/queries/1287539?d=1

bh2smith:
Nice job here - I guess my concatenation trick did not work out as expected. I will look into this and see what can be done about it... Worst case we will have to include `evt_index` as a column and include it in the unique ID... however, it would be nice if we could sum up the amounts. This also looks like an unintended bug in order execution.

bh2smith:
OK. I have found the issue - it is with this block constructing a map between order_uid and app_data

```sql
order_id_and_trade as (
    select
        orderUid as order_id,
        explode(trades) as trade
    from gnosis_protocol_v2_ethereum.GPv2Settlement_call_settle
     join (
        select evt_tx_hash, evt_index, orderUid
        from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Trade
                 distribute by evt_tx_hash
            sort by evt_index
        ) as _
    on call_tx_hash = evt_tx_hash
    -- ADDED FOR TESTING
   WHERE call_block_time >= '{{start_ts}}'
      AND call_block_time < '{{end_ts}}'
),
-- TODO - replace this with abstraction `cow_protocol_ethereum_order_id_app_data_map
uid_to_app_id as (
    select
        distinct order_id as uid, -- remove duplicate occurrences of order_id (when order is partially fillable)
                 get_json_object(trade, '$.appData') as app_data,
                 get_json_object(trade, '$.receiver') as receiver
    from order_id_and_trade
),
```

This was a syntax change that was different from postgres I must have gotten incorrect

bh2smith:
I have managed to fix the app data mapping query! https://dune.com/queries/1287714

Will update the PR very soon.

bh2smith:
Should be ready now @soispoke! Sorry about the issue before.

soispoke (commented):
@bh2smith Still seeing some issues with duplicate rows (less than before but still some in there)
https://dune.com/queries/1287539?d=1

bh2smith:
So sorry about that. I will try to have something that works for us tomorrow.

bh2smith:
Ok - so it looks like its because there are duplicated tokens (i.e. we were narrowing our search down to tokens_ethereum, but rather all networks).

Hopefully this is the final iteration.

soispoke:
just pushed a quick fix but otherwise LGTM !

soispoke (approved):

soispoke:
all good @bh2smith 
Capture d’écran 2022-09-20 à 18 09 37

bh2smith:
@soispoke 

> Do you think we could build the other table suitable for dex.trades right away or in a following PR (but keeping this one as well) then ?
> `dex.trades` is ready, we just need to actually add protocols now, so I'm thinking this is a good opportunity to get the ball rolling

We started looking into this, but it looks like the dex.trades table has been made exclusively for DEXes (while CoW Protocol is better categorized as an **Aggregator**). There was some discussion in your Discord in #spellbook ([around here](https://discord.com/channels/757637422384283659/999683200563564655/1011695268237934622))and more recently [here](https://discord.com/channels/757637422384283659/1012706316755664926/1022091693308264488).

soispoke:
@jeff-dude @dsalv is there a template for trade aggregators we could use for this ?

jeff-dude:
i followed up in discord, but we can use CoW as the example to introduce ```dex_aggregator.trades``` spell. the initial plan is to follow ```dex.trades``` with an additional column to support "aggregators of aggregators" via a string value of which aggregator pushed it through. any other tweaks we can manage in the build phase
https://github.com/duneanalytics/spellbook/pull/1353

PR #1581, opened 2022-09-19
PR Job Dependency Management

dot2dotseurat:

dbt builds all tables and views in dev under the users "personal schema" (my_schema). so if a spell (spell b) depends on another spell (spell a), it will need to exist in the user's personal schema. 

dbt achieves this by rebuilding the table under the personal schema even if the table exists in prod using the `+` command in the dbt run command. This is very time-consuming and wasteful of resources especially for solana tables. 

To get around this, my proposal is we use the prod and pr manifest files to parse what the dependent models and seeds are and then build them as views from the prod versions. 

the python script pr_job_utils.py parses dependencies and builds a macro that creates views under the expected personal schema for the dbt job and points them to the prod versions. 

then I remove the `+` indicators from the dbt pr job. 

WARNING this pr includes some junk changes in existing spells so I can trigger "modified" states. These changes will need to be removed before we merge this work in. 

My tragically long list of commits is because our action runner apparently runs a very old version of python and it hates typing hints which were introduced to python in v3.5 :/


dot2dotseurat:
Screen Shot 2022-09-19 at 5 10 28 PM
this seems to be working as I was able to build models and run tests with seeds based on dependencies. 
we still are hitting a fairly high number of 502s but that needs to be addressed in a different PR.

soispoke (approved):
lgtm !

PR #1584, opened 2022-09-20
Manifest file shrinker

dot2dotseurat:
Sometimes we use SQL files to store long lists of labels, addresses, etc. This has resulted in a bloated manifest file which we use to generate the documentation. This script which will be called in the deployment process will cut off the raw_sql component after 500 lines. 

This should only affect long lists of values. If a query is logic and not lists and it would be cut off by this... we need to refactor that query!

### DIFF
--- a/scripts/docs_shrink_manifest.py
+++ b/scripts/docs_shrink_manifest.py
@@ -1,0 +1,36 @@
+import json
+import sys
+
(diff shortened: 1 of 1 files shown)


philippWassibauer (approved):
lgtm

PR #1587, opened 2022-09-21
V2: Add CoW Protocol Batches

bh2smith:
Migrating the `gnosis_protocol_v2.batches` from [V1 Abstractions](https://github.com/duneanalytics/spellbook/blob/main/deprecated-dune-v1-abstractions/ethereum/gnosis_protocol_v2/insert_batches.sql) table to V2 Spellbook under `cow_protocol_ethereum.batches`. 

cc @gentrexha as cow organization reviewer.

**Note that**
- This table involves a join of `ethereum.transactions` (on `hash`) and `prices.usd`

For our own personal records, this table is slightly different than the V1 abstraction 
- we do not include `gas_per_trade` column since it can easily be computed from `gas_used / num_trades`
- `gas_price_gwei` has been changed to just `gas_price` (which is in WEI)
- Concatenated `solver_name` (environment + name) is no longer here, to get solver names for batches one must join on solvers table (this was an awkward feature of the v1 abstraction where we found ourselves joining on the solver view anyway to get the (non-concatenated form of the solver names). 

Finally, *This table still depends on the solvers view* for the purpose of computing `dex_swaps`. Unfortunately for batches settled via aggregators we are at the mercy of using `dex.trades` to count the number of dex swaps. Since the dex trades table doesn't yet exist, this has been left as a TODO here in the SQL file.

*For Dune Engine V2*
I've checked that:

**General:**
---> The PoC Query https://dune.com/queries/1290518 is slightly modified from the compiled version (including block time bounds for avoiding timeouts)
---> NOT APPLICABLE 

**Join logic:**

**Incremental logic:**
---> ONLY WITHIN INCREMENTAL
---> ONLY WITHIN INCREMENTAL


bh2smith:
Note that the failing tests are not at all related to the changes made here: 

```
ERROR foundation_ethereum_assert_trades ................................. [ERROR in 0.14s]
```

gentrexha (approved):

bh2smith:
What happened to my force push... I had squashed this all into a single commit?

models/cow_protocol/ethereum/cow_protocol_ethereum_batches.sql
```diff
@@ -0,0 +1,93 @@
...
+            WHEN name ilike '%1inch'
+               OR name = '%ParaSwap'
+               OR name = '%0x'
+               OR name = 'Legacy'
+               THEN NULL
+            -- TODO: We can't rely on dex.trades table here yet,
+            -- thus we cannot know how many dex swaps are happening in settlements via aggregators
+            -- V1: (select count(*) from dex.trades where tx_hash = evt_tx_hash and category = 'DEX')
+            ELSE dex_swaps
+        END                                              as dex_swaps,
```
gentrexha, line 71: Question: What's the difference between `batch_counts.dex_swaps` and `combined_batch_info.dex_swaps`?

bh2smith: Great question: `batch_counts.dex_swaps` will record a value of 1 for batches settled through aggregators, because there is only 1 contract interaction with the aggregator’s router (the router is responsible for routing all the trades through different AMMs) and we don’t have that viability from the high level call data (all those AMM interactions are nested inside the call data being forwarded to the aggregation router).

So, In order to get the correct number of `dex_swaps` for aggregator batches, we count the DEX interactions on `dex.trades` for the batch's `tx_hash`. 
While this is a bit of a cheeky hack, it might actually be possible to decode the interaction call data deeper instead, but we would have to work on this as independent for a while.

models/cow_protocol/ethereum/cow_protocol_ethereum_batches.sql
```diff
@@ -0,0 +1,93 @@
+{{  config(
+        alias='batches',
+        materialized='incremental',
+        unique_key = ['tx_hash'],
+        on_schema_change='fail',
```
soispoke, line 5: I'm getting the following error after trying to update the model incrementally:
Capture d’écran 2022-09-21 à 09 38 27

it seems to be related to the on_schema_change settings here

bh2smith: Based on the suggestion names it seems 
- `append_new_columns` seems weird because why would there be new columns if the table structure hasn't changed?
- `sync_all_columns` also seems intense especially if this means the entire table.

Maybe I should be thinking of the term `column` transposed to `row` here because this is a column oriented database, in which case this would mean... "append new rows" which sounds pretty reasonable or "sync all rows" which would also make sense.

The situation we face here is that the trades table doesn't always have the prices until a while later (the sync schedule of `prices.usd`), thus there are batch records being created for which we can not yet evaluate the "batch_value" so we do want sync some of these "columns" (although not necessarily all - because there comes a point when the table contents before a certain block time become static). 

I suppose I need to learn a bit more on how all this works to give the correct answer here (suggestions welcome).

bh2smith: Ok - after taking a closer look at the docs https://docs.getdbt.com/docs/building-a-dbt-project/building-models/configuring-incremental-models

I have changed `on_schema_change` to `sync_all_columns` and also noticed that `incremental_strategy` should probably be `delete+insert` instead. 

I hope that `sync_all_columns` isn't to computationally intensive (although I don't expect we will be changing the schema).

soispoke: @bh2smith thanks, will take a look now, I also think sync_all_columns is the way to go (and I don't think it would be too intensive in this particular case but let me check)

soispoke:
Everything LGTM, thanks @bh2smith for your submission  🤝

soispoke (approved):

PR #1583, opened 2022-09-20
Add Curvefi to dex.trades on Spellbook

dsalv:
Brief comments on the purpose of your changes:
- Add support for Curvefi on dex.trades on Arrakis

Tasks:
* [x] Port Curvefi model from Caladan to Arrakis
* [x] Integrate into dex.trades
* [x] Update metadata
* [x] Write tests

*For Dune Engine V2*
I've checked that:
General checks:
* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [x] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [x] if adding a new model, I added a test
* [x] the filename is unique and ends with .sql
* [x] each sql file is a select statement and has only one view, table or function defined  
* [x] column names are `lowercase_snake_cased`
* [x] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [x] if adding a new model, I edited the alter table macro to display new database object (table or view) in UI explorer
* [x] if adding a new materialized table, I edited the optimize table macro

Join logic:
* [x] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

Incremental logic:
* [x] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [x] where block_time >= date_trunc("day", now() - interval '1 week')
* [x] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [x] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #1588, opened 2022-09-21
NIT: Use trailing commas instead of preceding

bh2smith:
Uber nit, but I noticed the commas had an awkward pattern and that other protocols inserting themselves in this table we following suit (e.g. #1583). Thought I might suggest a more visibly appealing standard before it gets out of control.


soispoke (approved):
haha lgtm :)

jeff-dude:
ah man, i'm team left comma club, we can't give in!

PR #1591, opened 2022-09-21
[CoW] Consolidate <network>_trades table and conform to dex_trades

gentrexha:

We create a consolidated cow...trades table with columns conforming to that of dex trades. This will make is easier to insert trade records into the aggregator table later. cf #1600 

*For Dune Engine V2*
I've checked that:
General checks:
~~* [ ] if adding a new materialized table, I edited the optimize table macro~~

~~Join logic:~~
Not Applicable

~~Incremental logic:~~
Not Applicable


macros/alter_table_properties.sql
```diff
@@ -884,7 +893,6 @@ ALTER TABLE cow_protocol_ethereum.batches SET TBLPROPERTIES ('dune.public'='true
 {% do run_query(transfers_ethereum_erc20) %}
 {% do run_query(seaport_ethereum_view_transactions) %}
 {% do run_query(seaport_ethereum_transfers) %}
-{% do run_query(uniswap_trades) %}
```
gentrexha, line 887: The run_query(uniswap_trades) was here twice. Let me know if you want to remove that in a separate PR.

@soispoke

models/cow_protocol/cow_protocol_trades.sql
```diff
@@ -0,0 +1,38 @@
...
+            token_pair,
+            units_bought AS token_bought_amount,
+            units_sold AS token_sold_amount,
+            atoms_bought AS token_bought_amount_raw,
+            atoms_sold AS token_sold_amount_raw,
+            usd_value AS amount_usd,
+            buy_token_address AS token_bought_address,
+            sell_token_address AS token_sold_address,
+            trader AS taker,
+            '' AS maker,
```
gentrexha, line 25: @bh2smith Need some of your input in which is which on these two, as I'm not really sure. I've been reference the Uniswap implementation so far.

models/cow_protocol/cow_protocol_trades.sql
```diff
@@ -0,0 +1,38 @@
...
+            project_contract_address,
+            tx_hash,
+            trader AS tx_from,
+            receiver AS tx_to,
+            '' AS trace_address,
+            evt_index,
+            'cow_protocol' ||'-'|| tx_hash ||'-'|| IFNULL(evt_index, '') AS unique_trade_id
```
gentrexha, line 32: @soispoke Is there any guide on how to create this unique_trade_id or can I leave it like this?

models/cow_protocol/cow_protocol_trades.sql
```diff
@@ -0,0 +1,38 @@
...
+            atoms_sold AS token_sold_amount_raw,
+            usd_value AS amount_usd,
+            buy_token_address AS token_bought_address,
+            sell_token_address AS token_sold_address,
+            trader AS taker,
+            '' AS maker,
+            project_contract_address,
+            tx_hash,
+            trader AS tx_from,
+            receiver AS tx_to,
```
gentrexha, line 29: @bh2smith Also please double check if these are correct as well.

macros/alter_table_properties.sql
soispoke: please see this very recent PR to get rid of the alter_table_properties macro #1592

models/cow_protocol/cow_protocol_trades.sql
soispoke: This looks ok to me, no strict guidelines as long as it makes sense and doesn't create duplicates

models/cow_protocol/cow_protocol_trades.sql
bh2smith: If I remember correctly, for Uniswap, the AMM is considered the "maker" (aka market maker). In our protocol there is no maker. Might be worth looking at what 1inch considers the "maker" for comparison. 

Maybe we want to leave this NULL instead of an empty string?

Also, @josojo might have something to add here.

models/cow_protocol/cow_protocol_trades.sql
bh2smith: Can't we just use the OrderUid concatenated with the txhash? Like we do for unique key in the trades table?

models/cow_protocol/cow_protocol_trades.sql
bh2smith: That is a very interesting question. Technically I would think `tx_from` is actually the solver address (if one wanted to compare / join on `ethrereum.transactions`). I kinda wish these column names were trade from and trade to (if that is what they are intended to mean). Do we still have `trader_a` and `trader_b` because this would be redundant and I would actually propose the following:

```suggestion
            solver AS tx_from,
            <SETTLEMENT_CONTRACT_ADDRESS> AS tx_to,
```

Note that the `SETTLEMENT_CONTRACT_ADDRESS` shouldn't need to be hardcoded as indicated here (since there are two versions of this deployed, we can actually use `tx.from` and `tx.to` from `ethereum.transactions` table for this.

models/cow_protocol/cow_protocol_trades.sql
bh2smith:
```suggestion
            'cow_protocol' ||'-'|| tx_hash ||'-'|| order_uid AS unique_trade_id
```

models/cow_protocol/cow_protocol_trades_schema.yml
```diff
@@ -0,0 +1,86 @@
...
+        description: "Project contract address which executed the trade on the blockchain"
+      - &tx_hash
+        name: tx_hash
+        description: "Unique transaction hash value tied to each transaction on the DEX"
+      - &tx_from
+        name: tx_from
+        description: "Address which initiated the transaction"
+      - &tx_to
+        name: tx_to
+        description: "Address which received the transaction"
```
bh2smith, line 77: Here I think we are using the term "transaction" to mean something different than usual.

This is a "trade" and `tx_from` is the trader address owning the order being filled and the `tx_to`(although I find these column names odd) is the recipient of the buy token. Not sure we want to mix up the term "transaction" to be analogous with "trade" here.

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -15,6 +15,8 @@ WITH
 trades_with_prices AS (
     SELECT evt_block_time            as block_time,
            evt_tx_hash               as tx_hash,
+           evt_index,
+           contract_address          as project_contract_address,
```
bh2smith, line 19: If this was introduced only for the purpose of defining a Unique ID, we may not need it. Unless it is required by dex.trades table elsewhere.

I do, however agree that evt_index would ultimately ensure uniqueness of the key (in the event that a single partially fillable order is touched twice in the same transaction).

bh2smith (commented):
A couple of things I noticed here.

It seems as if we are creating a new model with nearly identical namespace (i.e. without _ethereum) and slightly different column naming conventions in order to conform to the `dex.trades` table. 

It was my impression that we could simply make our select statement directly in `dex_trades.sql` without having to create an intermediary table. I am ok with doing this if it is deemed necessary by the dune team cc @soispoke (to keep dex.trades clean) although I suspect this may not be upheld by new projects in the future.

If we do wind up sticking with this somewhat duplicated table, do you think we could come up with a slightly different naming convention for one of them? Or am I just being too picky?

gentrexha:
> A couple of things I noticed here.
> 
> It seems as if we are creating a new model with nearly identical namespace (i.e. without _ethereum) and slightly different column naming conventions in order to conform to the `dex.trades` table.

As far as I understood this table is supposed to keep all trades in it independent of the chain they're on. The same holds true for `dex.trades`.

> It was my impression that we could simply make our select statement directly in `dex_trades.sql` without having to create an intermediary table. I am ok with doing this if it is deemed necessary by the dune team cc @soispoke (to keep dex.trades clean) although I suspect this may not be upheld by new projects in the future.

That still works, and we can do that. But on the other hand, with this setup we keep a trades table agnostic of the chain the trades are on and there is less work to implement a new chain once the time comes.

> If we do wind up sticking with this somewhat duplicated table, do you think we could come up with a slightly different naming convention for one of them? Or am I just being too picky?

I've based this on the implementation done from Uniswap and referred to that. It might be possible that we don't need all the extra tables as well. In my opinion, it makes the extension of new chains easier and if everyone is going to follow that trend I think we should do so too. What do you think?

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
gentrexha: > If this was introduced only for the purpose of defining a Unique ID, we may not need it. Unless it is required by dex.trades table elsewhere.
> 
> I do, however agree that evt_index would ultimately ensure uniqueness of the key (in the event that a single partially fillable order is touched twice in the same transaction).

As far as I'm aware it is part of the `dex.trades` model and that's why I've added it.

Maybe @soispoke can elaborate further if this is required or optional.

macros/alter_table_properties.sql
gentrexha: Thanks! I've added the post hook to the model I'm adding. 

Not sure if I need to do anything else.

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
bh2smith: Do we know yet about whether this should belong in a `dex.aggregators` table or if the `category` field will be added?

models/cow_protocol/cow_protocol_trades.sql
bh2smith:
```suggestion
            trader AS taker,
            NULL AS maker,
```

models/cow_protocol/cow_protocol_trades_schema.yml
gentrexha: I've updated the column descriptions. Let me know what you think of them now.

bh2smith:
Linking @jeff-dude'a comment here

https://github.com/duneanalytics/spellbook/pull/1580#issuecomment-1253737449

And here

https://discord.com/channels/757637422384283659/1012706316755664926/1022139370205089834

It looks like the idea is to create a new table dex_aggregator.trades

And have a column "category" (I guess in both) then build something out of the two... I think. 

We can use this PR, or possibly sneak one in between, to define the structure.

bh2smith:
So the plan for this now is only to expose the `cow_protocol_trades` table here whose columns are aligned with that of `dex.trades`. We can get this merged into main and then rebase #1600 onto it so that it can start picking up the records from the newly introduced table.

bh2smith (approved):
This appears to be passing the build now so we should be in the clear to merge without having touched dex.trades

The follow up #1600 could then be rebased onto this and we should be able to introduce `dex_aggregator_trades` while simultaneously populating it with cow protocol trades.

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -1,7 +1,7 @@
 {{  config(
         alias='trades',
         materialized='incremental',
-        unique_key = ['tx_hash', 'order_uid'],
+        unique_key = ['tx_hash', 'order_uid', 'evt_index'],
         on_schema_change='fail',
```
soispoke, line 5: pls update this to `sync_all_columns`

soispoke (commented):
lgtm after this small change

soispoke (changes requested):

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -28,7 +30,7 @@ trades_with_prices AS (
            feeAmount                 as fee_amount,
            ps.price                  as sell_price,
            pb.price                  as buy_price
-    FROM {{ source('gnosis_protocol_v2_ethereum', 'GPv2Settlement_evt_Trade') }}
+    FROM {{ source('gnosis_protocol_v2_ethereum', 'GPv2Settlement_evt_Trade') }} settlement
```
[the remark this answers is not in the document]
bh2smith: I could do that here. Not sure I understand the decoded sources comment, could you elaborate?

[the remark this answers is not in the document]
soispoke: Yeah sorry, so basically we have to set up incremental filters using this syntaxt
`{% if is_incremental() %}  x.blocktime >= date_trunc("day", now() - interval '1 week') {% endif %}`

It allows to speed up incremental runs a lot because it's only looking to append/merge data for the last week instead of scanning the entire table.

My comment was about you including more filters like these, on sources (e.g., `{{ source('gnosis_protocol_v2_ethereum', 'GPv2Settlement_evt_Trade') }}`), prices, etc...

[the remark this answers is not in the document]
soispoke: If you look at this opensea spell, we apply incremental filters a lot on different downstream dependencies https://github.com/duneanalytics/spellbook/blob/main/models/opensea/ethereum/opensea_v1_ethereum_events.sql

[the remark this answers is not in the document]
bh2smith: Ok I have added several incremental filters here (found lots of places). I hope it still compiles.

soispoke (approved):

soispoke:
all tests you good, thanks @bh2smith and @gentrexha !

models/cow_protocol/cow_protocol_trades.sql
```diff
@@ -0,0 +1,42 @@
...
+        )
+}}
+SELECT *
+FROM
+(
+        SELECT
+            'ethereum' AS blockchain,
+            'CoW Protocol' AS project,
+            '1' AS version,
+            TRY_CAST(date_trunc('DAY', block_time) AS date) AS block_date,
```
jeff-dude, line 16: @bh2smith @gentrexha what do you think about moving this block_date logic to the base object ```cow_protocol_ethereum.trades``` and add partition in the config block? this has been an approach we've taken on dex trades, with the idea that the column will be used in downstream queries/dashboards often to filter on date ranges

our base tables (like ethereum transactions) are partitioned in this format, so i thought it'd make sense to match that -- could help query performance

bh2smith: Sounds like a good idea. I had noticed this pattern lingering around the project and totally makes sense.

bh2smith: is there any good reason for using "try_cast" here?

bh2smith:
There seems to be an issue here populating the `evt_index` and `project_contract_address` or this table:

There are 384705 records with empty value here...

models/cow_protocol/cow_protocol_trades.sql
jeff-dude: no particular reason, an old habit i guess. we would actually want it to fail if incompatible, but we know the block_time coming through will be fine to cast as date, so no big deal currently

PR #1596, opened 2022-09-21
dont make views of new dependencies

dot2dotseurat:
https://github.com/duneanalytics/spellbook/pull/1583

I saw in [Dias Alymbekov](https://dune.height.app/user/dias)'s pr that the dependency creation was failing on dependencies that were introduced in the pr.

curvefi_ethereum.trades is created by this pr and then it's a depedency later. we were trying to reference the prod version of this table but it didn't exist yet.

I added a commit to his pr to resolve this by removing new models from the refs list.

I'm opening a new pr with just those changes since they should take effect before the curvefi pr is merged.


jeff-dude (approved):
👍

dsalv:
Thanks! 👍

PR #1600, opened 2022-09-22
Introduce DEX Aggregator Trades

bh2smith:
This is just a proposed structure for the separation of `dex_trades` and `aggregator_trades`. One key point for discussion about the changes here:

I propose changing the name of the `dex_trades` to `amm_trades` and reserve `dex_trades` for the combined table (which includes both `amm_trades` and `aggregator_trades` along with a `category` column). Here I have used a longer form `dex_amm_trades` and `dex_aggregator_trades` but I consider this redundant and would like to change it. 

This would also mean that anyone already having written a query on dex_trades would be guaranteed to be getting similar results without breaking their existing query (and it would include aggregator trades as well). 

If this proposal is indeed accepted, I am wondering if we actually have to move the files elsewhere because the names don't align with the directory structure.

The select statement for aggregator trades has been left commented because I wanted to get the structure right before we rebase or merge the contents of #1591


models/dex/dex_aggregator_trades.sql
```diff
@@ -0,0 +1,43 @@
...
+--             token_bought_amount,
+--             token_sold_amount,
+--             token_bought_amount_raw,
+--             token_sold_amount_raw,
+--             amount_usd,
+--             token_bought_address,
+--             token_sold_address,
+--             taker,
+-- -- I don't believe this is necessary here... unless for 1inch, the maker is the router contract..
+--             NULL as maker,
```
bh2smith, line 30:
```suggestion
--             maker,
```

models/dex/dex_aggregator_trades.sql
```diff
@@ -0,0 +1,43 @@
...
+--             tx_to,
+--             trace_address,
+--             evt_index,
+--             unique_trade_id
+--         FROM {{ ref('cow_protocol_trades') }}
+        /*
+        UNION
+        <add future protocols here>
+        */
+    )
```
bh2smith, line 43:
```suggestion
    )

```

models/dex/dex_amm_trades.sql
```diff
@@ -0,0 +1,42 @@
+{{ config(
+        alias ='amm_trades',
+        post_hook='{{ expose_spells(\'["ethereum"]\',
+                                "sector",
+                                "dex",
+                                \'["jeff-dude", "bh2smith"]\') }}'
```
bh2smith, line 6: This file was actually renamed and the only "actual diff" is that alias has been changed to 'amm_trades' and I added myself to the contributors list (which isn't all that necessary).

models/dex/dex_schema.yml
```diff
@@ -5,36 +5,39 @@ models:
     meta:
       blockchain: ethereum
       sector: dex
-      contributors: jeff-dude
+      contributors: jeff-dude, bh2smith
     config:
-      tags: ['ethereum','uniswap','trades', 'dex','cross-chain']
+      tags: ['ethereum', 'trades', 'dexes', 'aggregators', 'cross-chain']
```
bh2smith, line 10: I found the fact that "uniswap" was tagged specifically in the "dex_trades" table to be biased and removed it. Unless we expect every protocol to tag themselves here as well.

models/dex/dex_schema.yml
```diff
@@ -5,36 +5,39 @@ models:
...
-        description: "UTC event block date of each DEX trade"
+        description: "UTC event block date of each trade"
       - &block_time
         name: block_time
-        description: "UTC event block time of each DEX trade"
+        description: "UTC event block time of each trade"
       - &token_bought_symbol
         name: token_bought_symbol
-        description: "Token symbol for token bought in the transaction"
+        description: "Token symbol for token bought in the trade"
```
bh2smith, line 34: I noticed that the term transaction was being used, but this doesn't generalize to transactions with multiple trades at once, so I changed all occurrences of "transaction" to "trade" which I believe to be the most appropriate term for this (since its a trades table).

models/dex/dex_amm_trades.sql
soispoke: You should definitely add your name in there 👍

models/dex/dex_trades.sql
```diff
@@ -3,40 +3,10 @@
...
-        /*
-        UNION
-        <add future protocols here>
-        */
-)
\ No newline at end of file
+
+SELECT "AMM" as category, * from {{ ref('dex_amm_trades') }}
+UNION
+SELECT "Aggregator" as category, * from {{ ref('dex_aggregator_trades') }}
```
bh2smith, line 12: Here is the ❤️ of the proposed structure.

`dex_trades` becomes this table that include an additional column "category", we refer to what were previously "dex" in v1 as "amm" (aka individual liquidity pool) and "aggregator" for aggregator. 

To defend the rational for using AMM here, is that when one goes to a platform like 1inch, in the interface they might refer to this as a DEX (a decentralized platform/protocol facilitating exchange). Similarly for CoW Protocol. Both of these fall under the category of aggregator, but it would not be incorrect to refer to them as DEXes. However, it would be incorrect to refer to them as AMMs.

models/dex/dex_schema.yml
soispoke: If we want to follow what was done with nft trades, we would want protocol to tag themselves in there but it might not be very easy to maintain so let's keep this as is

soispoke (commented):
Hey @bh2smith ! Thanks a lot for all the work you put into this, laying the foundations of dex/aggregator trades is super important 🙏

I think I like the dex_amm_trades / dex_aggregator_trades (instead of amm_trades, aggregator_trades) UNIONed into a dex_trades table better even if it's a bit redundant because we have to account for other aggregators (e.g., NFT aggregators) and I think it makes sense to have dex_aggregators and nft_aggregators clearly separated (in the repo but also in their actual names)

would love to have @jeff-dude @dsalv @0xBoxer take a look at this as well

models/dex/dex_amm_trades.sql
bh2smith: ok then I will put it back...

models/dex/dex_trades.sql
jeff-dude: i may have misunderstood the original conversation around this, but i thought the goal was to split "amm" and "aggregator" -- which we are -- but then not to end up union'ing back into one table, since the confusion on v1 engine was having both in one table. i must have interpreted it as the two objects stay separate in v2 & you can join them together on certain fields to get a full picture.

i'm no expert in the dex sector, so happy to hear otherwise.

models/dex/dex_aggregator_trades.sql
```diff
@@ -0,0 +1,42 @@
+{{ config(
+        alias ='aggregator_trades',
```
jeff-dude, line 2: a more minor comment, which do we prefer?
1. dex_aggregator.trades
2. dex.aggregator_trades

i think the first option is more in line with the majority of the repo, but maybe @soispoke can help me confirm that?

bh2smith: Yea right. I just thought we would the other part had to remain dex to stay online with the directory structure.

bh2smith:
Hey @jeff-dude tha is for your thoughts here. 

Your suggestions make sense and I am now wondering if the aggregator trades should go in a different directory “aggregators” instead of having the extended names like dex amm trades and dex aggregator trades. 

The idea was to at least keep something called dex trades while also maintains two separate tables (I.e giving people the choice and not changing things too much out of their comfort zone). If we do leave this here in the dex directory, then I would insist that dex trades still be renamed to either amm trades or dex amm trades. Reason for this is that people querying dex trades might wonder why there are no aggregator records (or maybe they will learn quick) I guess we could go with leaving dex trades as is and creating a new aggregator trades (however then I suppose it should be in a different directory WDYT?)

Mostly I’d like to know if we should simply remove the unioned table from what’s proposed here, or completely move aggregator trades to a different directory.

bh2smith:
I have reworked this PR to be as minimally invasive as possible. That is, 

- Only adding a new table `dex_aggregator.trades` an inserting `cow_protocol_trades` as the first collection of records
- Minor adjustments to the field descriptions in the `dex.trades` table (I couldn't help myself - but willing to revert as well).

There seems to be an issue trying to make the namespace convention `dex.aggregator_trades` using the subdirectory structure and I am unable to tell what the CI error is. Otherwise, I would move this out of draft state...

cc @soispoke and @jeff-dude

models/dex/dex_aggregator_trades.sql
bh2smith: I have been trying to make this `dex_aggregator.trades` work (analogous to the `_blockchain` pattern) with the directory structure, but there must be something I am missing because I can't seem to get it to work. The current diff state of the PR shows my attempt so far.

soispoke:
tagging @dot2dotseurat for visibility (CI errors)

models/dex/dex_aggregator_trades.sql
jeff-dude: i pushed a change with an edit to dbt project file (which handles the directories and schema names) and edited the table alias -- does this work now?

dbt_project.yml
```diff
@@ -46,6 +46,9 @@ models:
     dex:
       +schema: dex
       +materialized: view
+      aggregator:
+        +schema: dex_aggregator
+        +materialized: view
```
bh2smith, line 51: Yes this is what I was looking for. Thanks for clearing that up!

models/dex/dex_aggregator_trades.sql
bh2smith: Sure does! Awesome!

bh2smith:
Ready on my side (but I cant approve).

cc @gentrexha (if you'd like to take a look).

models/dex/dex_trades.sql
bh2smith: This union has been removed.

jeff-dude (approved):
thank you @bh2smith -- great next step towards advancing our dex spells!

bh2smith:
Thanks @jeff-dude - I noticed a few other PRs trying to add aggregator trades. I could try and find some and link the updated approach so they can get merged. Not sure if the creator wants me messing with their code, so I'll leave it to them.

jeff-dude:
> Thanks @jeff-dude - I noticed a few other PRs trying to add aggregator trades. I could try and find some and link the updated approach so they can get merged. Not sure if the creator wants me messing with their code, so I'll leave it to them.

yeah, nothing active at the moment (at least PR wise). i'll be sure to link to this foundation for others to use.

bh2smith:
Hey - so this is merged, but the table introduced here doesn't yet appear query-able in the interface..

soispoke:
I'll take a look :)

models/dex/aggregator/dex_aggregator_trades.sql
```diff
@@ -0,0 +1,42 @@
+{{ config(
+        alias ='trades',
+        post_hook='{{ expose_spells(\'["ethereum"]\',
+                                "sector",
+                                "dex",
```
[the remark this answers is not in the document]
bh2smith: That could be it. Should I submit a PR?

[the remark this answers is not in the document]
soispoke: yes please :)

PR #1472, opened 2022-08-31
Create a perpetuals swaps/trades aggregated table on Dune V2

RplusT:
Brief comments on the purpose of your changes:
This PR creates the initial foundation of a perpetuals swaps/trades aggregated table on Dune V2 patterned after `dex.trades`. Currently contains perpetuals protocols/markets on Optimism, namely Perpetual V2, Synthetix/Kwenta, and Pika v1, v2, v3. Structured in such a way that perpetuals markets on other chains can be added later on.

*For Dune Engine V2*
I've checked that:

* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [x] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [x] if adding a new model, I added a test
* [x] the filename is unique and ends with .sql
* [x] each sql file is a select statement and has only one view, table or function defined  
* [x] column names are `lowercase_snake_cased`

PR #1601, opened 2022-09-22
Add spell for sudoswap pool stats

Jam516:
Adding a spell that provides stats on all sudoswap poola

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:

### DIFF
--- a/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
+++ b/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
@@ -76,3 +76,4 @@ FROM
   pool_stats ps
 INNER JOIN pool_balance pb ON pb.pool_address = ps.pool_address
 INNER JOIN pool_trades pt ON pt.pool_address = ps.pool_address
+;
--- a/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
+++ b/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
@@ -1,0 +1,275 @@
+{{
+ config(
+       alias='pools',
+       post_hook='{{ expose_spells(\'["ethereum"]\',
+                                   "project",
+                                   "sudoswap",
+                                   \'["niftytable"]\') }}')
+}}
+
+WITH
+  pairs_created AS (
+    SELECT
+      CASE
+        WHEN _bondingCurve = '0x5b6ac51d9b1cede0068a1b26533cace807f883ee' THEN 'linear_bonding'
+        ELSE 'exp_bonding'
+      END as pricing_type,
+      _delta / 1e18 as delta,
+      CASE
+        WHEN _poolType = 0 THEN 'token'
+        WHEN _poolType = 1 THEN 'nft'
+        WHEN _poolType = 2 THEN 'trade'
+      END AS pooltype,
+      _spotPrice / 1e18 AS spotprice,
+      _nft AS nftcontractaddress,
+      _initialNFTIDs AS nft_ids,
+      _fee AS initialfee,
+      output_pair AS pair_address,
+      call_block_time AS block_time,
+      tx.FROM AS creator_address,
+      date_trunc('day', now()) - date_trunc('day', call_block_time) AS days_passed,
+      cardinality(_initialNFTIDs) AS initial_nft_count,
+      tx.value / 1e18 AS initial_eth
+    FROM
+      {{ source('sudo_amm_ethereum','LSSVMPairFactory_call_createPairETH') }} cre
+      LEFT JOIN {{ source('ethereum','transactions') }} tx ON tx.hash = cre.call_tx_hash
+    WHERE
+      call_success
+  ),
+  most_recent_spot_delta AS (
+    SELECT
+      COALESCE(del.contract_address, spot.contract_address) AS pair_address,
+      del.delta,
+      spot.spotprice
+    FROM
+      (
+        SELECT
+          *
+        FROM
+          (
+            SELECT
+              contract_address,
+              newDelta / 1e18 AS delta,
+              row_number() over (
+                PARTITION BY
+                  contract_address
+                ORDER BY
+                  evt_block_number DESC,
+                  evt_index DESC
+              ) as most_recent
+            FROM
+              {{ source('sudo_amm_ethereum','LSSVMPair_general_evt_DeltaUpdate') }}
+          ) a
+        WHERE
+          most_recent = 1
+      ) del
+      FULL OUTER JOIN (
+        SELECT
+          *
+        FROM
+          (
+            SELECT
+              contract_address,
+              newSpotPrice / 1e18 AS spotprice,
+              row_number() over (
+                PARTITION BY
+                  contract_address
+                ORDER BY
+                  evt_block_number DESC,
+                  evt_index DESC
+              ) AS most_recent
+            FROM
+              {{ source('sudo_amm_ethereum','LSSVMPair_general_evt_SpotPriceUpdate') }}
+          ) a
+        WHERE
+          most_recent = 1
+      ) spot ON del.contract_address = spot.contract_address
+  ),
+  erc721_balances AS (
+    SELECT
+      holder_address,
+      sum(value) AS tokens_held
+    FROM
+      (
+        SELECT
+          to AS holder_address,
+          count(evt_tx_hash) AS value
+        FROM
+          {{ source('erc721_ethereum','evt_transfer') }}
+        WHERE
+          to IN (
+            SELECT
+              pair_address
+            FROM
+              pairs_created
+          )
+        GROUP BY
+          1
+        union all
+        SELECT
+        FROM
+          AS holder_address,
+          count(evt_tx_hash) * -1 as value
+        FROM
+          {{ source('erc721_ethereum','evt_transfer') }}
+        WHERE
+        FROM
+          IN (
+            SELECT
+              pair_address
+            FROM
+              pairs_created
+          )
+        GROUP BY
+          1
+      ) a
+    GROUP BY
+      1
+  ),
+  eth_balances AS (
+    WITH
+      eth_in AS (
+        SELECT
+          tr.to AS holder_address,
+          SUM(tr.value / 1e18) AS eth_funded
+        FROM
+          {{ source('ethereum','traces') }} tr
+          JOIN pairs_created pc ON pc.pair_address = tr.to
+        WHERE
+          tr.block_time > '2022-04-23'
+          AND tr.success = true
+          AND tr.type = 'call'
+          AND (
+            tr.call_type NOT IN ('delegatecall', 'callcode', 'staticcall')
+            OR tr.call_type IS null
+          )
+        GROUP BY
+          1
+      ),
+      eth_out AS (
+        SELECT
+          tr.FROM AS holder_address,
+          SUM(tr.value / 1e18) AS eth_spent
+        FROM
+          {{ source('ethereum','traces') }} tr
+          JOIN pairs_created pc ON pc.pair_address = tr.FROM
+        WHERE
+          tr.block_time > '2022-04-23'
+          AND tr.success = true
+          AND tr.type = 'call'
+          AND (
+            tr.call_type NOT IN ('delegatecall', 'callcode', 'staticcall')
+            OR tr.call_type IS null
+          )
+        GROUP BY
+          1
+      )
+    SELECT
+      eth_in.holder_address,
+      eth_in.eth_funded,
+      eth_out.eth_spent,
+      COALESCE(eth_funded, 0) - COALESCE(eth_spent, 0) as eth_balance
+    FROM
+      eth_in
+      LEFT JOIN eth_out ON eth_in.holder_address = eth_out.holder_address
+  ),
+  all_pairs_cleaned AS (
+    SELECT
+      pricing_type,
+      pooltype,
+      nftcontractaddress,
+      creator_address,
+      COALESCE(mr.delta, pc.delta) AS delta --if delta was never updated, just keep original deploy delta
+,
+      round(COALESCE(mr.spotprice, pc.spotprice), 4) as spotprice --same logic as above
+,
+      nft_bal.tokens_held,
+      round(COALESCE(eth_bal.eth_balance, 0), 4) as eth_balance,
+      pc.pair_address AS raw_pair_address,
+      days_passed,
+      pc.spotprice AS initial_price,
+      pc.initial_nft_count,
+      pc.initial_eth
+    FROM
+      pairs_created pc
+      LEFT JOIN most_recent_spot_delta mr ON pc.pair_address = mr.pair_address
+      LEFT JOIN erc721_balances nft_bal ON nft_bal.holder_address = pc.pair_address
+      LEFT JOIN eth_balances eth_bal ON eth_bal.holder_address = pc.pair_address
+  ),
+  trading_totals as (
+    SELECT
+      CASE
+        WHEN trade_category = 'Sell' THEN buyer
+        ELSE seller
+      END AS pair_address,
+      sum(amount_original) AS eth_volume,
+      sum(amount_usd) AS usd_volume,
+      sum(number_of_items) AS nfts_traded,
+      sum(pool_fee_amount) AS owner_fee_volume_eth,
+      sum(platform_fee_amount) AS platform_fee_volume_eth,
+      sum(
+        CASE
+          WHEN trade_category = 'Sell' THEN -1 * amount_original
+          ELSE amount_original
+        END
+      ) AS eth_change_trading,
+      sum(
+        CASE
+          WHEN trade_category = 'Sell' THEN number_of_items
+          ELSE -1 * number_of_items
+        END
+      ) AS nft_change_trading
+    FROM
+      (
+        SELECT
+          trade_category,
+          buyer,
+          seller,
+          amount_original,
+          amount_usd,
+          number_of_items,
+          pool_fee_amount,
+          platform_fee_amount
+        FROM
+          ({{ ref('sudoswap_ethereum_events') }})
+        WHERE
+          buyer IN (
+            SELECT
+              pair_address
+            FROM
+              pairs_created
+          )
+          OR seller IN (
+            SELECT
+              pair_address
+            FROM
+              pairs_created
+          )
+      ) a
+    GROUP BY
+      1
+  )
+SELECT
+  acc.raw_pair_address AS pool_address,
+  nftcontractaddress,
+  creator_address,
+  spotprice,
+  COALESCE(tokens_held, 0) AS nft_balance,
+  COALESCE(eth_balance, 0) AS eth_balance,
+  COALESCE(trade.eth_volume, 0) AS eth_volume,
+  COALESCE(trade.nfts_traded, 0) AS nfts_traded,
+  COALESCE(trade.usd_volume, 0) AS usd_volume,
+  COALESCE(trade.owner_fee_volume_eth, 0) AS owner_fee_volume_eth,
+  COALESCE(trade.platform_fee_volume_eth, 0) as platform_fee_volume_eth,
+  pooltype,
+  pricing_type,
+  delta,
+  days_passed,
+  initial_price,
+  initial_nft_count,
+  initial_eth,
+  COALESCE(trade.eth_change_trading, 0) AS eth_change_trading,
+  COALESCE(trade.nft_change_trading, 0) AS nft_change_trading
+FROM
+  all_pairs_cleaned acc
+  LEFT JOIN trading_totals trade ON trade.pair_address = acc.raw_pair_address
--- a/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
+++ b/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
@@ -97,9 +97,10 @@ WITH
         FROM
           {{ source('erc721_ethereum','evt_transfer') }}
         WHERE
-          to IN (
+          evt_block_time > '2022-04-23'
+        AND (contract_address, to) IN (
             SELECT
-              pair_address
+              (nftcontractaddress, pair_address)
             FROM
               pairs_created
           )
@@ -113,10 +114,11 @@ WITH
         FROM
           {{ source('erc721_ethereum','evt_transfer') }}
         WHERE
-        FROM
+          evt_block_time > '2022-04-23'
+        AND (contract_address, from)
           IN (
             SELECT
-              pair_address
+              (nftcontractaddress, pair_address)
             FROM
               pairs_created
           )
--- a/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
+++ b/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
@@ -212,7 +212,7 @@ WITH
       sum(
         CASE
           WHEN trade_category = 'Sell' THEN -1 * amount_original
-          ELSE amount_original
+          ELSE (amount_original-platform_fee_amount)
         END
       ) AS eth_change_trading,
       sum(
--- a/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
+++ b/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
@@ -32,7 +32,7 @@ WITH
       tx.value / 1e18 AS initial_eth
     FROM
       {{ source('sudo_amm_ethereum','LSSVMPairFactory_call_createPairETH') }} cre
-      LEFT JOIN {{ source('ethereum','transactions') }} tx ON tx.hash = cre.call_tx_hash
+      INNER JOIN {{ source('ethereum','transactions') }} tx ON tx.hash = cre.call_tx_hash
     WHERE
       call_success
   ),
--- a/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
+++ b/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
@@ -1,279 +1,78 @@
 {{
- config(
-       alias='pools',
-       partition_by = ['pool_address'],
-       materialized = 'incremental',
-       file_format = 'delta',
-       incremental_strategy = 'merge',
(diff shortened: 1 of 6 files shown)


models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,275 @@
...
+              pair_address
+            FROM
+              pairs_created
+          )
+        GROUP BY
+          1
+      ) a
+    GROUP BY
+      1
+  ),
```
0xRobin, line 128: Should probably also filter for the correct erc721 contract address here to not include other transfers 
(airdrops or other stuff)

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,275 @@
...
+      sum(amount_usd) AS usd_volume,
+      sum(number_of_items) AS nfts_traded,
+      sum(pool_fee_amount) AS owner_fee_volume_eth,
+      sum(platform_fee_amount) AS platform_fee_volume_eth,
+      sum(
+        CASE
+          WHEN trade_category = 'Sell' THEN -1 * amount_original
+          ELSE amount_original
+        END
+      ) AS eth_change_trading,
```
0xRobin, line 215: In the case of a `Buy` (pool sells NFT, receives ETH)
The amount of ETH received should be `(amount_original - platform_fee_amount - royalty_fee_amount)`
royalty can be ignored cause 0 but this block should be changed to:
```suggestion
      sum(
        CASE
          WHEN trade_category = 'Sell' THEN -1 * amount_original
          ELSE (amount_original-platform_fee_amount)
        END
      ) AS eth_change_trading,
```

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
0xRobin: Maybe something like this then (not sure this tuple syntax is supported in spark sql though)

```suggestion
  erc721_balances AS (
    SELECT
      holder_address,
      sum(value) AS tokens_held
    FROM
      (
        SELECT
          to AS holder_address,
          count(evt_tx_hash) AS value
        FROM
          {{ source('erc721_ethereum','evt_transfer') }}
        WHERE
          evt_block_time > '2022-04-23'
        AND (contract_address, to) IN (
            SELECT
              (nftcontractaddress, pair_address)
            FROM
              pairs_created
          )
        GROUP BY
          1
        union all
        SELECT
        FROM
          AS holder_address,
          count(evt_tx_hash) * -1 as value
        FROM
          {{ source('erc721_ethereum','evt_transfer') }}
        WHERE
          evt_block_time > '2022-04-23'
        AND (contract_address, from)
          IN (
            SELECT
              (nftcontractaddress, pair_address)
            FROM
              pairs_created
          )
        GROUP BY
          1
      ) a
    GROUP BY
      1
  ),
```

0xRobin (changes requested):
2 small remarks on the data correctness, bet the rest is looking good!

I just finished this dash today that covers a lot of the same metrics
https://dune.com/0xRob/sudo-pool

models/sudoswap/ethereum/sudoswap_ethereum_schema.yml
```diff
@@ -222,4 +222,55 @@ models:
...
+      project: sudoswap
+      contributors: [niftytable]
+    config:
+      tags: ['ethereum','sudoswap','pools']
+    description: >
+        sudoswap pools on Ethereum
+    columns:
+      - name: pool_address
+        description: "Address of the pool"
+      - name: nftcontractaddress
```
jeff-dude, line 239: minor nit: can we ensure all column names are using '_' to separate words? this would reflect here and the query in the model. there are a few instances here, but i'll just leave the comment here

jeff-dude (changes requested):

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,253 @@
...
+      _fee AS initialfee,
+      output_pair AS pair_address,
+      call_block_time AS block_time,
+      tx.FROM AS creator_address,
+      date_trunc('day', now()) - date_trunc('day', call_block_time) AS days_passed,
+      cardinality(_initialNFTIDs) AS initial_nft_count,
+      tx.value / 1e18 AS initial_eth
+    FROM
+      {{ source('sudo_amm_ethereum','LSSVMPairFactory_call_createPairETH') }} cre
+      INNER JOIN {{ source('ethereum','transactions') }} tx ON tx.hash = cre.call_tx_hash
```
jeff-dude, line 35: this model will be materialized as a view (based on dbt_project yaml file and no override in config block in this model). with this being a view, a join on a base table like ethereum transactions on only tx_hash is not efficient on spark and will take a long time to query each execution.

do you think this should be materialized as a delta table and incrementally update?

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,253 @@
...
+      1
+  ),
+  eth_balances AS (
+    WITH
+      eth_in AS (
+        SELECT
+          tr.to AS holder_address,
+          SUM(tr.value / 1e18) AS eth_funded
+        FROM
+          {{ source('ethereum','traces') }} tr
```
jeff-dude, line 114: same idea here, this will be less performant over time as the date range grows

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,253 @@
...
+          )
+        GROUP BY
+          1
+      ),
+      eth_out AS (
+        SELECT
+          tr.FROM AS holder_address,
+          SUM(tr.value / 1e18) AS eth_spent
+        FROM
+          {{ source('ethereum','traces') }} tr
```
jeff-dude, line 132: same idea

jeff-dude (changes requested):

models/sudoswap/ethereum/sudoswap_ethereum_schema.yml
Jam516: will make those changes now

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
Jam516: I'll try to make it a delta table with increments instead based on the example I see in sudoswap_ethereum_events

Jam516: First time, doing one of these delta tables. In the config, would it be fine to only use the pool address as the 'unique_key'? Is using the block_date a must?

jeff-dude: ```block_date``` is useful in the ```partition_by``` property, as it can help downstream query performance if you use date filter on dashboards. it's similar to how base tables (like ```ethereum.transactions```) is partitioned by block_date. 

it's not necessarily needed for ```unique_key``` -- the question to answer for this property is "what fields in my target table make it unique, so when i perform a merge statement, it will properly insert new and update existing?"

Jam516: I see, thanks for the clarification. I used `block_date` in `partition_by`. 

I used ['pool_address', 'nft_contract_address', 'creator_address'] for my unique key since the final table is pool based instead of time based. So new pools will be added and existing pools will be updated. Hope that was the right logic 😅

Jam516: Ah, my test failed because I dont have `block_date` in my final table. Thought it was an internal variable, didnt realise it needed to be in there.

Would it work if I partitioned by the date of the pools creation? 
Since its not an event based table there isnt really other date related variables there

jeff-dude: yeah, definitely partition by whatever makes the most sense for your target object. the goal should be a column which is frequently used in filters and/or joins downstream

jeff-dude:
now that we've upgraded this from a view to an incremental table, there is a bit more work to do to get it standardized with other incremental objects we're building. i'll try to point out where we can continue to enhance in a few comments

models/sudoswap/ethereum/sudoswap_ethereum_schema.yml
```diff
@@ -222,4 +222,55 @@ models:
...
+
+  - name: sudoswap_ethereum_pools
+    meta:
+      blockchain: ethereum
+      project: sudoswap
+      contributors: [niftytable]
+    config:
+      tags: ['ethereum','sudoswap','pools']
+    description: >
+        sudoswap pools on Ethereum
```
jeff-dude, line 235: now that you've defined your unique key in the model config, it's good practice to apply a test to this table in the schema file on those columns. please see for example:
https://github.com/duneanalytics/spellbook/pull/1472/files#diff-7da167e3f4a73d5c5c2e3d245dc8c73c88619f49d08d0fea792a4bd8c54382aeR14

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,275 @@
...
+      tx.FROM AS creator_address,
+      date_trunc('day', now()) - date_trunc('day', call_block_time) AS days_passed,
+      cardinality(_initialNFTIDs) AS initial_nft_count,
+      tx.value / 1e18 AS initial_eth
+    FROM
+      {{ source('sudo_amm_ethereum','LSSVMPairFactory_call_createPairETH') }} cre
+      INNER JOIN {{ source('ethereum','transactions') }} tx ON tx.hash = cre.call_tx_hash
+    WHERE
+      call_success
+    {% if is_incremental() %}
```
jeff-dude, line 44: approach for incremental logic:
1. use in where clause on the table in the 'from' -- i.e. ```cre``` alias in this query
2. use in the join conditions on joined tables -- i.e. ```trx``` alias in this query (use in 'and' logic rather than where)

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,275 @@
...
+              newDelta / 1e18 AS delta,
+              row_number() over (
+                PARTITION BY
+                  contract_address
+                ORDER BY
+                  evt_block_number DESC,
+                  evt_index DESC
+              ) as most_recent
+            FROM
+              {{ source('sudo_amm_ethereum','LSSVMPair_general_evt_DeltaUpdate') }}
```
jeff-dude, line 71: i realize not all decoded tables need an incremental filter, just checking if it's needed here or not

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,275 @@
...
+    WITH
+      eth_in AS (
+        SELECT
+          tr.to AS holder_address,
+          SUM(tr.value / 1e18) AS eth_funded
+        FROM
+          {{ source('ethereum','traces') }} tr
+          JOIN pairs_created pc ON pc.pair_address = tr.to
+        WHERE
+          tr.block_time > '2022-04-23'
```
jeff-dude, line 131: rather than having this filter every time, we've approached project start date filters in "not incremental" logic:
          {% if not is_incremental() %}
          AND tr.block_time > '2022-04-23'
          {% endif %}

jeff-dude: this "not incremental" logic can then also be used on all instances where "is incremental" is used, to help expedite initial build and any potential full refreshes in the future

jeff-dude: base tables, decoded tables, etc

jeff-dude: one more thing: you can also apply '2022-04-23' as a variable at the top of the model, then reference in these joins. an example:
https://github.com/duneanalytics/spellbook/pull/1472/files#diff-8be4351b64f13bdb3ced1d5365aa0de0fb03ecafcbeb1540561d00998748a1cbR12

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,275 @@
...
+          trade_category,
+          buyer,
+          seller,
+          amount_original,
+          amount_usd,
+          number_of_items,
+          pool_fee_amount,
+          platform_fee_amount
+        FROM
+          ({{ ref('sudoswap_ethereum_events') }})
```
jeff-dude, line 234: imagine this needs incremental filters as well

jeff-dude:
our goal is also to build a unit test for new incremental tables. are there clear test cases for this object you can think of?

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
Jam516: pool contract address is the most frequently used to filter and join. Will try using that to partition by

models/sudoswap/ethereum/sudoswap_ethereum_schema.yml
Jam516: Ok, adding that dbt_utils.unique_combination_of_column test

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
Jam516: cool, in that case its not needed here right? will remove

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
Jam516: Im not sure, is it needed here?

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
```diff
@@ -0,0 +1,275 @@
...
+              newSpotPrice / 1e18 AS spot_price,
+              row_number() over (
+                PARTITION BY
+                  contract_address
+                ORDER BY
+                  evt_block_number DESC,
+                  evt_index DESC
+              ) AS most_recent
+            FROM
+              {{ source('sudo_amm_ethereum','LSSVMPair_general_evt_SpotPriceUpdate') }}
```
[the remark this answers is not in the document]
Jam516: Im not sure, is it needed here?

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
Jam516: ok ive made the start date a variable and added is not incremental logic

models/sudoswap/ethereum/sudoswap_ethereum_pools.sql
Jam516: ok ive added one now

0xRobin:
Ok, a couple of concerns making this a materialized model:

So what this query currently is doing is calculating all-time aggregations on different pools. (like total ETH volume)
This is definitely possible to do in a view model (although it's indeed slow to compute every time).

When we want to materialize the table as an incremental delta table though, we need to take more things into account.
**Basically you want your query to make sens when ran on only the last couple days of data**.

_scenario_:
When we keep the query as is, on the initial full refresh we'll calculate the correct all-time ETH volume.
On the first incremental run that result will be overwritten with the new ETH volume that was calculated solely on the events from the last couple days. So that's a problem.

**For this to work we need to store "delta's" or _state changes_ in the materialized table**, and then make a new view model on top that aggregates to all-time stats. I hope I'm making sense here.

---

For this specific model this means rewriting it so that we calculate **daily** aggregations. 
basically something like this:

**daily_pool_stats.sql**
_materialized incremental delta table_
```sql
SELECT
  pool_address
  ,date_trunc('day',block_time) as day
  ,sum(amount_original) as volume
  ,concat(day, pool_address) as unique_id
FROM sudoswap.trades
{% if is_incremental() %}
    -- this filter will only be applied on an incremental run. We only want to update with new transfers.
    WHERE block_time >= date_trunc("day", now() - interval '1 week')
{% endif %}
GROUP BY pool_address, day
```
**pool_stats_view.sql**
_non-materialized view_
```sql

SELECT
  pool_address
  ,sum(volume) as all_time_volume
FROM daily_pool_stats
GROUP BY pool_address
```

I hope this makes sense @Jam516 


Please correct me @jeff-dude if I'm anywhere mistaken :smile:

Jam516:
> Ok, a couple of concerns making this a materialized model:
> 
> So what this query currently is doing is calculating all-time aggregations on different pools. (like total ETH volume) This is definitely possible to do in a view model (although it's indeed slow to compute every time).
> 
> When we want to materialize the table as an incremental delta table though, we need to take more things into account. **Basically you want your query to make sens when ran on only the last couple days of data**.
> 
> _scenario_: When we keep the query as is, on the initial full refresh we'll calculate the correct all-time ETH volume. On the first incremental run that result will be overwritten with the new ETH volume that was calculated solely on the events from the last couple days. So that's a problem.
> 
> **For this to work we need to store "delta's" or _state changes_ in the materialized table**, and then make a new view model on top that aggregates to all-time stats. I hope I'm making sense here.
> 
> For this specific model this means rewriting it so that we calculate **daily** aggregations. basically something like this:
> 
> **daily_pool_stats.sql** _materialized incremental delta table_
> 
> ```sql
> SELECT
>   pool_address
>   ,date_trunc('day',block_time) as day
>   ,sum(amount_original) as volume
>   ,concat(day, pool_address) as unique_id
> FROM sudoswap.trades
> {% if is_incremental() %}
>     -- this filter will only be applied on an incremental run. We only want to update with new transfers.
>     WHERE block_time >= date_trunc("day", now() - interval '1 week')
> {% endif %}
> GROUP BY pool_address, day
> ```
> 
> **pool_stats_view.sql** _non-materialized view_
> 
> ```sql
> SELECT
>   pool_address
>   ,sum(volume) as all_time_volume
> FROM daily_pool_stats
> GROUP BY pool_address
> ```
> 
> I hope this makes sense @Jam516
> 
> Please correct me @jeff-dude if I'm anywhere mistaken 😄

That makes sense. In that case, should I split this into two tables: 
1/ One that is a materialized table uses increments and gets the trading related stats for each day
2/ One that is a view that gets calculates current eth and NFT balances + refers to table 1 to calculate total trading stats

Or would it be fine to just make this current table a view @jeff-dude? Happy to go whichever way, just let me know and I'll get it done today, keen to put this together for a work thing.

jeff-dude:
> > Ok, a couple of concerns making this a materialized model:
> > So what this query currently is doing is calculating all-time aggregations on different pools. (like total ETH volume) This is definitely possible to do in a view model (although it's indeed slow to compute every time).
> > When we want to materialize the table as an incremental delta table though, we need to take more things into account. **Basically you want your query to make sens when ran on only the last couple days of data**.
> > _scenario_: When we keep the query as is, on the initial full refresh we'll calculate the correct all-time ETH volume. On the first incremental run that result will be overwritten with the new ETH volume that was calculated solely on the events from the last couple days. So that's a problem.
> > **For this to work we need to store "delta's" or _state changes_ in the materialized table**, and then make a new view model on top that aggregates to all-time stats. I hope I'm making sense here.
> > For this specific model this means rewriting it so that we calculate **daily** aggregations. basically something like this:
> > **daily_pool_stats.sql** _materialized incremental delta table_
> > ```sql
> > SELECT
> >   pool_address
> >   ,date_trunc('day',block_time) as day
> >   ,sum(amount_original) as volume
> >   ,concat(day, pool_address) as unique_id
> > FROM sudoswap.trades
> > {% if is_incremental() %}
> >     -- this filter will only be applied on an incremental run. We only want to update with new transfers.
> >     WHERE block_time >= date_trunc("day", now() - interval '1 week')
> > {% endif %}
> > GROUP BY pool_address, day
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **pool_stats_view.sql** _non-materialized view_
> > ```sql
> > SELECT
> >   pool_address
> >   ,sum(volume) as all_time_volume
> > FROM daily_pool_stats
> > GROUP BY pool_address
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I hope this makes sense @Jam516
> > Please correct me @jeff-dude if I'm anywhere mistaken 😄
> 
> That makes sense. In that case, should I split this into two tables: 1/ One that is a materialized increment table that gets the trading related stats for each day using sudoswap.trades or nft.trades 2/ One that is a view table that calculates current eth and NFT balances (using erc721 transfers and ethereum.traces) + refers to table 1 to calculate total trading stats
> 
> Or would it be fine to just make this current table a view @jeff-dude? Happy to go whichever way, just let me know and I'll get it done today, keen to put this together for a work thing.

we can certainly allow this to be a view to simplify, but i'd highly suggest running the query that builds the view and monitor performance. is it sufficient or does it drive you crazy?

if it's fine, feel free to make a view

if it's not fine, then we can essentially add a day column (or week, or month, whatever timeframe) as part of the unique key and group by in the query. then a second view on top to sum it as necessary

Jam516:
> > > Ok, a couple of concerns making this a materialized model:
> > > So what this query currently is doing is calculating all-time aggregations on different pools. (like total ETH volume) This is definitely possible to do in a view model (although it's indeed slow to compute every time).
> > > When we want to materialize the table as an incremental delta table though, we need to take more things into account. **Basically you want your query to make sens when ran on only the last couple days of data**.
> > > _scenario_: When we keep the query as is, on the initial full refresh we'll calculate the correct all-time ETH volume. On the first incremental run that result will be overwritten with the new ETH volume that was calculated solely on the events from the last couple days. So that's a problem.
> > > **For this to work we need to store "delta's" or _state changes_ in the materialized table**, and then make a new view model on top that aggregates to all-time stats. I hope I'm making sense here.
> > > For this specific model this means rewriting it so that we calculate **daily** aggregations. basically something like this:
> > > **daily_pool_stats.sql** _materialized incremental delta table_
> > > ```sql
> > > SELECT
> > >   pool_address
> > >   ,date_trunc('day',block_time) as day
> > >   ,sum(amount_original) as volume
> > >   ,concat(day, pool_address) as unique_id
> > > FROM sudoswap.trades
> > > {% if is_incremental() %}
> > >     -- this filter will only be applied on an incremental run. We only want to update with new transfers.
> > >     WHERE block_time >= date_trunc("day", now() - interval '1 week')
> > > {% endif %}
> > > GROUP BY pool_address, day
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > **pool_stats_view.sql** _non-materialized view_
> > > ```sql
> > > SELECT
> > >   pool_address
> > >   ,sum(volume) as all_time_volume
> > > FROM daily_pool_stats
> > > GROUP BY pool_address
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I hope this makes sense @Jam516
> > > Please correct me @jeff-dude if I'm anywhere mistaken 😄
> > 
> > 
> > That makes sense. In that case, should I split this into two tables: 1/ One that is a materialized increment table that gets the trading related stats for each day using sudoswap.trades or nft.trades 2/ One that is a view table that calculates current eth and NFT balances (using erc721 transfers and ethereum.traces) + refers to table 1 to calculate total trading stats
> > Or would it be fine to just make this current table a view @jeff-dude? Happy to go whichever way, just let me know and I'll get it done today, keen to put this together for a work thing.
> 
> we can certainly allow this to be a view to simplify, but i'd highly suggest running the query that builds the view and monitor performance. is it sufficient or does it drive you crazy?
> 
> if it's fine, feel free to make a view
> 
> if it's not fine, then we can essentially add a day column (or week, or month, whatever timeframe) as part of the unique key and group by in the query. then a second view on top to sum it as necessary

Running it as a query is slow, takes about 10min last time I tried. But the daily stats version of this query takes so long to run that it just times out 😅. 

Is there any chance the view would be a little bit faster than a normal query because of caching or something along those lines? Or would is it best to try making a spellbook of the daily stats version with increments even though it times out in regular Dune?

0xRobin:
The reason you're seeing a time-out is because when you run the fully compiled query you are actually simulating the first full refresh of that table. Calculating all the daily stats is more resource heavy then directly calculating all the aggregates, so it's normal that this takes longer than the view query. 

**It's not a problem if that query times out in the regular dune editor.**

To still be able to test this query just need to add a couple 
`where block_time > date_trunc('day',now()) - interval 1 week`
clauses on the source tables (in the dune query editor, not your source files).
Do this on the places where you've got the `if_incremental` blocks in your source files.
Then you simulate what an incremental run would run like, and that should run pretty quickly.

Jam516:
> The reason you're seeing a time-out is because when you run the fully compiled query you are actually simulating the first full refresh of that table. Calculating all the daily stats is more resource heavy then directly calculating all the aggregates, so it's normal that this takes longer than the view query.
> 
> **It's not a problem if that query times out in the regular dune editor.**
> 
> To still be able to test this query just need to add a couple `where block_time > date_trunc('day',now()) - interval 1 week` clauses on the source tables (in the dune query editor, not your source files). Do this on the places where you've got the `if_incremental` blocks in your source files. Then you simulate what an incremental run would run like, and that should run pretty quickly.

Thanks for the guidance 🙂. I tried that and the new version version I pushed last night has the materialised  incremental table of daily stats in it. 

Broke the daily stuff into two separate incremental tables 

And made one view table that sums the result of those two tables to get overall stats.

jeff-dude (approved):
thank you @Jam516 for the build and @0xRobin for the feedback. this design makes sense to me & the testing holds up. i'll get this deployed 👍

Jam516:
> thank you @Jam516 for the build and @0xRobin for the feedback. this design makes sense to me & the testing holds up. i'll get this deployed 👍

Awesome! Thanks a lot 🙂

PR #1610, opened 2022-09-22
temporarily removing ovm1

dot2dotseurat:
@chuxinh @MSilb7 we are going to move this work to a better place for deploying large files. @soispoke is working on this now. The tables are deployed so they should still be queryable. 

Unfortunately, the size of the files is causing issues downstream including memory errors in our production deployment jobs, documentation generation, and automated testing. 

As context, dbt generates this file called the manifest that copies the entire project as raw and compiled sql so the size of the file sort of exploded. 

We'll resolve this soon but for now, we need to move these files out of the project.


jeff-dude (approved):

chuxinh:
Thanks for keeping us in the loop! 
Yea can see the data still in explorer which is great. 

Wonder does it create issues on building other spells on top of these tables tho since i will probably be referencing them as some sources for other dependencies?

soispoke:
@chuxinh We're still working out the details, but in any case you'll always be able to reference the table itself (e.g., `ovm1_optimism.contracts` instead of using refs (e.g., `{{ ref('ovm1_optimism_contracts') }}`) 👍

PR #1623, opened 2022-09-23
Create nft.wash_trades

hildobby:

Label wash trading for all NFT marketplaces on Ethereum using 4 different critieria:
* Were the buyer and seller the same wallet?
* Was it traded another time with buyer and seller exchanged (within the timeframe of ± 1 week)?
* Did the buyer buy that nft 3+ times (only for erc721s)?
* Were the buyer and seller's wallets funded by the same wallet (excluding CEX and bridge wallets) or by one another?

Tests are yet to be written :)

*For Dune Engine V2*
I've checked that:
General checks:
* [X] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributors)

Join logic:

Incremental logic:


dot2dotseurat:
I see this is failing on `Table or view not found: addresses_ethereum.cex; line 5 pos 14`. I think this has been renamed labels.cex_ethereum?

hildobby:
I really don't understand the issue here, why would addresses be struct (i didn't even know that data type) and not strings when I'm pulling them straight from nft.trades. also, tests are ofc missing but I wanted to ensure the query runs well before I get onto those.

@jeff-dude do you think you could help me understand what's causing the error here? Thanks!

jeff-dude:
> I really don't understand the issue here, why would addresses be struct (i didn't even know that data type) and not strings when I'm pulling them straight from nft.trades. also, tests are ofc missing but I wanted to ensure the query runs well before I get onto those.
> 
> @jeff-dude do you think you could help me understand what's causing the error here? Thanks!

i pushed a fix for this, it was due to a ```.``` in your select columns rather than ```,```

hildobby:
> @hildobby i want to clarify the end goal for these two models. `first_funded_by` this is being built as a regular table, each time it runs, it'll read all history of `traces` and do a `create or replace table` -- is this the intention, or are we able to incrementally read from `traces`? if so, we'd need all the unique key info, tests, etc.

Yes exactly, unfortunately there is no way that I could think of to make this table incremental (unless it can call itself but as far as I know that is not possible?). I added unique key there: `address` since there should only ever be one line per address.

> `nft_wash_trades` this is built as a view as of now, however, there is incremental syntax included. i don't believe those will ever be applied in the view. do you intend to make this incremental table too? if so, similar to above changes

I changed this table's config to make it an incremental table, I also added unique key there, should be `unique_trade_id`, just like all the other tables under the `nft` schema name.

> the query for wash trades is still failing. it's an issue with the case statements. please use this as latest query to continue to debug and push fixes:

I changed the query to that one now for debugging. What's the issue with case statements? Is there no way to check for those working properly currently?

For the tests I do indeed plan on adding them, I just wanted to make sure that github's continuation integration gives a green light on the queries before I go about making tests for them. Will start writing them and should have them pushed very soon.

jeff-dude:
> > @hildobby i want to clarify the end goal for these two models. `first_funded_by` this is being built as a regular table, each time it runs, it'll read all history of `traces` and do a `create or replace table` -- is this the intention, or are we able to incrementally read from `traces`? if so, we'd need all the unique key info, tests, etc.
> 
> Yes exactly, unfortunately there is no way that I could think of to make this table incremental (unless it can call itself but as far as I know that is not possible?). I added unique key there: `address` since there should only ever be one line per address.
> 
> > `nft_wash_trades` this is built as a view as of now, however, there is incremental syntax included. i don't believe those will ever be applied in the view. do you intend to make this incremental table too? if so, similar to above changes
> 
> I changed this table's config to make it an incremental table, I also added unique key there, should be `unique_trade_id`, just like all the other tables under the `nft` schema name.
> 
> > the query for wash trades is still failing. it's an issue with the case statements. please use this as latest query to continue to debug and push fixes:
> 
> I changed the query to that one now for debugging. What's the issue with case statements? Is there no way to check for those working properly currently?
> 
> For the tests I do indeed plan on adding them, I just wanted to make sure that github's continuation integration gives a green light on the queries before I go about making tests for them. Will start writing them and should have them pushed very soon.

please see this model from 0xRob on how he uses logic in lines 132-167 for first-time run/full refresh vs. incremental and if you can reference your own model with {{ this }} syntax.
note: i did not look closely to see if this applies here, but thought i'd suggest. otherwise, it's likely fine to leave as regular table (although our current prod job won't run in the half-hourly refresh, only incrementals)

as for the wash trading query issues, i meant to suggest that you take the query i used above and run on dune until the case statements are fixed, then apply fixes here. i didn't intend for that query to be applied here

hildobby:
Ah very interesting! Unfortunately I dont think this would apply here as the entire history since Ethereum's first block is needed within the run to get an accurate list. I tried to think of a way for it to be incremental but all I could think of was a recursive method and I'm not sure if that's possible and also probably wouldn't be the best idea so I chose to not go with that route.

models/nft/nft_wash_trades.sql
```diff
@@ -0,0 +1,112 @@
...
+    , nftt.tx_from
+    , nftt.tx_to
+    , nftt.block_time
+    , date_trunc('day', nftt.block_time) AS block_date
+    , nftt.block_number
+    , nftt.tx_hash
+    , nftt.unique_trade_id
+    , filter_funding_buyer.first_funded_by AS buyer_first_funded_by
+    , filter_funding_seller.first_funded_by AS seller_first_funded_by
+    , count(1)
```
jeff-dude, line 36: the commented out CASE statements below still need resolved, and this count can be removed. that was only used as reference for me to share where the query stood.

you'll need these case output columns for final spell i imagine

models/nft/nft_wash_trades.sql
```diff
@@ -0,0 +1,112 @@
...
+    --         filter_funding_buyer.first_funded_by = filter_funding_seller.first_funded_by
+    --         AND filter_funding_buyer.first_funded_by NOT IN (SELECT DISTINCT address FROM labels.bridges)
+    --         AND filter_funding_buyer.first_funded_by NOT IN (SELECT DISTINCT address FROM labels.cex)
+    --     )
+    --     OR filter_funding_buyer.first_funded_by = nftt.seller
+    --     OR filter_funding_seller.first_funded_by = nftt.buyer
+    --     THEN true
+    --     ELSE false
+    --     END AS is_wash_trade
+FROM nft.trades nftt
```
jeff-dude, line 69: looks like all tables were reverted to hardcode, should be back to refs/sources

models/nft/nft_wash_trades.sql
```diff
@@ -0,0 +1,112 @@
...
+    , nftt.aggregator_address
+    , nftt.tx_from
+    , nftt.tx_to
+    , nftt.block_time
+    , nftt.block_number
+    , nftt.tx_hash
+    , nftt.unique_trade_id
+    , filter_funding_buyer.first_funded_by
+    , filter_funding_seller.first_funded_by
+ order by count(1) desc
```
jeff-dude, line 111: order by likely can be removed as well

jeff-dude (changes requested):

jeff-dude:
> Ah very interesting! Unfortunately I dont think this would apply here as the entire history since Ethereum's first block is needed within the run to get an accurate list. I tried to think of a way for it to be incremental but all I could think of was a recursive method and I'm not sure if that's possible and also probably wouldn't be the best idea so I chose to not go with that route.

we can explore leaving this one as just a table, we'll see how that works

hildobby:
Not sure why the tests job was aborted, I've added tests (and a 4th criteria), the PR should be ready. Not sure how to properly test it given that case statements seem to fail either way every time here, any suggestions @jeff-dude? If it's all good, it is ready to merge.

jeff-dude:
i pushed a commit to try and make ```first_funded_by``` incremental -- what do you think? basically just says if already exists, filter out in the subquery.

as for ```nft_wash_trades``` -- we're getting that ambiguous "job aborted" error again, similar to the old days of tornado cash spells 🤔 the issue went away previously when we switched to a SQL endpoint compute, but it appears to have resurfaced. i'll have to dig further on that one

hildobby:
Ohhh I didn't know you could make recursive calls in speelbook. I would have just added `WHERE et.to NOT IN (SELECT address FROM {{this})` to my table and made it incremental if I knew `{{this}}` was a thing but I think your method also works fine!

hildobby:
> as for `nft_wash_trades` -- we're getting that ambiguous "job aborted" error again, similar to the old days of tornado cash spells 🤔 the issue went away previously when we switched to a SQL endpoint compute, but it appears to have resurfaced. i'll have to dig further on that one

Ah damn that's annoying, let me know if I can help somehow! I would love to have this merged soon so I can start publishing some stats using it, it's holding back most of the dashboards I want to publish at the moment

jeff-dude:
i'm trying to deep dive into this wash trades query to see the issue(s). please see query here with different CTEs viewing things differently.
https://dune.com/queries/1458213?d=1

after looking into first few CTEs, i ended up in ```final_model_query``` CTE which is your query with filters for incremental and focused on one specific ```unique_trade_id``` which is returned in the prior CTEs (which i assume means a candidate for wash trading).

the query runs if i comment out the final two case statements. when uncommenting, it fails. can you focus in on those two closer?

here is error message in databricks directly:
```
Job aborted due to stage failure: Couldn't find exists#172891 in [blockchain#163634,project#163635,version#163636,nft_contract_address#163652,token_id#163638,token_standard#163641,trade_category#163644,buyer#163647,seller#163646,project_contract_address#163653,aggregator_name#163654,aggregator_address#163655,tx_from#163658,tx_to#163659,block_time#163637,block_number#163656,tx_hash#163657,unique_trade_id#163660,first_funded_by#172105,first_funded_by#172110,count(filter_baf.block_number#172883)#172335L,count(filter_bought_3x.block_number#172884)#172336L,first(count(filter_baf.block_number)#172887L) ignore nulls#172888L,first(count(filter_bought_3x.block_number)#172889L) ignore nulls#172890L]
```

my assumption is those case statements using ```NOT IN``` may be the issue? or the ```OR``` statements not used correctly? not 100% sure, but that's where to focus.

please try to see where you can take this next 🙏

hildobby:
Hey @jeff-dude thanks for looking into it! I altered the abstraction to remove the need for those `NOT IN` and `OR` statements in the lat 2 CTEs anymore, hopefully the tests pass this time around 🙏

hildobby:
Hey @jeff-dude thanks a lot for looking into it! I altered the abstraction to remove the need for those `NOT IN` and `OR` statements in the lat 2 CTEs. It seems to compile properly now.

The tests are failing on uniqueness test for the `nft.wash_trades`'s `unique_trade_id` but I am fairly certain this comes from uniqueness test also failing for the `nft.trades`'s `unique_trade_id` since `nft.wash_trades` relies on `nft.trades`.

Let me know if there still are some issues I failed to spot here!

jeff-dude:
> Hey @jeff-dude thanks a lot for looking into it! I altered the abstraction to remove the need for those `NOT IN` and `OR` statements in the lat 2 CTEs. It seems to compile properly now.
> 
> The tests are failing on uniqueness test for the `nft.wash_trades`'s `unique_trade_id` but I am fairly certain this comes from uniqueness test also failing for the `nft.trades`'s `unique_trade_id` since `nft.wash_trades` relies on `nft.trades`.
> 
> Let me know if there still are some issues I failed to spot here!

ah amazing, you're past the model creation 🙌  did you get a chance to query the table under the github test schema? it only persists 24 hours, so if not, you'd have to rerun the gh action and see. then you can grab table name and validate anything.

i'll look closer at the tests in the meantime

tests/addresses_events/addresses_events_ethereum_first_funded_by_assert_unique.sql
```diff
@@ -0,0 +1,6 @@
+-- Check for multiple addresses
+
+SELECT address
```
jeff-dude, line 3: you don't need this test at all, you can simply apply the built-in unique test on the schema file for this model (tests: -unique) under the column ```address```

tests/nft/nft_wash_trades_assert_unique.sql
```diff
@@ -0,0 +1,8 @@
+-- Check for multiple unique_trade_id
+
+SELECT unique_trade_id
```
jeff-dude, line 3: this test makes more sense as you are filtering for last two days. we could also build a generic test which is a dbt macro that accepts table name, column name, number of days to filter back. note that i already have a unique_filtered generic macro under generic tests, you could apply similarly to my other comment, but use - unique_filtered instead (assuming a week is okay)

jeff-dude: https://github.com/duneanalytics/spellbook/blob/main/tests/generic/is_unique_filtered.sql

jeff-dude: we can edit that macro to be hardcoded to last two days vs. the select max(), as there have been issues with that approach. i don't believe this macro is used much, so not a harm to edit

hildobby:
> ah amazing, you're past the model creation 🙌 did you get a chance to query the table under the github test schema? it only persists 24 hours, so if not, you'd have to rerun the gh action and see. then you can grab table name and validate anything.
> 
> i'll look closer at the tests in the meantime

Yes I looked and it looks good to me :)

hildobby:
So you suggest removing the test I made and using that macro instead? How do I do the latter? @jeff-dude

jeff-dude:
> So you suggest removing the test I made and using that macro instead? How do I do the latter? @jeff-dude

i'll push something in a sec :)

hildobby:
> i'll push something in a sec :)

Nice ty @jeff-dude! It looks like it's failing but I'm fairly certain that is because of `nft.trades` failing unique tests as well. PR LGTM, lmk if there is something else that needs to be done :)

jeff-dude:
> > i'll push something in a sec :)
> 
> Nice ty @jeff-dude! It looks like it's failing but I'm fairly certain that is because of `nft.trades` failing unique tests as well. PR LGTM, lmk if there is something else that needs to be done :)

looks like you're right on the nft trades 🤔 i'll have to see, otherwise i do believe this is good

jeff-dude (approved):

jeff-dude:
ignore that last commit, was from meghan and i messing around :)

PR #1628, opened 2022-09-24
[Prices] Add wstETH, rETH and TEMPLE

bh2smith:
While preparing a query to display Balancer Pool Balances, I noticed that some of the high valued pools had missing token prices. So far only selected the obvious ones (wstETH, rETH, TEMPLE) - it also appears we will need to do something about the Balancer Boosted Aave Pools because coin paprika only seems to have one of the three (`bb-a-USDC`, `bb-a-USDT`, `bb-a-DAI`).

Resources for comparison:

**Balancer Pool Page:** https://app.balancer.fi/#/

Screenshot 2022-09-24 at 8 51 16 AM

**Dune Query:** https://dune.com/queries/1303739?d=1

Screenshot 2022-09-24 at 8 52 51 AM

**Coin Paprika Links:**

wstETH: https://coinpaprika.com/coin/wsteth-wrapped-liquid-staked-ether-20/
rETH: https://coinpaprika.com/coin/reth-rocket-pool-eth/
TEMPLE: https://coinpaprika.com/coin/temple-templedao/

### DIFF
--- a/models/prices/prices_tokens.sql
+++ b/models/prices/prices_tokens.sql
@@ -581,7 +581,9 @@ VALUES
     ("eure-monerium-eur-emoney", "ethereum", "EURe", "0x3231Cb76718CDeF2155FC47b5286d82e6eDA273f", 18),
     ("fort-forta", "ethereum", "FORT", "0x41545f8b9472D758bB669ed8EaEEEcD7a9C4Ec29", 18),
     ("caps-ternoa", "ethereum", "CAPS", "0x03be5c903c727ee2c8c4e9bc0acc860cca4715e2", 18),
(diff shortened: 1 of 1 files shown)


soispoke (approved):
lgtm, thanks @bh2smith

bh2smith:
Just wondering.... how many tokens do you think we could add at once? I recall there was an issue with the legacy platform and scalability when trying to add 700 tokens, is this something the new engine could handle?

soispoke:
@bh2smith 700 tokens should be fine, we're working on specific processes to handle 10k + files right now but it doesn't seem to fall into that category so a normal submission should do it :)

aalan3:
@bh2smith @soispoke if you plan to add 700 tokens, please do so in batches. It will make reviewing easier and there might be some downstream limits that we need to test.

PR #1629, opened 2022-09-24
[ERC20] Add Missing Tokens

bh2smith:
- Adds 13 missing tokens on V1, fetched from https://dune.com/queries/857522
- Adds 28 missing tokens on V2, fetched from https://dune.com/queries/1317323


jeff-dude:
@bh2smith looks like you may be adding duplicates
```
select
    contract_address as unique_field,
    count(*) as n_records

from test_schema.git_78d40b9c_tokens_ethereum_erc20
where contract_address is not null
group by contract_address
having count(*) > 1
```

the gh action linked created that test table, then the applied unique test on address built this query, which results in addresses with duplicates

bh2smith:
Yea I think it's because of the issue earlier with the mixed case addresses. Will update and resubmit again tomorrow morning.

bh2smith:
Thanks @jeff-dude - good thing we have these checks in place (it was indeed an issue with the mixed case tokens and its all good now). The reduced list of V2 tokens added is 28.

jeff-dude (approved):
thanks @bh2smith 👍

PR #1634, opened 2022-09-26
Change sector to dex_aggregator

bh2smith:
In response to suggestion made in https://github.com/duneanalytics/spellbook/pull/1600#discussion_r979634669

we change the sector to `dex_aggregator`


soispoke (approved):
thanks, let's see if we can now expose it in the Dune App

PR #1638, opened 2022-09-26
[Bug] select LOWER missing on token tables

bh2smith:
[This query](https://dune.com/queries/1317291)  shows that there are hundreds of erc20 tokens with mixed case addresses (making integrity of token data inconsistent). This makes it consistent with the other token tables.

### DIFF
--- a/models/prices/prices_tokens.sql
+++ b/models/prices/prices_tokens.sql
@@ -7,7 +7,7 @@
                                     \'["aalan3", "jeff-dude"]\') }}'
         )
 }}
-SELECT token_id, blockchain, symbol, contract_address, decimals from (
+SELECT token_id, blockchain, symbol, LOWER(contract_address) as contract_address, decimals from (
 VALUES
     ("ada-cardano", null, "ADA", null, null),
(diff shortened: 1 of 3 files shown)


soispoke (approved):
Thanks @bh2smith 

tagging @aalan3 for visibility

soispoke (approved):
Good catch, thanks @bh2smith

bh2smith:
How long until we see these changes in the interface?

jeff-dude:
while not for lower checks, we have tests to validate no duplicates in the manual token views (as well as accepted values) for fields. these are built-in dbt tests.
https://github.com/duneanalytics/spellbook/blob/main/models/tokens/ethereum/tokens_ethereum_schema.yml

we could likely introduce a pretty simple generic custom test for lower function on the address too. you can find other generic tests here:
https://github.com/duneanalytics/spellbook/tree/main/tests/generic

then can apply the new test into respective schema file and column. you're already doing so much for us, it's understandable to make this one lower priority 👍 we may be able to do at some point soon too, just need to find a good time.

PR #1641, opened 2022-09-26
[Easy] Removed post_hook from `cow_protocol_ethereum_trades.sql` to not have double columns

gentrexha:
Removed post_hook from `cow_protocol_ethereum_trades.sql` to not have double columns in `cow_protocol.trades`. E.g.: 
image

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
@@ -5,11 +5,7 @@
         unique_key = ['tx_hash', 'order_uid', 'evt_index'],
         on_schema_change='sync_all_columns',
         file_format ='delta',
-        incremental_strategy='merge',
-        post_hook='{{ expose_spells(\'["ethereum"]\',
-                                    "project",
-                                    "cow_protocol",
-                                    \'["bh2smith", "gentrexha"]\') }}'
+        incremental_strategy='merge'
     )
 }}
 


bh2smith:
Looks alright to me... I am not entirely sure how the name space separation works just yet so, I will withhold approval.

jeff-dude:
@gentrexha i see your point to the explorer showing fields more than once, but removing post-hook isn't the solution. we will need to look closer at how this happened. my assumption is: first deploy ran, post-hook ran. changes were made, post-hook ran again, but rather than overwrite, it appended

jeff-dude:
tagging @dot2dotseurat and @soispoke for awareness, we will need to check on our side

soispoke:
@gentrexha After some discussions we think removing the post hook from the `cow_protocol_ethereum_trades` model might actually be the right call and solve the problem, since what was happening is that we attributed the exact same metadata (tblproperties) to two distinct tables. I'll merge this and deploy, hoping this fixes the issue but this isn't 100% guaranteed to work 🙏

soispoke (approved):

soispoke:
Alright, looks like this worked (but please feel free to check in the data explorer to make sure everything looks good to you guys) @gentrexha @bh2smith

bh2smith:
I am not sure this is looking like we would expect... now the only remaining table is the new one. I was hoping that there would be two - One which conforms to the columns in dex_trades and one with the columns we have specified. Now there is only the conforming one.

Screenshot 2022-09-28 at 6 49 28 PM

Screenshot 2022-09-28 at 6 49 39 PM


@gentrexha - I think we are going to have to give it a different name to end this.

soispoke:
@bh2smith Ok, wasn't sure you wanted to keep both tables visible. If that's the case and it's on the same blockchain, then the most straightforward way to expose both is indeed to have a different name for the underlying table

bh2smith:
I am going to submit a PR to fix this, because it is causing problems for others now. cc @fleupold

PR #1645, opened 2022-09-27
sudoswap pool spells [rework]

0xRobin:
I noticed some bugs in the recently deployed sudoswap pool models.
Fixing them turned into a rework of the different sub tables.

1. split of the pool creation info in its own `pool_creations` table for storage optimization (this data is static so should not be duplicated)
2. split of the calculation of the latest pool settings into its own `pool_settings_latest` table.
This uses an experimental structure with self-reference, not sure it'll work..
([This line](https://github.com/duneanalytics/spellbook/blob/0c5b439f2006c96f045c4c50feabcb82a3629209/models/sudoswap/ethereum/sudoswap_ethereum_pools.sql#L26) was the origin of the bug where previous design only returned 1 row.)
3. some renaming of tables/columns for consistency over spells.
~~`pair_address`~~ -> `pool_address`
~~`owner_fee`~~ -> `pool_fee`

**models**:
new: `sudoswap_ethereum.pool_creations`
new: `sudoswap_ethereum.pool_settings_latest`
rework: ~~`sudoswap_ethereum.pool_balance_changes`~~ -> `sudoswap_ethereum.pool_balance_agg_day`
rework: ~~`sudoswap_ethereum.pool_trades`~~ -> `sudoswap_ethereum.pool_trade_agg_day` 
rework: ~~`sudoswap_ethereum.pools`~~ -> `sudoswap_ethereum.pool_view` 

---

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


0xRobin:
Ok this was an uphill battle, now ready for review. :heavy_check_mark:
Not sure why the dbt_utils rowcount tests didn't work out.

cc: @jeff-dude, @Jam516

models/sudoswap/ethereum/sudoswap_ethereum_pool_settings_latest.sql
```diff
@@ -0,0 +1,168 @@
...
+        ,coalesce(t1.latest_update_time,t3.creation_block_time) as latest_update_time
+        from latest_settings t1
+        full outer join initial_settings t3
+            ON t1.pool_address = t3.pool_address
+        left join {{ this }} t2
+            ON t1.pool_address = t2.pool_address
+    ) foo
+    where bonding_curve is not null --temp hack to exclude updates form erc20 pools
+)
+{% endif %}
```
0xRobin, line 148: @jeff-dude please take a closer look at this setup as it uses the existing table to fill empty columns on an incremental update.

Jam516:
> Ok this was an uphill battle, now ready for review. ✔️ Not sure why the dbt_utils rowcount tests didn't work out as I checked the compiled queries manually and they should have succeeded.
> 
> cc: @jeff-dude, @Jam516

Looks great! Thanks for this fix Rob 🙏
Please take a look @jeff-dude

jeff-dude:
> > Ok this was an uphill battle, now ready for review. ✔️ Not sure why the dbt_utils rowcount tests didn't work out as I checked the compiled queries manually and they should have succeeded.
> > cc: @jeff-dude, @Jam516
> 
> Looks great! Thanks for this fix Rob 🙏 Please take a look @jeff-dude

sounds good, will look this afternoon

models/sudoswap/ethereum/sudoswap_ethereum_pool_creations.sql
```diff
@@ -0,0 +1,57 @@
+{{ config(
+        alias = 'pool_creations',
+        partition_by = ['pool_address'],
```
jeff-dude, line 3: i don't think it makes sense to partition by ```pool_address``` here, as it's unique and only one row per pool. spark doesn't love a ton of small files. we can remove partition altogether or look to use a different field, which would be commonly used in joins/filters for this table

jeff-dude (changes requested):

models/sudoswap/ethereum/sudoswap_ethereum_pool_settings_latest.sql
```diff
@@ -0,0 +1,168 @@
+{{ config(
+        alias = 'pool_settings_latest',
+        partition_by = ['pool_address'],
```
jeff-dude, line 3: similar idea here

jeff-dude: although the test is failing on this model for unique, so it's not exactly the same as the other

models/sudoswap/ethereum/sudoswap_ethereum_pool_creations.sql
0xRobin: Ah yes, those shouldn't have stayed in.
Would be cool if we could partition on unique columns in the future with something like a mod 256 or so.

models/sudoswap/ethereum/sudoswap_ethereum_pool_settings_latest.sql
0xRobin: Ok, this should be unique.
Any idea why the general unique tests don't show up in the dbt slim ci?

0xRobin: ah, ~~test~~ -> tests. got it

0xRobin:
@jeff-dude
Any idea why the dbt smil ci could be hanging (after 50+min) on 
`START incremental model test_schema.git_d7685baf_sudoswap_ethereum_pool_balance_agg_day`

The compiled query for `pool_balance_agg_day` runs in ~6min on Dune..
https://dune.com/queries/1328722

(this query also includes the compiled pool_creations subquery)

jeff-dude:
> @jeff-dude Any idea why the dbt smil ci could be hanging (after 50+min) on `START incremental model test_schema.git_d7685baf_sudoswap_ethereum_pool_balance_agg_day`
> 
> The compiled query for `pool_balance_agg_day` runs in ~6min on Dune.. https://dune.com/queries/1328722
> 
> (this query also includes the compiled pool_creations subquery)

that's definitely odd. we'll look closer as we continue to review this

macros/optimize_tables.sql
```diff
@@ -27,8 +27,12 @@ OPTIMIZE safe_ethereum.safes;
 OPTIMIZE seaport_ethereum.transfers;
 {% endset %}
 
-{% set sudoswap_ethereum_events %}
+{% set sudoswap_ethereum %}
 OPTIMIZE sudoswap_ethereum.events;
+OPTIMIZE sudoswap_ethereum.pool_creations;
```
jeff-dude, line 32: i wish this worked, but i'm being told from the team that each set has to include only one optimize statement, which is why all others in the file are set that way :(

jeff-dude (changes requested):

macros/optimize_tables.sql
0xRobin: Ah, makes sense since you can only execute 1 statement in a do_run.

jeff-dude:
i still see duplicates when running the test:
```dbt test --select sudoswap_ethereum_pool_settings_latest```
are you sure the ```unique_key``` in the model is as expected?

**Comment by jeff-dude** [comment]
> i still see duplicates when running the test: `dbt test --select sudoswap_ethereum_pool_settings_latest` are you sure the `unique_key` in the model is as expected?

ah, i found the fix. check my latest commit on fix to join condition. that was resulting in blowing up to 450 million+ rows

**Comment by jeff-dude** [comment]
as for the ```pool_balance_agg_day``` performance, i'm messing with the below (focused on traces CTE). it seems when the join condition includes both ```to``` OR ```from``` it just hangs. when i do only one of them in the join, it returns in reasonable time. if we broke them out into multiple CTEs, then it'd read from traces twice, which may not be ideal but could be an approach to speed it up regardless

```
WITH
  pools AS (
    SELECT
         pool_address
        ,nft_contract_address
    FROM dbt_jeff_sudoswap_ethereum.pool_creations
  ),

  -- erc721_deltas AS (
  --   SELECT
  --     date_trunc('day', et.evt_block_time) AS day,
  --     pool_address,
  --     SUM(CASE WHEN et.to = p.pool_address THEN 1 ELSE -1 END) AS nft_balance_change
  --   FROM
  --     erc721_ethereum.evt_transfer et
  --     INNER JOIN pools p ON p.nft_contract_address = et.contract_address
  --     AND (et.to = p.pool_address OR et.from = p.pool_address)
    
  --   WHERE et.evt_block_time >= '2022-09-23'
    
    
  --   GROUP BY
  --     1,2
  -- ),

  eth_deltas AS (
    SELECT
      date_trunc('day',tr.block_time) AS day,
      pool_address,
      SUM(CASE WHEN tr.to = pc.pool_address THEN tr.value/1e18 ELSE -1*tr.value/1e18 END) AS eth_balance_change
    FROM
      ethereum.traces tr
      -- INNER JOIN pools pc ON (pc.pool_address = tr.to OR pc.pool_address = tr.from)
      INNER JOIN pools pc ON pc.pool_address = tr.to
      -- INNER JOIN pools pc ON pc.pool_address = tr.from
    WHERE tr.success = true
      AND tr.type = 'call'
      AND (
        tr.call_type NOT IN ('delegatecall', 'callcode', 'staticcall')
        OR tr.call_type IS null
      )
      
      AND tr.block_time > '2022-09-23'
      
      
    GROUP BY
      1,2
  )

select count(1)
from eth_deltas
;
```

0xRobin:
@jeff-dude ok thanks for looking into this!
I'm really constrained on debugging this since I can't run it locally.
I've introduced the `to OR from` condition especially to only scan the traces table once, which should (in theory) improve performance.

To move this PR forward it's probably best to split it up and continue.

But it will definitely be worth it to extract a minimal example and try to get to the bottom of this why it hangs on dbt run. (as it performs fine as a dune query)
I'll take this back up on monday, enjoy the weekend. :)

0xRobin:
@jeff-dude, tests passing now. Should be clear to merge.

jeff-dude (approved):
think we're all set 👍

Jam516:
> think we're all set 👍

Awesome! Thanks so much guys 🙇🏾‍♂️

PR #1646, opened 2022-09-27
Add Rarible to nft.trades on Dune v2

hildobby:

Add Rarible NFT Marketplace data to nft.trades on Dune v2.

*For Dune Engine V2*
I've checked that:
General checks:
* [X] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributors)

Join logic:

Incremental logic:


soispoke:
I've fixed a number of different small issues, but there is still one you should take a look at @hildobby 

We have duplicates generated by the same trade (row) being attributed two different `trade_category` (Buy and Accepted Offer), see here for an example: https://dune.com/queries/1320040?d=1

hildobby:
> I've fixed a number of different small issues, but there is still one you should take a look at @hildobby
> 
> We have duplicates generated by the same trade (row) being attributed two different `trade_category` (Buy and Accepted Offer), see here for an example: https://dune.com/queries/1320040?d=1

That query shows a pool from curve, don't think that's what you wanted to show, right?

hildobby:
I'm a little confused as to why the tests stopped mid way through, but I think this PR is ready.

dot2dotseurat:
> I'm a little confused as to why the tests stopped mid way through, but I think this PR is ready.

It times out after 60 min. :/

soispoke:
I pushed a small fix so tests should all pass now

hildobby:
Awesome! Rarible's main contract is upgradable and was upgraded on September 6th, I just resubmitted it for it to be decoded so I can add abstraction for txs after sept 6th.

dot2dotseurat:
> Awesome! Rarible's main contract is upgradable and was upgraded on September 6th, I just resubmitted it for it to be decoded so I can add abstraction for txs after sept 6th.

Does that mean I should wait to reach out to Rarible for a review?

hildobby:
> Does that mean I should wait to reach out to Rarible for a review?

The rest is ready to be checked already, just waiting for Antonio to decode the contract and I'll add the abstraction for post ABI upgrade, I'll keep you posted!

dot2dotseurat:
From Ekaterina at Rarible:

I've looked through git code 
and I have some questions : 
1. Rarible as aggregator - I see that there is the only way to identify it - by ethereum.transactions to , but as for seaport contracts (and for aggregators info)  we use have special marker in data: last 8 hex symbols: 9616c6c64617461 , could you add this into aggregators identification ? 
As I know other aggregators have the same issue to identify their transactions , how do they solve this issue? 
I liked one report based on nft.trades (from v2 ) , but I don't know how fairly all aggregators were marked, 
because if we use this table as is - share of aggregators is not very high, so it's interesting to investigate this assumption further 

2. I've tried to find some txn made through rarible, but they had lost, i.e. https://etherscan.io/tx/0x910f31c9f25649a4da468390819b689c01b8aa2c2803c9eb56045e5583c128ef 
actually I've just checked today, in ethereum.transactions  I could find it, but still I couldn't find in nft.trades (yesterday I couldn't find in both tables) 
 
I''m happy to jump into call also, and discuss these questions

PR #1653, opened 2022-09-28
add RPL and wstETH

markusbkoch:
So as to improve the accuracy of metrics in dex.trades


jeff-dude (approved):

jeff-dude:
fyi
- wstETH already existed in the file, so no new prices are added there
- RPL is still returning the old contract address in the API call it appears, those prices are still loading. you can see on etherscan the contract migrated, but i guess the API still pulls the old one

Issue #1651, opened 2022-09-28
V2: Multiple Duplicate Records in ens_resolver_addresses


bh2smith:
[This query](https://dune.com/queries/1324487?d=1&category=abstraction&namespace=ens&table=view_registrations&blockchains=ethereum) demonstrates 512 duplicates for a single ens name:

https://github.com/duneanalytics/spellbook/blob/main/models/ens/ens_resolver_addresses.sql

I haven't had chance to check the reason for this.

0xRobin:
> Seems to be related to #1569
Indeed related, I'm gonna get started on an ENS spell rework tomorrow that should fix it. 
I'll probably prioritize the reverse ens records first.

bh2smith:
@0xRobin - I see the relented branch was merged. Does this close the issue?

0xRobin:
@bh2smith yeah, reverse records should be fine now and shouldn't contain duplicates.
Still working on the resolve records (ens -> address)

bh2smith:
I think there still may be somethign subtly off about this table.... unless maybe I misunderstood the meaning. but this pair

`(bahman.eth, 0x428824314a823c89241EBCdfCccA8932C4573c4a)` doesn't show up in the table at all if one searches for either address or name.

Does the reverse lookup have somethign to do with wallets that registered the domain?

jeff-dude:
closing as this spell no longer exists (or was rebuilt / renamed)

Issue #1666, opened 2022-09-29
ENS spellbook integrations overview.


0xRobin:
Overview of ENS integrations and views.
Aggregated from discussions in https://github.com/duneanalytics/spellbook/pull/1569
There's already models for registrations/renewals/expires so we want to add resolver records and reverse records.
This can then be used in the v2 labels. 

---
### 1. Reverse names
**model**: `ens_reverse_latest`
**status**: Done ✔️ https://github.com/duneanalytics/spellbook/pull/1671 
| field | type |
| ------------- | ------------- |
| address | string **[unique]** |
| name | string (readable) |
| address_node | string (namehash of address reverse record) |
| latest_tx_hash | string |
| latest_tx_block_time | datetime |

! currently this does not check correctness.
(we should check that `resolve(ens_name) = address` also holds true. )

### 2. Resolver records
**model**: `ens_resolver_latest`
**status**:   Done ✔️
| field | type |
| ------------- | ------------- |
| node | string (namehash) |
| name | string (readable) |
| address | string |
| latest_tx_hash | string |
| latest_tx_block_time | datetime |

**todo**:
- ~~update field names to be in line with https://docs.ens.domains/terminology~~

0xRobin:
Quick overview of what I found on the resolver record side.

1) The namehash function was made available in v1 [here](https://github.com/duneanalytics/spellbook/blob/main/deprecated-dune-v1-abstractions/ethereum/public/namehash.sql)
BUT, it wasn't used for any v1 abstractions. Meaning this would only be used to parse ENS names as query parameters. There were however no real ENS tables with node information available in v1, so I don't think this actually got any use.
@hildobby, did you see any other uses of the namehash in v1?

2) The old ENS labels (ens name) in v1 were **just based on registrations**, meaning they didn't track the right current resolver records. This can be seen [here](https://github.com/duneanalytics/spellbook/blob/main/deprecated-dune-v1-abstractions/labels/ethereum/ens_names.sql). Although this probably captures a lot of the correct records, it's not functionally correct.

I'm still figuring out if we can capture all the correct `<address , node (namehash), name (readable)>` records from just the contract calls and events, which means we won't need the namehash function to provide resolver records as labels. 

Tagging for visibility so we can concentrate any future ENS discussion here. :)
CC: @hildobby @jeff-dude @soispoke @makoto

PR #1671, opened 2022-09-29
ENS reverse records

0xRobin:

Introduces a new ENS reverse record view that only shows the latest reverse records. 
This does not currently check for backward correctness (whether `resolve(ens) = address`)
Uses explicit traces filtering to make it more performant (alternative is unioning a bunch of event tables and joining on traces to retrieve trace.to)

Fixes issue https://github.com/duneanalytics/spellbook/issues/1651
First part of epic: https://github.com/duneanalytics/spellbook/issues/1666

---

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


catherine-2:
Thank you @0xRobin for putting this together!! 🙏  Was just looking to see if this existed this morning

0xRobin:
Tests look good.
This can be merged and https://github.com/duneanalytics/spellbook/pull/1569 should be closed then.
cc: @soispoke @jeff-dude, @hildobby

models/ens/ens_ethereum_reverse_latest.sql
```diff
@@ -0,0 +1,62 @@
+{{ config(
```
soispoke, line 1: I think the model  should be named `ens_reverse_latest`

If you prefer to keep the ethereum part in the name, then there needs to be a modification in the schema added in the dbt_project.yml file

0xRobin: ok, I tried to be consistent here with other model definitions where we use `<project>_<chain>_<table>`

0xRobin: Should we change the other models to use `ens_ethereum` and adapt the dbt_project.yml, or revert here to `ens_`?

soispoke: yeah I think it was a good idea, I'm actually not sure if ENS is only on the mainnet Ethereum chain ? If that's the case, we can just get rid of ethereum in the name. If not, we should just add it to the dbt_project.yml

0xRobin: For now seems to just be on mainnet eth. I'll revert the change.

soispoke (approved):
lgtm ! thanks @0xRobin

0xRobin:
@hildobby just realized this change is probably going to break your tornadoxxed queries.

bh2smith:
Hey there! congrats on landing this (it looks like its working well). Maybe I am confusing reverse with resolver but I made this query based on this table and am unable to find an entry I was expecting to see:

[This Query](https://dune.com/queries/1332079) whose default values are the record I was expecting to find do not give any results in either direction (specifying the name or wallet address that it resolves to).

Note that this is a "kinda" special case where the domain was purchased originally by someone and then the wallet which it resolves to was set to a different account.

Hopefully this helps with testing/debugging.

0xRobin:
@bh2smith you're indeed confusing reverse/resolver,
resolve record: (`bahman.eth` -> `0x428824314a823c89241ebcdfccca8932c4573c4a`)
We don't have a working model for this on v2 yet.

reverse record: (0x428824314a823c89241ebcdfccca8932c4573c4a -> null)
As we can see here:
https://app.ens.domains/address/0x428824314a823c89241ebcdfccca8932c4573c4a
The reverse record for this address is not set, which is why the query returns no results.

0xRobin:
Reverse records are mainly used for display only, to give an address a readable web3 'username'.
They are limited to 1 reverse record for each address, while there can be multiple ENS names that resolve to the same address.

So if you want to make a table that includes a bunch of user addresses more readable, you can join on reverse ENS names.

If you want to have an ENS as a user input parameter for a dashboard, you need to lookup what address it resolves to. This is not possible (and not the goal) with this model. Hope that clears things up.

bh2smith:
This is what I was hoping to get an end name from and address. In the sample query I gave, this address is not listed the reverse record table.

PR #1680, opened 2022-09-30
optimize spell macro

dot2dotseurat:

The purpose of this change to to replace the optimize_table macro with a macro we can apply at the dbt_project level that runs on all models conditionally. 

The goal is to remove the manual step of adding a macro for each table/incremental table.

We can't* shouldn't* use the githib action because this change affects all models.


dot2dotseurat:
Screen Shot 2022-09-30 at 2 40 07 PM

I can see in databricks the command is being trigger from a local run

jeff-dude:
> Screen Shot 2022-09-30 at 2 40 07 PM
> 
> I can see in databricks the command is being trigger from a local run

```tokens_avalanche_c.nft``` is supposed to be a view though 🤔  i wonder if it's because in the dbt_project file, that object doesn't have default materialization set? it's not set in the model either (as expected, since we assume view unless overwritten)

dot2dotseurat:
Screen Shot 2022-09-30 at 5 00 51 PM
confirming test with gh action ran the correct command. 

I'm going to remove the last commits that I used to trigger the test.

jeff-dude (approved):
nice simple solution, i like it. maybe we should wait until monday to deploy?

dot2dotseurat:
> nice simple solution, i like it. maybe we should wait until monday to deploy?

sounds good to me.

soispoke:
looks good to me!

soispoke (approved):

jeff-dude:
> looks good to me! 🎉 The only question I have is: If I understand this properly, the optimize command is going to run every time we update incremental models ? If that's the case, it might be a bit of an overkill (and a waste of computational ressources) ? It might not be, as I know the optimize command takes a long time/lot of compute the first time it runs on a model, but much less afterwards, just something we might want to consider

let's deploy today and see what happens over a 24-hour time period for performance/time of job? maybe first half-hour job is slow, then speeds up, as you mentioned?

soispoke:
@jeff-dude Yes, sounds good :)

dot2dotseurat:
> @jeff-dude Yes, sounds good :)

@jeff-dude @soispoke if it does, we can add in a incremental flag if statement and skip the command during the incremental runs

PR #1705, opened 2022-10-05
dex.trades Clipper integration

0xRobin:
Integrates Clipper DEX into the v2 dex.trades table.

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


0xBoxer:
lgtm, but wondering if we should keep this convention alive @jeff-dude?

https://github.com/duneanalytics/spellbook/blob/56350a3cf83ca631114fd3a39512aaddb1a904cd/deprecated-dune-v1-abstractions/ethereum/dex/trades/insert_clipper.sql#L69-L75

0xBoxer:
theoretically this problem of accounting for protocols that have raw ETH as trading pair arises the first time in uniswap_v1 here:

https://github.com/duneanalytics/spellbook/blob/56350a3cf83ca631114fd3a39512aaddb1a904cd/deprecated-dune-v1-abstractions/ethereum/dex/trades/insert_uniswap_v1.sql#L70

here we just swap the address with WETH, which doesn't seem like the optimal solution IMO, we should be able to differentiate between WETH and ETH within dex.trades.

0xRobin:
Since we're using strings instead of bytea in spellbook we could even do something like '0xETH' as address.
Just need to have it consists between dex.trades, nft.trades, token and prices tables.
Preferably something that even isn't a correct hex address so there's 0 confusion.

Swapping in the WETH address seems like a suboptimal solution..

models/clipper/ethereum/clipper_ethereum_trades.sql
```diff
@@ -0,0 +1,21 @@
+{{ config(
+        schema = 'clipper_ethereum_trades',
```
jeff-dude, line 2: please remove ```schema``` property here, the default from dbt_project yaml will take care of it. we don't want to override.

models/clipper/ethereum/clipper_v1_ethereum_trades.sql
```diff
@@ -0,0 +1,104 @@
+{{ config(
+    schema = 'clipper_v1-3_ethereum',
```
jeff-dude, line 2: any reason we don't follow the standard set in uniswap for a schema per version?
```clipper_v1_ethereum``` on the v1 model file here, same idea for 2 & 3 files

File: models/clipper/ethereum/clipper_v1_ethereum_trades.sql [hunk 3]
```diff
@@ -0,0 +1,104 @@
...
+    alias = 'trades_v1',
```
jeff-dude, line 3: following schema comment, this ideally is always just ```trades```

models/clipper/ethereum/clipper_v1_ethereum_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+    {% if is_incremental() %}
+    AND evt_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
+
+)
+
+SELECT
+    'ethereum' AS blockchain
+    ,'clipper' AS project
+    ,'v1' AS version
```
jeff-dude, line 42: can we make this just ```1```? it would follow approach from other models

jeff-dude: same to 2 & 3

seeds/dex/trades/dex_trades_seed.csv
```diff
@@ -22,4 +22,7 @@ ethereum,Curve,Regular,2022-09-03,0x0fc5910c411bdbbf1827fa278c22b3351892e8931e7c
...
 ethereum,Curve,Factory V1 Meta,2022-09-23,0x182e11e393388ea9365b1e566bfc66754bbdfb7f83344f6e1e4f05cde97ea66b,11,0x6c3f90f043a72fa612cbac8115ee7e52bde6e490,429398385988372154944,0xa693b19d2931d498c5b318df961919bb4aee87a5,10259457322
 ethereum,Curve,Factory V1 Plain,2022-09-19,0xe31b85292938984c61b4a9d8d2eb781f20bfa22c068a380a67fce94ef8ec450c,60,0xd533a949740bb3306d119cc777fa900ba034cd52,742474512665704726710,0x62b9c7356a2dc64a1969e19c23e4f579f9810aa7,774567425737246266396
-ethereum,Curve,Factory V2,2022-09-03,0x02fe546401dd52e07528532ed1ac4ed1d4f525118836f71a4014282fb6a06839,4,0xaf5191b0de278c7286d6c7cc6ab6bb8a73ba2cd6,51356173218735436675801,0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48,33333000000
\ No newline at end of file
+ethereum,Curve,Factory V2,2022-09-03,0x02fe546401dd52e07528532ed1ac4ed1d4f525118836f71a4014282fb6a06839,4,0xaf5191b0de278c7286d6c7cc6ab6bb8a73ba2cd6,51356173218735436675801,0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48,33333000000
+ethereum,clipper,v1,'2022-07-14',0x8d246ba8ec0775e6b7cff03010b858e22a9295d24b7fd08a8f705bbd3d333dab,301,0xdac17f958d2ee523a2206206994597c13d831ec7,165.266255,0x0000000000000000000000000000000000000000,0.082849
```
jeff-dude, line 26: make sure versions are updated here too

jeff-dude (changes requested):
thanks for the great PR! few minor comments below

models/clipper/ethereum/clipper_ethereum_schema.yml
```diff
@@ -0,0 +1,133 @@
+version: 2
+
+models:
+  - name: clipper_v1-3_ethereum
```
jeff-dude, line 4: i believe this would apply to model name (aka file name), not schema applied, right?

jeff-dude: if so, would need split out for each version

models/clipper/ethereum/clipper_ethereum_schema.yml
```diff
@@ -0,0 +1,133 @@
...
+    meta:
+      blockchain: ethereum
+      sector: dex
+      project: clipper
+      contributors: 0xRob
+    config:
+      tags: ['ethereum','clipper','trades','dex','0xRob']
+    description: >
+        Clipper trades on Ethereum
+    tests:
```
jeff-dude, line 96: also think we should apply these tests at each incremental model, rather than on the view, helps us catch it faster and easier to diagnose which version has issues

models/clipper/ethereum/clipper_ethereum_schema.yml
```diff
@@ -0,0 +1,133 @@
...
+      - *token_sold_symbol
+      - *token_pair
+      - *token_bought_amount
+      - *token_sold_amount
+      - *token_bought_amount_raw
+      - *token_sold_amount_raw
+      - *amount_usd
+      - name: token_bought_address
+        description: "Contract address of the token bought"
+        tests:
```
jeff-dude, line 122: similar concept on test here

jeff-dude (changes requested):
few more comments on schema file

jeff-dude:
> Since we're using strings instead of bytea in spellbook we could even do something like '0xETH' as address. Just need to have it consists between dex.trades, nft.trades, token and prices tables. Preferably something that even isn't a correct hex address so there's 0 confusion.
> 
> Swapping in the WETH address seems like a suboptimal solution indeed..

not sure the best path here @0xBoxer & @0xRobin, this is a dex trades sector wide design decision. how do you think we should go about approaching an answer?

we can always proceed with this PR, then have an overall PR which captures the concept in any existing dex trades entries thus far

models/clipper/ethereum/clipper_v1_ethereum_trades.sql
0xRobin: schema is the same for v1 to v3, so I wanted to reuse.
Will duplicate it if that doesn't work.

0xRobin:
@jeff-dude resolved the comments, all tests passing. :heavy_check_mark:

jeff-dude:
> @jeff-dude resolved the comments, all tests passing. ✔️

thank you! pushed a final commit:
1. remove tests from the clipper ethereum view -- no need to repeat after all three versions have the tests
2. add schema to config block on each version -- versions only time as of now we need to apply schema outside of dbt_project file, otherwise directories can get too nested potentially
3. fix block_date on v1 seed row

the gh action can be deceiving on a few of these things. for instance:
- since it uses different standards for schema/table name (test dummy), there were no issues, but in the real scenario, they'd all share the same schema/table name without it and fail, sadly only i can see that running locally right now
- the seed tests return false positive when join conditions don't work, fixing the block_date ensured the test ran -- you're aware of this with convo on your PR for enhanced test

jeff-dude (approved):

0xRobin:
@jeff-dude thanks! 

> 1. remove tests from the clipper ethereum view -- no need to repeat after all three versions have the tests

Technically if we want the test to be complete we should repeat it on the full view.
Right now we could have an edge case that the different version queries report the same trade but all tests pass.
This setup is fine though, just wanted to clarify. 

> since it uses different standards for schema/table name (test dummy), there were no issues, but in the real scenario, they'd all share the same schema/table name without it and fail, sadly only i can see that running locally right now

I had a feeling this would be the case, can you expand on how the setup differs?
How are the table names sourced in the git CI and on production, what exactly caused the collision here?


There's lots of different naming going on:
- filename
- alias (in config block)
- schema name (in .._schema.yml)
- project name (meta in schema.yml)
- spell name (in expose_spells(...))
- model name (in dbt_project)
It's not always clear how they are used and integrated (cause they are often the same)

jeff-dude:
> @jeff-dude thanks!
> 
> > 1. remove tests from the clipper ethereum view -- no need to repeat after all three versions have the tests
> 
> Technically if we want the test to be complete we should repeat it on the full view. Right now we could have an edge case that the different version queries report the same trade but all tests pass. Current setup is fine though, just wanted to clarify.
> 
> > since it uses different standards for schema/table name (test dummy), there were no issues, but in the real scenario, they'd all share the same schema/table name without it and fail, sadly only i can see that running locally right now
> 
> I had a feeling this would be the case, can you expand on how the setup differs? How are the table names sourced in the git CI and on production, what exactly caused the collision here?
> 
> There's lots of different naming going on:
> 
> * filename
> * alias (in config block)
> * schema name (in .._schema.yml)
> * project name (meta in schema.yml)
> * spell name (in expose_spells(...))
> * model name (in dbt_project)
>   It's not always clear how they are used and integrated (cause they are often the same)

- filename only used for ref() and default table/view name if alias is left out of config block
- schema yaml file only references file/model name, but doesn't have impact on schema name in db. ultimate use case of this is documentation generated
- project name (meta property) is for documentation as well, also used in post-hook config on how to display in data explorer on dune ui
- expose spells / spell name only help dune ui data explorer again
- dbt_project yaml doesn't affect model names, just directories in project -- then within, we apply default schema name

for your PR here, since schema was left out of config block in version models, it would default to:
v1: ```clipper_ethereum.trades``` due to dbt_project yaml file default schema name, alias in config block for table name
v2: ```clipper_ethereum.trades``` due to dbt_project yaml file default schema name, alias in config block for table name
v3: ```clipper_ethereum.trades``` due to dbt_project yaml file default schema name, alias in config block for table name

since all versions are in same project directory. only way to differentiate is to override. 

hope this all helps a bit

PR #1707, opened 2022-10-05
Better generic seed checking tests

0xRobin:
When working on the v2 integration of Clipper in dex.trades I noticed the current setup with `dex_trades_token_bought` only checks for:

> **IF** the matching columns of a seed record are present in the model records
> -> does it have the same `token_bought_address` as the seed record?

This really is only an edge case and misses the following checks. 
- If the record is NOT present in the model.
- If any other values present in the seed are incorrect in the model.

This PR introduces a generic `check_seed` test that can be applied at the top level (ex on `dex.trades`).
The test checks that 
1) all records specified in the seed file are present in the model.
2) column values are correct.

To do this you need to specify the matc_columns and the check_columns. 
example: 
```yml
tests:
  - check_seed:
      seed_file: ref('dex_trades_seed')
      match_columns:
        - block_date
        - blockchain
        - project
        - version
        - tx_hash
        - evt_index
      check_columns:
        - token_bought_address
        - token_bought_amount
        - token_sold_address
        - token_sold_amount
```
This is very readable and checks what I assume it would check.
If accepted we should probably also use the same construct for the general `nft.trades` model.

The result of the test also gives a clear view on what's missing/incorrect.


0xRobin:
Results from this test:
https://dune.com/queries/1355102?d=1

1) The Curve records don't show up in the model -> **something wrong with model?**
2) The Uniswap records don't match cause they have insufficient decimals -> **adapt the seed file**

tests/generic/check_seed.sql
```diff
@@ -0,0 +1,73 @@
+-- this tests checks a model for every row in a seed file.
+-- you need to specify the matching columns and the columns to check for equality.
+
+{% test check_seed(model, seed_file, match_columns=[], check_columns=[]) %}
+    {# --jinja comment
+--    I wish we could do something dynamic like this below, but it is prohibited by having a local database connection.. :(
```
dot2dotseurat, line 6: :(

0xRobin: Looks like we only need 'read_metadata' permissions to do more fancy compile stuff. 😉

dot2dotseurat: I *think* that exposes s3 buckets so it's probably a no go. :(

jeff-dude:
thanks for introducing the enhancement ideas. i haven't looked close enough to run and see myself, but wanted to share initial thoughts:
- the intention of this generic test was to set the foundation for how to set up & also knock out a common issue found in v1 engine dex trades, where token bought and token sold are flipped, which is why it simply checks token bought address
- if the token bought address is correct, the thought was it's not necessary to set up processing & compute to check token sold address
- amount isn't checked since the fields pulled/logic correlate closely to the addresses (an assumption, of course, but didn't get much feedback of this issue on v1)
- i've also seen issues with the test where the join "fails" i.e. the matching conditions didn't work as expected, then the test returns a false positive, since it's returns no rows (which it seems you've built for this here)

with that said, i don't see an issue with looking to increase our testing scope, as i don't envision the seed file ever being that large of a file and cause performance issues. i'm happy to take a closer look at this soon & see how we can finalize it

tests/generic/check_seed.sql
0xRobin: Just had a quick check, `DESCRIBE TABLE EXTENDED` does indeed expose data location. :(
https://docs.databricks.com/security/access-control/table-acls/object-privileges.html#operations-and-privileges
https://docs.databricks.com/spark/latest/spark-sql/language-manual/sql-ref-syntax-aux-describe-table.html#examples

0xRobin:
@jeff-dude Thanks! 
Having a minimal set of columns to check is definitely ok, It was just a bit confusing having more columns in the seed file which were not used in any tests.. 
I'll push a performance improvement in a minute to reduce the amount of joins.
But, since we're joining on the partitioned column and the amount of records in a seed file will always be small-ish, I think performance increase should be negligible. :ok_hand:

btw, any idea why dbt slim ci is not triggered on this PR?

0xRobin:
ℹ️ checks only pass currently cause it's setup as `warning`.

jeff-dude:
hey @0xRobin -- sorry for somewhat deprioritizing this, trying to move forward a bunch of other PRs. i still like the ideas here, but i'm wondering if you can explore how to do via inner join -- or at least flip to where left join is the smaller seeds? first glance, that may be counterproductive to the goal, but we have performance issues left joining large spells such as dex trades (that'll continue to grow)

0xRobin:
@jeff-dude no problem on pushing this down the line. 

Changing the join construction won't be possible without breaking the test logic..
Only thing I can imagine doing is something like:

```sql
select
...
from seeds
left join (
  select ...
  from seeds
  inner join model
  )
``` 
Do you think that could improve performance? :thinking: 
It does feel very counter-intuitive

tests/generic/check_seed.sql
```diff
@@ -0,0 +1,79 @@
...
+        seed.{{column_name}} as seed_{{column_name}},
+        model.{{column_name}} as model_{{column_name}} {% if not loop.last %},{% endif %}
+        {% endfor -%}
+        from {{seed_file}} seed
+        left join {{model}} model
+        ON 1=1
+        {%- for column_name in seed_matching_columns %}
+        AND seed.{{column_name}} = model.{{column_name}}
+        {% endfor -%}
+    ),
```
Hosuke, line 31: I tested the performance on different join logic of seeds and the model:

1.  [seeds left join the model](https://dune.com/queries/1508699)
2. [seeds inner join the model](https://dune.com/queries/1508752)
3. [seeds left join (unique keys from seeds inner join the model)](https://dune.com/queries/1508732)

Seeds represent the smaller dataset, while models(in this case ethereum.transactions) represent the larger dataset/table.
If we use the [seeds inner join the model] approach, we can add another counter test at the same time.

Hosuke:
> @jeff-dude no problem on pushing this down the line.
> 
> Changing the join construction won't be possible without breaking the test logic.. Only thing I can imagine doing is something like:
> 
> ```sql
> select
> ...
> from seeds
> left join (
>   select ...
>   from seeds
>   inner join model
>   )
> ```
> 
> Do you think that could improve performance? 🤔 It does feel very counter-intuitive

From the tests above this approach actually improves the performance.

jeff-dude:
pushed a commit with my latest thoughts:
- apply test a child model level (i.e. project / version level, similar to now). this will require us to edit each child schema file, but we'd have to either approach, to remove the old test
- remove from parent dex level
- add conditional filters for incoming args of blockchain/project/version to filter seed down to intended records

this allows the gh action to run the tests as-is. if we do at parent dex level, we need to consider opening up running downstream objects from edited child models. early tests of that show we run more models than intended and it can be tough to get granular, therefore the gh action may get overworked unnecessarily.

curious to hear thoughts @0xRobin @Hosuke

Hosuke:
I agreed to add filters for blockchain/project/version(or something corresponding to dbt_project.yml models' structure) to filter seed down to intended records. As many models are developed from buttom-up, and submit child-level models PR first then later merge into a meta-table if applicable. Fined filter of the seeds allows utilizing this test in the child-level model development, and with the aggregated seed file(like dex_trade_seed).

jeff-dude:
> @jeff-dude yeah this is great. I pushed some final commits with a couple optimizations.
> 
> * `check_seed` logic is moved to a macro under `macros/test-helpers/`, the old test is now a wrapper around this macro. This allows us to reuse the logic in different tests.
> * the blockchain/project/version parameters are moved under a `filter` dictionary, this makes it generic and not dex.trades specific
> * introduced a `check_dex_seed` test, a wrapper that abstracts the static parameters for dex.trades (seed file and column names).
> 
> this results in just having to specify:
> 
> ```
>     tests:
>       - check_dex_seed:
>           blockchain: ethereum
>           project: uniswap
>           version: 1
> ```
> 
> ~Still need to fix the pre-commit, not sure why the yml doesn't get picked up..~

thanks for the continual enhancements 🤝 working through testing now to get deployed

jeff-dude:
okay, i have it tested and applied to applicable models. tests pass as expected. please do look through latest code to see it all applied.

one caveat: models which use multiple versions within one model (rather than split out one version per model, then union'd together) maintained the old generic test for now. i didn't want to hold up this PR to figure out how to pass multiple versions in.

quick first thought to do in new PR to handle this: we could look to make the filter be ```IN (x,y,z)``` rather than ```='v1'``` or something similar.

Hosuke:
> one caveat: models which use multiple versions within one model (rather than split out one version per model, then union'd together) maintained the old generic test for now. i didn't want to hold up this PR to figure out how to pass multiple versions in.

Or may use check_seed multiple times for different versions in the schema.yml

Hosuke:
@jeff-dude Current check_dex_seed applied PR list (Keep updating):

- https://github.com/duneanalytics/spellbook/pull/2008

0xRobin:
> quick first thought to do in new PR to handle this: we could look to make the filter be `IN (x,y,z)` rather than `='v1'` or something similar.

Ah yes, I thought models would either:
a) define 1 version
b) define all the versions (in this case you can just omit the version filter).

We could adapt it to use IN (x1,x2,x3), I can't quickly find an example though where it's needed.

Hosuke:
> > quick first thought to do in new PR to handle this: we could look to make the filter be `IN (x,y,z)` rather than `='v1'` or something similar.
> 
> Ah yes, I assumed models would either: a) define 1 version b) define all the versions (in this case you can just omit the version filter).
> 
> but there's indeed a case between that's not covered.
> 
> We could adapt it to use IN (x1,x2,x3), I can't quickly find an example though where it's needed. In the meantime we could define the test once per version in the schema file to circumvent the restrictions.

- https://github.com/duneanalytics/spellbook/pull/2045
Here may be an example, this is a woofi_bnb model, which contains both version 1 and version 2.

jeff-dude:
please see list on the dex trades model via comment:
https://github.com/duneanalytics/spellbook/blob/main/models/dex/dex_trades.sql#L10

jeff-dude:
we can also revisit the config to remove the warning, intend for it to fail instead

PR #1709, opened 2022-10-05
Add native (ie, non-platform) NFT mints on Ethereum to nft.mints

UmerHA:

Currently, native (ie, non-platform) NFT mints on Ethereum are not included in nft.mints or nft.trades.
This PR aims to change that.

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


hildobby:
Wait I'm confused @UmerHA, this would cause a lot of overlaps with mints from projects already abstracted into `nft.mints`, no? Also, I planned on adding abstractions for ENS, ArtBlocks and probably other minting events, but now that means there would be mint tx duplicates with this query right?

I like the idea of having all mints within `nft.mints` but I think there needs to be a filter where all other mint abstractions get priority and this one is the backup, just not sure how to achieve this without a recursive call (but I think recursive is a bad solution).

What do you think @soispoke @jeff-dude?

hildobby:
Also, instead of an empty project name, having the `namespace` from `ethereum.contracts` based on the address from `tx_to` could be a good idea imo. The tx to address is the one the user interacted with to realise the tx so most of the times it should be the right project namespace. It would only be inaccurate if the user used an nft mint aggregator or another router contract, which unless some probable edge cases, I am not aware of any.

hildobby:
Also, instead of an empty project name, having the `namespace` from `ethereum.contracts` based on the address from `tx_to` could be a good idea imo. The tx to address is the one the user interacted with to realise the tx so most of the times it should be the right project namespace. It would only be inaccurate if the user used an nft mint aggregator or another router contract, which unless some probable edge cases, I am not aware of any.

UmerHA:
Yes, this would overlap with projects already abstracted into `nft.mints`, but as far as I see, these can be divided into 2 categories:

1. Projects initially minted on a marketplace's shared storefront
=> We can exclude the shared storefronts easily. The number of shared storefronts is (i) rather static and (ii) low

2. Native projects (e.g, ENS, CryptoPunks, ...)
=> If these projects are already abstracted into `nft.mints`, they need to be excluded manually. So far, this is only the case for CryptoPunks (which I have excluded).
Going forward I think it's better to have all native projects automatically abstracted into `nft.mints` instead of having to abstract each project separately.

Happy to hear your thoughts!

UmerHA:
> Wait I'm confused @UmerHA, this would cause a lot of overlaps with mints from projects already abstracted into `nft.mints`, no? Also, I planned on adding abstractions for ENS, ArtBlocks and probably other minting events, but now that means there would be mint tx duplicates with this query right?
> 
> I like the idea of having all mints within `nft.mints` but I think there needs to be a filter where all other mint abstractions get priority and this one is the backup, just not sure how to achieve this without a recursive call (but I think recursive is a bad solution).
> 
> What do you think @soispoke @jeff-dude?

Yes, this would overlap with projects already abstracted into nft.mints, but as far as I see, these can be divided into 2 categories:

    Projects initially minted on a marketplace's shared storefront
    => We can exclude the shared storefronts easily. The number of shared storefronts is (i) rather static and (ii) low

    Native projects (e.g, ENS, CryptoPunks, ...)
    => If these projects are already abstracted into nft.mints, they need to be excluded manually. So far, this is only the case for CryptoPunks (which I have excluded).
    Going forward I think it's better to have all native projects automatically abstracted into nft.mints instead of having to abstract each project separately.

Happy to hear your thoughts!

UmerHA:
> Also, instead of an empty project name, having the `namespace` from `ethereum.contracts` based on the address from `tx_to` could be a good idea imo. The tx to address is the one the user interacted with to realise the tx so most of the times it should be the right project namespace. It would only be inaccurate if the user used an nft mint aggregator or another router contract, which unless some probable edge cases, I am not aware of any.

I don't understand that. Can you explain a bit more (or share resources I can read to get a better understanding)?

UmerHA:
> > Wait I'm confused @UmerHA, this would cause a lot of overlaps with mints from projects already abstracted into `nft.mints`, no? Also, I planned on adding abstractions for ENS, ArtBlocks and probably other minting events, but now that means there would be mint tx duplicates with this query right?
> > I like the idea of having all mints within `nft.mints` but I think there needs to be a filter where all other mint abstractions get priority and this one is the backup, just not sure how to achieve this without a recursive call (but I think recursive is a bad solution).
> > What do you think @soispoke @jeff-dude?
> 
> Yes, this would overlap with projects already abstracted into `nft.mints`, but as far as I see, these can be divided into 2 categories:
> 
>     1. Projects initially minted on a marketplace's shared storefront
>        => We can exclude the shared storefronts easily. The number of shared storefronts is (i) rather static and (ii) low. But you're right, still have to do that.
> 
>     2. Native projects (e.g, ENS, CryptoPunks, ...)
>        => If these projects are already abstracted into `nft.mints`, they need to be excluded manually. So far, this is only the case for CryptoPunks (which I have excluded). Going forward I think it's better to have all native projects automatically abstracted into `nft.mints` instead of having to abstract each project separately.
> 
> 
> Happy to hear your thoughts!

An update re point 1: When building non-native mints the way proposed in this PR, no mints from aggregators are included - see https://dune.com/queries/1374255

hildobby:
Btw, I think that it's very misleading to make this abstraction. Looking into all time minting stats by volume, we have:
- 65% from Uniswap LP NFTs
- 8% from Alpha Homora
- 3% from Curve finance

So that is already over 75% of NFT minting volume all from DeFi, those NFTs are definitely not seen in the same light as other NFTs and I think having a global minting abstraction like this would lead to a lot of false information.

Source: https://dune.com/queries/1341447

FYI: This is only looking into NFTs minted with ETH, if you look at ERC20s too, you can bet that Uniswap will take a lot bigger of an NFT minting volume chunk.

So in general I would advise against having this abstraction, or at least not having it as a part of `nft.events`/`nft.mints`, but happy to hear other differing opinions.

UmerHA:
> Btw, I think that it's very misleading to make this abstraction. Looking into all time minting stats by volume, we have:
> 
>     * 65% from Uniswap LP NFTs
> 
>     * 8% from Alpha Homora
> 
>     * 3% from Curve finance
> 
> 
> So that is already over 75% of NFT minting volume all from DeFi, those NFTs are definitely not seen in the same light as other NFTs and I think having a global minting abstraction like this would lead to a lot of false information.
> 
> Source: https://dune.com/queries/1341447
> 
> FYI: This is only looking into NFTs minted with ETH, if you look at ERC20s too, you can bet that Uniswap will take a lot bigger of an NFT minting volume chunk.
> 
> So in general I would advise against having this abstraction, or at least not having it as a part of `nft.events`/`nft.mints`, but happy to hear other differing opinions.

Good catch, I didn't know that. I agree that a global `nft.mints` dominated by Uniswap LP NFTs would be misleading.

Still, I think it would be good to have a more complete `nft.mints`. I'll think about it and get back to you!

UmerHA:
> > > > Also, instead of an empty project name, having the `namespace` from `ethereum.contracts` based on the address from `tx_to` could be a good idea imo. The tx to address is the one the user interacted with to realise the tx so most of the times it should be the right project namespace. It would only be inaccurate if the user used an nft mint aggregator or another router contract, which unless some probable edge cases, I am not aware of any.
> > > 
> > > 
> > > I don't understand that. Can you explain a bit more (or share resources I can read to get a better understanding)?
> > 
> > 
> > When you interact with a smart contract, it shows up as the `to` address in `ethereum.transactions`. A lot of smart contracts have been decoded and given a `namespace` in dune. For example all artblocks smart contracts now have the `artblocks` namespace. So using that method I proposed, all artblocks mints would be labelled as such. Here this probably helps: https://dune.com/queries/1341447
> 
> Sounds good, will change

Done

UmerHA:
> Btw, I think that it's very misleading to make this abstraction. Looking into all time minting stats by volume, we have:
> 
>     * 65% from Uniswap LP NFTs
> 
>     * 8% from Alpha Homora
> 
>     * 3% from Curve finance
> 
> 
> So that is already over 75% of NFT minting volume all from DeFi, those NFTs are definitely not seen in the same light as other NFTs and I think having a global minting abstraction like this would lead to a lot of false information.
> 
> Source: https://dune.com/queries/1341447
> 
> FYI: This is only looking into NFTs minted with ETH, if you look at ERC20s too, you can bet that Uniswap will take a lot bigger of an NFT minting volume chunk.
> 
> So in general I would advise against having this abstraction, or at least not having it as a part of `nft.events`/`nft.mints`, but happy to hear other differing opinions.

Okay, so I understand there are basically two types of NFTs: Collectible NFTs (eg BAYC) and Functional NFTs (eg Uniswap LP). The naming of `nft.mints` / `nft.events` suggests these tables will consider all NFTs, regardless of their function. But I agree most users will only be interested in collectible NFTs, and adding functional NFTs will paint a very misleading picture.

Still, I think we should try to add all collectible NFTs to `nft.mints` / `nft.events`. So I suggest we exclude functional NFTs from this query. For this, I'd add an `addresses_ethereum_defi` table analogous to `addresses_ethereum_cex` and exclude those addresses.

See most recent commit. What do you think @hildobby, @jeff-dude, @soispoke?

UmerHA:
@jeff-dude can you remove the WIP tag? From my pov, it's ready to review. Thanks!

jeff-dude:
tagged @soispoke to see if he can help review

UmerHA:
Awesome. I'll add the discord discussion: https://discord.com/channels/757637422384283659/757893948428517376/1031858560579158036

models/nft/nft_mints.sql
```diff
@@ -1,221 +1,254 @@
 {{ config(
         alias ='mints',
         post_hook='{{ expose_spells(\'["ethereum","solana"]\',
-                                    "sector",
-                                    "nft",
-                                    \'["soispoke"]\') }}')
+                    "sector",
+                    "nft",
+                    \'["soispoke"], ["umer_h_adil"]\') }}')
```
soispoke, line 6: should be `'["soispoke"], "umer_h_adil"]\'` here

models/nft/nft_mints.sql
```diff
@@ -1,221 +1,254 @@
...
+        aggregator_name,
+        aggregator_address,
+        tx_hash,
+        block_number,
+        tx_from,
+        tx_to,
+        unique_trade_id
+    FROM {{ ref('zora_ethereum_mints') }}
+),
+general_mints AS (
```
soispoke, line 220: How about `native_mints` ?

models/nft/nft_mints.sql
```diff
@@ -1,221 +1,254 @@
...
-                aggregator_address,
-                tx_hash,
-                block_number,
-                tx_from,
-                tx_to,
-                unique_trade_id
-        FROM {{ ref('zora_ethereum_mints') }}
-)
\ No newline at end of file
+WITH project_specific_mints AS (
```
soispoke, line 9: Could we just name this `projects_mints` maybe ?

models/nft/ethereum/nft_ethereum_schema.yml
```diff
@@ -16,3 +16,89 @@ models:
...
+        description: "Symbol of the token in which fees are paid out"
+      - name: royalty_fee_amount_raw
+        description:  "Raw numerical amount for royalty fees"
+      - name: royalty_fee_amount
+        description:  "Royalty fee amount in original token currency (properly formatted in decimals)"
+      - name: royalty_fee_amount_usd
+        description:  "Royalty fee amount in USD"
+      - name: royalty_fee_percentage
+        description:  "Royalty fee in % of the amount paid for a given trade"
+      - name: unique_trade_id
```
soispoke, line 103: Can you add a uniqueness test on the `unique_trade_id` column ? 🙏

soispoke (commented):
First pass here, lmk what you think @UmerHA

soispoke:
@UmerHA dbt slim ci job is failing because of duplicate rows: 
Capture d’écran 2022-10-19 à 18 03 32

UmerHA:
@soispoke have fixed the duplicate issue

UmerHA:
‼️Don't merge yet. I realized the tests are wrong. I used `< 0.001` instead of `> 0.001`. Will push a commit to show it's not ready yet.

UmerHA:
> ‼️Don't merge yet. I realized the tests are wrong. I used `< 0.001` instead of `> 0.001`. Will push a commit to show it's not ready yet.

✅ fixed now

models/nft/ethereum/nft_ethereum_native_mints.sql
```diff
@@ -0,0 +1,201 @@
...
+  ercs.to as project_contract_address,
+  '' as aggregator_name,
+  '' as aggregator_address,
+  ercs.evt_tx_hash as tx_hash,
+  ercs.evt_block_number as block_number,
+  ercs.`from` as tx_from,
+  ercs.`to` as tx_to,
+  ercs.contract_address || '-' || ercs.evt_tx_hash || '-' || coalesce(ercs.tokenId, '') || '-' || ercs.`from` || '-' || ercs.evt_index
+  as unique_trade_id,
+  -- Temporary fix: {{ source('ethereum','contracts') }} can contain duplicates, which would result in duplicate mints.
```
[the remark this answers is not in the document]
UmerHA: This only affects ~20 mints with total value ~$150, so I thought it's negligible. See https://dune.com/queries/1433939.
Also, we can take `first` because namespace (the only column we're taking from `ethereum.contracts`) is identical partitioned by address.

But yes, this could be a source of bugs in the future. So I'll dive into `ethereum.contracts` to find out why there are duplicates. 🕵️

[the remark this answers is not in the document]
UmerHA: @soispoke I confirmed with boxer in the office hours this is a bug in `ethereum.contracts`, meaning we have to either include a temp fix or wait until `ethereum.contracts` is fixed.

Because this only affects ~20 rows with total value  < $150 and the fix is trivial (and thus can easily be removed by anyone when no longer necessary), I'd say we include the temp fix and ship now.

What do you think?

soispoke:
@UmerHA alright sounds good to me thanks for taking another look 🙏👍

lgtm on my end

soispoke (approved):

PR #1714, opened 2022-10-06
update source freshness on all warning/error models

jeff-dude:
using prod as starting point, fix any source freshness issues in full dbt project

### DIFF
--- a/models/archipelago/ethereum/archipelago_ethereum_sources.yml
+++ b/models/archipelago/ethereum/archipelago_ethereum_sources.yml
@@ -3,9 +3,7 @@ version: 2
 sources:
   - name: archipelago_ethereum
     freshness:
-      warn_after: { count: 12, period: hour }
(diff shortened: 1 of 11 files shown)


jeff-dude:
@dot2dotseurat @soispoke plz take a look, this is a step towards getting our daily views/seeds/tests job to run in prod successfully

jeff-dude:
agreed. each decoded table can vary wildly in terms of freshness. sometimes it's simply a "pool added" to a dex & that can go months without a new pool. it requires users to really dig deep into each one, then at that point, is it worth it?

i'll push this through for now, but a larger conversation should be had. this doesn't feel like the best approach.

jeff-dude:
also should add source freshness dbt command to our slim ci job, will help PRs catch them up front

soispoke:
👍 thanks for looking into this @jeff-dude

PR #1724, opened 2022-10-07
Bnb chain and Ethereum gas spells

soispoke:

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


soispoke:
fyi @augustog @dot2dotseurat test is failing just because it takes more than 1 hour to build the intial gas spells but it's actually running fine

models/gas/ethereum/gas_ethereum_fees.sql
```diff
@@ -0,0 +1,58 @@
+{{ config(
+    alias = 'fees',
+    partition_by = ['block_date'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['block_time','tx_hash','tx_amount_native']
```
masquot, line 7: Question to help my understanding: Why does the unique key consist of these fields?

models/gas/ethereum/gas_ethereum_fees.sql
```diff
@@ -0,0 +1,58 @@
...
+     gas_price /1e9 AS gas_price_gwei,
+     gas_price / 1e18 * p.price AS gas_price_usd,
+     txns.gas_used,
+     txns.gas_limit,
+     txns.gas_used / txns.gas_limit * 100 AS gas_usage_percent,
+     difficulty,
+     type AS transaction_type
+FROM {{ source('ethereum','transactions') }} txns
+JOIN {{ source('ethereum','blocks') }} blocks ON blocks.number = txns.block_number
+{% if is_incremental() %}
```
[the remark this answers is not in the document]
masquot: The same comments also apply to BNB.

soispoke:
Thanks @masquot for taking the time to review this! Will definitely add a seed file 👍
2. Yeah you're right, we've been trying to stay on the safe side be setting 1 week as a default but I'll reduce it to two days for this particular case

models/gas/ethereum/gas_ethereum_fees.sql
soispoke: This was just chosen because I don't expect to have duplicate rows using these columns, did you have something else in mind ?

masquot: Maybe I'm overlooking something, but couldn't `tx_hash` be sufficient?

soispoke: yeah, that definitely makes sense :)

soispoke:
@masquot changed incremental filters to 2 days and unique_key to 'tx_hash', will add a seed file a bit later

masquot:
> @masquot changed incremental filters to 2 days and unique_key to 'tx_hash', will add a seed file a bit later

👍 lfg

models/gas/gas_fees_schema.yml
```diff
@@ -0,0 +1,70 @@
+version: 2
+
+models:
+  - name: gas_fees
```
augustog, line 4: I see several columns here that seem to me to be evm- and/or eth-specific (e.g. does solana have a concept similar to gas limit? do other evm chains implement a burning mechanism like eip-1559?).

If all these columns remain relevant to all other chains, all my points below are moot.

Otherwise: I think we have an interesting design decision here:
- do we want to make this table fully encompassing of all different chains (like dex.trades does with protocols)?
- or do we want to have more dedicated views (e.g. gas-evm, gas-solana) to cover nuances of the protocol and have a more "abstracted out" perspective of all chains?

For this I would ask: what kind of analyses do we envision will happen cross-chain (particularly between diff technologies like evm v solana)

augustog (commented):
added a few comments regarding structure (which I believe is the main point you wanted my feedback on @soispoke).

I like this framework for designing novel spells!

models/gas/gas_fees_schema.yml
```diff
@@ -0,0 +1,70 @@
...
+      contributors: soispoke
+    config:
+      tags: ['ethereum', 'bnb', 'gas', 'fees']
+    description: >
+        Gas Fees across chains
+    columns:
+      - &blokchain
+        name: blockchain
+        description: "Blockchain name"
+  - &block_time
```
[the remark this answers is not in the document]
soispoke: thankss! Good catch, this is now fixed

models/gas/gas_fees_schema.yml
soispoke: So maybe it's even more obvious in the Avalanche PR here: #1751 but 
- Yes Solana and other EVM compatible blockchains share some (not all) properties and post EIP 1559 mechanisms. 
- I think a good compromise is too include all columns that can be shared across chains, but leave specific/interesting columns in individual models, and then expose both

soispoke:
Planning on merging this so I can test deployment + have the right links/table names for queries and research post, but I'll open another PR soon adding seed for all gas spells 👍

models/gas/gas_fees_schema.yml
augustog: > I think a good compromise is too include all columns that can be shared across chains, but leave specific/interesting columns in individual models, and then expose both

Agree! 

we could have the structure be like:
- gas.fees
- gas.ethereum_fees
- gas.avalanche_fees
- gas.solana_fees
- (etc)

as for naming convention for columns in gas.fees we could:
- choose a generic name
- stick to evm names and explain mapping in schema file (=>documentation)
- _other ideas?_

Personally I gravitate to the second, since evm chains and associated concepts are pretty dominant in the web3 space, I would be surprised if users were confused by evm naming.

soispoke: Yes exactly, if you look at my Avalanche PR this is exactly what I've done there: #1751
Side note: I actually don't think Solana will work atm, too many transactions = too much compute/time spent time generate this spell. Will focus on shipping the three EVM compatible L1s (Ethereum, BNB, Avalanche) + blog post and then I'll work on Arbitrum/Optimism/Gnosis for L2s

PR #1648, opened 2022-09-27
Add SuperRare trades

catherine-2:
Hi! This is my first non-static table PR so please let me know if there are things I need to fix / any questions! I looked at other trade table files to try to match the format so we can add SuperRare trades.

Brief comments on the purpose of your changes:

*For Dune Engine V2*
I've checked that:
General checks:
* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each sql file is a select statement and has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributors)
* [ ] if adding a new materialized table, I edited the optimize table macro

Join logic:
* [ ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

Incremental logic:
* [ ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ ] where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #1725, opened 2022-10-07
enhance performance for SuperRare nft events

jeff-dude:
using recently merged #1648 as reference, update query to be more performant on spark. this came up when monitoring in production.


jeff-dude:
@catherine-2 please see this PR for attempt at enhanced performance for SuperRare events. in prod, it's taking a bit more time than it likely should each run.

i've written a query to hopefully expedite your review:
https://dune.com/queries/1369408

in there i have the ```original``` CTE which is your model, then ```jeff_test``` CTE which is the rewrite. i wrote an except between ```original``` and ```jeff_test``` CTE which returned no rows (should be good sign)

the main approach changes:
- inner joins to base tables where possible, i.e. ```ethereum.transactions```
- avoid using ```IN``` statement on query returned results, especially string fields like ```hash``` or such
- read from large tables as little as possible

spark seems to like this approach more. hope this makes sense and i didn't break anything!

jeff-dude:
fyi @soispoke

jeff-dude:
> @catherine-2 please see this PR for attempt at enhanced performance for SuperRare events. in prod, it's taking a bit more time than it likely should each run.
> 
> i've written a query to hopefully expedite your review: https://dune.com/queries/1369408
> 
> in there i have the `original` CTE which is your model, then `jeff_test` CTE which is the rewrite. i wrote an except between `original` and `jeff_test` CTE which returned no rows (should be good sign)
> 
> the main approach changes:
> 
> * inner joins to base tables where possible, i.e. `ethereum.transactions`
> * avoid using `IN` statement on query returned results, especially string fields like `hash` or such
> * read from large tables as little as possible
> 
> spark seems to like this approach more. hope this makes sense and i didn't break anything!

although the gh action failed on ```unique_trade_id``` so something must be off 🤔 likely if i open the filter up to more than 1 week, we'd see the issues (gh action was full history)

models/superrare/superrare_ethereum_events.sql
```diff
@@ -1,6 +1,6 @@
 {{ config(
     alias = 'events',
-    partition_by = ['block_time'],
+    partition_by = ['block_date'],
```
jeff-dude, line 3: another important minor tweak to call out: the partition field on date vs. time

if it's time, then we have a massive amount of directories in data lake with tiny files vs. less directories with larger files. spark prefers the latter. this also matches our base table design on tables such as ```ethereum.transactions```

catherine-2:
Thank you so much @jeff-dude ! This is super helpful and definitely much cleaner than my previous version (and thank you for cleaning up some duplicate comments throughout too!). Will make a note to myself to make sure I'm better about using inner joins in spells. 

I went through all of the changes + compared the output and agree I think everything looks good! Though I'm also not sure what is causing the error with unique_trade_id.

jeff-dude:
@catherine-2 i made another round of tweaks & the job runs fine with test passing for no dupes on ```unique_trade_id``` -- do you want to give a last glance before potentially merging and refreshing prod?

catherine-2:
@jeff-dude I think this looks great! 

Thank you again for your help improving this spell -- really helpful to have a second set of eyes and gives me ideas to make my future queries better :)

soispoke (approved):
lgtm!

PR #1731, opened 2022-10-08
OP DEX Trades - Add Velo, Curve, Update Uni

MSilb7:

- Add Curve and Velodrome to dex.trades
- Update Uniswap to pull from the pools ref rather than factory events, since pools created in OVM1 do not have a create event.
   - Since all prior Uniswap trades would miss these pools, the whole Optimism Uniswap table should be dropped and reloaded
   - The "launch date" parameterized for Uniswap was also later than actual launch, which likely causes a need for re-inserting as well.

~Curve still has an issue since bytea2numeric results seem to return as a multi-row subquery? This needs to be fixed in that function for the curve insert to work. Otherwise, I just submitted a Curve "basic pool" for decoding that we'll be able to pull events from.~ Resolved, but see comment below about bytea2numeric result oddities.

As a next step, we can add Uniswap OVM1 trades to this if it's possible to decode those contracts in legacy OVM

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


MSilb7:
Fixed the curve issue - using conv(<column>,16,10) rather than bytea2numeric(<column>). Weird that the function has issues.

models/curvefi/optimism/curvefi_optimism_pools.sql
```diff
@@ -0,0 +1,127 @@
+{{ config(
+    alias = 'pools',
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['version', 'tokenid', 'token', 'pool'],
```
jeff-dude, line 6: @MSilb7 looks like incremental logic is complaining about the ```unique_key``` being all four fields in the table -- seems it has nothing to insert/update based on that. can we reduce fields in ```unique_key``` or change this from incremental to another format?

jeff-dude (changes requested):

jeff-dude:
velo & updated uniswap seem good to go ✅

models/curvefi/optimism/curvefi_optimism_pools.sql
MSilb7: Hey, sorry just getting back to this. The updates to this table should be fairly rare (just when new pools are created).

Do you have a suggestion for another format type?

jeff-dude: incremental is fine, you just need to reduce the keys in unique keys. all columns can't be included it appears

MSilb7: cool, removed 'version' from the key.

jeff-dude: it seems ```version``` should actually be part of the key since it's hardcoded and never changes -- are there no other columns to remove from key? you mentioned it only changes when new pools are created. whichever column indicates that changing pool, that should be the non-key value(s)

think of the unique keys as the columns to include in join conditions when writing an update statement. if this is an incremental model, the question is:
- which field(s) used in join conditions to itself? those are unique keys
- which field(s) get updated? those are non-keys

jeff-dude:
closing due to inactivity, please do open again when ready

MSilb7:
> closing due to inactivity, please do open again when ready

Hi, what was the issue with getting this merged?

models/curvefi/optimism/curvefi_optimism_pools.sql
jeff-dude: @MSilb7 this comment here

MSilb7: ah got it. I wouldn't expect any of these fields to change. So would it make sense to not do a "merge" strategy then?

jeff-dude: yeah, i'd revise and look at other options such as append only or insert overwrite: https://docs.getdbt.com/docs/build/incremental-models#about-incremental_strategy

PR #1732, opened 2022-10-09
Fix: typos

omahs:
Fix: typos


masquot (approved):
Thank you for these fixes @omahs

PR #1747, opened 2022-10-11
Add ens old pre images

soispoke:

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


soispoke:
@0xRobin and @bh2smith, would love to know what you guys think

bh2smith (commented):
Looks good to me - although I noticed the data file is a zip file so I can't actually look at it. was that intended? I don't see any place in the code that is responsible for unzipping it...

0xRobin:
@soispoke I don't think this will work.
Unless you reprocessed it somehow, the old file contains labels and labelhashes.

What we need to do is have a table of names and nodes and union them to ens_node_names.
And then expand the ens_resolver_records to catch the resolver settings of old addresses.
Refresher:
vitalik -> label
namehaseh(vitalik) -> labelhash
vitalik.eth -> name
namehash(vitalik.eth) -> node

When a new address is set on an ENS name we get event:
newAddr(node, addr)

This is the reason we have to build a name <> node database to make the nodes readable names.

0xRobin:
I tried to look at some old example transaction to see if we can extract node information but this get's really tricky and might break in a lot of edge cases.
I believe we should focus the further efforts in making the `namehash()` function available as a UDF.
This will allow us to remove the sketchy event matching currently in place and to process a list of plain labels.

soispoke:
got it, closing this for now then thanks for the feedback !

soispoke:
wait, would it work to use a hardcoded to reformat `labels` (e.g., vitalik) to names (e.g., `vitalik.eth`) for now ?
Capture d’écran 2022-10-11 à 12 19 10

0xRobin:
That still leaves us with labelhashes instead of nodes. 
What would work is generating the sql file with a python script.
something like
```python
from namehash import namehash
for label in labels:
    name = label+".eth"
    yield (name , namehash(name), label, namehash(label))
```
It's a bandaid solution but gives us something to work with.
When we've got namehash available as a UDF we can do this logic in a model and just need a seed file with the labels we want to process.

0xRobin:
nothing is so permanent as a temporary fix though.. :sweat_smile:

Issue #1713, opened 2022-10-06
[Bug] cow protocol missing batches


bh2smith:
This [adapted version of the PoC Query](https://dune.com/queries/1361238) for batches demonstrates that the select query should return these missing values records but they do not appear in the table.

There are actually around 578 batches not appearing in this table over all time - see [here](https://dune.com/queries/1360768?interval_e15077=days&N_n26d66=1000&Start_d83555=2021-01-01+00%3A00%3A00)

In attempts to fix this, I might suggest rebuilding the table from scratch... is it possible that whatever update scheduler was failing around the time of the missing records?

bh2smith:
I will add this query as a test to the batches.

bh2smith:
Closed by #1752

PR #1752, opened 2022-10-11
[Testing] CoW Protocol Batches

bh2smith:
Testing the issue raised in #1713 

PoC Query introduced in unit test (does not currently return no results as expected):

https://dune.com/queries/1389623

### DIFF
--- a/tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql
+++ b/tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql
@@ -1,0 +1,14 @@
+
+with batches_with_trade_count as (
+    select
+        s.evt_tx_hash,
+        s.evt_block_time,
+        count(t.evt_index) as num_trades
+    from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Trade t
+    inner join gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Settlement s
+        on s.evt_tx_hash = t.evt_tx_hash
+    group by s.evt_tx_hash, s.evt_block_time
+)
+
+select * from batches_with_trade_count
+where evt_tx_hash not in (select tx_hash from {{ ref('cow_protocol_ethereum_batches' )}})
--- a/tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql
+++ b/tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql
@@ -4,11 +4,14 @@ with batches_with_trade_count as (
         s.evt_tx_hash,
         s.evt_block_time,
         count(t.evt_index) as num_trades
-    from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Trade t
-    inner join gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Settlement s
+    from {{ source('gnosis_protocol_v2_ethereum','GPv2Settlement_evt_Trade') }} t
+    inner join {{ source('gnosis_protocol_v2_ethereum','GPv2Settlement_evt_Settlement') }} s
         on s.evt_tx_hash = t.evt_tx_hash
     group by s.evt_tx_hash, s.evt_block_time
 )
 
 select * from batches_with_trade_count
 where evt_tx_hash not in (select tx_hash from {{ ref('cow_protocol_ethereum_batches' )}})
+-- The reference table is only refreshed once in a while,
+-- so we impose a time constraint on this test.
+and evt_block_time < date(now())
--- a/tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql
+++ b/tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql
@@ -1,4 +1,4 @@
-
+-- Try it out here: https://dune.com/queries/1389623
 with batches_with_trade_count as (
     select
         s.evt_tx_hash,
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_batches.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_batches.sql
@@ -50,7 +50,7 @@ batch_values as (
         sum(fee_usd)    as fee_value,
         price           as eth_price
     from {{ ref('cow_protocol_ethereum_trades') }}
-        join {{ source('prices', 'usd') }} as p
+        left outer join {{ source('prices', 'usd') }} as p
             on p.contract_address = '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'
             and p.minute = date_trunc('minute', block_time)
             and blockchain = 'ethereum'
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
@@ -11,8 +11,8 @@
 
 -- Find the PoC Query here: https://dune.com/queries/1283229
 WITH
--- First subquery joins buy and sell token prices from prices.usd.
--- Also deducts fee from sell amount.
+-- First subquery joins buy and sell token prices from prices.usd
+-- Also deducts fee from sell amount
 trades_with_prices AS (
     SELECT try_cast(date_trunc('day', evt_block_time) as date) as block_date,
            evt_block_time            as block_time,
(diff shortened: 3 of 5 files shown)


tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql
```diff
@@ -0,0 +1,14 @@
+
+with batches_with_trade_count as (
+    select
+        s.evt_tx_hash,
+        s.evt_block_time,
+        count(t.evt_index) as num_trades
+    from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Trade t
```
dot2dotseurat, line 7: can you reference these tables as sources?

dot2dotseurat:
(looking into why the ref isn't being generated)

dot2dotseurat:
@bh2smith I edited the script to pick up the ref from a modifed test. It's creating the dependency now but the test itself seems to be failing.

bh2smith:
The test is expected to fail because there is an issue with the data itself. This is why I added the test. There is more context in the issue linked in the description.

tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql
bh2smith: Yes absolutely.

bh2smith:
So this is one reason why there is a missing trade and a resulting missing batch:

https://github.com/duneanalytics/spellbook/issues/1765

bh2smith:
Now that #1765 has been fixed, this table is still showing thousands of missing records. However, the last one appeared 2 days ago. Could someone please rebuild this table and see if its really fixed? cc @dot2dotseurat

dot2dotseurat:
> Now that #1765 has been fixed, this table is still showing thousands of missing records. However, the last one appeared 2 days ago. Could someone please rebuild this table and see if its really fixed? cc @dot2dotseurat
> 
> This table is a critical part of us continuing our V2 migration.

I can rebuild it @bh2smith. We just migrated the locations of our tables and I need to sort out some remaining permissions errors but I will rebuild it asap.

dot2dotseurat:
hi @bh2smith I've just rebuilt the table.

bh2smith:
Thanks very much @dot2dotseurat! I'll check back on this once CI has finished!

bh2smith:
Unfortunately it seems that the Record from September 14 is still missing: I will have a look back and see if there could be any other reason for this:

cf the query with the test - https://dune.com/queries/1389623 

PS - I imagine the later missing records are because of the spells update schedule (or at least I hope so).

bh2smith:
> @bh2smith my hunch would be could be missing from prices but I have no validated. 

Perhaps for the latest records, but I don't believe it's missing prices for sept 14. otherwise we would be seeing a lot more missing than just a single one. Anyway I'll try and look into it soon. Thanks for your help today.

bh2smith:
Ok, I have discovered a new fact that is causing this particular record to be missing:

This Spell Depends on the trades table (which was also missing this records due to this https://github.com/duneanalytics/spellbook/issues/1765). This [missing trades query](https://dune.com/queries/1540086?d=1) shows that there is only one rogue missing entry - all others are from within the last hour (so I chalk this up to scheduling). 

I guess this means the trades table needs to be rebuilt (followed by this one).

[Here is a query](https://dune.com/queries/1539986?+source%28%27prices%27%2C+%27usd%27%29+_t6c1ea=prices.usd&+ref%28%27tokens_ethereum_erc20%27%29+_t6c1ea=tokens_ethereum.erc20&+source%28%27gnosis_protocol_v2_ethereum%27%2C+%27GPv2Settlement_evt_Trade%27%29+_t6c1ea=gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Trade&+source%28%27gnosis_protocol_v2_ethereum%27%2C+%27GPv2Settlement_call_settle%27%29+_t6c1ea=gnosis_protocol_v2_ethereum.GPv2Settlement_call_settle) demonstrating that this trade should be contained in the table.

Note that this is this only "stray" missing record (all others are within the past week). I wonder if this has something to do with the way updates are arranged. Batches depend on Trades how do we ensure that batches are updated after trades or something like that? Or that changes to the trades table trigger updates to the batches?

cc @dot2dotseurat

jeff-dude:
> Batches depend on Trades how do we ensure that batches are updated after trades or something like that? Or that changes to the trades table trigger updates to the batches?

normally in prod, since you use ref() syntax, it would look for trades to refresh first since batches is dependent on it. on the gh action, to save processing time, rather than run all dependencies, we create views of the prod version. that means trades isn't refreshed in this action. the "hacky" way would be to add a minor tweak to that model to get it to run in the same action, then in theory should have updated data.

bh2smith:
Right but this particular transaction is missing in the production trades table so it needs to be rebuilt anyway. Followed by batches (which depends on it) unless that will be triggered anyway.

jeff-dude:
> Right but this particular transaction is missing in the production trades table so it needs to be rebuilt anyway. Followed by batches (which depends on it) unless that will be triggered anyway.

yes, rebuild in prod will happen regardless, but just curious if we should test it now in the gh action too. let me try something.

jeff-dude:
check out the gh action now -- it looks like tests passed on initial build (with incremental run going now). is this expected? previously, you expected a failure with one row returned to find that sept 14th tx.

bh2smith:
If the trades table has been rebuilt then I would expect the missing tx to be fixed. I'm not sure I understand the question.

jeff-dude:
the gh action built trades & batches. then the tests ran and succeeded. then incremental ran successfully and tests succeeded again. if you're happy with the result, i can merge and prod will rebuild

jeff-dude (approved):

bh2smith:
Are you sure the rebuild will actually be triggered since these changes don't touch the necessary tables?

jeff-dude:
aren't the two issues trades & batches (i.e. edited models in this PR)?
prod ran full refresh on both:
image

bh2smith:
Oh yes! I see you added a dot 8in trades! Thanks. Also checked it out. Looks good so far only missing records that are very recent (which I assume will get covered in the next update). This is incredible thanks a million!

Cc @gentrexha and @josojo.

bh2smith:
Unfortunately, [this query](https://dune.com/queries/1389623) shows that there are STILL missing records from yesterday.

dot2dotseurat:
I can see when your query was run that there were 126 missing trades. When I forked and reran there were 97. To me, this looks like late-arriving data. 

Your model is set up to keep checking for data up to one week back and merging it in every incremental run (we've scheduled that hourly) so my guess is this will resolve in a day. 

```
    {% if is_incremental() %}
    WHERE block_time >= date_trunc("day", now() - interval '1 week')
    {% endif %}
```

Screen Shot 2022-11-08 at 9 44 48 AM

Looking at the pct arrived, looking at this [query](https://dune.com/queries/1548038), it looks like we are at 97.4% for yesterday and 100% for all prior days but only 83.4% for today.  

Screen Shot 2022-11-08 at 9 55 43 AM

What's your use case for the data? Are you using it for programmatic decisions or for monitoring/analysis? I'm trying to understand how serious the implications are for late arriving data.

bh2smith:
We are using this data to reimburse protocol operators for their operational costs on a weekly basis. The payout transaction is compiled (partially) from this table about 7 hours past the the previous day.

dot2dotseurat:
I'm not sure. It doesn't seem like we should be more than an hour off. It looks like that is true for trades which is the only other spellbook ref this model is relying upon. 
```
select date_trunc('hour', evt_block_time), count(et.evt_tx_hash), count(t.tx_hash), count(t.tx_hash)/count(et.evt_tx_hash)::float
from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Trade et
left join cow_protocol_ethereum.trades t 
on et.evt_tx_hash = t.tx_hash
group by 1
order by 1 desc
``` 
Screen Shot 2022-11-08 at 11 18 56 AM

Btw I do see that the spell does rely on one hard coded table that should be a ref (cow_protocol_ethereum.solvers). But that's a view that only uses upstream tables so I don't think the latency could be introduced there.

dot2dotseurat:
This has to be something in the query not the way we run the updates because I am rebuilding it locally with my own schema and it hasn't resolved any issues with late data.

bh2smith:
Found the bug and squashed it! Problem and solution described in https://github.com/duneanalytics/spellbook/pull/1930

Thanks @dot2dotseurat for all your assistance here.

PR #1761, opened 2022-10-12
T 17638 pre-push hooks

dot2dotseurat:

This is not ready for merge but I requested reviews because it's a broader question with respect to changing workflows for wizards and adding another setup step. 

I'm proposing "pre-push" hooks not pre-commit hooks despite my branch title. Most of the useful checks require dbt compile to be run which is too slow for every commit. 

These are the available pre written hooks https://github.com/marketplace/actions/pre-commit-dbt
I selected a limited number of them and am actively avoiding the modifier ones because they don't seem to work well. 

The first two check-script-has-no-table-name and check-script-ref-and-source do not require dbt compile but if they fail, your commit fails. I am guessing this will be an annoyance because I can see some people writing the query on dune.com and pasting it in and using commits to save their progress. Imho, it only *needs* to be addressed when pushing. 

When these checks fail, they will literally not be able to push. I can see this being frustrating if you haven't worked with hooks before because it's not super obvious that your push failed unless you read the messages. We will have to do some education if we go down this route but I think it's still worth it. 

A sticking point on this is the wizard will need to run `pre-commit install --hook-type pre-push` to get this set up. I'm trying to think of a way we can enforce that automatically but I'm not sure how. That command generates a file that includes info about the user paths so we can't just commit a copy of the file since it will be different for everyone. I don't think it's a huge deal to ask people to do this but it would still be nice to try to avoid disrupting people's workflow. 

cow protocol file is just in here because I am messing with it in order to trigger the checks. it won't be included when this is merged.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,4 @@
-default_stages: [merge-commit, manual]
+default_stages: [push, manual]
 exclude: ^deprecated-dune-v1-abstractions/
 repos:
 - repo: https://github.com/offbi/pre-commit-dbt
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,4 @@
+default_stages: [push, manual]
 exclude: ^deprecated-dune-v1-abstractions/
 repos:
 - repo: https://github.com/offbi/pre-commit-dbt
@@ -5,14 +6,12 @@ repos:
   hooks:
   - id: check-script-has-no-table-name
   - id: check-script-ref-and-source
+  - id: check-script-ref-and-source
   - id: dbt-compile
-    stages: [push]
   - id: check-model-has-description
-    stages: [push]
   - id: check-model-has-meta-keys
     args: ['--meta-keys', 'blockchain', 'project', 'contributors', '--']
-    stages: [push]
   - id: check-model-has-properties-file
-    stages: [push]
+
 
 
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,3 @@ repos:
   - id: check-model-has-meta-keys
     args: ['--meta-keys', 'blockchain', 'project', 'contributors', '--']
   - id: check-model-has-properties-file
-
-
-
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,6 @@ repos:
   hooks:
   - id: check-script-has-no-table-name
   - id: check-script-ref-and-source
-  - id: check-script-ref-and-source
   - id: dbt-compile
   - id: check-model-has-description
   - id: check-model-has-meta-keys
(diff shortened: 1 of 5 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,14 @@
+default_stages: [push, manual]
+exclude: ^deprecated-dune-v1-abstractions/
+repos:
+- repo: https://github.com/offbi/pre-commit-dbt
+  rev: v1.0.0
+  hooks:
+  - id: check-script-has-no-table-name
+  - id: check-script-ref-and-source
+  - id: check-script-ref-and-source
```
soispoke, line 9: duplicate `check-script-ref-and-source` here ?

soispoke (commented):
This looks good, thanks @dot2dotseurat 

Maybe I didn't spend enough time testing this out or didn't understand how the check-scirpt-has-no-table-name prehook works but when I tried to intentionally cause an error by setting 
`ethereum.transactions` instead of `{{ source('ethereum', 'transactions') }}` I was able to commit and push without any error message ?

dot2dotseurat:
> This is cool, thanks @dot2dotseurat
> 
> Maybe I didn't spend enough time testing this out or didn't understand how the check-scirpt-has-no-table-name prehook works but when I tried to intentionally cause an error by setting `ethereum.transactions` instead of `{{ source('ethereum', 'transactions') }}` I was able to commit and push without any error message ?

Did you install the pre-push hooks with `pre-commit install --hook-type pre-push` after reinstalling your pip env?

dot2dotseurat:
> No I did not :) Ok now it actually works, also tested it with Github Desktop, works as well Capture d’écran 2022-10-13 à 14 22 07
> 
> @dot2dotseurat It does make every single push way longer though (even the successful ones), which is quite frustrating tbh

yes, its because of the dbt compile command.

bh2smith:
I think there may be an issue with the hooks:

```
  - id: check-model-has-description
  - id: check-model-has-properties-file
```
Since none of the tests contain these, we are seeing lots of errors like [this one](https://github.com/duneanalytics/spellbook/actions/runs/3409911862/jobs/5672249882)

```
 Check the model has description..........................................Failed
- hook id: check-model-has-description
- exit code: 1

tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql: does not have defined description or properties file is missing.
tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_trades.sql: does not have defined description or properties file is missing.

Check the model has properties file......................................Failed
- hook id: check-model-has-properties-file
- exit code: 1

tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_batches.sql: does not have model properties defined in any .yml file.
tests/cow_protocol/ethereum/cow_protocol_ethereum_assert_trades.sql: does not have model properties defined in any .yml file.
```


Would it make sense to exclude `tests/` in a similar way as you have "deprecated-dune-v1-abstractions"?

I am not overly familiar with the syntax requirements for https://github.com/offbi/pre-commit-dbt but it should be possible...

bh2smith:
Update: [I made a PR](https://github.com/duneanalytics/spellbook/pull/1907) excluding tests directory from two hooks (like so)
```
  - id: check-model-has-description
    exclude: ^tests/
  - id: check-model-has-properties-file
    exclude: ^tests/
```

PR #1767, opened 2022-10-13
[CoW Protocol] Add Solver List for Gnosis Chain

bh2smith:
Nearly identical to #1563 except for Gnosis Chain.

[PoC Query](https://dune.com/queries/1399494)

### DIFF
--- a/models/cow_protocol/gnosis/cow_protocol_gnosis_solvers.sql
+++ b/models/cow_protocol/gnosis/cow_protocol_gnosis_solvers.sql
@@ -1,0 +1,66 @@
+{{ config(materialized='view', alias='solvers',
+        post_hook='{{ expose_spells(\'["gnosis"]\',
+                                    "project",
+                                    "cow_protocol",
+                                    \'["bh2smith", "gentrexha"]\') }}'
+)}}
+
+-- Find the PoC Query here: https://dune.com/queries/1276806
+WITH
+-- Aggregate the solver added and removed events into a single table
+-- with true/false for adds/removes respectively
+solver_activation_events as (
+    select solver, evt_block_number, evt_index, True as activated
+    from {{ source('gnosis_protocol_v2_gnosis', 'GPv2AllowListAuthentication_evt_SolverAdded') }}
+    union
+    select solver, evt_block_number, evt_index, False as activated
+    from {{ source('gnosis_protocol_v2_gnosis', 'GPv2AllowListAuthentication_evt_SolverRemoved') }}
+),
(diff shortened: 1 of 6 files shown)


augustog (changes requested):
Overall looks good!

Only minor hiccup: 
model wasn't added to [dbt_project.yml](https://github.com/duneanalytics/spellbook/blob/47b4d4f58835d9411cf10388f074ca79d2b2e3d6/dbt_project.yml#L299) ⚠️ 

Otherwise all good
- Directory and file format correct ✅ 
- Source and Schema yml files complete ✅ 
- Jinja syntax used on all joins ✅ 
- Config blocks include alias and post hook to expose model on UI ✅ 
- CTEs made reading it a breeze <3
- Unit tests following pattern for ethereum ✅

bh2smith:
Awesome will fix tomorrow morning.

augustog:
I'm also happy to fix that myself if you're okay with it, it's a fairly trivial change :)

bh2smith:
> I'm also happy to fix that myself and merge if you're okay with it, it's a fairly trivial change :)

Fixed it! Thanks though (didn't see your comment till now)

augustog (commented):
perfect, looks good to go cc @jeff-dude

augustog (approved):

models/cow_protocol/gnosis/cow_protocol_gnosis_solvers.sql
```diff
@@ -0,0 +1,66 @@
+{{ config(materialized='view', alias='solvers',
```
jeff-dude, line 1: one very minor nitpick: we don't need to use ```materialized``` as view here, since dbt_project file defaults to view. i've tried to somewhat stay on top of that for consistency.
@augustog can you push a quick commit to remove? then if auto-test passes again (which it should), you can go ahead and merge 👍

jeff-dude (approved):

models/cow_protocol/gnosis/cow_protocol_gnosis_solvers.sql
bh2smith: In that case we should also be removing this from the other solver file on mainnet

bh2smith: Done.

jeff-dude: it's really not a big deal, but thanks for adding it

PR #1757, opened 2022-10-12
dex.trades shibaswap integration

0xRobin:
Brief comments on the purpose of your changes:
Resolves GH-1656
Also introduces a more dynamic dex_trades.sql setup, less code duplication.

*For Dune Engine V2*
General checks:
* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [x] if adding a new model, I added a test

PR #1773, opened 2022-10-14
add sandwich attacker labels

AlexTheLion123:

Add sandwich attacker labels

*For Dune Engine V2*
I've checked that:
General checks:
* [ x] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ x] if adding a new model, I added a test
* [ x] the filename is unique and ends with .sql
* [ x] each sql file is a select statement and has only one view, table or function defined  
* [ x] column names are `lowercase_snake_cased`
* [ x] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ x] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributors)

Join logic:
* [ x] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

Incremental logic:
* [ x] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ x] where block_time >= date_trunc("day", now() - interval '1 week')
* [ x] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ x] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')


augustog (changes requested):
Hello @AlexTheLion123, thanks for this submission. Having a labels dataset for sandwichers will be very useful.

have you checked the `flashbots` dataset on dune? 

The flashbots team contributed a very thorough dataset mapping many MEV behaviours including sandwich attacks, and is being maintained constantly to detect them reliably. I'd suggest you use that as a source (example query: https://dune.com/queries/1404958)

As for the spell itself, the only thing you're missing is adding it to `dbt_project.yml`. Otherwise it looks good.

If you rely on flashbots data and materialise as a view, it's probably fine to push without any unit tests, but if you're applying more convoluted logic we would also ask you to add a few.

Please let us know if you need any support!

models/labels/sandwich_attackers/ethereum/labels_sandwich_attackers_ethereum.sql
```diff
@@ -0,0 +1,32 @@
+{{config(alias='sandwich_attackers_ethereum')}}
```
augustog, line 1: minor style comment, just for readability, I'd add a bit more whitespace in the config blocks

```
{{ config(
      alias=...
  )
}}
```

augustog:
Sure, @AlexTheLion123, in that case, could you please tell us a bit more about the logic you're using to detect sandwich traders? Have you validated the logic against false positives & false negatives? i.e. 
- Are there any non-sandwich trade transactions that could meet these criteria?
- Are there any sandwich trades that may be missed?

As for the project.yml file, you can see an example in [this PR here ](https://github.com/duneanalytics/spellbook/pull/1757/files#diff-27386df54b2629c1191d8342d3725ed8678413cfa13b5556f59d69d33fae5425)

Lastly, [the build is failing](https://github.com/duneanalytics/spellbook/actions/runs/3277520323/jobs/5394904042), likely due to a data type issue. I'll highlight the fix inline.

soispoke:
@augustog When an issue has been fixed all we have to do it run the dbt slim ci again (I just did), and it should work :)

AlexTheLion123:
@augustog 

The logic to identify a sandwich attacker is as follows:

1. Two separate trades, t1 and t2 in no particular order, in the same block initiated by the same address
2. Both trades inititated on the same project
3. Token bought in t1 = token sold in t2
4. Token sold in t1 = token bought in t2
5. Amount bought in t1 = Amount sold in t2
x 6. Amount sold in t1 < amount bought in t2 (excluded)
7. (index of t1 >= index of t2 + 2) or (index of t2 >= index of t1 + 2)
8. exclude uniswap v2 router address since this gets included as a false positive for some reason

- This logic should work to include both buy-first and less common sell-first sandwiches
- By excluding point 6, we also include sandwich attacks that made a loss
- By using '>=' instead of just '=' in point 7, we can also include sandwich attackers that are perhaps not using flashbots and not tightly bundling their transactions

Also, I don't see the additions for specific labels on the dbt_project.yml file.

augustog:
@AlexTheLion123 thanks for clarifying.

I did a handful of [validations against flashbots' dataset](https://dune.com/queries/1461930?d=1&category=third_party_data&namespace=flashbots&table=sandwiches) and indeed this logic does cover a handful of cases (~500) they don't.

I spot checked transactions from that subset against etherscan and they do indeed look clearly like sandwiches, so this should be good to be merged ✅ 

I'd suggest we document the logic within the spell (e.g. under description in schema.yml), but shouldn't be a blocker to merge (can be done in another PR, or I can do it later this week).

PS: Indeed you are correct, you should not alter dbt_project.yml for this PR, my mistake.

augustog (approved):

PR #1784, opened 2022-10-17
add hashflow to dex.trades

justabi:
Resolves 
- https://app.dework.xyz/dune/dextrades-migration?taskId=ae2e93c2-9046-487a-a27b-d48508ffa30f
- https://github.com/duneanalytics/spellbook/issues/1655

Adds hashflow into dex.trades (also ports over the model that decodes hashflow transactions)

For Dune Engine V2
General checks:

- [x]  I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
- [x]  if adding a new model, I added a test


augustog (dismissed):
Hello @justabi, thank you for this contribution! It is probably one of the most complex models I have seen, so this is an incredible feat.

I added a few minor comments inline mostly regarding style. One of the tests is failing likely due to a join causing duplications, I marked the likely cause inline too. Other than that it looks very good!

Lastly, given the complex logic involved here, I would like to have a bit more thorough testing on this one before merging. @0xBoxer do we have someone at hashflow that can propose a few more testing cases? Since this is a migration we can get them directly from V1.

justabi:
@augustog thanks so much! will look at the feedback, the dupes and fix those conflicts this afternoon (just at a conference this week so a little intermittently available)

augustog:
No worries! I hear Coalesce is happening these days? I missed it sadly :(

augustog:
No worries! Enjoy the conference :)

Hosuke:
> Hello @justabi, thank you for this contribution! It is probably one of the most complex models I have seen, so this is an incredible feat.
> 
> I added a few minor comments inline mostly regarding style. One of the tests is failing likely due to a join causing duplications, I marked the likely cause inline too. Other than that it looks very good!
> 
> Lastly, given the complex logic involved here, I would like to have a bit more thorough testing on this one before merging. @0xBoxer do we have someone at hashflow that can propose a few more testing cases? Since this is a migration we can get them directly from V1.

I will try to add some hashflow test cases here.

Hosuke:
Hi @justabi, it's a great job with high quality codes.

I am digging out the [test failure info](https://github.com/duneanalytics/spellbook/actions/runs/3261550545/jobs/5356703915), and have found [some duplicate records generated by the query](https://dune.com/queries/1461326) that caused the test to fail.

We can try to fix the dupes in subquery( hashflow_ethereum_trades ) first and see how it affects the outer query.
 
I try to paste the `tx_hash` to [etherscan](https://etherscan.io), it seems like records having `composite_index` with value `1` are transactions that failed. 

However, transaction
[0xfcda3d7612d1ec7dc0bb96460a56522b48c27354528582033782d20e0202b8e0](https://etherscan.io/tx/0xfcda3d7612d1ec7dc0bb96460a56522b48c27354528582033782d20e0202b8e0),
[0x0d138951d21d3cd32de9606740a75707ba0aa94261cbcfe4bad8c73d5392ef01](https://etherscan.io/tx/0x0d138951d21d3cd32de9606740a75707ba0aa94261cbcfe4bad8c73d5392ef01)
and [0x55ff314eafe7a61f31554ef8c0a4be77d5942faebaddcad3150e48036aeae4a9](https://etherscan.io/tx/0x55ff314eafe7a61f31554ef8c0a4be77d5942faebaddcad3150e48036aeae4a9)
are successful transactions.

I will continue to try to figure out the reason that causes the dupes, and see how we can get this great work merged.

Hosuke:
GM, I have checked [some specific dupes records](https://dune.com/queries/1462580).
It shows that dupes with the same unique keys are having different ['method_id', 'pool', 'source'] values.

models/hashflow/ethereum/hashflow_ethereum_trades.sql
```diff
@@ -0,0 +1,407 @@
...
+
+    union all
+
+    select * from legacy_router_w_integration
+
+    union all
+
+    select * from new_router
+
+    {% endif %}
```
Hosuke, line 386: Since `new_router` and `new_pool` CTEs have overlapping transaction events in the source, 

- we can add another UNION CTE that unions `new_router` and `new_pool`
- and group by the unique keys `['block_date', 'composite_index', 'tx_hash']` 
- (with `coalesce(collect_list(...))` or `first_value(..., true)` on columns like `['method_id', 'pool', 'source', ...]`) 

before the all_trades UNION.

Also, in this complex model, the incremental process may need to be handle with care.

justabi: that makes sense - will update to remove the overlapping elements this way

r.e. your incremental comment -> did you have specific parts of my current approach that you were referring to?

Hosuke: Only `new_router` this part, as I see it only runs when full refresh. So if we do the CTE union before `all_trades` with `new_pool`, the incremental logic also need to be migrated to that CTE.

justabi:
@augustog Heya Augusto, I can't see any inline comments by you btw - just wanted to confirm whether there was any feedback you had waiting to be actioned (apart from stuff related to duplicates which i'm about to commit a fix for)

justabi:
@Hosuke [0xfe23b3c11be28d682af05dcd0a5b1cb408b3b0c878cd7bcc95a88ba1708af5d9 is an example](https://dune.com/queries/1491278) where there are dupe rows from the new_router cte = it seems this transaction failed - i'm looking into it a bit more - but would you know how we'd expect this data to show up in the final table?

do we want to keep one of the rows - if so which, the latest?

also - pre-commit will keep failing due to the use of a for loop + variable in the ref part of the dex_trades.sql model

Hosuke:
> @Hosuke [0xfe23b3c11be28d682af05dcd0a5b1cb408b3b0c878cd7bcc95a88ba1708af5d9 is an example](https://dune.com/queries/1491278) where there are dupe rows from the new_router cte = it seems this transaction failed - i'm looking into it a bit more - but would you know how we'd expect this data to show up in the final table?
> 
> do we want to keep one of the rows - if so which, the latest?
> 
> also - pre-commit will keep failing due to the use of a for loop + variable in the ref part of the dex_trades.sql model

Maybe we can filter out the unsuccessful trades first, and only use the successful trades to join other tables. You can also include all unique keys in the test to see whether there are truly dupes.

models/hashflow/ethereum/hashflow_ethereum_trades.sql
```diff
@@ -0,0 +1,424 @@
...
+                                \'["justabi", "jeff-dude"]\') }}'
+    )
+}}
+
+{% set project_start_date = '2021-04-28' %}
+
+with ethereum_traces as (
+    select *
+    from {{ source('ethereum', 'traces') }}
+    where `to` in ('0x455a3B3Be6e7C8843f2b03A1cA22A5a5727ef5C4','0x9d4fc735e1a596420d24a266b7b5402fe4ec153c',
```
Hosuke, line 20: Also the first address can be in lower cased

justabi: oop - missed this! on it

Hosuke:
I think we can try another [approach](https://dune.com/queries/1492715).
Make use of `hashflow_ethereum.Pool_evt_Trade` and `hashflow_ethereum.HashflowFactory_evt_PoolCreated` decoced tables.

justabi:
@Hosuke - please may you elaborate? are you referring to replacing new_router cte specifically or a new approach to the entire query

my understanding is data before april is only in router_call_tradesinglehop, which is why it's used -> https://dune.com/queries/1492849

Hosuke:
> @Hosuke - please may you elaborate? are you referring to replacing new_router cte specifically or a new approach to the entire query
> 
> my understanding is data before april is only in router_call_tradesinglehop, which is why it's used -> https://dune.com/queries/1492849
> 
> i believe we may need to either:
> 
> * dedupe this table on the 3 (previously) defined unique keys -> composite index, block time and tx hash
>   or
> * like you suggested, update the definition of uniqueness to include call trace address - but from looking at the data this doesn't seem intuitive to me: https://dune.com/queries/1493138
>   or
> * simply filter out failed transactions (this would also work, but wary of doing this as it may be useful to be able to report on failed transactions at times)

You are right, just noticed the [earliest record](https://dune.com/queries/1493239) in `hashflow_bnb.Pool_evt_Trade` is on 2022-01-29 22:34.

My suggestion is that we filter out failed transactions to build an version that can be passed the uniqueness test first(like a MVP version), and then we can develop the sub spell that can show up all trades with failed transactions if we have time.

justabi:
@Hosuke oki that makes sense!! will do that now

justabi:
yay @Hosuke - tests pass now

justabi:
what happens now @Hosuke

justabi:
@jeff-dude i think we're good for a final review + merge

jeff-dude:
thanks all for the help pushing this forward. i pushed some final edits, hopefully good with all:
- rename files / db objects to ```raw_trades``` and ```trades``` so that our final object is consistent with other dexs. i know hashflow is an outlier due to no decoded, but i think this flows better downstream
- update not incremental logic to use ```{% if not is_incremental() %}``` to be consistent throughout spellbook. if/else does work, and may be cleaner, but i opted for consistency with other models
- added filters to tokens/prices views for blockchain = ethereum, this has caused dupes on other dexs without

i believe that's it, we'll see how the gh action tests run 🤞

jeff-dude (approved):

PR #1792, opened 2022-10-18
OP Contracts - Comment out Null

MSilb7:
Small change, just commenting out one contract creator that was null, because we never want it mapped. But we don't want the project to show as "null."

Is there a good way to add a check to make sure that this address is never entered in the table? cc @chuxinh 

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:


jeff-dude:
i think since this is just a select from hardcoded values list, you can simply apply a where clause at the end. try to compile your query locally, run on dune, then add where filter, ensure it works, bring it back to your code and submit 👍 

the alternative testing approach (probably not as good):
you can add a test which writes a query to return false if that address is present in the view. it won't prevent the load up front though, just tell you afterwards and raise an error in automation.

two options:
1. write test macro in generic folder, apply test in schema file of your model here (example of this is used on dex trades token bought address)
2. write test in tests/contracts/optimism that is specific to this use case, can be a normal query format vs. macro in first option

PR #1794, opened 2022-10-18
remove STARL due to bad price feed

bh2smith:
This Price is WAY OFF on coin Paprika for the date October 17 (and continuing today).

CP is quoting 2.44$ but its far below 0.000

https://coinpaprika.com/coin/starl-starlink/

https://etherscan.io/token/0x8e6cd950ad6ba651f6dd608dc70e5886b1aa6b24

Here is an example trade from yesterday

https://etherscan.io/tx/0xaba68eef58a89ae4d740281d7565951488a38c633415e246cff645e63ffc75af

This is affecting Dex trades and all other exchange related metrics!

Should probably also run some script to purge the prices from the tables as well.

### DIFF
--- a/deprecated-dune-v1-abstractions/prices/ethereum/coinpaprika.yaml
+++ b/deprecated-dune-v1-abstractions/prices/ethereum/coinpaprika.yaml
@@ -1780,11 +1780,6 @@
   symbol: ARCH
   address: 0x1f3f9d3068568f8040775be2e8c03c103c61f3af
   decimals: 18
-- name: starl_starlink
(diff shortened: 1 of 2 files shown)


bh2smith:
Looks like they have fixed it:

https://coinpaprika.com/coin/starl-starlink/

Unfortunately it is still affecting the dex.trades table.

bh2smith:
Hey Dune. we are still seeing bogus prices in `dex.trades` table related to that token we dropped yesterday. 

Query: https://dune.com/queries/1427499

PR #1806, opened 2022-10-19
Add ens_poapid_devcon6

makoto:

*For Dune Engine V2*
I've checked that:
General checks:
* [x ] the filename is unique and ends with .sql
* [x ] each sql file is a select statement and has only one view, table or function defined  

Join logic:

Incremental logic:


augustog:
Hello @makoto, thanks for the addition! Apologies for the slow response from my side, I wanted to ensure we didn't have a better source for this information. 

From what I see in the dataset this should be fairly easy to obtain as a more generalised abstraction on top of our nft.events dataset. Are you missing any of the data points you need on there?

The location of the dataset may perhaps be not ideal, if what you want is to extend POAP's with metadata we don't currently have on nft.events or its children, may I suggest creating a new project directory for POAP specifically? 

We are discussing the option of implementing NFT metadata datasets, but we have no concrete plans at the moment, so if that's something you need, hardcoding is a good way to go

Lastly, if it is to be merged, two comments:
- Please make sure that all columns created by the sql code have a definition on the schema.yml file
- While we don't have a strong policy against this, I'd say having a field for the ENS domain without TLD hardcoded onto the file should be avoided unless there's a strong need for it

Please let me know if you have any questions or if what I proposed above wouldn't solve your use case.

augustog:
Hello @makoto!

Please let us know if you need any help regarding my comments above. Otherwise, please let us know if this PR is no longer relevant.

Thanks!

augustog:
Closing this for lack of activity. If it's still relevant please reopen it and we will find the best way to get these addresses onto dune.

Issue #1785, opened 2022-10-17
Token Prices Not in Table (wstETH & rETH)


bh2smith:
https://github.com/duneanalytics/spellbook/blob/df795b398746c0fb132db32a8f63e8414301a85c/models/prices/prices_tokens.sql#L590-L591

Among these two tokens both rETH and wstETH are not appearing in the table at all:

1. [rETH](https://coinpaprika.com/coin/reth-rocket-pool-eth/) appears to be "inactive" on coin paprika (which makes sense).
2. [wstETH](https://coinpaprika.com/coin/wsteth-wrapped-liquid-staked-ether-20/) is NOT inactive

cf: this query

```sql
select count(*) from prices.usd where
contract_address in (
    '0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0',
    '0xae78736cd615f374d3085123a210448e74fc6393'
)
```

jeff-dude:
this appears resolved
```
select
  contract_address,
  count(*)
from
  prices.usd
where
  contract_address in (
    0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0,
    0xae78736cd615f374d3085123a210448e74fc6393
  )
group by
  contract_address
```

PR #1644, opened 2022-09-27
[Prices] Add 387 Missing Token Prices

bh2smith:
Similar to yesterday's addition of 50 tokens (#1639), today we attempt to add another 387.

This was built with a script that fetches all active [tokens from coin paprika](https://api.coinpaprika.com/v1/coins) and "joining" on the results from their [contracts endpoint](https://api.coinpaprika.com/v1/contracts/eth-ethereum). 

We then take the results of this [tokens with missing prices query](https://dune.com/queries/1317238) and match whatever we can (uniquely by address).

In the future we can clean up the script and generalize to make this work for multiple networks (at least those which [coin paprika supports](https://api.coinpaprika.com/v1/contracts/))

cc @soispoke

PR #1807, opened 2022-10-19
add 450 ethereum token prices

bh2smith:
Run using the same script at in #1644. However, I must make a note that I had to manually remove the entry:

("wsteth-wrapped-liquid-staked-ether-20", "ethereum", "wstETH", "0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0", 18),

Which has already been added but still not appearing in the price feed!

There was a reported issue about this in #1785

Script to generate these results is:

```
python -m src.missing_prices
```

from https://github.com/cowprotocol/data-misc (soon we will make a docker file so it can be run without cloning and installing the repo).


soispoke:
tagging @aalan3 for visibility, but lgtm 👍

soispoke (approved):

bh2smith:
Note that there is still an open issue with the wstETH price (which exists in this token file but not yet in the price table!). https://github.com/duneanalytics/spellbook/issues/1785

PR #1809, opened 2022-10-19
Fix Gnosis Chain MIST token Symbol

bh2smith:
While working on a ["cross chain token map"](https://dune.com/queries/1415800) - I noticed that the MIST token (which is the emoji ⚗️) appears as ?? in V2 engine. Since I had the ethereum mainnet mapping of the token it was clear to see that this is the MIST token.

Screenshot 2022-10-19 at 1 07 24 PM


soispoke (approved):
lgtm, thanks @bh2smith

PR #1824, opened 2022-10-20
Fix LooksRare Duplicate

hildobby:

Fix duplicate `unique_trade_id`, see issue here: https://dune.com/queries/1434604?d=1
Tx here which shows its actually distinct trades not dups: https://etherscan.io/tx/0x7fa82bbe1dbca851e153f03237ebc6e2e8e0aeb3dffc90c9f02cde62ba4c2c11

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:

### DIFF
--- a/models/looksrare/ethereum/looksrare_ethereum_events.sql
+++ b/models/looksrare/ethereum/looksrare_ethereum_events.sql
@@ -170,7 +170,7 @@ SELECT DISTINCT
     royalty_fee / looks_rare.price * 100 as royalty_fee_percentage,
     royalty_fee_receive_address,
     royalty_fee_currency_symbol,
-    'looksrare' || '-' || looks_rare.tx_hash || '-' ||  token_id::string || '-' ||  seller::string || '-' || COALESCE(erc.contract_address, nft_contract_address) || '-' || looks_rare.evt_index::string || '-' || COALESCE(evt_type::string, 'Other')  || '-' || erc.evt_index  || '-' || COALESCE(case when erc.value_unique::string is null then '0' ELSE '1' end, '1') as unique_trade_id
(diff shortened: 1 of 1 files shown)


hildobby:
PR needed for https://github.com/duneanalytics/spellbook/pull/1536

jeff-dude:
the successful run/test from gh action is a bit deceiving this time around. check out this query:
https://dune.com/queries/1434821?d=1

where:
- prod shows the issue we're aware of
- dbt_builtin_test shows the default query dbt builds for ```unique``` property assigned in schema file. notice the where clause to filter out null ```unique_trade_id``` -- dbt thinks it's fine
- remove_filter_from_builtin shows the second option without the filter. we see there are 79 rows with a null ```unique_trade_id``` -- as of now, incremental runs, but this shows in the future we can have nulls and it'll break. also the history for those 79 would be invalid

looks like the field needs fixed further?

i'd also like to explore if there is any way we can mimic dex trades approach of unique keys / unique combo of columns test. in dex trades, all models use the same keys, but they seem to vary in nft trades. ideally there's a subset of columns in nft trades which are unique regardless of which model feeds it

hildobby:
> the successful run/test from gh action is a bit deceiving this time around. check out this query: https://dune.com/queries/1434821?d=1
> 
> where:
> 
> * prod shows the issue we're aware of
> * dbt_builtin_test shows the default query dbt builds for `unique` property assigned in schema file. notice the where clause to filter out null `unique_trade_id` -- dbt thinks it's fine
> * remove_filter_from_builtin shows the second option without the filter. we see there are 79 rows with a null `unique_trade_id` -- as of now, incremental runs, but this shows in the future we can have nulls and it'll break. also the history for those 79 would be invalid
> 
> looks like the field needs fixed further?
> 
> i'd also like to explore if there is any way we can mimic dex trades approach of unique keys / unique combo of columns test. in dex trades, all models use the same keys, but they seem to vary in nft trades. ideally there's a subset of columns in nft trades which are unique regardless of which model feeds it

Hmm weird, I'm not sure where the empty ones are coming from so I issued a new commit with all the fields used for `unique_trade_id` coalesced (defaults to -1). It either fixes it or at least should help us identify where those empty lines are coming from by just searching for `--1`

jeff-dude:
okay, that does technically remove the dupes (can refer back to linked query above, updated).

i guess we can push this through, fyi @soispoke 

in the longer term, @soispoke also mentioned he would offer up his services to help determine if there are a set of unique columns in nft trades we can use universally & apply unique keys/unique combo of columns tests similar to dex trades

jeff-dude (approved):

PR #1833, opened 2022-10-24
Extend nft_ethereum.native_mints to before block 14801608

UmerHA:
A small correction to PR 1709: Extend nft_ethereum.native_mints to before block 14801608.

*For Dune Engine V2*
I've checked that:
General checks:

Join logic:

Incremental logic:
* [] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables


soispoke:
In this case I think we have to start from the beginning, but yeah tests need to pass before we can look to merge anything

UmerHA:
Hey @augustog, @soispoke

I fixed the two bugs:
- also removed block_number limit from test ✅
- implemented better duplication test for `ethereum.contracts` ✅ (due to a bug `ethereum.contracts` contains duplicates, ie same address has multiple namespaces. In PR 1709 I assumed they at least have a unique `created_at` value, which turned out to be false. Therefore a better duplication test was needed)

The tests run locally:
- unique col test: https://dune.com/queries/1454781 ✅
- custom test: https://dune.com/queries/1454797 ✅

So I believe the code is correct. The timeout makes sense to me, as removing the block number limit increases the blocks to process by ~15X (from 6.5% of all blocks to 100%,  see https://dune.com/queries/1454666).

Is there a way to remove or increase the timeout limit for this kind of PRs?

soispoke:
@UmerHA all good, needed to merge last changes from the main branch to get dbt slim ci running correctly, but everything lgtm now

soispoke (approved):

PR #1834, opened 2022-10-24
Move docs on Optimism Contract Mapping to yaml

augustog:
This is a very small change to keep our documentation consolidated on YAML files, should be treated as low priority.

This way we can have spell-specific documentation (like in the case of optimism contracts) living as a github-friendly README.md file, *and* as part of dbt-docs.

image

What I did:
- [x] add docs block to yml file
- [x] add docs block marker to readme file
- [x] validate `dbt docs generate` locally


augustog:
@team not at all a high priority here, but I'd like y'all's take of having this as our way of keeping spell-specific docs in the future

soispoke (approved):

soispoke:
lgtm

dbt_project.yml
```diff
@@ -15,7 +15,6 @@ test-paths: ["tests"]
 seed-paths: ["seeds"]
 macro-paths: ["macros"]
 snapshot-paths: ["snapshots"]
-docs-paths: ["docs"]
```
davidkell, line 18: AFAIK if you remove this, we need to move the `docs/overview.md` file into the `models` directory. This file is used to render the custom welcome page in the docs site

jeff-dude (approved):
i haven't dug too deep into the docs side yet, but in theory this lgtm. i think it's certainly an area we can experiment with

dbt_project.yml
augustog: let me test it locally to ensure it still works!

augustog (commented):
Latest commit addresses @davidkell 's point. (indeed the config change reverted the docs to the default).

jeff-dude:
@augustog please let me know if i need to monitor anything execution-wise of this code in prod. otherwise, it's merged and should be set 👍

PR #1847, opened 2022-10-28
Add Zigzag Arbitrum to dex.trades

mtitus6:

*For Dune Engine V2*
I've checked that:
General checks:
* [ X] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ X] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [X ] if adding a new model, I added a test
* [ X] the filename is unique and ends with .sql
* [X ] each sql file is a select statement and has only one view, table or function defined  
* [ X] column names are `lowercase_snake_cased`
* [ X] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [X ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributors)

Join logic:
* [ X] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

Incremental logic:
* [X ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ X] where block_time >= date_trunc("day", now() - interval '1 week')
* [ X] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ X] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')


models/zigzag/arbitrum/zigzag_v1_arbitrum_trades.sql
```diff
@@ -0,0 +1,106 @@
...
+    select
+      call_block_time as block_time,
+      zzmo.makerOrder:sellToken as token_sold_address,
+      zzmo.makerOrder:buyToken as token_bought_address,
+      zzmo.output_matchedFillResults:takerSellFilledAmount as token_bought_amount_raw,
+      zzmo.output_matchedFillResults:makerSellFilledAmount as token_sold_amount_raw,
+      NULL AS amount_usd,
+      zzmo.makerOrder:user as maker,
+      zzmo.takerOrder:user as taker,
+      call_tx_hash as tx_hash,
```
Hosuke, line 29: Hi @mtitus6 , I have reviewed the code, and overall it seems legitimate to me.
I am not sure why the `dbt slim ci` test have not been triggered.
I am guessing is because of the advanced sql syntax you are using here(get attributes out of a json).
So here is the (uniqueness test)[https://dune.com/queries/1517700] I manually ran

Hosuke:

- Inner join to large tables(arbitrum.transactions) as possible ✅

Hosuke:
[Sample query results for this model](https://dune.com/queries/1517719)

models/zigzag/arbitrum/zigzag_v1_arbitrum_trades.sql
mtitus6: I am not sure either.  

Is there something I need to change or should I just remain patient with the existing code?

jeff-dude: hi both -- please ensure to merge dune main branch into working branch(s) to keep up-to-date. i pushed a commit with the merge of main and the tests are now running

jeff-dude:
pushed a few commits:
- fixed missing comma issue in add to tokens file (also shifted to left comma to make copy/paste easier in the future)
- removed schema property in config block, as dbt_project file covers schema name
- removed ```'``` from ```block_date``` in seed file
- removed not null test from ```output_matchedFillResults``` column as the test doesn't filter where call_success = false, like the models does, so we get false positive failed tests
- forced ```evt_index``` to be int to match other dexs

jeff-dude:
please review the gh action test now. in the dbt slim ci job, you will see failed test after incremental run. it appears no value for ```evt_index``` may be an issue. i've noticed all other dex models have a value. are we certain it can't be populated? if not, what is workaround for passing unique combo of columns test? do we have to force a value such as ```1``` default or something?

Hosuke:
> please review the gh action test now. in the dbt slim ci job, you will see failed test after incremental run. it appears no value for `evt_index` may be an issue. i've noticed all other dex models have a value. are we certain it can't be populated? if not, what is workaround for passing unique combo of columns test? do we have to force a value such as `1` default or something?

@mtitus6 `evt_index` can be obtained from the `arbitrum.logs` table by joining on `tx_hash` and (topics = `taker` or `maker`). Another approach is that we submit more Zigzag contracts to decode and obtain a trade events table(usually with `evt_index`). Currently I see there is only one source for Zigzag (`zigzag_test_v6_arbitrum`) which may be insufficient for the model.

mtitus6:
Thanks @Hosuke  for the suggestion.  The logs appear to be emitted from the erc20 transfers and not the swap.  The zigzag contract does not have any events or logs and uses all functions.  Maybe because it is native to SOL?  See this tx --> https://arbiscan.io/tx/0x41343dc4169bd83e83793598585fd9e2c916d0db698cb544d6a3a30962bd3d68 .  All logs from USDT and USDC contracts.  Am I missing something?  I suggest we proceed with 0 or 1.

Hosuke:
> Thanks @Hosuke for the suggestion. The logs appear to be emitted from the erc20 transfers and not the swap. The zigzag contract does not have any events or logs and uses all functions. Maybe because it is native to SOL? See this tx --> https://arbiscan.io/tx/0x41343dc4169bd83e83793598585fd9e2c916d0db698cb544d6a3a30962bd3d68 . All logs from USDT and USDC contracts. Am I missing something? I suggest we proceed with 0 or 1.

It seems you are right.
image
image
We can try with `0` or `1` for `evt_index` to see if it works for incremental test.

mtitus6:
I adjusted evt_index = 1, and adjusted dex_trades_seed file accordingly.

Hosuke:
> I adjusted evt_index = 1, and adjusted dex_trades_seed file accordingly.

Seems like this works for incremental tests.

Hosuke:
> please review the gh action test now. in the dbt slim ci job, you will see failed test after incremental run. it appears no value for `evt_index` may be an issue. i've noticed all other dex models have a value. are we certain it can't be populated? if not, what is workaround for passing unique combo of columns test? do we have to force a value such as `1` default or something?

GM @jeff-dude , we have set the default value for `evt_index` as `1`, and it seems like this modification passes the dbt slim ci job. The `evt_index` for Zigzag Arbitrum is hard to define in this case, as shown in https://github.com/duneanalytics/spellbook/pull/1847#issuecomment-1308165396.

augustog:
At first glance seems like hardcoding evt_index in this case will not be an issue since it's only used as a uniqueness key and not necessary for any join (as is the case with nft.*), so I'd agree w @Hosuke on the solution here. 

Only question is: do we envision a case where this could throw duplicates? e.g. transactions with multiple trades perhaps? In that case, do we think decoding more zigzag contracts would enable us to avoid this workaround?

mtitus6:
Hi @augustog.  The exchange contract uses only functions not events which creates a unique situation for defining evt_index 😬.  I'm not sure about duplicate potential but if you have an idea on how to verify it, I would be glad to look into it.  One idea would be to rank the evt_index partitioned by the tx_hash.  This would ensure no dupes but not sure if value added. Thoughts?

jeff-dude:
> Hi @augustog. The exchange contract uses only functions not events which creates a unique situation for defining evt_index 😬. I'm not sure about duplicate potential but if you have an idea on how to verify it, I would be glad to look into it. One idea would be to rank the evt_index partitioned by the tx_hash. This would ensure no dupes but not sure if value added. Thoughts?

i like the approach of rank partitioned by tx_hash rather than original idea of hardcoded int -- want to push that change?

mtitus6:
Added the ranking logic for evt_index to prevent duplicates and synced to main.

jeff-dude (approved):
think we're set here -- thanks again @mtitus6 🤝

jeff-dude:
> think we're set here -- thanks again @mtitus6 🤝

and @Hosuke as always for the support!

PR #1848, opened 2022-10-28
[Easy] Add missing tokens

gentrexha:
Usual token update taken from missing tokens query:

V1: https://dune.com/queries/236085
V2: https://dune.com/queries/984709


augustog:
- Spot checked token addresses ✅ 

This is making me think we should consider a more automated approach to token metadata. @gentrexha could you point me to the script that y'all use to find and populate these? 

Nothing set in stone yet, but I want to explore if we could streamline this process from our side :)

augustog (approved):
Spot checked token metadata ✅

jeff-dude:
> * Spot checked token addresses ✅
> 
> This is making me think we should consider a more automated approach to token metadata. @gentrexha could you point me to the script that y'all use to find and populate these?
> 
> Nothing set in stone yet, but I want to explore if we could streamline this process from our side :)

script is in description :)

jeff-dude:
@gentrexha in the v1 file, please ensure to use proper spacing. postgres is finicky with that. i pushed a commit for reference. an easy way to ensure is to copy from a previous row and paste.

jeff-dude (approved):

bh2smith:
> @gentrexha in the v1 file, please ensure to use proper spacing. postgres is finicky with that. i pushed a commit for reference. an easy way to ensure is to copy from a previous row and paste.

Also @gentrexha  - the reason the spacing didn't work for you is because PyCharm formats the output different that what the script writes. If you run the script in your shell you can directly copy and paste the output into your PR. Sorry I didn't mention this earlier.

gentrexha:
> @gentrexha in the v1 file, please ensure to use proper spacing. postgres is finicky with that. i pushed a commit for reference. an easy way to ensure is to copy from a previous row and paste.

Thanks for fixing merging this @jeff-dude and @augustog. I was on vacation.

> @gentrexha could you point me to the script that y'all use to find and populate these?

Here's the script we use to fetch the token data: https://gist.github.com/gentrexha/7edf90ab334a3dcb1468880461d1df8f

PR #1869, opened 2022-11-02
Move nexusmutual items from v1 to v2

sharkxff:

*For Dune Engine V2*
I've checked that:
General checks:

Pricing checks:

Join logic:

Incremental logic:


sharkxff:
I've checked that General checks

models/nxm/ethereum/nxm_ethereum_schema.yml
```diff
@@ -0,0 +1,103 @@
...
+  - name: nxm_ethereum_quotation_trades
+    meta:
+      blockchain: ethereum
+      project: nxm
+      contributors: [zxf, jonah]
+    config:
+      tags: ['ethereum','nexusmutual','trades']
+    description: >
+        nexusmutual quotation trades on Ethereum
+    columns:
```
Hosuke, line 13: Hi @sharkxff , you can add an uniqueness test here.
```suggestion
    description: >
        nexusmutual quotation trades on Ethereum
    tests:
      - dbt_utils.unique_combination_of_columns:
          combination_of_columns:
            - evt_block_number
            - evt_tx_hash
            - evt_index
    columns:
```

models/nxm/ethereum/nxm_ethereum_quotation_trades.sql
```diff
@@ -0,0 +1,79 @@
+{{ config(
+	alias ='trades',
+	partition_by = ['block_date'],
+	materialized = 'incremental',
+	file_format = 'delta',
+	incremental_strategy = 'merge',
+	unique_key = ['block_date', 'cover_block_number', 'statusNum', 'evt_tx_hash', 'evt_index'],
```
Hosuke, line 7: Hi @sharkxff , the unique_key list here should match the columns in uniqueness test in schema.yml

Hosuke:
Hi @sharkxff, as this is a new model, we may need to have some more tests on it.
You can add some test cases(which have been manually verified on etherscan.io) in the seed file.
The seed file directory would be like `seeds/nxm/ethereum/nxm_ethereum_trades_seeds.csv`.
You can check these example seeds for reference: [perpetual_trades_seeds](https://github.com/duneanalytics/spellbook/blob/main/seeds/perpetual/trades/perpetual_trades_seed.csv), [x2y2_seeds](https://github.com/duneanalytics/spellbook/blob/main/seeds/x2y2/ethereum/x2y2_ethereum_trades_etherscan.csv), [dex_trade_seeds](https://github.com/duneanalytics/spellbook/blob/main/seeds/dex/trades/dex_trades_seed.csv).
Then we can use these seeds to write more specific tests in the schema.yml.

Hosuke:
If you can add some tests on specific columns using [generic test](https://github.com/duneanalytics/spellbook/tree/main/tests/generic) that would be better.

dbt_project.yml
```diff
@@ -582,6 +589,16 @@ seeds:
...
+    nxm:
+      +enabled: true
+      +schema: test_data
+      ethererum:
+        nxm_ethereum_trades_seed:
+          +column_types:
+            evt_tx_hash: string
+            block_date: date
+            #TODO Please fill in column name and its type for the seed
+
```
Hosuke, line 601: Hi @sharkxff , please follow the pattern and fill in the data type for the columns in the seed file.

models/nxm/ethereum/nxm_ethereum_schema.yml
```diff
@@ -0,0 +1,120 @@
...
+        description: "evt_tx_hash"
+      - &curr
+        name: curr
+        description: "curr"
+      - &premium
+        name: premium
+        description: "premium"
+        tests:
+          - nxm_trades_column_values:
+              seed_file_location: ref('nxm_ethereum_trades_seed')
```
Hosuke, line 48: And we can test specific column values like this way

Hosuke (dismissed):
Hi @sharkxff , please make the seed available for the project by adding it to dbt_project.yml

dbt_project.yml
```diff
@@ -469,6 +469,13 @@ models:
         +schema: hashflow_ethereum
         +materialized: view
 
+    nomad:
+      +schema: nexusmutual
+      +materialized: view
+      ethereum:
+        +schema: nexusmutual_ethereum
+        +materialized: view
+
```
Hosuke, line 478: You may mistakenly name the model as `nomad`, which is another existing model. I will commit a fix.

Hosuke:

- Metadata looks good ✅
- Some column names in the schema file are not lowercased, but match the names in the source decoded table.
- Inner join to large tables(ethereum.transactions) as possible ✅

jeff-dude:
thank you @Hosuke & @sharkxff -- below are my notes:
- renamed anything ```nxm``` to ```nexusmutual``` to match standard at decoded table level, let's strive for consistency
- moved forward with column rename to lowercase/underscore, i see hosuke mentioned, but i think we should follow through
- moved the date filters to within join conditions themselves for consistent approach across spellbook
- added blockchain filter on join to erc20 tokens
- removed descriptions in schema file which didn't add value past column name. they aren't a must if we don't have good descriptions

with the above changes, i didn't make huge logic changes outside of the blockchain filter on join to erc20 tokens (which also is an inner join, usually we do left join since it's manually maintained and can get behind -- should this update?). however, the gh action is now failing on duplicates post-incremental run. can someone validate the dupes and see why?

Hosuke:
> with the above changes, i didn't make huge logic changes outside of the blockchain filter on join to erc20 tokens (which also is an inner join, usually we do left join since it's manually maintained and can get behind -- should this update?).

I also noticed is the inner join with erc20 tokens, maybe nexusmutual only supports limited erc20 tokens and all listed in the table. I think we need @sharkxff to confirm this.

sharkxff:
After the compiled sql runs group by unique_key(block_date,cover_block_number,status_num,evt_tx_hash,evt_index ) in the DUNE UI, the result is" No results from query". However, dbt-test shows that !=0 unique_key results can be found. This part needs help to check.

sharkxff:
image

sharkxff:
image

models/nexusmutual/ethereum/nexusmutual_ethereum_quotation_trades.sql
```diff
@@ -0,0 +1,99 @@
...
+    {% if is_incremental() %}
+    WHERE evt_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
+),
+quo_cse_data as (
+    select * FROM (
+        select  *,
+            row_number() over(partition by cid order by evt_block_time desc) as time_rank
+        FROM {{ source('nexusmutual_ethereum', 'QuotationData_evt_CoverStatusEvent') }}
+    ) as quo_cse
```
Hosuke, line 45: Please try to avoid using `select *` for performance gain, unless all columns needed(In that case you still can list out all columns and make it clearer for review and maintainance).

Hosuke:
> image

I write the query to show out [all grouped results with count(*)](https://dune.com/queries/1584259?d=1). It seems each row appears only 1 time. I will have a closer look for this issue.

Hosuke:
I think the [test failure] is caused by the unique keys is not strong enough for the incremental merge, which means there are overlapping part during the incremental update.

sharkxff:
But before yesterday, the unique value judgment was success

Hosuke:
> But before yesterday, the unique value judgment was success

Can you add more descriptions about columns in the schema.yml and sources.yml files for the model? As we need to understand this model better to help to solve the issue.

Hosuke:
Wait, the checks passed!

sharkxff:
> > But before yesterday, the unique value judgment was success
> 
> Can you add more descriptions about columns in the schema.yml and sources.yml files for the model? As we need to understand this model better to help to solve the issue.

I changed the table structure and removed some fields, and the compile has been successful so far

jeff-dude (approved):
thank you @sharkxff and @Hosuke 🤝 
i pushed one last fix for seed file. the join wasn't returning any rows, giving us false positive test results, due to bad seed data.

PR #1873, opened 2022-11-02
changed UNION to UNION ALL

sohwak:

I checked the current scripts, `UNION` is used too much. I think this will significantly impact the performance aspect of the current DB.

In most cases, using `UNION ALL` is highly recommended rather than using `UNION` except in some cases on purpose.

`UNION ALL` simply merges two tables. So, it checks whether only the number of columns and data type are the same then merges them immediately.
On the other hand, `UNION` looks similar but additionally, it removes all duplicated rows. To this, DB performs sorting or something equivalent to find a duplicate value. All volumes (columns * rows) are included in this heavy operation.

This is a SQL grammar, so I think there is no exception regardless of DB type or architecture. At least every DB I used did.

For this PR, I have modified only the parts where the logic is clear and most large performance degradation is expected.

*For Dune Engine V2*
I've checked that:
General checks:

Pricing checks:

Join logic:

Incremental logic:

### DIFF
--- a/models/dex/dex_trades.sql
+++ b/models/dex/dex_trades.sql
@@ -51,7 +51,7 @@ FROM (
         evt_index
     FROM {{ ref(dex_model) }}
     {% if not loop.last %}
-    UNION
+    UNION ALL
     {% endif %}
     {% endfor %}
 )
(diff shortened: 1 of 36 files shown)


sohwak:
fyi.
https://stackoverflow.com/questions/49925/what-is-the-difference-between-union-and-union-all
https://stackoverflow.com/questions/46260599/performance-of-union-vs-union-all

sohwak:
I will close this and make another PRs. Thanks.

PR #1875, opened 2022-11-02
Add polygon to post hook for explorer

aalan3:
Forgot to add this one so that this shows up correctly in the explorer


jeff-dude (approved):
good catch

jeff-dude:
our prod job runs views only if new or edited, i'll be curious to see if schema change kicks it off. if you don't see this in ~1-2 hours, we can manually run

PR #1880, opened 2022-11-02
Fix wash filter

hildobby:

Added case columns to group by and also added an anti left join for incremental runs to make sure trades arent abstracted twice.

*For Dune Engine V2*
I've checked that:
General checks:

Pricing checks:

Join logic:

Incremental logic:


hildobby:
I just pushed and was trying to look into the test that errors in dune but it doesn't work 🤔 

Screenshot 2022-11-03 at 21 07 49

jeff-dude:
> I just pushed and was trying to look into the test that errors in dune but it doesn't work 🤔
> 
> Screenshot 2022-11-03 at 21 12 21

yeah, if the step where it actually creates that table fails, it won't exist. if you haven't already, i'd recommend compiling locally and running on dune -- does that work?

hildobby:
> yeah, if the step where it actually creates that table fails, it won't exist. if you haven't already, i'd recommend compiling locally and running on dune -- does that work?

Ah you're right, found the issue and fixed it that way! I also ran some tests on dune.com and ran the abstraction on the past 1 month of data yields no duplicates, see here: https://dune.com/queries/1522465 . So I am pretty sure duplicates come from rerunning the table on incremental runs. I added a recursive check, this should now make the abstraction work as intended:

```
{% if is_incremental() %}
LEFT ANTI JOIN {{this}} nft_wtf ON nftt.unique_trade_id = nft_wtf.trade_unique_id
{% endif %}
```

jeff-dude (approved):
i think this is good, i'll merge it & we can monitor

PR #1886, opened 2022-11-03
UNION to UNION ALL for nft related views

sohwak:

Objective: https://github.com/duneanalytics/spellbook/pull/1873

Change `UNION` to `UNION ALL` for nft views.
the changes in this PR are only for views related `nft` spell; this will not take much time to modify, but I expect a remarkable performance improvement.

*For Dune Engine V2*
I've checked that:
General checks:

Pricing checks:

Join logic:

Incremental logic:


sohwak:
fyi, I compared each of performance:
https://dune.com/dune_hack/union-vs-unionall

dot2dotseurat:
Hi @sohwak, I did some experimentation with the table produced by your PR `test_schema.git_923fc0e3_nft_trades` and `nft.trades` and I don't see a performance improvement with respect to query time. 

When I do a limit 100 on each view, `nft.trades` had a query time of 6.66M and 1.47h task time and `test_schema.git_923fc0e3_nft_trades` had a query time of 7.43m and 1.47h of task time. I don't think we are seeing a difference in query time because we are not doing much (if any) deduplication. We use UNION instead of UNION ALL in case we have duplicates in the view but there should theoretically be no duplicates based on how this table is set up. 

What I do see is the task time on element_bnb.events is WAY higher than all the other underlying tables and I've just verified that the table seems to be partitioned on block_time and it should be block_date. I think we should keep using UNION instead of UNION ALL but look to improve the query performance by fixing the element_bnb.events table.  

I'm going to close these UNION v UNION ALL PRs if that is ok with you and open a fix for the element_bnb.events table. Thank you for raising this PR because I don't I would have seen that issue otherwise and it should help querytime on nft.trades significantly.

sohwak:
@dot2dotseurat, thank you for the test. I was worried that it wouldn't make much difference because the current DB performs well, and it seems this happened. Still, I'm glad you found improvements in other parts and found them helpful.

Sorry for bothering you but could you do one more test? The difference between `UNION` vs `UNION ALL` is well-known query performance degradation, so the result is a little bummer.

Could you please take `EXPLAIN` for the below queries?
https://dune.com/queries/1525390
https://dune.com/queries/1525493

It is put `EXPLAIN EXTENDED` at the beginning of the query, but it is not possible for the general user who does not have permission. https://docs.databricks.com/sql/language-manual/sql-ref-syntax-qry-explain.html

Actually, the query itself was written to take a lot of loads. The current DB performance is good, so I put a lot of performance load on purpose because I thought it would be done quickly if I didn't do this.

I wonder how DB handles this query at least.

dot2dotseurat:
@sohwak there's some sensitive data in EXPLAIN (mainly s3 buckets). I started to redact them but it could be easier to summarize why one of those queries is much faster than the other (afaik). The UNION ALL query is fully supported by [Photon](https://www.databricks.com/product/photon) whereas the UNION query is not. 

When I do the same test with nft.trades (union versus union all). Both are fully supported by Photon. I think that's why didn't see the gains you saw on your sample queries. 

To be honest, I don't see why the UNION query isn't supported by Photon but the UNION ALL one is. The error message is `Photon scan does not support the columns having data types: [(access_list,array)]` which seems weird to me because the tables in both queries have the same column types so it might have something to do with applying DISTINCT to that data type which UNION would need to do. 

This does make me think that we should be surfacing if Photon is supporting queries back to wizards. That difference in performance for your examples is shocking.

dot2dotseurat:
This is really good work and I'm sorry we aren't giving you good tools to dig into this yourself. It's really useful that you are looking into this.

sohwak:
@dot2dotseurat 
Yeah, well, thank you for testing. I wanted to help with performance personally, and I also curious how it was handled here with databricks. Anyway, the Photon engine seems to handle it well; I'll close this up.

And by the way, It wasn't asking to use a tool. I think the policy here, Dune v2 is much better than Dune v1. I hope that another feature (even if for a paid-version) will come out.

dot2dotseurat:
It did give me the idea that I should really find a way to easily expose if the query is supported by Photon or not. It seems like it could be useful. Thanks again, we really appreciate your looking into performance stuff.

jkylling:
Link T-20616

soispoke:
@sohwak after looking into it more it seems like improvements are actually quite big using the UNION ALL instead of UNION approach so I'm going to merge this, thanks again for contributing !

soispoke (approved):

sohwak:
Oh, that's also a good conclusion. I'm glad it was helpful.

PR #1871, opened 2022-11-02
Create nft.transfers: a table combining erc721, erc1155single & erc1155batch transfers

hildobby:
Brief comments on the purpose of your changes:

Combining all nft transfer tables to more easily access all nft transfers with 1 join instead of 3 different joins

*For Dune Engine V2*
I've checked that:
General checks:
* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each sql file is a select statement and has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributors)

Pricing checks:
* [ ] `coin_id` represents the ID of the coin on coinpaprika.com
* [ ] all the coins are active on coinpaprika.com (please remove inactive ones)

Join logic:
* [ ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

Incremental logic:
* [ ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ ] where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #1897, opened 2022-11-04
Add polygon to tornado cash abstractions

hildobby:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


hildobby:
Current issue is from `polygon.transactions` not being defined in base tables which I fixed in this PR: https://github.com/duneanalytics/spellbook/pull/1871

hildobby:
Merged main within which polygon base tables were added, should fix undefined tables issue.

dot2dotseurat:
@sohwak there's some sensitive data in EXPLAIN (mainly s3 buckets). I started to redact them but it could be easier to summarize why one of those queries is much faster than the other (afaik). The UNION ALL query is fully supported by [Photon](https://www.databricks.com/product/photon) whereas the UNION query is not. 

When I do the same test with nft.trades (union versus union all). Both are fully supported by Photon. I think that's why didn't see the gains you saw on your sample queries. 

To be honest, I don't see why the UNION query isn't supported by Photon but the UNION ALL one is. The error message is `Photon scan does not support the columns having data types: [(access_list,array)]` which seems weird to me because the tables in both queries have the same column types so it might have something to do with applying DISTINCT to that data type which UNION would need to do. 

This does make me think that we should be surfacing if Photon is supporting queries back to wizards. That difference in performance for your examples is shocking.

hildobby:
Alright awesome ty for the clarification @dot2dotseurat, just changed all the UNION ALLs to UNIONs in a new commit 🤝

augustog (approved):
👍

jeff-dude (approved):

jeff-dude:
to @augustog point, i went ahead a used variables for date filters on historical loads. also added a tag to schema file.

PR #1906, opened 2022-11-07
labels added for Alameda

andrewhong5297:

added Alameda labels and one ftx label from Larry's thread https://twitter.com/lawmaster/status/1589510081694949377

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


andrewhong5297:
one thing I'm curious about is how to label addresses across all EVM chains - since technically they have control on all chains. Should I just list them all one by one in the array? or is there some jinja macro we can create for `{{ all_evm_chains() }}`?

augustog:
are there any addresses here not covered by #1906?

andrewhong5297:
> are there any addresses here not covered by #1906?

What do you mean? This is PR #1906

augustog:
Oh sorry!! My bad, I meant #1904 🤦

andrewhong5297:
Ah I didnt see that one - i think its the same, just the one CEX difference and using macro instead.

andrewhong5297:
Okay I checked against existing, this PR adds about 20 more addresses given by The Block yesterday. 

I also still don't understand why this is labeled as ethereum, funds control the address across all EVM chains so I think we should use the macro to have `{ allEVMchains() }` instead of `array('ethereum')` for the funds and cex tables.

andrewhong5297:
ah it looks like those are in #1904 but changes aren't merged yet 😓 

guess we made PR at the same time haha

soispoke:
haha just saw this, looks like we had the same idea @andrewhong5297 :)

1chioku:
@andrewhong5297

Just a note on the schemas, as the files get thicker, it is easier to make mistakes and add in duplicates.
Is there not a rationale to add a `not unique` test to the schemas on the address column? (For all the labels in general)

I thought a bit about it and I do not see obvious scenarios of duplication, safe for the case of having one address with multiple names, tags, e.g a multi sig contract with more than 1 owner, e.g a DAO treasury safe owned by many protocols collectively.

soispoke (commented):
The only thing we have right now for all labels is the test defined here @1chioku, is this what you mean ?

https://github.com/duneanalytics/spellbook/blob/main/models/labels/labels_all_schema.yml

soispoke (commented):
I merged main into this branch since most changes were already integrated in #1904 

I'm not sure I understand the logic/value of the macro you created @andrewhong5297 could you give me a bit more context ? 🙏

1chioku:
> https://github.com/duneanalytics/spellbook/blob/main/models/labels/labels_all_schema.yml

Yes, I was thinking to add the following to the sub folder schemas:

```
- name: address
        description: "Label address"
        tests: 
        - unique
```

Here is the context: yesterday when I got the addresses of FTX and wanted to add them to the CEX, but I had to check one by one to make sure it was not already added. Because the table can get really crowded, it is easy to add a duplicate. If we were to add the test above, then the dbt test would actually fail when a contributor was to add a duplicate into the table.

However there could be issues with adding that test at the labels_all_schema.yml level, the reason being that an address could haver multiple labels. e.g let's say you had a hacker which a "fund" which is also a "tornado cash user", so the address has 2 labels.

https://github.com/duneanalytics/spellbook/blob/main/models/labels/labels_all.sql

In the code above you use `UNION` and not `UNION ALL` which I think prevents an address from having 2 labels. Note that on Etherscan, many addresses have multiple labels, so there might be a need to rethink labels as plural and not singular.

soispoke:
Ok @1chioku thanks for taking the time to give context !

I do agree it should be more clear how we deal with multiple labels attributed to the same address. But I think the current design does allow that, since the UNION is based on all columns, so if the name or category is different the address will definitely have the two labels attributed to it

andrewhong5297:
> I merged main into this branch since most changes were already integrated in #1904 
> 
> I'm not sure I understand the logic/value of the macro you created @andrewhong5297 could you give me a bit more context ? 🙏

Yeah sure - mainly its that these funds addresses exist across all chains so I think the table/schema should reflect that. macro just makes it so you dont have to type out the array

jeff-dude (approved):

jeff-dude:
looks like #1963 was merged quick and added the FTX US address. we'll keep your name in there for credit 🤝

PR #1907, opened 2022-11-07
Exclude Tests From some Hooks

bh2smith:
Extending on https://github.com/duneanalytics/spellbook/pull/1761/files

I have noticed that these hooks fail in many cases on the test directory (a place model properties and description aren't relevant - or at least -  aren't used)

See also [this failing run](https://github.com/duneanalytics/spellbook/actions/runs/3409911862/jobs/5672249882):where only test files were changed.

This is related to a comment made [here](https://github.com/duneanalytics/spellbook/pull/1761#issuecomment-1305483882)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,4 +7,6 @@ repos:
   - id: dbt-compile
   - id: check-script-ref-and-source
   - id: check-model-has-description
+    exclude: ^tests/
   - id: check-model-has-properties-file
+    exclude: ^tests/


bh2smith:
The failing dbt-here doesn't make any sense (this change is based on main) which does not include those test files. I don't understand how those tests are even being considered.

dot2dotseurat:
@bh2smith thank you, we appreciate this fix and agree! the hooks are intended to be in beta and I was hoping to do more work on them last week but my week filled up with other tasks. please feel free to submit more changes as you see fit.

dot2dotseurat (approved):

PR #1856, opened 2022-10-31
All new Seaport Trades

sohwak:
Brief comments on the purpose of your changes:

Fully newly created `seaport.trades` spell table with increased accuracy

Feature: 
- Only event table is used to increase accuracy and reduce error probability

Next plans:
- When data integrity is verified then integrate into the following data:
  - `nft.trades`
  - `seaport.transfers` (refer to `seaport.trades`, and remain for backward-compatibility)
  - `seaport.transactions` (refer to `seaport.trades`, and remain for backward-compatibility)
- Create and deploy other chains as well including Avalanche, Arbitrum, Polygon, Optimism, etc.

*For Dune Engine V2*
I've checked that:
General checks:
* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [x] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [x] if adding a new model, I added a test
* [x] the filename is unique and ends with .sql
* [x] each sql file is a select statement and has only one view, table or function defined  
* [x] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributors)

Pricing checks:
* [ ] `coin_id` represents the ID of the coin on coinpaprika.com
* [ ] all the coins are active on coinpaprika.com (please remove inactive ones)

Join logic:
* [x] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

Incremental logic:
* [x] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [x] where block_time >= date_trunc("day", now() - interval '1 week')
* [x] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [x] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #1911, opened 2022-11-07
Seaport.trades on Polygon

sohwak:

`Seaport.trades` on Polygon
related to All new Seaport table : https://github.com/duneanalytics/spellbook/pull/1856

- create Seaport tables on each chain
- and then will integrate into related views after this. (`seaport.trades`, `opensea.trades`, `nft.trades`,... )

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


sohwak:
One exception exists:
https://polygonscan.com/tx/0x74f3fbd3d5fa2ff45fc7d76d60ffb0e68aa2efdc1f673c0fd25665399e856ee7
same tx_hash + evt_index + token_contract + token_id...
I will think how to make `unique_id` again

augustog:
@sohwak do we know why this is the case? if those four fields don't uniquely identify a trade, I'm pretty lost...

cc @0xBoxer more erc1155 edge cases 🎉, can you think of a reason why [this transaction](https://polygonscan.com/tx/0x74f3fbd3d5fa2ff45fc7d76d60ffb0e68aa2efdc1f673c0fd25665399e856ee7) has duplicates with
same tx_hash + evt_index + token_contract + token_id ?

sohwak:
I'm just looking for a moment to see why this case is happening. To change `unique_id` if it is a typical case.

Hosuke:
> One exception exists: https://polygonscan.com/tx/0x74f3fbd3d5fa2ff45fc7d76d60ffb0e68aa2efdc1f673c0fd25665399e856ee7 same tx_hash + evt_index + token_contract + token_id... I will think how to make `unique_id` again

Hi @sohwak , can we add the `evt_index` for `TransferSingle` log event as part of the unique id (for erc1155)?
image
image

sohwak:
Thanks, @Hosuke; it didn't use trace, only using the decoded table itself. I added `sub_idx` to solve the conflict.

Hosuke:
> Thanks, @Hosuke; it didn't use trace, only using the decoded table itself. I added `sub_idx` to solve the conflict.

Do you think the `sub_idx` can be applied to all 5 PRs, and also for all erc1155 tokens in this case?

sohwak:
@Hosuke , Yep I believe this works well. I also added this to other Seaport sources just in case.

sohwak:
Please merge https://github.com/duneanalytics/spellbook/pull/1944 before this.
It relies on the `token.erc20` that will be marged above.

Hosuke:
After tests, the minimum set of columns for uniqueness test would be ['tx_hash', 'evt_index', 'nft_contract_address', 'token_id', 'sub_idx']. `sub_idx` in this case is derived from the `offer_idx` of column `offer` in `seaport_polygon.Seaport_evt_OrderFulfilled` table([sample query](https://dune.com/queries/1607803?d=1)). Moreover for incremental updates of the view, I add 'block_date' into the unique key list in the config block.

Hosuke:
For the `seaport_polygon_base_pairs` view, as the records increase over the time, we can modify it to an incremental view, to avoid repeating calculation.

sohwak:
> After tests, the minimum set of columns for uniqueness test would be ['tx_hash', 'evt_index', 'nft_contract_address', 'token_id', 'sub_idx']. `sub_idx` in this case is derived from the `offer_idx` of column `offer` in `seaport_polygon.Seaport_evt_OrderFulfilled` table([sample query](https://dune.com/queries/1607803?d=1)). Moreover for incremental updates of the view, I add 'block_date' into the unique key list in the config block.

Yep, these would be good.

sohwak:
Yeah, please proceed with this. I also have a plan to merge this into `opensea.trades` with others

Hosuke:
Hi @sohwak , do you think the modification in this PR is applicable to other PRs as well?

Hosuke:
> Hi @Hosuke, I think `sub_idx` is enough, but also good to add `sub_type` to be safe. applying for other PR is good, too.
> 
> It is hard to explain (my poor Eng and also I haven't fully understood how seaport works), but in Seaport, it looks like all items in `offer` equal all items in `consideration`. So it has to be unique with only ['tx_hash', 'evt_index', 'nft_contract_address', 'token_id'] I think, but there was an odd case. I also want to test in OS front how this is possible, but I have no time to do that this week.

We may remove `sub_type` in the future, or just rename `sub_idx` to `offer_idx`(which is its origin). Because I am not sure about 'consideration' and 'offer', so I keep the `sub_type` info for the downstream as we may use this info later.

Also, I think we need to find some test cases/seeds for this PR, and make some tests in the schema file.

sohwak:
> We may remove `sub_type` in the future, or just rename `sub_idx` to `offer_idx`(which is its origin). Because I am not sure about 'consideration' and 'offer', so I keep the `sub_type` info for the downstream as we may use this info later.

Then let's leave this `sub_idx` itself. It is just an index of `explode` and if it is 'offer' then `offer_idx`, 'consideration' then `consideration_idx`.
And I will let `base_pairs` as a view (or table), so it is good to let the same column name `sub_idx` refer to it.

> Also, I think we need to find some test cases/seeds for this PR, and make some tests in the schema file.

Yep, but let's do it future

Hosuke:
Hi @sohwak @jeff-dude , I have added seeds and tests to this model.
For the test, I use the ideas in this PR https://github.com/duneanalytics/spellbook/pull/1707.

Besides, I have reviewed this PR and here is the summary:
- Metadata and naming convention looks good ✅
- Inner join to large tables(polygon.transactions) as possible ✅

Since there are other 4 PRs similar to this one. If we can make this PR robust, we can keep the code consistency for other 4 PRs. Also a good example to follow for other nft.trades models in future.

jeff-dude (approved):
thank you @sohwak & @Hosuke -- some great collaboration to get this in a prod-ready state 🙌 

i pushed a final few commits to remove ```unique_trade_id``` and add ```block_date``` to use in joins rather than ```block_time``` since that is partition column. i'm sure spark would be able to translate the time easily, but better to play it safe. if we end up pushing this to ```nft.trades``` before we make changes on unique keys there, we can always add back a ```unique_trade_id``` in that script

if either of you want to apply all the changes to the other seaport PRs to fully match this one, i'm happy to merge through after.

thanks again!

sohwak:
Thank you @Hosuke and @jeff-dude. The reply was late because I was OOO last week. I checked the changes, and thank you for doing everything well. I agreed with all the changes, and I will modify the other 4 PRs as similarly as possible and re-submit them soon.

PR #1916, opened 2022-11-07
added three utility macros

andrewhong5297:

Spoke with @soispoke about this earlier, created two macros I want as UDFs for getting the symbol/prices_symbol/prices_address or block explorer base url for a given chain. I also added an href creator so people can stop doing the long `concat <a href = "" target='_blank'> </a>` thing.

Not sure if these need tests or how to test them.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


henrystats:
this is amazingggg and going to be really helpful, thank you serrr!

jeff-dude:
did we determine how these can be used post-merge? i was thinking the macro will exist in spellbook, but can only referenced in future spell models, not necessarily directly in the db (i.e. when querying dune)

andrewhong5297:
> did we determine how these can be used post-merge? i was thinking the macro will exist in spellbook, but can only referenced in future spell models, not necessarily directly in the db (i.e. when querying dune)

@soispoke mentioned the macro would be usable in the query interface too

andrewhong5297:
Ah also not by computer but the explorer url macro should only return url, none of the other columns

sohwak:
I happened to see this post, and I also think it would be very useful. I was just dealing with the price of the native tokens on the multi-chain spells, so I am more interested.

I looked at the tokens that were added, and all tokens used as alternatives in other native tokens are the same as mine except for the ones in Polygon. I was referring to MATIC `0x0000000000000000000000000000001010` for the native token on Polygon. I was curious and looked up WMATIC `0x0d500b1d8e8ef31e21c99d1db9a6444d3adf1270` this PR mentioned, but it is not currently in `prices.usd`.

So I think it would be better to use MATIC for Polygon.

andrewhong5297:
Makes sense to me!

andrewhong5297:
okay fixes are in, this should be ready to merge or test now

macros/public/get_chain_native_token.sql
```diff
@@ -0,0 +1,15 @@
+{% macro get_chain_native_token(chain, column) %}
+   SELECT
```
jeff-dude, line 2: will this fail with two selects?

jeff-dude:
if you want to test pre-merge, you can add sql files under ```tests/macros/public/``` which call each macro within a query. can be a basic query, but you'd at least see results and can force result to return nothing to pass -- i.e. filter out expected result

macros/public/get_chain_native_token.sql
andrewhong5297: haha yes oops

andrewhong5297:
can't seem to call them inside tests.

https://github.com/duneanalytics/spellbook/actions/runs/3421135094/jobs/5696864292

Screen Shot 2022-11-08 at 12 12 44 PM

andrewhong5297:
Okay, the three tests are passing now!

jeff-dude (approved):
nice work!

dot2dotseurat:
@andrewhong5297 looks like this is causing issues for `dbt compile`. I'm going to revert and see what we can do to exclude this automatically from compilation.

PR #1918, opened 2022-11-07
element_bnb.events partition on block date

dot2dotseurat:
After taking a look at @sohwak's contribution on swapping union for union all, I saw that querytime was especially high on element_bnb.events and I believe that is due to selection block_time instead of block_date as the partition key. 

This will require a full refresh of the tables if it works.

### DIFF
--- a/models/element/avalanche_c/element_avalanche_c_events.sql
+++ b/models/element/avalanche_c/element_avalanche_c_events.sql
@@ -1,7 +1,7 @@
 {{ config(
     schema = 'element_avalanche_c',
     alias = 'events',
-    partition_by = ['block_time'],
+    partition_by = ['block_date'],
     materialized = 'incremental',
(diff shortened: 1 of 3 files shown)


dot2dotseurat:
@jeff-dude I ran some test queries between this test_schema version and the prod version. Comparing a limit 100 on each version of the table reduces the files read from 104 to 1. I think this change will improve query time on nft.trades because we were creating too many partitions before.

jeff-dude:
bnb is the outlier for sure, likely due to tx volume. most spells on bnb take a ton of time.

dot2dotseurat:
@jeff-dude good point, I missed ethereum and pushed an update to that chain as well.

jeff-dude:
oops, i didn't see you already had avalance too

jeff-dude (approved):

PR #1930, opened 2022-11-09
[CoW Protocol] Fix Batches Table!

bh2smith:
Based on the ongoing discussion in #1752, we took a very careful look at the query building this table, to discover that the incremental change logic was not including records which were contained in the "outer" part of the outer join. In brief, what we had was

```sql
select 
from table_a 
left outer join table_b 
    join on condition
-- this was not captured by the outer join because the field values were null.
where condition on table_b.block_time 
```

What we have here is 

```sql
select 
from table_a 
left outer join table b 
    join on condition
       -- Now is captured
       and condition on table_b.block_time 
```

This [PoC query](https://dune.com/queries/1290518) demonstrates that the [missing records](https://dune.com/queries/1389623) are contained in the new query!


soispoke (approved):
lgtm!

jeff-dude:
good stuff, ben 🙌

bh2smith:
Everything looks great! So glad to have overcome this hurdle. Thanks all for your shared persistence!

PR #1938, opened 2022-11-09
fix dbt slim ci action to only run incremental steps when needed

jeff-dude:
looks like gh action runs even when no incremental models return, as ```wc -l``` appears to count blanks as a line


jeff-dude:
closing for now. low priority.

PR #1940, opened 2022-11-10
Add Uniswap DAO votes and proposals

soispoke:
New spell category alert!

DAO votes and proposals for Uniswap v3

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


jeff-dude:
i think we've asked this before: are we sure we can update pipfile objects without hurting users local env's?

models/uniswap/ethereum/uniswap_ethereum_schema.yml
```diff
@@ -219,4 +219,89 @@ models:
...
+      blockchain: ethereum
+      sector: dao
+      contributors: soispoke
+    config:
+      tags: ['ethereum', 'proposals', 'dao', 'cross-chain']
+    description: >
+      DAO proposals on all chains across all contracts and versions
+    columns:
+      - *blockchain
+      - *project
```
jeff-dude, line 283: do we know if this ```*``` usage will pickup from the dao one above and not a dex usage? it may flow to the docs reusing the values from the dex one. not sure how we handle similar column names across different models types in the same schema file

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
```diff
@@ -0,0 +1,71 @@
+{{ config(
+    schema = 'uniswap_v3_ethereum',
+    alias = 'proposals',
+    materialized = 'incremental',
```
jeff-dude, line 4: would it make sense to partition by ```block_date``` like other models? i.e. will downstream dashboards filter on or group by date? it would help performance in that case. if not, all good

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
```diff
@@ -0,0 +1,71 @@
...
+
+SELECT 
+    '{{blockchain}}' as blockchain,
+    '{{project}}' as project,
+    '{{project_version}}' as version,
+    pcr.evt_block_time as block_time,
+    pcr.evt_tx_hash as tx_hash, -- Proposal Created tx hash
+    '{{dao_name}}' as dao_name,
+    '{{dao_address}}' as dao_address,
+    proposer,
```
jeff-dude, line 48: can we look to add table alias to all columns in select? it helps for quick readability/review

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
jeff-dude: same concept to other model

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
jeff-dude: same concept to other model

soispoke:
pre commit is failing because of for loops (known issue) but looks good otherwise

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
soispoke: Yeah I thought about that but tables are really small (e.g., proposals is like 25 rows) so I didn't think it would make a difference, will still add it in case it does when a lot of DAOs are added in the future

soispoke: Done

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
soispoke: Done, if I understood your comment well :)

models/uniswap/ethereum/uniswap_ethereum_schema.yml
soispoke: Done, you were right, rewrote the columns so it doesn't pick up names/descriptions from dex models

models/dao/dao_schema.yml
```diff
@@ -0,0 +1,95 @@
...
+      contributors: soispoke
+    config:
+      tags: ['ethereum', 'proposals', 'dao', 'cross-chain']
+    description: >
+      DAO proposals on all chains across all contracts and versions
+    columns:
+      - *blockchain
+      - *project
+      - *version
+      - *block_time
```
augustog, line 73: For votes it seems fairly obvious that block_time = time the vote was cast
Probably here it means time when the proposal was created? Or does it mean this is the time as-of-when the vote counts are the ones below? Maybe we should rename to clarify?

models/dao/dao_schema.yml
```diff
@@ -0,0 +1,95 @@
...
+    description: >
+      DAO votes on all chains across all contracts and versions
+    columns:
+      - &blockchain
+        name: blockchain
+        description: "Blockchain which the DAO is deployed"
+      - &project
+        name: project
+        description: "Project name of the DAO"
+      - &version
```
augustog, line 20: How do protocol versions map here? Does e.g. uniswap have different DAO contracts for each version of the protocol?

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
```diff
@@ -0,0 +1,73 @@
...
+with cte_support as (SELECT 
+        voter as voter,
+        CASE WHEN support = 0 THEN sum(votes/1e18) ELSE 0 END AS votes_against,
+        CASE WHEN support = 1 THEN sum(votes/1e18) ELSE 0 END AS votes_for,
+        CASE WHEN support = 2 THEN sum(votes/1e18) ELSE 0 END AS votes_abstain,
+        proposalId
+FROM {{ source('uniswap_v3_ethereum', 'GovernorBravoDelegate_evt_VoteCast') }}
+GROUP BY support, proposalId, voter),
+
+cte_sum_votes as (
```
augustog, line 31: How will this work for active proposals where vote counts are rolling? Since the unique key refers to proposal creation exclusively (thus 1 row = 1 proposal), it seems like this would update the row while the prop is rolling with every run of the incremental job, and if a proposal lives for longer than a week (not sure if this edge case is reasonable) we may have inconsistencies?

I'm thinking perhaps it makes more sense to keep both datasets solely focused on what they represent, and have a `view` on top that calculates latest status of votes.

augustog (commented):
I'm really looking forward to explore possible use cases of these spells, they look super promising.

I made a few comments inline, mostly for discussion.

- I'm thinking we should use more context-specific naming for generic stuff like block_time, since it's not directly obvious (esp. for proposals) if it's the creation date, last update date, or date as-of-which the votes are counted.
- Similarly, I'm wondering if the concept of protocol version makes sense here in the same way than in `dex.*` / `nft.*`? 
    - Are DAO contracts similarly versioned?
    - Do they map 1-1 to the project version? (e.g. dao votes for uni v3 come from the contract version 3, and only pertain to uni v3)
- Lastly, I'm concerned about the edge case where a proposal lives longer than our incremental  update window (thus may leave vote counts off?)
    - Is this a reasonable edge case to worry about? (seems props on uni live on avg 5-6 days so it should be fine, but I'm wondering if this is going to be the case across the board)
    - If this is a valid concern, perhaps it makes sense to have either
        - A more explicit approach to counting votes to ensure we always capture all votes of a proposal before we stop updating it (e.g. incremental run includes all actives)
        - Or we spearate the votes entirely from the proposals table, or have a view built on top of these two tables to facilitate running the counts

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
jeff-dude: i meant the table alias prefix to the column in select/where/group by/etc -- i.e. ```pcr.votes_far``` so we know which table it comes from. no need to alias the column name, especially when it's the same

jeff-dude: sorry for confusion!

soispoke: ohhh that makes more sense 👍, done

models/dao/dao_schema.yml
soispoke: Yes, agree it would make more sense to have the column renamed as "created_at" (open to other suggestions if you don't like it) for proposals 👍 Done

models/dao/dao_schema.yml
soispoke: I'm confident the active, tally contract is specifically related to uniswap v3, and DAO contracts existed for previous versions, see screenshot of governoralpha contract for uniswap_v2

I think those should be added as well, but prob in a different PR
Capture d’écran 2022-11-14 à 08 40 55

models/uniswap/ethereum/uniswap_v3_ethereum_proposals.sql
soispoke: thanks, definitely a good catch you're right we have to be pretty careful about this. Tbh one straightforward option would just be to remove the incremental clause and refresh the spells on each run since the tables are really small?

soispoke:
thanks @jeff-dude and @augustog for your feedback!

Pipfile
```diff
@@ -4,7 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-dbt-databricks = "1.2.1"
+dbt-databricks = "1.3.0"
```
dot2dotseurat, line 7: Let's make this update in a separate PR. This change needs to be coordinated with upgrading the GH action runner.

models/uniswap/ethereum/uniswap_ethereum_schema.yml
```diff
@@ -219,4 +219,94 @@ models:
...
+      - &votes_share
+        name: votes_share
+        description: "Votes share in percent for a given proposal"
+      - &token_symbol
+        name: token_symbol
+      - &token_address
+        name: token_address
+      - &votes_value_usd
+        name: votes_value_usd
+        description: "Value in USD corresponding to the governance tokens used to vote on a given proposal"
```
dot2dotseurat, line 263: is this at the time of the vote or current USD value?

models/uniswap/ethereum/uniswap_ethereum_schema.yml
```diff
@@ -219,4 +219,94 @@ models:
...
+      - &voter_address
+        name: voter_address
+      - &support
+        name: support
+        description: "For, Against or Abstain votes"
```
dot2dotseurat, line 268: Is this a text field? Can we add an accepted_values test if so?

models/uniswap/ethereum/uniswap_ethereum_schema.yml
```diff
@@ -219,4 +219,94 @@ models:
...
+      - &votes_total
+        name: votes_total
+        description: "Total number of governance tokens used to vote on a given DAO proposal"
+      - &number_of_voters
+        name: number_of_voters
+      - &participation
+        name: participation
+        description: "Participation in percent: Number of governance tokens used to vote / Total token supply"
+      - &status
+        name: status
```
dot2dotseurat, line 308: Similar, can we use an accepted values test to ensure these are the only responses?

Pipfile
soispoke: Sounds good 👍

models/uniswap/ethereum/uniswap_ethereum_schema.yml
soispoke: At the time of the vote, this is consistent with what we do for nft.trades or dex.trades but I modified the description

models/uniswap/ethereum/uniswap_ethereum_schema.yml
soispoke: It's not, it's either `0`,`1` or `2` but I can add a test for these values (although this isn't always consistent across contracts)

soispoke: Done

models/uniswap/ethereum/uniswap_ethereum_schema.yml
soispoke: Done

models/uniswap/ethereum/uniswap_ethereum_schema.yml
dot2dotseurat: Can we define how each option maps to the ints in the description?
0 = Against, 1 = For, 2 = Abstain ?

dot2dotseurat:
I started to write some simple unit tests against the proposals table [here](https://dune.com/queries/1512113?d=1) and I see an issue with duplicated rows for "queued" proposals. 

I wanted to validate vote count for [proposal 16](https://www.tally.xyz/governance/eip155:1:0x408ED6354d4973f66138C91495F2f2FCbd8724C3/proposal/16) (selected randomly) and I see that there are 12 duplicated rows showing a queue status. Additionally, the proposal has already been executed so it should be complete. 

I checked the underlying source tables and I see single row for all of them for that proposal.  (I was hypothesizing it could have been queued multiple times or something). But I see more proposals with duplicated rows (12 in each instance). 

Can we add a test to check for a unique proposal id and work on a fix for this? I think it might be incremental logic since the query logic looks sound. TBH when I first read the PR I thought it would be silly to ask for that test since the SQL looks like it would proposal_id would obviously be unique. 

Screen Shot 2022-11-14 at 11 42 54 AM

dot2dotseurat (changes requested):
Lets add unit tests for both votes and proposals. I think it would be good to test:
- The outcome of two proposals (canceled and executed) to check the status, votes, and participation. (I think we can do this in SQL without adding a new seed)
- Uniqueness of proposal id in the proposals table
- A small number of votes (3?) just to confirm if they voted the way we think and with the number of tokens we expect. 
This should prevent future regressions and serve as a nice example for wizards adding new daos.

soispoke:
@dot2dotseurat thanks for catching this! Done :)

soispoke:
@dot2dotseurat @jeff-dude @augustog let me know when you guys think we can go ahead and merge this, so I can start working on adding new ones

jeff-dude (approved):
good with me outside of the few remaining comments (i.e. union vs union all) & pipfile. as mentioned, i'm not 100% sure the impact of that

models/dao/dao_proposals.sql
```diff
@@ -0,0 +1,40 @@
...
+        votes_against,
+        votes_abstain,
+        votes_total,
+        number_of_voters,
+        participation,
+        status,
+        description
+    FROM {{ ref(dao_model) }}
+    {% if not loop.last %}
+    UNION
```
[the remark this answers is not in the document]
soispoke: 👍

soispoke:
@jeff-dude thanks, I missed the UNION one (now done) but I did change the pipfile back to what it was 👍

jeff-dude:
> @jeff-dude thanks, I missed the UNION one (now done) but I did change the pipfile back to what it was 👍

i still see it in the edited files in this PR, only reason i mentioned it

soispoke:
@jeff-dude oh right the pipfile.lock one! Done :)

models/dao/dao_schema.yml
soispoke: Sounds good, yeah I think it actually might depend on what projects choose since this contract can be modified to suit particular need.

soispoke:
@dot2dotseurat I think I've made all the changes now but merging is still blocked 

@henrystats that sounds good, I think I'd like the merge the PR as is but adding uniswap_v3 dao address to the folder you're mentioning makes a lot of sense, please tag me in the PR if you end up opening one for this

dot2dotseurat (approved):

PR #1941, opened 2022-11-11
aztec_v2 connect spell [Ready for Review]

henrystats:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


henrystats:
Hola @jeff-dude, @augustog. I'm done with this query and I can't seem to figure out what is wrong with and why the test is failing. 

This was the query I converted to spell book (from dune) and it has no dupes - <https://dune.com/queries/1582754>


This was a query to test dupes using the test schema created by gh action and it has a lot of dupes - <https://dune.com/queries/1587639>

I'm quite confused on what could be wrong here, it's the same code (except the labels which I converted into a spell in the same pr.

But I can't pinpoint why the tests are failing, I'd really appreciate your help sers. Thank you very much.

augustog:
Hmm... I had a quick pass and nothing jumps out.

- Have you checked individual base tables to see if there were any dupes there? 
    - Since you're explicitly aggregating prices on a subset of fields, I would not worry about prices.usd (which has misbehaved in the past)
- Did you ensure that all joins are performed on all the necessary fields?

henrystats:
Hola @augustog, thank you sir! Yeah, it isn't prices.usd. The only view that is causing dupes is the rollup bridge transfers table. 

And I'm only left joining with the bridges label table, I'm using the same code, there are no dupes on the query itself so im very confused on what could be wrong.

henrystats:
I hardcoded the labels and didn't reference a spell and the tests are still failing. Not sure what to do anymore at this point.

henrystats:
this is the query without dupes - https://dune.com/queries/1582754

I really don't know what is wrong here 🥲, this is so frustrating!

henrystats:
bump, pls help ser @augustog @jeff-dude. The compiled query runs without dupes and the tests are failing in gh action. 

I was looking to get this finalised last week but the tests failing have slowed me down as I'm way behind on submitting this : (

jeff-dude:
hi @henrystats -- my first guess is no dupes in query, run initial load, tests pass as expected. incremental run starts, one of the unique_key values has nulls or blanks, which breaks the join on merge statement, then inserts duplicates. then the test fails after incremental run. can you double check unique_key values are valid across the board?

you can also push a commit to force gh action to run again, see test fail after incremental, then grab table name from gh action and query it on dune with a group by on keys/having count > 1 to find the problem rows to narrow it down.

henrystats:
Ermm, I just tried to run this and it looks like some transactions are missing. 

https://dune.com/queries/1640448?d=1

only 5 results in 2hrs and the original query returns more than 200 rows. (with 2hrs filter). 

I'm frustrated 🥲

henrystats:
cc @jeff-dude

jeff-dude:
@henrystats okay, i've had time to dig in now. thanks for the patience. [this query](https://dune.com/queries/1644948?d=1) is intended to mimic an incremental run on a model.
- CTE for incoming source data, filtered to a specific tx_hash which is causing a failure
- CTE for target data after initial run (**_note_**: it's important to recognize this is only one run from source, no incremental data ran or added yet)
- CTE to join source to target, to mimic a merge statement, which is what is used in this model (incremental, merge, insert new rows, update existing rows based on unique keys determined in config block)

notes to look at:
- source has one row (as expected)
- target has one row (as expected)
- join on unique keys, **no result returned** -- which means merge will **insert** the row again -- this is bad
- comment out a few join conditions on unique keys, one row is returned -- this is good, this is what we want

question becomes:
- why are these three columns causing issues in the join? is it due to null / blank space values? if so, do we need to default those values to a hardcoded value so the join works?
- better yet, can the unique keys be cut down to avoid these columns? there seems to be a lot of columns to determine unique. if its required, that's fine, revisit above point instead

i hope this helps you to continue testing. i realize you don't have local access to the db to test the same way i do, but hopefully you can mimic this approach through the gh action (i.e. try above approaches and see if it works). it's not quite as simple, as the gh action will run incremental right away, which makes your query harder to view as you'd want to analyze prior to an incremental run

henrystats:
Thank you very much ser @jeff-dude , I'll work on this later today and see if I can identify where the dupes are.

henrystats:
or sorry, fix what's causing this.

henrystats:
It finally worked!!!!! thank you soo soo much ser @jeff-dude 

I hardcoded values for when there are null values on the join and it works! gracias ser! This is now readyyy for review. 

Also yes, we do need a lot of columns for the unique tests, there are usually a bunch of transactions at the same transaction hash with the same contracts interacting with each other and some times same amounts being transferred.

henrystats:
@augustog pls can you help add the `ready for review` label tag ser!

jeff-dude (approved):

jeff-dude:
glad it helped, thanks for applying the fix! it's possible using just ```coalesce(column, '')``` into blank space would work too, to keep clean vs. a hardcoded word -- do you want to try that or good with the default 'Blank' value?

henrystats:
thanks ser @jeff-dude, I think '' would be better. 

I'll push a fix now.

jeff-dude:
gh action is a bit clogged from merging a bunch of PRs the last few hours, it should catch up soon. i'll check back tomorrow to see if this works as intended

henrystats:
okaie ser 🥳, the checks already completed.

PR #1960, opened 2022-11-15
Creating Inorganic Volume Filter Spell

henrystats:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


henrystats:
This is now ready for review ser @augustog (if you can help update the label)

models/inorganic_volume_filter/blur/inorganic_volume_filter_blur.sql
```diff
@@ -0,0 +1,324 @@
...
+        (
+        SELECT 
+            *, 
+            ROW_NUMBER() OVER (PARTITION BY collection ORDER BY evt_block_time DESC) as ordering 
+        FROM 
+        (
+        SELECT 
+            * 
+        FROM 
+        {{ source('looksrare_ethereum', 'RoyaltyFeeRegistry_evt_RoyaltyFeeUpdate') }}
```
augustog, line 45: Is this the right source event?

models/inorganic_volume_filter/blur/inorganic_volume_filter_blur.sql
```diff
@@ -0,0 +1,324 @@
...
+        {{ ref('tokens_erc20') }} erc20 
+            ON t.currency_contract = erc20.contract_address
+            AND erc20.blockchain = 'ethereum'
+        WHERE t.project = 'blur'
+),
+
+filtered_trades as (
+        SELECT 
+            t.*, 
+            CASE WHEN mt.filter IS NOT NULL THEN true ELSE FALSE END as mt_filter,
```
augustog, line 275: Minor comment: `mt.filter is not null as mt_filter` should give the same result (without the case statement)

augustog (commented):
Hello @henrystats. I had a thorough read of the PR and I understand why it is structured as it is. I left a few small comments on the code, let me know what you think.

Regarding the structure of this on spellbook, this is a bit of a new layer of abstraction (I'd think of it in a similar category as `nft.wash_trades`, an interpretative layer), and it seems the analysis is primarily referring Opensea. I would suggest we move it under the opensea models (please correct me if I'm wrong). Based on the same assumption, I agree this shouldn't be spread into individual projects' schemas.

Does the model need to be materialised? I understand from the TG conversation that this may not be ready to be on the data explorer, perhaps it makes sense to keep it as a view and drop the post hooks? That way the model is still accessible to query from dune but is not on the data explorer.

Tell me if this makes sense and I'd be happy to help with the changes and please feel free to reach me on Discord if you need any help..

models/inorganic_volume_filter/blur/inorganic_volume_filter_blur.sql
henrystats: this was the equivalent on v1 `looksrare."RoyaltyFeeRegistry_evt_RoyaltyFeeUpdate"` and I also tried querying on v2 annd it returned results, why do you think it isn't the right source event? incase I need to double check something.

henrystats:
Hola ser @augustog, I didn't know one could query even without posthooks wao. That's amazing!!

So yeah, I agree to dropping the post hooks. Lemme try updating and doing that. Would I need to remove the file from dbt project?

models/inorganic_volume_filter/blur/inorganic_volume_filter_blur.sql
henrystats: ah yes ty, that doesn't make any sense haha, I guess it was to avoid null values? or do I need to put the true & false statements as strings?

thank you sooo soo much for your help ser!

augustog:
Not from the dbt_project.yml file, all models need to be referenced there so dbt can build them.

But if we put the model under opensea and merge the schema files you don't need to reference it again.

models/inorganic_volume_filter/blur/inorganic_volume_filter_blur.sql
augustog: Oh it's a high level check, since we have the models split by project and both models are nearly identical otherwise, I was asking in case a copypasta error left both models referring to looksrare.

henrystats:
okie ty ser, will update later tonight. 

Also, I think it would be better being separated from opensea per discussions on tg.

models/inorganic_volume_filter/blur/inorganic_volume_filter_blur.sql
henrystats: ah okay, ty ser, it was intentional.

henrystats:
How do I keep it as a view and drop the post hooks?, I was thinking to just remove the post hooks line, but if I change `materialised` to view then that would break the incremental structure, except if by view you just meant removing the post hooks.

henrystats:
Updated now ser @augustog

augustog:
Hey @henrystats!

The post hook (as we use it on spellbook) is used to expose the spell on the Dune UI, in itself has nothing to do with materialization.

To change from incremental to view, you should:
- set `materialized = 'view'` on all the config blocks
- remove the incremental filters from all your `where` and `join` statements
- when a spell is a view, you no longer need to specify `partition`, `unique_key`, `file_type` or `incremental_strategy` so you should drop those too

I also added a comment inline, we should avoid representing columns as string when they have a better fitting native type (in this case, boolean).

models/inorganic_volume_filter/looksrare/inorganic_volume_filter_looksrare.sql
```diff
@@ -272,13 +269,13 @@ trades_enrich as (
...
         SELECT 
             t.*, 
-            CASE WHEN mt.filter IS NOT NULL THEN true ELSE FALSE END as mt_filter,
-            CASE WHEN sb.filter IS NOT NULL THEN true ELSE FALSE END as sb_filter,
-            CASE WHEN lv.filter IS NOT NULL THEN true ELSE FALSE END as lv_filter,
-            CASE WHEN hp.filter IS NOT NULL THEN true ELSE FALSE END as hp_filter,
-            CASE WHEN wf.filter IS NOT NULL THEN true ELSE FALSE END as wf_filter,
-            CASE WHEN cb.filter IS NOT NULL THEN true ELSE FALSE END as cb_filter,
-            CASE WHEN cs.filter IS NOT NULL THEN true ELSE FALSE END as cs_filter,
+            CASE WHEN mt.filter IS NOT NULL THEN 'true' ELSE 'false' END as mt_filter,
```
augustog, line 272: We should avoid representing these as strings when they are boolean values.

The code before was correct. My comment meant to say that the `case` was redundant and the code could be more succinct, but it was correctly implemented :) 

(sorry for the confusion!)

I'd make these either:

- `CASE WHEN mt.filter IS NOT NULL THEN true ELSE FALSE END as mt_filter,`
- or `mt.filter IS NOT NULL as mt_filter`

Some people choose the first option to be more explicit, others prefer the one below. Not a big deal IMO, but whichever we do, data type should be bool.

henrystats:
Thank you very much ser @augustog, a question about changing from incremental to view. 

The query is running the ethereum.transactions table as well, wouldn't changing it from incremental to view make the query run a lot slower?

models/inorganic_volume_filter/looksrare/inorganic_volume_filter_looksrare.sql
henrystats: okaiee serrr, tyvm will update!

henrystats:
ah okay, I'll make the changes. Also if I don't need to specify the `unique keys`, wouldn't that affect the tests in the schema file?

@augustog

henrystats:
okaie ser, ty. It's updated. Also I already merged the schema files and moved inorganic_volume under opensea

jeff-dude (approved):
thanks team, looks good

jeff-dude (approved):
thanks team, looks good

PR #1962, opened 2022-11-15
CryptoPunks - fixes + new tables

catherine-2:

WIP - opening PR to start tracking progress / any error checks. Will tag and add more context once ready for review!

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


catherine-2:
Hi @jeff-dude ! This is ready for review.

This PR contains a few different tables all related to the CryptoPunks collection. Because they have their own marketplace and are an ERC20 they require some specific queries to pull data. There are also some nuances due to the June 2017 migration to the current contract that aren't supported in current tables. These new tables will make it easy for any wizard to pull stats on the collection.

Here's a summary of each file added or changed in the PR:

**Tables Edited**

1/ events
- the original query did not accurately show the buyer for bid sales (was sometimes NULL)
- this query addresses that so that this table + nft.trades are now accurate

**New Tables Added**

1/ current_listings
- select * from this new table will let a wizard pull all currently available public listings in the punks marketplace
- query looks at all listing and de-listing events per punk id, and where the most recent event was a listing it shows it as active in the final table

2/ current_owners_per_token
- select * will show a wizard the current owner of each token
- ERC20 balances tables don't work for CryptoPunks because of the 2017 contract migration. I pulled that into this query via the call table data in V1 which I confirmed is accurate

3/ listings_over_time
- show the number of active listings per day over time
- similar logic to current listings table

4/ owners_over_time
- show the number of unique wallets that owned a CryptoPunk per day over time
- similar logic to current owners table

5/ metadata
- punk type, number of attributes, and attributes array for each punk ID 

Please let me know of any suggestions or questions! Thank you!

jeff-dude:
thanks for the add & details @catherine-2 !
i pushed a final commit with few minor fixes, we'll see how the gh action runs go & then we can merge if successful and you agree with changes 🤝

catherine-2:
Thank you @jeff-dude ! These changes all look great to me

jeff-dude (approved):

jeff-dude:
fyi, moving forward, to help us meet a future-state requirement for ansi-sql compliance -- can you ensure to use cast() for type conversions, rather than ```::``` usage? i know many times code is copied from v1, where it works, but we need to continue to enhance for future architecture. existing models should get updated soon.

catherine-2:
> fyi, moving forward, to help us meet a future-state requirement for ansi-sql compliance -- can you ensure to use cast() for type conversions, rather than `::` usage? i know many times code is copied from v1, where it works, but we need to continue to enhance for future architecture. existing models should get updated soon.

Yes will do!

PR #1966, opened 2022-11-15
[CoW Protocol] Update solver Quasimodo submission address and name

harisang:
(Created a new clean PR)

This PR updates the solver Quasimodo's name to the more explicit Gnosis_Quasimodo, and also updates its submission addresses on Ethereum Mainnet that were just changed.

cc @bh2smith

PR #1976, opened 2022-11-16
[CoW Protocol] -- Correction to PR#1966 : Update solver info

harisang:
(Apologies for reopening this after it was merged as PR #1966 , but there needs to be a correction)

This PR updates the info of Quasimodo, which from now on is to be called Quasilabs. This applies to both mainnet and Gnosis Chain.

cc @bh2smith


bh2smith (changes requested):
Are you sure you changed the right one? This diff does not contain a removal of anything under the name "Gnosis_Quasimodo"

This doesn't look right.

harisang:
You were right, the local branch I was working on was outdated, and I did a reset as an easy way out, but didn't realize this would close this PR. Is there a way to reopen this instead of creating a new one?

bh2smith:
I think they may have some configuration to not allow force push (but I don't have any visibility on that). Might as well just create a new one.

PR #1980, opened 2022-11-16
Fix for duplications of ERC1155 transactions on Wyvern

augustog:
This is a draft PR to address an issue where some transactions from wyvern were getting duplicated due to issues in the join to erc1155.

Background:
- The primary reason why these duplicates occur is that some erc1155 transactions don't deduplicate on token_id
- The only other way we have to deduplicate these is by evt_index, but wyvern does not trigger events (so we join calls)
- At the moment we don't decode calls to erc1155 standard, so, as a workaround, @0xBoxer figured out we can join directly with ethereum.traces -- since we join on block time, the join's performance should not be that horrible

I added a very basic test to validate if this works, we should extend that test to account for other possibly duplicated txns.


jeff-dude:
closing due to #2138

PR #1939, opened 2022-11-10
continuing discussing on macros -> UDFs (fixing and testing with run-operations)

andrewhong5297:
Brief comments on the purpose of your changes:

**For Dune Engine V2**

I've checked that:

### General checks:
* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each sql file is a select statement and has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [ ] `coin_id` represents the ID of the coin on coinpaprika.com
* [ ] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:
* [ ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:
* [ ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ ] where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #1986, opened 2022-11-17
add native chain tokens as table

andrewhong5297:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:

### DIFF
--- a/models/tokens/tokens_schema.yml
+++ b/models/tokens/tokens_schema.yml
@@ -34,3 +34,24 @@ models:
         description: "NFT Token's symbol"
       - name: decimals
         description: "NFT Token's number of decimals"
+
+  - name: tokens_native
+    meta:
+      blockchain: avalanche_c, bnb, ethereum, gnosis, optimism, ethereum, arbitrum, solana
+      sector: tokens
+      contributors: ilemi
+    config:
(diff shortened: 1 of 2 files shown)


andrewhong5297:
added a new tokens model that gives us the native token and corresponding token symbol/address on prices.usd. Going to use this in a UDF in other PR #1939

models/tokens/tokens_schema.yml
```diff
@@ -33,4 +33,25 @@ models:
...
+    description: >
+        The chain native token and symbol, and the prices.usd symbol/address that is closest.
+    columns:
+      - name: chain
+        description: "the chain, as it's named in Dune" 
+      - name: symbol
+        description: "Token's symbol"
+      - name: price_symbol
+        description: "Token's prices.usd symbol"
+      - name: price_address
```
jeff-dude, line 54: would it make sense to put a unique check on the address? or maybe unique combo of columns that include address and chain?

andrewhong5297: It should always be only one row for chain, so unique on chain should be good enough. Maybe future chains will adopt the "deploy thing to same address everywhere" practice so not really necessary rn

jeff-dude (approved):

PR #1998, opened 2022-11-18
[Easy]Modify pre-commit-config to point to dune's repo

Hosuke:

Modify pre-commit-config to point to dune's pre-commit-dbt repo.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 default_stages: [push, manual]
 exclude: ^deprecated-dune-v1-abstractions/
 repos:
-- repo: https://github.com/offbi/pre-commit-dbt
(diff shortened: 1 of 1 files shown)


Hosuke:
In order to make this code work, we need to draft/make a release with tag v1.0.0 like [this](https://github.com/offbi/pre-commit-dbt/releases/tag/v1.0.0) for [dune's pre-commit-dbt repo](https://github.com/duneanalytics/pre-commit-dbt)

jeff-dude:
i wonder if we can just use ```main``` branch rather than tags? then we aren't changing it every time?

dot2dotseurat:
I feel good about using main instead of the tags. Could mess us up every once in a while but the trade off for minimizing steps to update feels worth it.

jeff-dude (approved):

PR #2011, opened 2022-11-19
Make nft.wash_trades easier to use + final small tweaks

hildobby:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


hildobby:
⚠️ MANUAL TIMEOUT CHANGE TO BE REMOVED BEFORE MERGING ⚠️

jeff-dude:
it's a bit worrisome to see the query still timing out. i think it will need to be refined further

hildobby:
> it's a bit worrisome to see the query still timing out. i think it will need to be refined further

It runs for every single NFT trade on Ethereum so it has quite a bit to run on so runtime does make a bit of sense to me. I can also run it on dune.com for the past day and it takes less runtime than running OpenSea's abstraction for the past day on dune.com 🤷 

I did add one more Join for increased accuracy but otherwise it's just variable name changes (so they're easier for others to find in the table)

You don't think it's good to go?

dot2dotseurat:
@hildobby we can't merge this because our incremental job would completely break down if we added a model that was this compute-intensive. But I think there are things we can do to refactor. 

refactoring is imho completely hellish/impossible if your model can't finish in five hours. So what I am suggesting if you use some sampled data and a "refactor test" which is just the current state rather than a test where we know all the values are 100% correct. Then you can make changes to your query without worrying about inadvertently breaking your past sql logic. 

I think this will be helpful to make the development cycle much shorter and make refactoring less time-consuming. When you are done just replace the sampled tables with the real ones and see if runtime has improved. 

I made two new sources. 
table_sample_nft.trades_2022_11_20
(nft trades from 2022-11-20)
table_sample_nft.addresses_events_ethereum_first_funded_by_GT_2022_11_10
(addresses_events_ethereum_first_funded_by greater than 2022-11-10)

I added a seed which is your existing logic but with the sampled tables above
nft_wash_trades_refactor_seed.csv

And finally, a test nft_wash_trades_refactor_test
which compares the wash trades columns from your model to the seed. 

As for refactoring, this is my suggestion. But don't hate me if it doesn't work. 
Convert all the joins to inner joins and create CTEs for each filter, merge them together, do a single left join to nft.trades at the end. 

so CTE_1 -> only trades with same buyer seller
...
CTE_MERGE 
UNIQUE_KEY, FILTER_1, FILTER_2... FILTER_N
...
nft.trades left join CTE_MERGE

PR #2016, opened 2022-11-20
Archipelago buyer/seller fix

hildobby:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:

### DIFF
--- a/models/archipelago/ethereum/archipelago_ethereum_events.sql
+++ b/models/archipelago/ethereum/archipelago_ethereum_events.sql
@@ -8,7 +8,7 @@
         post_hook='{{ expose_spells(\'["ethereum"]\',
                                     "project",
                                     "archipelago",
-                                    \'["0xRob"]\') }}'
+                                    \'["0xRob", "hildobby"]\') }}'
         )
 }}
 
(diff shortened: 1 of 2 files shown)


models/archipelago/ethereum/archipelago_ethereum_events.sql
```diff
@@ -195,39 +195,57 @@ SELECT
...
+    , te.platform_fee_percentage
+    , te.royalty_fee_amount
+    , te.royalty_fee_amount_usd
+    , te.royalty_fee_amount_raw
+    , te.royalty_fee_currency_symbol
+    , te.royalty_fee_receive_address -- null here
+    , te.royalty_fee_percentage
+    , te.unique_trade_id
+from trades_enhanced te
+left join {{ ref('nft_ethereum_transfers') }} buyer_fix on buyer_fix.block_time=te.block_time
```
jeff-dude, line 234: is it feasible to make these inner joins?

hildobby: No unfortunately this fix is only sometimes needed and thus often doesn't find anything to join on. So changing to INNER JOINs would remove all trades where this fix isn't needed. Why would INNER JOINs be better here?

jeff-dude: it's a large table to join & we've seen iffy performance on left joins. that's fine though, makes sense here

jeff-dude (approved):

PR #2028, opened 2022-11-21
T-20805 - Get nft.trades to work with Trino SQL

belen-pruvost:
Our attempt with @jkylling to make `nft.trades` work on Trino SQL.

The problem laid mainly on implicit casting of types that Databricks was doing. We tested that the new super explicit view definition runs nicely on Trino.


jeff-dude (approved):
sounds good @belen-pruvost 🤝

soispoke:
I'm worried this table UNIONed with other table in nft.trades will fail if we include the changes introduced by this PR, since the other ones will still have columns of type `string` ?

belen-pruvost:
> I'm worried this table UNIONed with other table in nft.trades will fail if we include the changes introduced by this PR, since the other ones will still have columns of type `string` ?

@soispoke that's a good point, maybe I can include similar changes on this PR for the other tables (`magiceden`, `looksrare_ethereum `, etc). WDYT?

soispoke:
@belen-pruvost Yes, I think it's more work but worth doing, making sure we have consistent types across nft.trades models now becomes super crucial since we can't rely on Databricks SQL to implicitly standardize it (which is actually probably a good thing to take care of anyways).

dot2dotseurat (approved):

denz-e:
I'm not 100% sure if it's because of this change, but some values seem incorrect now, e.g. OpenSea's platform_fee_percentage is now returning 3 instead of 2.5

models/opensea/ethereum/opensea_ethereum_events.sql
```diff
@@ -36,11 +36,11 @@ FROM
                 platform_fee_amount_raw,
                 platform_fee_amount,
                 platform_fee_amount_usd,
-                platform_fee_percentage,
+                CAST(platform_fee_percentage AS DECIMAL(38,0)) platform_fee_percentage,
```
jkylling, line 39: Probably this and other casts of percentages should be to Double and not Decimal(38, 0).

models/opensea/ethereum/opensea_ethereum_events.sql
```diff
@@ -77,11 +77,11 @@ FROM
                 ,platform_fee_amount_raw
                 ,platform_fee_amount
                 ,platform_fee_amount_usd
-                ,case when amount_raw > 0 then platform_fee_amount_raw / amount_raw * 100 end platform_fee_percentage
+                ,case when amount_raw > 0 then CAST ((platform_fee_amount_raw / amount_raw * 100) AS DECIMAL(38,0)) end platform_fee_percentage
```
jkylling, line 80: Should be Double here

models/opensea/ethereum/opensea_ethereum_events.sql
belen-pruvost: I'll fix this 👍

belen-pruvost:
> I'm not 100% sure if it's because of this change, but some values seem incorrect now, e.g. OpenSea's platform_fee_percentage is now returning 3 instead of 2.5

@denz-e thanks for reporting this - I have a fix incoming!

denz-e:
> > I'm not 100% sure if it's because of this change, but some values seem incorrect now, e.g. OpenSea's platform_fee_percentage is now returning 3 instead of 2.5
> 
> @denz-e thanks for reporting this - I have a fix incoming!

awesome!

PR #2029, opened 2022-11-21
T-20626 - Compatible null to varchar cast

belen-pruvost:
Replacing `NULL::string` to be `CAST(NULL AS VARCHAR)` on Spellbook views.


jeff-dude (approved):
@belen-pruvost thank you for the submission 🤝 as i write this, the gh action to test is still running & there's a decent chance it hits it's timeout at an hour due to the number of models edited (it'll run each). i do have confidence this is a "simple" change though and would be confident to push either way. we could also test by just doing a few models first to ensure it runs successfully before timeout, then add the rest. let's let it ride for a bit

belen-pruvost:
Sounds good @jeff-dude! 🙌 if it timeouts I can split it in a few PRs to let the check take it easy 😄

jeff-dude:
this may fall into the same category of the other PR, where we want to validate downstream objects (although this one is just string --> varchar, should be less risky for downstream failure)

PR #2032, opened 2022-11-21
T-20626 - Compatible null to varchar cast for aave

belen-pruvost:
Subset of changes originally added on https://github.com/duneanalytics/spellbook/pull/2029


jeff-dude (approved):
this should be good, i don't believe there is any downstream impact to consider

PR #2048, opened 2022-11-22
Chainlink Polygon Oracle Price Feed Spell

0xroll:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


Hosuke:

- Metadata and naming convention looks good ✅
- Inner join to large tables(polygon.logs) as possible ✅

jeff-dude:
hi @Hosuke & @0xroll -- thanks for the work so far. all looks good, up to the test:
```
with unit_test as (
    select
        case when test.oracle_price = actual.oracle_price then true else false end                                          AS oracle_price_test,
        case when test.proxy_address = actual.proxy_address then true else false end                                        AS proxy_address_test,
        case when test.aggregator_address = actual.aggregator_address then true else false end                              AS aggregator_address_test,
        case when test.underlying_token_address = actual.underlying_token_address then true else false end                  AS underlying_token_address_test
    from       test_schema.git_3b028b6f_chainlink_polygon_price_feeds actual
    INNER JOIN test_schema.git_3b028b6f_chainlink_polygon_get_price test 
    ON (actual.blockchain = test.blockchain 
    AND actual.block_time = test.block_time
    AND actual.feed_name = test.feed_name)

)

select * from unit_test
-- where (oracle_price_test = false                OR 
--       proxy_address_test   = false             OR
--       aggregator_address_test = false          OR 
--       underlying_token_address_test = false
--       )
```
looks like it may be returning a false positive, where the CTE produces no results at all. when adding seeds & tests, please do either write the query to catch this false positive or run ```dbt compile``` and run the query on dune to validate, like the above

can you tweak the test to return as expected?

jeff-dude:
other notes on final commit from my side, please do apply to all PRs moving forward:
- ensure dbt_project has the directory called out, otherwise schema name will default to something we don't want
- use the new, cleaner format for ```union all``` models with loop
- avoid usage of ```::``` for casting datatypes, this is a newer ask to help with future state architecture that doesn't support that syntax -- look to use cast() instead

Hosuke:
So the current [pre-commit hooks failure](https://github.com/duneanalytics/spellbook/actions/runs/3529870875/jobs/5921277446) can be ignored, as it is caused by [the looping variable here](https://github.com/duneanalytics/spellbook/pull/2048/files#diff-c2e3a1254d9870d88999b1ade5563948c8579e41307dc9d4a03a7449f3570422R17).

Hosuke:
> hi @Hosuke & @0xroll -- thanks for the work so far. all looks good, up to the test:
> 
> ```
> with unit_test as (
>     select
>         case when test.oracle_price = actual.oracle_price then true else false end                                          AS oracle_price_test,
>         case when test.proxy_address = actual.proxy_address then true else false end                                        AS proxy_address_test,
>         case when test.aggregator_address = actual.aggregator_address then true else false end                              AS aggregator_address_test,
>         case when test.underlying_token_address = actual.underlying_token_address then true else false end                  AS underlying_token_address_test
>     from       test_schema.git_3b028b6f_chainlink_polygon_price_feeds actual
>     INNER JOIN test_schema.git_3b028b6f_chainlink_polygon_get_price test 
>     ON (actual.blockchain = test.blockchain 
>     AND actual.block_time = test.block_time
>     AND actual.feed_name = test.feed_name)
> 
> )
> 
> select * from unit_test
> -- where (oracle_price_test = false                OR 
> --       proxy_address_test   = false             OR
> --       aggregator_address_test = false          OR 
> --       underlying_token_address_test = false
> --       )
> ```
> 
> looks like it may be returning a false positive, where the CTE produces no results at all. when adding seeds & tests, please do either write the query to catch this false positive or run `dbt compile` and run the query on dune to validate, like the above
> 
> can you tweak the test to return as expected?

Also @0xroll has changed the joining condition to `actual.block_number = test.block_number` and solved the issue.
The sample query is [here](https://dune.com/queries/1647497?d=1).

jeff-dude (approved):

PR #2057, opened 2022-11-22
[CoW Protocol] start counting dex_swaps from dex.trades

bh2smith:
Previously we were using NULL for our "single order solvers" in the `dex_swaps` field. This now populates the field with a count from dex.trades.

This has been tested out for the last week of batches here in this PoC Query: https://dune.com/queries/1290518

It is not yet entirely accurate since not all projects are part of `dex.trades` yet, however it is MUCH better than null.

Test Plan:

Existing unit test for this table should suffice.


jeff-dude (approved):
sounds good, thanks for adapting tests to ensure 🤝

PR #2058, opened 2022-11-23
Changes workflow so it fails immediately on first error

andre-dune:
This prevents a user from not getting feedback from errors if the workflow timesout. It also prevents a long running workflow that will inevitably have to be run again.


a-monteiro:
Oops, test-user for the workflow action 😅

PR #2059, opened 2022-11-23
Report errors from DBT on Github action timeout

a-monteiro:
This prevents a user from not getting feedback from errors if the workflow timesout. It also prevents a long running workflow that will inevitably have to be run again.

Link T-21390

### DIFF
--- a/.github/workflows/dbt_slim_ci.yml
+++ b/.github/workflows/dbt_slim_ci.yml
@@ -53,7 +53,7 @@ jobs:
         run: "dbt seed --select state:modified --state ."
 
       - name: dbt run initial model(s)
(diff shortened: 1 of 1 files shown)


jkylling (approved):
`-x` means fail fast?

jeff-dude (approved):
interesting thought, these are errors which dbt compile doesn't catch? was it tested locally and/or through gh action with any model (edited to fail on purpose)?

a-monteiro:
> interesting thought, these are errors which dbt compile doesn't catch? was it tested locally and/or through gh action with any model (edited to fail on purpose)?

It seems so. This was flagged previously, See the height task for more details.

PR #2070, opened 2022-11-24
[CoW Protocol] Fix V1 Dex Swaps

bh2smith:
Similar to the subtle change made for v2 here #2057, the matching by solver names is made more correct (since they have changed) and are currently inaccurate.

If possible, this would require rebuild of this table from probably ~5 months back.


masquot (approved):
Thank you @bh2smith. 👍

Table rebuilt going forward from 1 June 2022

PR #2076, opened 2022-11-24
[Easy] Update cow_protocol_labels to include all dexes instead of just cow_protocol trades

gentrexha:
We want to extend our labels to other dex-es and have updated our labels query to label all dex trade tx_hashes, not just our own trades.


models/cow_protocol/tx_hash_labels/stable_to_stable/ethereum/cow_protocol_tx_hash_labels_stable_to_stable_ethereum.sql
```diff
@@ -7,11 +7,22 @@
...
-    from {{ ref('cow_protocol_ethereum_trades') }}
-    where buy_token_address in (select contract_address from {{ ref('tokens_ethereum_erc20_stablecoins') }})
-    and sell_token_address in (select contract_address from {{ ref('tokens_ethereum_erc20_stablecoins') }})
+        tx_hash
+    from (
+        select distinct tx_hash
+        from {{ ref('dex_aggregator_trades') }}
+        where blockchain = 'ethereum'
+        and token_bought_address in (select contract_address from {{ ref('tokens_ethereum_erc20_stablecoins') }})
+        and token_sold_address in (select contract_address from {{ ref('tokens_ethereum_erc20_stablecoins') }})
```
bh2smith, line 16: You might be able to do something more efficient here (I am not entirely sure if the performance is better, but intuitively it might be). 

Something like this:
```sql
with stable_coins as (
    select collect_list(contract_address) as stables
    from {{ ref('tokens_ethereum_erc20_stablecoins') }}
)
```

Then later, 

```sql
and array_contains(stables , from)
and array_contains(stables, to)
```

Although its a bit weird because you can't actually set some variable to be the list collected and it needs to be "selected" out of there... maybe this won't work unless there is something to join on...

Anyway, the idea was to collect the list only once (since its fixed) instead of multiple times.

Maybe it doesn't matter because its such a short list.

bh2smith (approved):
Looks good! I mentioned one suggestion, but not sure how to make it work properly without defining a constant.

It seems, however, that this entire file no longer belongs in the "cow_protocol" namespace.

jeff-dude:
assuming WIP due to feedback comments, please do let me know if you'd like to proceed

models/cow_protocol/tx_hash_labels/stable_to_stable/ethereum/cow_protocol_tx_hash_labels_stable_to_stable_ethereum.sql
bh2smith: Unfortinately, I am not sure how to isolate the set in a way that makes this more efficient, so instead...

[this fork](https://dune.com/queries/1745906) shows what can be done (with repitition)

gentrexha:
@jeff-dude This should be ready to merge now.

models/cow_protocol/tx_hash_labels/stable_to_stable/ethereum/cow_protocol_tx_hash_labels_stable_to_stable_ethereum.sql
```diff
@@ -7,11 +7,22 @@
...
+        UNION
```
jeff-dude, line 17: can we leverage `union all` here? it's universally beneficial on the query engine, if we can ensure the all doesn't introduce duplicates (which it shouldn't from what i can assume)

jeff-dude (approved):

PR #2078, opened 2022-11-24
Add GMX GLP on Avalanche

theachenyj:
Created models /gmx/avalanche_c/glp , and I added 2 spells in this folder.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


models/gmx/avalanche_c/glp/gmx_avalanche_c_glp_schema.yml
```diff
@@ -0,0 +1,196 @@
...
+        description: "Net asset under management for the WETH.e pool"
+        tests:
+          - not_null
+      - &wbtc_e_aum
+        name: wbtc_e_aum
+        description: "Net asset under management for the WBTC.e pool"
+        tests:
+          - not_null
+      - &btc_b_aum
+        name: wbtc_aum
```
Hosuke, line 193: Hi @theachenyj, you may have a typo here, do you mean `btc_b_aum`.

theachenyj: Nice catch! I'd committed a fix. Thx.

Hosuke:
Hi @theachenyj , since these are some new models, could you possibly make a seed file with a few records like [this](https://github.com/duneanalytics/spellbook/blob/main/seeds/jarvis_network/polygon/jarvis_network_polygon_view_transactions_seed.csv) for the `glp_components` and the `glp_aum` model?

theachenyj:
> Hi @theachenyj , since these are some new models, could you possibly make a seed file with a few records like [this](https://github.com/duneanalytics/spellbook/blob/main/seeds/jarvis_network/polygon/jarvis_network_polygon_view_transactions_seed.csv) for the `glp_components` and the `glp_aum` model?

Since the data is a aggregated result, it's not easy to find exact comparison. Besides the result would  be slightly different with different calculation method. I think I can get some seed datas from [official web site](https://stats.gmx.io/avalanche) , but in a different granularity, and compare them like this: if the difference between them within a certain value, like 0.1%, I'd say the result is correct. Do you think this way works out?

Hosuke:
> > Hi @theachenyj , since these are some new models, could you possibly make a seed file with a few records like [this](https://github.com/duneanalytics/spellbook/blob/main/seeds/jarvis_network/polygon/jarvis_network_polygon_view_transactions_seed.csv) for the `glp_components` and the `glp_aum` model?
> 
> Since the data is a aggregated result, it's not easy to find exact comparison. Besides the result would be slightly different with different calculation method. I think I can get some seed datas from [official web site](https://stats.gmx.io/avalanche) , but in a different granularity, and compare them like this: if the difference between them within a certain value, like 0.1%, I'd say the result is correct. Do you think this way works out?

Yes, for comparison between double typed values we can accept a range, and usually we use discrete/string typed value like addresses in the seed for test. So do not worry the seed record need be exactly matched to the query result.

Hosuke:
Currently the [test for glp_aum fails](https://github.com/duneanalytics/spellbook/actions/runs/3599401637/jobs/6063005139#step:12:93), we can remove the test for doubled value first, or accept a larger range.

theachenyj:
> Currently the [test for glp_aum fails](https://github.com/duneanalytics/spellbook/actions/runs/3599401637/jobs/6063005139#step:12:93), we can remove the test for doubled value first, or accept a larger range.

I fix some 

> Currently the [test for glp_aum fails](https://github.com/duneanalytics/spellbook/actions/runs/3599401637/jobs/6063005139#step:12:93), we can remove the test for doubled value first, or accept a larger range.

I'm stuck in the test. It's weird. I've run the test SQL on Dune and it has passed. I have no idea which part has problem.

seeds/gmx/avalanche_c/gmx_glp_aum_seed.csv
```diff
@@ -0,0 +1,6 @@
+date,usdc_aum,usdc_e_aum
+2022-09-15,20359495.092885,18949462.097249
+2022-09-16,20327050.875974,18440021.732571
```
Hosuke, line 3: Hi @theachenyj, please have a look on this line.
Moreover this is your [test query result](https://dune.com/queries/1707843?d=1)
image
It seems there is a null value for this record when joining.

theachenyj: Finally! I've made a mistake in aum cte. Thx a lot.

Hosuke (approved):

This PR contains 2 new models:
- `gmx_avalanche_c_glp_components`
- `gmx_avalanche_c_glp_aum`

Both models have been modified to incremental views.
Since the seed checking columns are double-typed, the generic test is not applied to these models.

jeff-dude:
thank you both. please do note final commits:
- remove redundant unique test
- add seed to project file, for schema naming standards mostly

jeff-dude (approved):

PR #2081, opened 2022-11-24
Change one file to table

catherine-2:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


catherine-2:
@jeff-dude This one file is slow to query from (the rest are all running very fast). I looked at the file you sent + some others and read through the DBT docs and think this format may be what I want but let me know if it is off. 

I want to make sure (1) the latest data is always available (not sure how often tables re-run - is it a set cadence and/or triggered by queries?) and (2) it overrides any existing data with new data. The table should always have 10,000 rows - one for each punk id, and columns for the current owner and last time of transfer

catherine-2:
> Hey hey @catherine-2 :) If you want the table to be refreshed regularly (every hour) I suggest to use `incremental` as a strategy (not `table`). To do this, you need to define parameters in the jinja config part of the model (`unique_key`, `partition`, `file_format`, etc...). You can take uniswap_ethereum_v3_trades as an example: https://github.com/duneanalytics/spellbook/blob/main/models/uniswap/ethereum/uniswap_v3_ethereum_trades.sql
> 
> You also need to add incremental filters with the following format: `{% if is_incremental() %} WHERE block_time >= date_trunc("day", now() - interval '1 week') {% endif %}`
> 
> If you have questions or need me to help you make some of these changes lmk 👍

Thank you @soispoke !

Would be super helpful to get your advice on how to best implement for this specific table.

I initially started with an incremental type but switched to a table since I want to make sure it would replace everything in the table the way the query is written. The ideal output is 1 row for each punk_id with details on the current owner.

Would it make sense for me to instead change this to two separate files?

1. An incremental model that updates every hour with all punk transfers (i.e. what is currently in the "b" section of the query with the incremental filter added)
2. A second file that is a view based off the incremental table above that then does the final aggregation (i.e. take the most recent entry for each punk_id)

soispoke:
Hey @catherine-2 looking back at it I was wondering if you considered using the existing `cryptopunks_ethereum_trades` table ?https://github.com/duneanalytics/spellbook/blob/main/models/cryptopunks/ethereum/cryptopunks_ethereum_trades.sql

catherine-2:
@soispoke Just compared the two files

It wouldn't currently work since the trades table just has events from the CryptoPunks marketplace (no other transfers). I could rewrite the underlying events table (linked below) to include all transfers - though it may be a little complicated with the static 10,000 rows we need to base this off of to get the current holders. If we went that route I think we would need to separate the static table out as a new file to reference and add all erc20 transfers to the events table where the tx_hash is not in the sale queries already.

Let me know if you have a strong preference either way! 

https://github.com/duneanalytics/spellbook/blob/main/models/cryptopunks/ethereum/cryptopunks_ethereum_events.sql

catherine-2:
@soispoke @jeff-dude Wanted to bump this one for review and see if the new solution I built works to turn the underlying data into a table and make querying on current_owners_per_token faster on Dune? Thank you!

models/cryptopunks/ethereum/cryptopunks_ethereum_punk_transfers.sql
```diff
@@ -0,0 +1,10048 @@
+{{ config(
+        alias ='current_owners_per_token',
+        partition_by = ['evt_block_time'],
```
jeff-dude, line 3: i wouldn't partition by a timestamp, that is much too granular. consider creating a new column for day/week/month, then partition by that. it depends on data volume per each time segment. if not many tx's happen, i'd consider monthly. if it's heavy volume, maybe daily

jeff-dude (changes requested):

models/cryptopunks/ethereum/cryptopunks_ethereum_punk_transfers.sql
```diff
@@ -0,0 +1,10048 @@
...
+        , ('0x23170777466de6046ff42bd5abcf525077e3ce27',4166)
+        , ('0xf138eb67f008da11daba9bcffc7a21f31b8b2b14',4167)
+        , ('0x53ede7cae3eb6a7d11429fe589c0278c9acbe21a',4168)
+        , ('0xc272ed09c636edb6e858fa8702c182406185f1af',4169)
+        , ('0xf1ad946ae30a285d96ea396acd3fd22c5aa75ef8',4170)
+        , ('0x53ede7cae3eb6a7d11429fe589c0278c9acbe21a',4171)
+        , ('0x53ede7cae3eb6a7d11429fe589c0278c9acbe21a',4172)
+        , ('0xd4fa6e82c77716fa1ef7f5defc5fd6eeefbd3bff',4173)
+        , ('0x53ede7cae3eb6a7d11429fe589c0278c9acbe21a',4174)
+        , ('0
```
jeff-dude, line 10022: consider casting null as datatype, or it'll be a weird 'void' datatype that can cause issues

catherine-2: Sorry which null should I cast as datatypes? Just double checked every address and id in this cte is filled in so there won't be any null values

models/cryptopunks/ethereum/cryptopunks_ethereum_punk_transfers.sql
catherine-2: Got it just updated! Thank you

models/cryptopunks/ethereum/cryptopunks_ethereum_punk_transfers.sql
jeff-dude: sorry, not the cte, but this line: ```select  NULL as from``` -- when you force null into a column, databricks infers an odd datatype. i'd just cast it like: ```cast(NULL as varchar(5))``` or something similar

catherine-2: Ah got it - just fixed! Thank you!

jeff-dude:
the incremental model here should have the ```unique_key``` property, which i just removed from the model converted to view (not needd there, but unique test in schema file is fine).

can you add this in config block on which columns make punk_transfers unique? these will be the join conditions on the hourly merge statement. then in addition, add a unique_combo_of_columns test in the schema file on this model, which contain the same keys from config block, example here:
https://github.com/duneanalytics/spellbook/blob/main/models/uniswap/ethereum/uniswap_ethereum_schema.yml#L14

catherine-2:
Thank you @jeff-dude ! Just added the unique test

This PR is now failing because of an error in seaport_ethereum_assert_samples but I haven't touched that file so not sure why

jeff-dude:
looks like the unique keys are still off. when incremental runs/loads, the unique test fails 10k times (assuming every punk_id being repeated). i tried your initial set of keys, then tried another set, but both failed. please do continue to find the unique combo of keys

catherine-2:
@jeff-dude I just figured it out! The way I originally wrote the spell it was adding the cte twice. I added an incremental filter to that part of the query and it is now passing the checks. 

Going to make 1 other update I think before we merge -- double checking a few things now. Will let you know once ready to be merged on my end.

catherine-2:
@jeff-dude Think this is now ready for merge! Thank you again for all of your help on this.

jeff-dude (approved):

PR #2084, opened 2022-11-25
T-26026 - Change sushiswap trades model to be Trino compatible

belen-pruvost:


belen-pruvost:
I checked the union vs union all count and it's matching 😌

soispoke:
this one looks good ✅

soispoke (approved):

PR #2087, opened 2022-11-25
T-26026 - Change curvefi trades model to be Trino compatible

belen-pruvost:


soispoke:
pre commit failing for the wrong reason here (for loop in curve model), otherwise everything looks good

soispoke (approved):

PR #2138, opened 2022-11-28
OpenSea V1 [Wyvern] rewrite and fixes

0xRobin:

+ changes pricing logic to use OrderMatched event
- removed dependency on traces and erc20 parsing for fee calculations
- removed excluded txs
- removed dependency on calldata substring parsing
- Split out trades into 1 entry per token_id (we distribute the price and fees based on total nft amounts in the trade, it's the best we can do as an estimation)

**For Dune Engine V2**

I've checked that:

### General checks:


jeff-dude (approved):
with end date filter applied, does this actually need run incrementally? i wonder if we should look into a way to stop it from running, if no new data comes through?

0xRobin:
We could just wrap all the logic in a not-incrememtal filter and have it return nothing quickly on an incremental run?

I thought about making it a non-incremental table but then it would fully rerun every time, right?

0xRobin:
@jeff-dude 
It's a bit 'hacky' but get's the job done. 
Incremental now runs on 4sec instead of 2mins.
```
{% if is_incremental() %}
SELECT * from {{this}} WHERE false;
{% else %}
```

Let's merge :ship: :rocket:  (I don't have permissions)

jeff-dude:
hmm i added unique test on unique keys, looks like we have dupes

0xRobin:
> hmm i added unique test on unique keys, looks like we have dupes

Good to check! 
Fortunately no real duplicates, just forgot the `token_id` in the `unique_trade_id` field.

jeff-dude:
> @jeff-dude from the GH CI it looks like the model is considered a view and not materialized. I think we're missing a `materialized = table` config?

good call. sometimes we override in project file, but let's just include here in model. pushing commit now, should be final

PR #2142, opened 2022-11-29
prices.usd_forward_fill solution for lagging price feeds

0xRobin:

Price feeds are lagging quite a lot, which means usd_amount is unfinalized for the last ~7h and even completely null for the last +2 
See viz here for current status: https://dune.com/queries/1645384

This PR introduces the `prices.usd_forward_fill` table that can be used to forward fill the latest known price.
This gives you a good temporary estimate and when the prices later catch up the incremental run will overwrite your results with the exact prices. 

Includes a test to see we introduce no duplicate or null prices. 

**For Dune Engine V2**

I've checked that:

### General checks:


soispoke:
@0xRobin Hey, this is great! I was just looking into this after comments from @bernatfp about usd prices being null for the last couple of hours in nft.trades 🔥

0xRobin:
@soispoke cool, yeah it had been bothering me seeing queries misrepresent USD volume amounts of the last hours and it also messed with ERC balance queries. 

Just checked the test results and there are 5 duplicates in the source usd.prices table. So they also show up in this test.
I'd like to keep the test in place so configured it to only error on > 5 duplicates.

soispoke:
@0xRobin yeah that makes sense, we could also just have a warning for duplicates if they are expected

I guess you'll open a few more PRs so spells are using `prices.usd_forward_fill` instead of `prices.usd`?

0xRobin:
@soispoke we don't really want duplicates in the prices table cause that will result in unwanted duplicate records if it's ever inner joined instead of left joined onto a table.

And yes I'll open up a different PR to apply this table in relevant spells.

soispoke:
yeah agree we don't want any but this table is relatively difficult to fix/look at compared to other spells, tagging @aalan3 here for visibility

soispoke (approved):

aalan3:
>Price feeds are lagging quite a lot, which means usd_amount is unfinalized for the last ~7h and even completely null for the last +2

This is a bug somewhere. Prices should not be more than around ~10min behind in the usual case. I'll look into this.

PR #2160, opened 2022-11-30
Use prices.usd_forward_fill in all dexes of dex.trades

0xRobin:
Applying solution of https://github.com/duneanalytics/spellbook/pull/2142 to dex.trades


soispoke:
hey @0xRobin so apparently @aalan3 seems to think prices.usd should be up to date for (+10 minutes), let's wait for this one until there is a fix/solution on the prices.usd table itself?

0xRobin:
@soispoke ok no problem discarding this fix if the base table is fixed!
As of now it still has +1h of delay on all tokens

aalan3:
Yes please wait, this shouldn't be needed. Problem was a bit more messy than I thought so might take a couple more days to fix.

0xRobin:
closing this now, can always re-open should the problems with prices persist.

PR #2164, opened 2022-11-30
[DO NOT MERGE] T-21220 - Make token_id int across nft.fees related views

belen-pruvost:


0xRobin:
I believe token_id should be strings since we have token_ids which are to large to fit in a numeric spark datatype

belen-pruvost:
Closing for now

PR #2172, opened 2022-11-30
nft aggregated models code cleanup

0xRobin:
Applying some Jinja magic to clean up aggregated nft views.

- code deduplication and cleanup. :broom: 
- made nft.transfers a view (transfers were materialized twice)
- rewrote linked_addresses


jeff-dude:
before i look closely, wanted to comment on your description -- we actually have ```nft.transfers``` excluded in prod right now. about a week ago or so, it started to take an abnormally high amount of time to run, and prod was all thrown off waiting on it each run. we haven't had a chance to dig into the why yet.

0xRobin:
> before i look closely, wanted to comment on your description -- we actually have `nft.transfers` excluded in prod right now. about a week ago or so, it started to take an abnormally high amount of time to run, and prod was all thrown off waiting on it each run. we haven't had a chance to dig into the why yet.

Good to know, I did not really look in depth at it.
Just noticed both the chain specific definitions and the overall nft.transfers were materialized.

jeff-dude (approved):
looks good 🤝 
it looks like there isn't overlap of files with #2289 and #2290 -- i wonder if we should batch all three of these to merge at same time, then have prod refresh all nft sector at the same time

PR #2174, opened 2022-11-30
Create dex.prices table

henrystats:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


Hosuke (approved):
This model is a good/creative approach to find out prices for missing tokens in prices.usd table.

I have reviewed this PR and here is the summary:

Since the model is a derived model from dex.trades, and with double-typed values,
currently there is no seed or test added for this model.

However @henrystats you can still add a seed file as we can check the `sample_size`(int typed) within certain `hour` range.

[Sample query results](https://dune.com/queries/1719360?d=1).

MSilb7:
Thank you, we've needed this spell!

I wonder if you all think there should be a higher sample size on # trades within an hour for us to re-set the price (i.e. 3 - 5). I've had cases doing this in the past where illiquid tokens prices can do all kinds of weird stuff if we're not careful.

henrystats:
Hola ser 👋 - I agree that we should have a higher sample trade size. I had originally thought there would be no need to do that since the sample size is a column and anyone using the table could determine what the sample size they want to use would be. 

But I'll update that to 5 now. Also I love the >20% change suggestion and will be happy to update in the coming weeks.

jeff-dude (approved):

PR #2182, opened 2022-12-01
fix element trades

jeff-dude:
cc @belen-pruvost @dot2dotseurat @soispoke 

here is a general overview:
- move any `cast()` statement to the base child models which are incremental delta tables, i.e. `element_ethereum.events` or `element_bnb.events` -- this will keep things organized and views simple, should look to apply this to all other protocols feeding `nft.trades/events/fees/mints/burns`
- reformat views to make git / dbt recognize they need rebuilt, since child models are changing datatypes -- until we build a solution to run downstream dependencies, we need to trigger them somehow
- `element_bnb.events` was accidently casting `block_number` as decimal, while all other protocols feeding nft trades has this field as default based on source decoded tables (bigint) -- this caused downstream failures
- reformat `nft.trades` to use our simple formatting from `dex.trades`


models/nft/nft_trades.sql
```diff
@@ -6,366 +6,57 @@
...
+        amount_original,
+        amount_raw, --team-query cast as decimal: CAST(amount_raw AS DECIMAL(38,0)) AS amount_raw,
+        currency_symbol,
+        currency_contract,
+        nft_contract_address,
+        project_contract_address,
+        aggregator_name,
+        aggregator_address,
+        tx_hash,
+        block_number, --**not sure where** team-query cast as decimal: CAST(number_of_items AS DECIMAL(38,0)) AS number_of_items,
```
belen-pruvost, line 52: Not sure if these comments should stay

belen-pruvost (approved):
This looks good, thank you!

soispoke (approved):
lgtm, passing all local tests 🤞

models/nft/nft_trades.sql
jeff-dude: we should remove those comments in a future PR, that was just my guide and forgotten 🥲

PR #2185, opened 2022-12-01
Cast opensea.burns only only on base child models

belen-pruvost:


belen-pruvost:
While the CI runs, I'm also running `dbt run --select +opensea_burns` locally

belen-pruvost:
Closing this  to have a fresh start

PR #2202, opened 2022-12-02
[WIP] Model run tracking package

dot2dotseurat:
@jeff-dude @soispoke @0xRobin 

This isn't ready to merge but these changes add the [elementary package](https://docs.elementary-data.com/introduction) to our project. This should give us a lot more visibility into our runs over time because it handles organizing the dbt artifacts for us. 

I'll need to work through how to set this up on the GH runner and on DBT Cloud before merging but I was able to generate some local results after configuring the profile.yml (@0xRobin that will require local credentials unfortunately).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 manifest.json
 macros/dune/create_views_of_dependencies.sql
 macros/dune/alter_table_locations.sql
+elementary*
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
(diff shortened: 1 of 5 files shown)


dot2dotseurat:
Looks like this breaks our dependency management job as is because it's adhoc adding models that don't actually exist into our project. Not sure how to fix this yet.

dot2dotseurat:
I *think* this is solved by another upcoming change in deployment so I will close this PR

PR #2216, opened 2022-12-05
T-20626 - Hide balances_ethereum + balancer_ethereum + nomad_ethereum + keep3r_network_ethereum + ens_view_expirations + dex_aggregator_trades + seaport_ethereum.view_transactions views on Trino

belen-pruvost:


jeff-dude:
not sure why the action is failing where it is -- can you try to merge main back into this branch, see if it gets past that step?

belen-pruvost:
This is failing because it times out. However it's changing a new property on the views, and not changing anything on the models per se. Would it be possible to ship this?

belen-pruvost:
I'll close this PR as discussed and open a new one just with the macro change.

jkylling:
CI is likely timing out since we are possibly changing every spell which exists. Let's fix this by making a special macro which is used only for hiding certain spells. E.g.,

```
{% macro expose_spells_hide_trino(blockchains, spell_type, spell_name, contributors, trino_enabled=True) %}
{%- if target.name == 'prod'-%}
        ALTER {{"view" if model.config.materialized == "view" else "table"}} {{ this }}
        SET TBLPROPERTIES (
        'dune.public'='true',
         "'dune.disabled_datasets'='11',"
        ...
```

and do

```
expose_spells_hide_trino(...)
```

instead of

```
expose_spells(trino_enabled=False)
```

PR #2228, opened 2022-12-06
[CoW Protocol] Expose both trade tables in schema

bh2smith:
Based on an old issue raised in #1641 - we lost our trades table from the schema due to clashing names. 

This PR renames the dex.trades version of the table to `dex_trades` (or at least hopefully) so that both tables can be exposed in the schema on the dune interface. 

Tagging @fleupold - since he is currently trying to query this data and also noticed the missing table. Also tagging @gentrexha since he may want to know about this.


jeff-dude:
@bh2smith thinking out loud here, let me know what you think.

rather than rename model and have both spells in the hierarchy appear in explorer, any reason you can't add the expose spell macro to config on both, with same inputs (essentially as-is now) and just have the parent one show?
`cow_protocol_ethereum.trades` -> `cow_protocol.trades` -- both have same data, as parent reads all from child model
with both having the config, the child runs first, sets in explorer, then parent runs and overwrites it. then only the parent shows (but same data, should be fine to only show the parent).

in future, if adding additional blockchains into the parent model union, then you can edit config to include new blockchain, which will then create two entries into dune explorer (one with just ethereum logo for child, one with both for parent).

if the child model has additional columns that you don't use for parent model, and the parent only exists to feed into `dex.trades` and you don't use it, then we can likely disregard this thought. or could just remove the expose spell macro from the parent, so only the intended child one shows. that way the model names can remain.

bh2smith:
Yes @jeff-dude your suggestion is exactly what I wanted to have. However I don't know how to do it. Please feel free to show me.

jeff-dude:
please note that this is (and never will be) an issue with the tables themselves. this is simply how they display in dune explorer UI. you can always query both tables(`cow_protocol_ethereum.trades` and `cow_protocol.trades`), even if one doesn't show in explorer.

jeff-dude:
but now that you are adding gnosis chain data, i think we can simplify this. let me see what i can push to share.

bh2smith:
> please note that this is (and never will be) an issue with the tables themselves. this is simply how they display in dune explorer UI. you can always query both tables(cow_protocol_ethereum.trades and cow_protocol.trades), even if one doesn't show in explorer.

Yes, be we want to see it in the explorer.

bh2smith:
> but now that you are adding gnosis chain data, i think we can simplify this. let me see what i can push to share.

YES PLEASE!

jeff-dude:
oops -- not sure why all those commits just pushed :(

jeff-dude:
need to step away for a few, but we'll let that run & revisit shortly. if working as intended, i can explain (if not obvious)

jeff-dude (approved):

jeff-dude:
our gh action doesn't run the expose_spell macro, as we don't want test tables exposed, so not great way to test it via PRs. i went ahead and merged, and we can validate/cleanup as needed.

the idea is:
- `cow_protocol_ethereum.trades` has it's own "unique" metadata now passed into the macro, with project `cow_protocol` and blockchain `ethereum`
- same concept for gnosis
- the parent model, now that it is cross-chain, has the metadata as both blockchains, therefore the macro should expose it separately

ideal output:
three tables displayed, all named trades, one with eth logo, one with gnosis logo, one with both logos.

to reiterate on the issue above, since the child and parent had same metadata passed in (i.e. parent only had ethereum still), since the child model runs first and sets the property, the parent then runs after succession on the child, and it's alter table properties runs the same info and overrides it. i guess that's technically a bug, when a parent cross-chain model doesn't have multiple blockchains, therefore shares the metadata with the child. i've seen other models follow this pattern, just leave out that config on one or the other, since the data is ultimately the same. something for us to consider for sure.

models/cow_protocol/cow_protocol_trades.sql
```diff
@@ -32,11 +32,35 @@ FROM
...
-            evt_index,
-            'cow_protocol' ||'-'|| tx_hash ||'-'|| order_uid || CAST(evt_index AS VARCHAR(100)) AS unique_trade_id
+            evt_index
         FROM {{ ref('cow_protocol_ethereum_trades') }}
-        /*
-        UNION
-        <add future chains here>
-        */
+
+        UNION ALL
```
bh2smith, line 38: Hey! Thanks for adding this!

models/dex/dex_trades.sql
```diff
@@ -3,7 +3,7 @@
         post_hook='{{ expose_spells(\'["ethereum", "bnb", "avalanche_c", "gnosis", "optimism", "arbitrum"]\',
                                 "sector",
                                 "dex",
-                                \'["jeff-dude", "hosuke", "0xRob", "pandajackson42", "Henrystats", "scoffie", "zhongyiio", "justabi", "umer_h_adil", "mtitus6", "dbustos20", "tian7"]\') }}'
+                                \'["jeff-dude", "hosuke", "0xRob", "pandajackson42", "Henrystats", "scoffie", "zhongyiio", "justabi", "umer_h_adil", "mtitus6", "dbustos20", "tian7", "bh2smith"]\') }}'
```
bh2smith, line 6: Nice touch, thanks!

models/dex/dex_trades.sql
```diff
@@ -44,6 +44,7 @@ list of models using old generic test, due to multiple versions in one model:
 ,'wombat_bnb_trades'
 ,'iziswap_bnb_trades'
 ,'babyswap_bnb_trades'
+,'cow_protocol_trades'
```
bh2smith, line 47: Is this supposed to be here? I thought we were only part of aggregator trades...

bh2smith (commented):
I am beyond happy with this solution as long as all three tables appear in the schema. Thanks for your additions here @jeff-dude

bh2smith:
Hey - I am looking at the interface and still don't see the ethereum or gnosis chain tables? Is there some sort of indexing period we need to wait for? If they are not going to show up, I would rather rename this other trades table.

models/dex/dex_trades.sql
jeff-dude: ah, right, let me remove it

models/cow_protocol/cow_protocol_trades.sql
```diff
@@ -32,11 +32,35 @@ FROM
             trader AS tx_from,
             receiver AS tx_to,
             '' AS trace_address,
-            evt_index,
-            'cow_protocol' ||'-'|| tx_hash ||'-'|| order_uid || CAST(evt_index AS VARCHAR(100)) AS unique_trade_id
```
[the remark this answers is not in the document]
jeff-dude: nope, we rely on `unique_keys` in config block + unique combo of columns test applied in schema file(s). dbt will merge incremental data based on unique keys, then the test will run after each run to ensure unique.

PR #2234, opened 2022-12-06
Dune version of hooks

dot2dotseurat:
We reverted this [change](https://github.com/duneanalytics/spellbook/commit/3749a4f80753739f0e84405eeeeaf3748fae7526) about two weeks ago because we couldn't use main as the version. I've created a [new release](https://github.com/duneanalytics/pre-commit-dbt) on the  pre-commit hooks fork and now we should be able to point to it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 default_stages: [push, manual]
 exclude: ^deprecated-dune-v1-abstractions/
 repos:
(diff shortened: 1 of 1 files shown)


jeff-dude (approved):
ohhh, excited to try this again, thanks for revisiting

Hosuke:
I think the release version link to some zip/compressed files available on github file server, so the hidden script may extract the pointed zip file and get the certain/specific checking script based on the known/hard-coded file structure.

PR #2240, opened 2022-12-06
T 22081 automatic price checks

dot2dotseurat:
Github Action to automate checks against coinpaprika

### DIFF
--- a/.github/workflows/prices_check.yml
+++ b/.github/workflows/prices_check.yml
@@ -23,10 +23,14 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v2
 
-      - name: print-diff
+      # This step finds only new lines in prices_tokens.sql.
+      # New lines are expected to be added to the values list in the query
+      # Other edits to the prices query like changing the SELECT statement or config will cause
+      # this step to fail.
+      - name: Get prices diff
         shell: bash
         run: |
-          git fetch origin main
-          git diff origin/main..HEAD > mydiff
-          echo "mydiff = "
-          cat mydiff
+          echo "PRICES_DIFF=$(git diff origin/main..HEAD models/prices/prices_tokens.sql | grep '^\+ ')" >> $GITHUB_ENV
+
+      - name: Test Variable
+        run: echo "$PRICES_DIFF"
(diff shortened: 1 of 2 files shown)


dot2dotseurat:
Screen Shot 2022-12-07 at 5 59 55 PM
 This is what a failure looks like if the token id is not valid according to coinpaprika

dot2dotseurat:
@jeff-dude @Hosuke here's my WIP PR for price token checks. 

Like other Github Action checks, this is hard to test because I need to modify the files in the PR itself to trigger the action. I'm going to try tomorrow to see if there is a way for me to pull shas from old PRs to do some more extensive testing. I will also configure the pytest tests to run from the PR; I'm currently just running them locally. 

In the meantime, I wanted to give you a rundown on what this check *should* do:
- it should return an exception and fail is the token_id is not valid according to the to CoinPaprika API
- it should return an error if the Symbol does not match; this is currently case-sensitive. 
- it should return an error if the contract is not listed; this is currently NOT case-sensitive. this will probably trigger a lot since many examples I tested had this error. I don't think the contracts are necessarily wrong, just not in the API response.
- it does NOT check decimals because I never saw them listed in any of the API checks I ran.
-  it checks that the blockchain is the "same" as the platform. I'm trying to map the blockchain platform relationship in TokenChecker under self.chain_mapper but this isn't complete because I didn't find examples for arbitrum or avalanche. 

Anyway, still some more work but I wanted to give ya'll an update. And thank you @Hosuke I used your price checker script a lot when writing this.

Hosuke:
The coinpaprika API may limit the free requests from certain IP. The last testing script is not fully stable due to this issue.

dot2dotseurat:
> The coinpaprika API may limit the free requests from certain IP. The last testing script is not fully stable due to this issue.

We can use a self hosted runner with an API token. But it looks like 25,000 calls/month are free at 10 per second. I expect we'll be under 25k and we can add some sleeps in between calls to get below 10 per second.

jeff-dude (approved):
awesome! look forward to seeing it in action. i guess i could test off this branch, but you've done so pretty extensively. we can merge & then monitor it in action against incoming PRs

PR #2245, opened 2022-12-07
[Prices] Add Popular Tokens to Gnosis Chain

bh2smith:
[This query](https://dune.com/queries/1719783) shows tokens without prices along with their popularity by trade count.

I have selected a few very clear prices that should be mirrored (and noticed that they already have been on other networks).


bh2smith:
I think there is something wrong with your CI... It seems to think there is no schema for prices and I definitely didn't remove it.

soispoke:
@dot2dotseurat tagging for visibility, the CI is not working properly

PR #2247, opened 2022-12-07
T-21220 - Make opensea.events trino compatible

belen-pruvost:

PR #2248, opened 2022-12-07
T-21220 - Make nft.burns trino compatible

belen-pruvost:


belen-pruvost:
I've run all the models locally (now using s3 instead of dbfs for the base incremental tables) and then queried nft tables on databricks. They still work.

However, for this change to fully work on Trino we also need to have https://github.com/duneanalytics/spellbook/pull/2247 merged

dot2dotseurat (approved):

PR #2258, opened 2022-12-07
remove cow protocol from dex trades

jeff-dude:
cow protocol is in dex_aggregator, can remove here


bh2smith (approved):

bh2smith:
> cow protocol is in dex_aggregator

So far still the only one - but I think 0x is on it.

jeff-dude:
> > cow protocol is in dex_aggregator
> 
> So far still the only one - but I think 0x is on it.

we had some internal work on it as well with 1inch data, however we had some resource shortage, so it's delayed currently

PR #2266, opened 2022-12-08
T-21622 - New macro for hiding unsupported spells in Trino

belen-pruvost:


macros/expose_spells_hide_trino.sql
```diff
@@ -0,0 +1,15 @@
+{% macro expose_spells_hide_trino(blockchains, spell_type, spell_name, contributors, trino_enabled=True) %}
```
dot2dotseurat, line 1: Does it make sense to add trino_enabled=True to the existing expose_spells macro? Or am I misunderstanding something here?

dot2dotseurat: Sorry catching up and just seeing this https://github.com/duneanalytics/spellbook/pull/2216#issuecomment-1342384332

My preference is removing the time limit from the CI for the PR and modifying the existing macro instead of adding a new one. But it is what it is and I don't want to make you redo it.  

We can clean this up at a later time when we have more bandwidth.

dot2dotseurat (approved):

PR #2273, opened 2022-12-08
T-21220 - Make nft.fees Trino compatible

belen-pruvost:


belen-pruvost:
The test that's failing here (`dbt test --select sudoswap_ethereum_events`) is also failing on main branch, so I don't think it relates to my changes

belen-pruvost:
I built the models locally and I'm able to query them on databricks (`dbt_belen_nft.fees`, `dbt_belen_sudoswap_ethereum.events`, etc). I'm still moving the incremental base tables from dbfs to s3 to be able to query them on Trino.

jeff-dude (approved):
i think i'm comfortable to proceed on this one, even with failed tests. the one is known, as you mentioned, and the other two introduced are simply cause those models aren't in the PR. the unique test passes on the model edited.
i think we should wait until your other two PRs are ready related to nft, then do simultaneously.

belen-pruvost:
Sounds good! I'll ping you on those PRs as soon as I finish testing them (almost there for nft.events)

PR #2277, opened 2022-12-08
T-21220 - Fix aztec views to be Trino compatible

belen-pruvost:


belen-pruvost:
I've run the models locally and everything is fine (also queried this in databricks + trino after building)

soispoke (approved):

PR #2284, opened 2022-12-11
Fix for OS v1 fees

0xRobin:
Fixes for OpenSea v1 (Wyvern):
- Where taker/maker fee is sourced from depending on order type (Buy/Sell)
- Doesn't apply 2.5% OS fee on private sales. (this resulted in negative fees)
- Doesn't apply 2.5% OS fee when total fees are less then 2.5% (temp fix)
- Adds fee information to the seed file.

- [x] I've manually checked the example trades
- [x] I've manually checked there are no more negative fees.


jeff-dude (approved):
lgtm 🤝

PR #2289, opened 2022-12-12
fix x2y2 platform_fee_percentage & royalty_fee_percentage

pandajackson42:
Brief comments on the purpose of your changes:

- platform_fee_percentage & royalty_fee_percentage shall be percentage

PR #2290, opened 2022-12-12
Make nft.fees, nft.events and nft.mints Trino compatible

belen-pruvost:


belen-pruvost:
I'm still building these models locally. As expected, it takes forever.

A side note: if we merge this in one piece, we may want to treat the magiceden_solana incremental update separately as per today's learning 😄 (cc @jeff-dude @soispoke @dot2dotseurat)

belen-pruvost:
I've built all these models locally, and I'm able to query them under my schema on both databricks and trino.

belen-pruvost:
- `amount_raw` has always been casted to decimal on the trino compatibility transformation, because so has `number_of_items`
- `platform_fee_amount_raw` and `royalty_fee_amount_raw` have always been casted to double as part of this process, to follow what we have on other fee fields without rounding them

To choose the precision (and hence the type) I searched each field from the base table to the upper level view that I try to fix, and also search on the repo for ` as <field_name>` to see if there were other explicit castings, in order to keep them consistent

0xRobin:
Ok, @augustog we should probably change the raw fee values to decimal to be in line with amount_raw? (but not in this PR)

jeff-dude:
> Ok, @augustog we should probably change the raw fee values to decimal to be in line with amount_raw? (but not in this PR)

this is a good point. it may make sense to document each of the nft sector objects (`nft.trades`, `nft.events`, etc) in terms of intended schema and their respective data points. that'll make it clear for us all to see. is that something @0xRobin  is comfortable with doing and @augustog supporting?

jeff-dude (approved):
this PR will have prod working hard, i'd recommend we merge early morning EST (or you guys can european time, but would suggest doing it with plenty of time to monitor/support).

jeff-dude:
> this PR will have prod working hard, i'd recommend we merge early morning EST (or you guys can european time, but would suggest doing it with plenty of time to monitor/support).

we should also merge #2289 at the same time, as it's another nft events base child model

PR #2300, opened 2022-12-13
[CoW Protocol] Fix Receiver Field

bh2smith:
The smart contract sets the trade receiver as follows:

```
if receiver = ZERO_ADDRESS --> use order owner
```

There are only a handful of these types of orders in the history of the protocol, but enough, that this was causing an issue (not noticed until recently). 

This fixes the receiver field so that it acts like the contract:

https://github.com/cowprotocol/contracts/blob/ff6fb7cad7787b8d43a6468809cacb799601a10e/src/contracts/libraries/GPv2Order.sol#L112-L127

This change requires a complete rebuild of the trades table in both Dune V1 and V2.


jeff-dude (approved):
v1 running still, will merge when done

PR #2302, opened 2022-12-13
[CoW Protocol] Add parsed AppData Spell

bh2smith:
We have recently introduced a community source table `cowswap.raw_app_data` which is unstructured in the sense that the `data` field in this table is a struct that need to be "unpacked". This table, along with this [PoC Query](https://dune.com/queries/1751965) demonstrates how that unpacking works and exposes it as a "view". 

cc @dsalv - to ensure this is created in a way that will be updated whenever the reference table changes.

cc @josojo and @gentrexha for internal review.

PR #2303, opened 2022-12-13
[CoW Protocol] Add Parsed Order Rewards

bh2smith:
Similar to #2302, we create a view which parses the unstructured content from `cowswap.raw_order_rewards` in to a table with unpacked `data` field.

PoC Query - [here](https://dune.com/queries/1752782)

cc @josojo and @gentrexha for internal review.

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
@@ -1,0 +1,13 @@
+{{ config(alias='order_rewards',
+        post_hook='{{ expose_spells(\'["ethereum"]\',
+                                    "project",
+                                    "cow_protocol",
+                                    \'["bh2smith"]\') }}'
+)}}
+
+select
+    tx_hash,
+    data.amount as cow_reward,
+    cast(data.surplus_fee as double) as surplus_fee,
+    order_uid
+from cowswap.raw_order_rewards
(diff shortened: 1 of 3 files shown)


models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
```diff
@@ -0,0 +1,14 @@
...
+                                    \'["bh2smith"]\') }}'
+)}}
+
+-- PoC Query here - https://dune.com/queries/1752782
+select
+    tx_hash,
+    data.amount as cow_reward,
+    cast(data.surplus_fee as double) as surplus_fee,
+    order_uid
+from cowswap.raw_order_rewards
```
bh2smith, line 14: This is not using a ref call but CI does not complain... Not sure yet how to reference community sources...

bh2smith: According to this:

https://github.com/duneanalytics/spellbook/blob/ad2b91fef757f12a885fd2b71df293319e047f68/models/labels/flashbots/labels_flashbots_ethereum.sql#L15-L15

I should use source(`cowswap`, 'raw_order_rewards')

bh2smith: Fixed

jeff-dude (approved):
this is the right approach after reading in community data elsewhere 👍

PR #2308, opened 2022-12-13
Update the prices_tokens model to include the Xen Crypto tokens

nibty:

Add the Xen Crypto token for Ethereum, Avalanche, and BNB
https://coinpaprika.com/coin/xen-xen-crypto/

**For Dune Engine V2**

I've checked that:

### Pricing checks:
* [X] `coin_id` represents the ID of the coin on coinpaprika.com
* [X] all the coins are active on coinpaprika.com (please remove inactive ones)

### DIFF
--- a/models/prices/prices_tokens.sql
+++ b/models/prices/prices_tokens.sql
@@ -146,6 +146,7 @@ VALUES
     ("ptp-platypus-finance","avalanche_c","PTP","0x22d4002028f537599be9f666d1c4fa138522f9c8",18),
     ("grape-grape-finance","avalanche_c","GRAPE","0x5541d83efad1f281571b343977648b75d95cdac2",18),
     ("tusd-trueusd","avalanche_c","TUSD","0x1c20e891bab6b1727d14da358fae2984ed9b59eb",18),
+    ("xen-xen-crypto", "avalanche_c", "XEN", "0xC0C5AA69Dbe4d6DDdfBc89c0957686ec60F24389", 18),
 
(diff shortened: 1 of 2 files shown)


jeff-dude:
for reference, block explorer results:
https://snowtrace.io/token/0xc0c5aa69dbe4d6dddfbc89c0957686ec60f24389
https://bscscan.com/token/0x2ab0e9e4ee70fff1fb9d67031e44f6410170d00e
https://etherscan.io/token/0x06450dee7fd2fb8e39061434babcfc05599a6fb8

jeff-dude:
@dot2dotseurat another use case to consider in the prices test. i see all three addresses returned in the API, commented block explorers for reference too. looks like the results shows platform as the coinpaprika id value, rather than blockchain name. the test fails on incorrect blockchain name, but the PR here is actually right. tagging for fyi, but this could be merged

dot2dotseurat:
> @dot2dotseurat another use case to consider in the prices test. i see all three addresses returned in the API, commented block explorers for reference too. looks like the results shows platform as the coinpaprika id value, rather than blockchain name. the test fails on incorrect blockchain name, but the PR here is actually right. tagging for fyi, but this could be merged

This is happening because we don't have the chain mapping for avalanche in the script yet. We coinpaprika returns avax-avalanche whereas we are mapped to avalanche_c. I believe if we just add that mapping then this will pass.

dot2dotseurat (approved):

PR #2312, opened 2022-12-14
[CoW Protocol] Easy Add Solver Address to Reward Table 

bh2smith:
Didn't realize this column would be so useful to keep around. Updated the [PoC Query](https://dune.com/queries/1752782)


soispoke (approved):
lgtm + passed local testing ✅

PR #2313, opened 2022-12-14
Fix Source Chain ID/Name Hop Optimism

soispoke:

@MSilb7 Here is a fix for source chain ID/name for the Optimism hop protocol models. I think what happened is the source chain ID was wrongfully set to NULL when the source chain ID was 1 (Ethereum Mainnet). There are probably more elegant ways to solve this issue, lmk what you think!


soispoke:
Fixed the null source chain name/id issue

jeff-dude (approved):

PR #2316, opened 2022-12-14
[Easy] [Labels] Implement dex trader label

gentrexha:
Adding a dex trader label for addresses that have completed a trade on a dex.

Tagging @bh2smith for internal review

**For Dune Engine V2**

I've checked that:

### General checks:


gentrexha:
> Looks good to me, just wondering if we also consider aggregator traders as dex traders or will have a separate label for that.

We were thinking of having a separate label for that.

soispoke:
@gentrexha  thanks a lot for contributing to labels it's awesome! Would you consider having models similar to what was done for NFT traders (Top 1,5,10% DEX traders based on either volume or transactions), and then a model for dex "users" 
=> https://github.com/duneanalytics/spellbook/tree/main/models/labels/nft 

It would make it more granular/consistent with the existing NFT trades labels

gentrexha:
> @gentrexha thanks a lot for contributing to labels it's awesome! Would you consider having models similar to what was done for NFT traders (Top 1,5,10% DEX traders based on either volume or transactions), and then a model for dex "users" => https://github.com/duneanalytics/spellbook/tree/main/models/labels/nft
> 
> It would make it more granular/consistent with the existing NFT trades labels

That's a great idea. I think it's a great idea to add labels dynamically like that. 

Here are some discussion points regarding that:

1. One issue I have with that is that you don't see the actual value of the top 1% (e.g.: that the top 1 %'s average trade value is $50k). Is it possible to add the actual value to the label text?
2. Would you include all these new labels into this newly dex "users" models folder?

soispoke:
That's my personal opinion, but
1. I'd rather see `Top 1% DEX trader`, and the number of DEXs he traded on + his average trade value (two labels categories you already added) than the average traded value for all TOP 1% DEX traders?
2. In the example, DEX "users" is actually a single model, just explicitly printing the names of DEXs a particular address has traded on: https://github.com/duneanalytics/spellbook/blob/main/models/labels/nft/labels_nft_users_platforms.sql

gentrexha:
1. Would those all be separate labels? If so, it makes sense to me to have one for the top 1% percentiles and one with the average trade value. The only remaining issue I have is that if you want to categorize users, it's hard to observe how much average trade value they need to belong in each category, i.e., If I'm visualizing volume by average trade value percentiles for a project, where does the 1% average trade value user start becoming 1%?
2. I see. That makes more sense. If I were to add this new umbrella folder for this, should it be named dex or dex_aggregator labels? I would want to include both.

gentrexha:
Starting analyzing some of these labels, I think having a label printing the names of DEXs a particular address has traded on is also helpful.

Also, open to having this PR put everything into one folder, similar to NFTs.

soispoke:
@gentrexha Hi, what's your current opinion about this PR? Do you want to proceed with this?

gentrexha:
> @gentrexha Hi, what's your current opinion about this PR? Do you want to proceed with this?

Yeah, I think that makes sense.

PR #2319, opened 2022-12-14
[Easy] [Labels] Implement trader frequency labels

gentrexha:

Adding a trader frequency label for addresses that have completed at least two trades on a dex or dex aggregator.

**For Dune Engine V2**

I've checked that:

### General checks:


models/labels/trader_frequencies/ethereum/trader_frequencies_ethereum.sql
```diff
@@ -0,0 +1,40 @@
...
+    config(
+        alias='trader_frequencies_ethereum',
+    )
+}}
+
+with
+ trader_frequencies as (
+    select
+        taker as address,
+        count(tx_hash) / datediff(max(block_time), min(block_time)) as trades_per_day
```
bh2smith, line 11: Can we use block date on these? I remember that these columns were required for partitioning tables and it seems like a reasonable place to use it

models/labels/trader_frequencies/ethereum/trader_frequencies_ethereum.sql
```diff
@@ -0,0 +1,40 @@
...
+    having datediff(max(block_time), min(block_time)) > 0
+ )
+
+select
+  array("ethereum") as blockchain,
+  address,
+  case
+    when trades_per_day >= 1 then 'Daily Trader'
+    when trades_per_day >= 0.142857142857 then 'Weekly Trader'
+    else 'Monthly Trader'
```
bh2smith, line 32: What about putting a hard requirement on monthly and make the last category "infrequent" or "sparse"

bh2smith (approved):
Looks good to me. Just a couple of ideas. I think you may be over clarifying monthly by making it the final else. I would expect that many wallets who do not trade anywhere near monthly would wind up here.

models/labels/trader_frequencies/ethereum/trader_frequencies_ethereum.sql
gentrexha: I think that's a good idea! I added two new labels: `Yearly Trader` and `Sparse Trader`.

models/labels/trader_frequencies/ethereum/trader_frequencies_ethereum.sql
gentrexha: Wow! There's an enormous performance increase in my query. Thanks, Ben!

models/labels/trader_frequencies/ethereum/trader_frequencies_ethereum.sql
bh2smith: Nice! Note I'll will likely wind up in the early category instead of sparse!

bh2smith: Do you have some wallets in mind to test each category?

gentrexha: > Do you have some wallets in mind to test each category is working as expected?

Good point! I'll create a list of wallets and expected labels to test these.

dot2dotseurat (approved):

PR #2322, opened 2022-12-14
[CoW Protocol] Fix AppData <--> OrderUid Mapping

bh2smith:
We have recently discovered a bug in the OrderUid <--> AppData Mapping constructed as a combination of call data and event data for our trades table (it was known that this was a "dirty hack" of a query before and we finally came back to fix it).

This can be tested by examining the [PoC Sub Query](https://dune.com/queries/1759039?d=1) for [this Ethereum Transaction](https://etherscan.io/tx/0x828bad1003f0bb44cdf8a0eabe65dac57e81592c58f952a36d3f96e26a1cee0b#eventlog) 

Namely, that the [Trade Event Logs](https://etherscan.io/tx/0x828bad1003f0bb44cdf8a0eabe65dac57e81592c58f952a36d3f96e26a1cee0b#eventlog) align with the Input Data:

Screenshot 2022-12-14 at 20 49 13

The subqeury link above shows only the mapping for a given tx hash (with default as the one shown here). It is also linked directly in the SQL file update here.

We also update the [PoC Trade Query](https://dune.com/queries/1759305?start_ts_d83555=2022-05-13+00%3A00%3A00&end_ts_d83555=2022-05-15+00%3A00%3A00) (in the code as well because it had not been updated in a while).

cc @gentrexha and @josojo  for internal review.

# Additional Note

This bug would require a complete table rebuild.

### DIFF
--- a/.github/workflows/dbt_slim_ci.yml
+++ b/.github/workflows/dbt_slim_ci.yml
@@ -56,7 +56,7 @@ jobs:
         run: "dbt seed --select state:modified --state ."
 
       - name: dbt run initial model(s)
-        run: "dbt -x run --select state:modified --state ."
+        run: "dbt -x run --select state:modified --state . --full-refresh"
 
       - name: dbt test initial model(s)
         run: "dbt test --select state:new state:modified --state ."
--- a/.github/workflows/dbt_slim_ci.yml
+++ b/.github/workflows/dbt_slim_ci.yml
@@ -56,7 +56,7 @@ jobs:
         run: "dbt seed --select state:modified --state ."
 
       - name: dbt run initial model(s)
-        run: "dbt -x run --select state:modified --state . --full-refresh"
+        run: "dbt -x run --select state:modified --state ."
 
       - name: dbt test initial model(s)
         run: "dbt test --select state:new state:modified --state ."
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
@@ -13,7 +13,7 @@
     )
 }}
 
--- Find the PoC Query here: https://dune.com/queries/1283229
+-- Find the PoC Query here: https://dune.com/queries/1759305
 WITH
 -- First subquery joins buy and sell token prices from prices.usd
 -- Also deducts fee from sell amount
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
@@ -22,7 +22,7 @@ trades_with_prices AS (
            evt_block_time            as block_time,
            evt_tx_hash               as tx_hash,
            evt_index,
-           settlement.contract_address          as project_contract_address,
+           trade.contract_address          as project_contract_address,
            owner                     as trader,
            orderUid                  as order_uid,
            sellToken                 as sell_token,
@@ -32,7 +32,7 @@ trades_with_prices AS (
            feeAmount                 as fee_amount,
            ps.price                  as sell_price,
            pb.price                  as buy_price
-    FROM {{ source('gnosis_protocol_v2_ethereum', 'GPv2Settlement_evt_Trade') }} settlement
+    FROM {{ source('gnosis_protocol_v2_ethereum', 'GPv2Settlement_evt_Trade') }} trade
              LEFT OUTER JOIN {{ source('prices', 'usd') }} as ps
                              ON sellToken = ps.contract_address
                                  AND ps.minute = date_trunc('minute', evt_block_time)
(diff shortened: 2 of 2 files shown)


bh2smith:
Note that this change is failing CI on another unrelated table:

```
   Runtime Error
    [CANNOT_UP_CAST_DATATYPE] Cannot up cast spark_catalog.magiceden_solana.events.royalty_fee_amount_raw from "DOUBLE" to "BIGINT".
    The type path of the target object is:
    
    You can either add an explicit cast to the input data or choose a higher precision type of the field in the target object
```

jeff-dude:
> Note that this change is failing CI on another unrelated table:
> 
> ```
>    Runtime Error
>     [CANNOT_UP_CAST_DATATYPE] Cannot up cast spark_catalog.magiceden_solana.events.royalty_fee_amount_raw from "DOUBLE" to "BIGINT".
>     The type path of the target object is:
>     
>     You can either add an explicit cast to the input data or choose a higher precision type of the field in the target object
> ```

i manually canceled the ci job, with the hope that prod succeeding now clears up the queue. please try to merge main back into your branch, ideally the action runs smoother now

bh2smith:
I rebased instead. I hope that's not a problem!

bh2smith:
Still failing...

bh2smith:
@jeff-dude - so we are failing now on this uniqueness check:

```
22:13:35  3 of 4 FAIL 176 dbt_utils_unique_combination_of_columns_cow_protocol_ethereum_trades_tx_hash__order_uid__evt_index  [FAIL 176 in 11.25s]
```


Could it be that the collision between the old version of this table and the new on is the cause here? If the table were rebuilt from scratch, would it be possible that it does not fail this check? I am afraid the table might be too big to check all of this by taking counts in the interface, but I can give it a try...

bh2smith:
Ok, I believe I have a pretty good idea of why this is failing. The reason is that one of the columns `app_data` had a bug and was corrected here. The test for uniqueness is probably detecting a "new" row (with different `app_data`) and trys to insert, but fails because of conflicting unique identifier.

dot2dotseurat:
@bh2smith I assume the `here` above was meant to have a link attached to it?

bh2smith:
> @bh2smith I assume the `here` above was meant to have a link attached to it?

"Here" meaning, the changed made in the PR.

I think that this will not fail if the table is rebuilt from scratch.

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -97,57 +97,54 @@ trades_with_token_units as (
...
-        get_json_object(trades.col, '$.sellAmount') as limit_sell_amount,
-        get_json_object(trades.col, '$.buyAmount') as limit_buy_amount,
-        get_json_object(trades.col, '$.validTo') as valid_to,
-        get_json_object(trades.col, '$.flags') as flags
-    from reduced_order_ids order_ids
-             join trade_data trades
-                  on evt_tx_hash = call_tx_hash
-                      and order_ids.pos = trades.pos
+        uid,
+        get_json_object(trade, '$.appData') as app_data,
```
bh2smith, line 138: The primary purpose of this PR is to change the way app_data is constructed. This "fix" here is why I believe the incremental strategy is failing the uniqueness condition.

dot2dotseurat:
Ok, one second. Let me push a commit we will remove before merging.

dot2dotseurat:
@bh2smith we are still seeing the same test fail with a full refresh. I'll try digging into the table created by the CI to see if I can spot anything. test_schema.git_eca85085_cow_protocol_ethereum_trades

dot2dotseurat:
@bh2smith I see many rows with 82 duplicates with this query https://dune.com/queries/1764000?d=1

bh2smith:
> @bh2smith I see many rows with 82 duplicates with this query https://dune.com/queries/1764000?d=1

Oh no! This is bad... Very sorry about this, I will figure out whats going on here and get it fixed... seems to be duplicated across the board! thanks for looking into it. Will dig more.

bh2smith:
Found the issue. There is one "partially fillable" order (with the same ID) that has been matched 82 times (i.e. across 82 different transactions). Still need to sort out the appropriate fix.

dot2dotseurat:
awesome, glad you found it! I'll be afk for a bit but I'll check in when I'm back

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -97,57 +97,54 @@ trades_with_token_units as (
...
-        get_json_object(trades.col, '$.receiver') as receiver,
-        get_json_object(trades.col, '$.sellAmount') as limit_sell_amount,
-        get_json_object(trades.col, '$.buyAmount') as limit_buy_amount,
-        get_json_object(trades.col, '$.validTo') as valid_to,
-        get_json_object(trades.col, '$.flags') as flags
-    from reduced_order_ids order_ids
-             join trade_data trades
-                  on evt_tx_hash = call_tx_hash
-                      and order_ids.pos = trades.pos
+        uid,
```
bh2smith, line 137:
```suggestion
        distinct uid,
```

.github/workflows/dbt_slim_ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
         run: "dbt seed --select state:modified --state ."
 
       - name: dbt run initial model(s)
-        run: "dbt -x run --select state:modified --state ."
+        run: "dbt -x run --select state:modified --state . --full-refresh"
```
bh2smith, line 59: I suppose this can be removed here? I will make the suggested edit, and we can remove it once we see a passing CI.
```suggestion
        run: "dbt -x run --select state:modified --state ."
```

bh2smith:
Yay - Its passing, what a relief!

dot2dotseurat (approved):

PR #2328, opened 2022-12-15
[Easy] [Labels] Implement trader_portfolios label

gentrexha:

Adding a trader portfolio value label for addresses that have completed at least one trade on a dex or dex aggregator. Thresholds were based on percentiles gotten from here: https://dune.com/queries/1758351

cc @bh2smith 

**For Dune Engine V2**

I've checked that:

### General checks:


models/labels/trader_portfolios/ethereum/trader_portfolios_ethereum.sql
```diff
@@ -0,0 +1,45 @@
...
+        alias='trader_portfolios_ethereum',
+    )
+}}
+
+with
+ trader_portfolios as (
+    select
+        sum(amount_usd) as portfolio_value_usd,
+        wallet_address as address
+    from {{ ref('balances_ethereum_erc20_day') }}
```
bh2smith, line 12: I am a bit concerned about using this table as a foundation because of the fact that I fear there are untamed bugs here. For example in the latest balances table our settlement has 8 negative balances

https://dune.com/queries/1761899

Of course it is only natural to depend on this, but the underlying table will need to be debugged eventually.

bh2smith (approved):
So this is fine as long as you are aware how inaccurate this will be and prepared to tackle addressing these issues.

models/labels/trader_portfolios/ethereum/trader_portfolios_ethereum.sql
bh2smith: Note that numerical overflow is a very likely candidate for these bugs... but it may not be the only problem.

Screenshot 2022-12-15 at 11 23 26

gentrexha: I agree with your concerns and realize the flaws of the table. I'll keep them in mind when doing any analysis using this label. I've added an `amount > 0` condition to ensure I don't include these errors in the label calculation.

bh2smith: That is a good start, but just because we exclude negatives does not mean that we are excluding the bad accounting. There could be positive amounts that are still incorrect due to this very same bug...

models/labels/trader_portfolios/ethereum/trader_portfolios_ethereum.sql
```diff
@@ -0,0 +1,45 @@
...
+    group by wallet_address
+    -- For some reason there's negative portfolio values in here
+    having portfolio_value_usd > 0
+ )
+
+select
+  array("ethereum") as blockchain,
+  address,
+  case
+    when portfolio_value_usd > 90000 then '>$90k'
```
soispoke, line 32: Similar comment (than other PRs): Suggestions => Add `DEX portfolio value: '>$90k'`

dot2dotseurat (approved):

soispoke:
Same here, this wasn't supposed to be merged either, @gentrexha could you open a PR with a fix, including `DEX` in the label's name like: '>$90k DEX portfolio value'' instead of'>$90k portfolio value'🙏?

gentrexha:
> Same here, this wasn't supposed to be merged either, @gentrexha could you open a PR with a fix, including `DEX` in the label's name like: '>$90k DEX portfolio value'' instead of'>$90k portfolio value'pray?

I'm not sure I agree with this one as the portfolio value is independent of their DEX trades. What do you think? I've only limited the selection to which I apply the label.

soispoke:
oh right, but then shouldn't ERC721 and ERC1155 be included as well? I feel like this label is one that needs a bit more thought

gentrexha:
> oh right, but then shouldn't ERC721 and ERC1155 be included as well? I feel like this label is one that needs a bit more thought

That's right. I'm also missing out on ETH value too at the moment. Therefore, it needs more discussion.

PR #2329, opened 2022-12-15
[Easy] [Labels] Implement trader_dex_diversity label

gentrexha:

Adding a trader dex diversity label for addresses that have completed at least one trade on a dex or dex aggregator and counting the number of different projects they've used in the past 30 days.

**For Dune Engine V2**

I've checked that:

### General checks:


bh2smith (approved):
Cool label idea. So is the label an integer? under the category "dex diversity"?

gentrexha:
> Cool label idea. So is the label an integer? Under the category "dex diversity"?

Thanks! The label is an integer from 1 to N, where N is the total number of projects in the dex and dex aggregator tables.

models/labels/trader_dex_diversity/ethereum/trader_dex_diversity_ethereum.sql
```diff
@@ -0,0 +1,34 @@
...
+        from {{ ref('dex_trades') }}
+        where blockchain = 'ethereum' and block_time > (current_date - interval '30 days')
+    )
+    group by taker
+ )
+
+select
+  array("ethereum") as blockchain,
+  address,
+  dex_diversity as name,
```
soispoke, line 27: Maybe add: Number of DEXs traded on: `dex_diversity` ?

gentrexha: Let me know if the latest update to the name is what you had in mind.

soispoke: Yes looks good! Now I'm just thinking: do you think this info will be a bit redundant with the `dex_user` model in which we print the actual names of DEXs the address has traded on? It's not a big deal we could have both, just wanted to have your opinion on this

gentrexha: IMHO: Having just the number is better for visualizations as it indicates that a user uses multiple DEXs without mentioning which. Whereas, for having just the names, there are various combinations for users who use more than two DEXs and, therefore, hard to group by.

gentrexha:
Thanks @soispoke !

PR #2339, opened 2022-12-16
Add bend dao spells  - Ready for Review

henrystats:
Hola 👋 - I'd like to suggest that we create a spell under the NFTs category that would be about lending & borrowing of nfts. 

Some ideas around name would be, nft.loans or nft.finance or nft.lending etc 

I've added bend_dao's repay & borrowing events and looking forward to feedback on how I can improve this and then I can go ahead and create the new category and also add other projects.


dot2dotseurat:
@henrystats if you are looking for feedback on how to improve this before creating a new category, my first suggestion is always going to add a custom test with a seed to check a few sample rows as a unit test. I think that's the easiest and most extendable way to guarantee quality. 

Other than that, this is stellar and I am hyped on `nft.lending`.

henrystats:
Thank you very much @dot2dotseurat, I've now included a seed/test file. This is ready for review now.

I haven't created the nft.lending table yet though. Just looking to get feedback on the column names and the likes and if there would be some changes to that.

henrystats:
> @henrystats I think these columns make sense but I know @augustog has also been thinking about the lending sector so he might have some ideas as well.

okaiee! Thanks very much, I've now created the nft.lending file as well and looking forward and excited to this getting reviewed.

ccing @augustog again incase you want to check this out ser!

jeff-dude:
@henrystats thank you for the great submission -- i can see how many new learnings/enhancements from spellbook you've applied, love to see it. i pushed a few minor commits, but also have some larger scale questions since this could be foundational for lending/borrowing:
- [minor] since decoded tables use just `bend` in the name -- should we follow suit here and drop the `_dao`? i'm honestly good either way, just thinking out loud. if we do change, all directories/files/contents within need edited to reflect
- i think having two spells in the hierarchy structure don't necessarily make sense here. with `nft.trades`, the events break down into multiple categories with views to represent each for subset of data, but here you only have two categories and pull both into the parent view. i'd say we scrap the second model and just rename the base model to lending, if possible (unless there are future use cases you are preparing for that i'm unaware of)
- in `nft.trades` we're looking to get away from `unique_trade_id` column and just use combo of columns for keys/seed joins, like in `dex.trades`. i'd prefer we start this sector spell that way too. please do search for minimum number of columns to use (ideally blockchain/project/version/tx_hash/evt_index -- but whatever you need)
- another random thought: can we reorg the columns to make more logical sense? maybe move the primary key columns towards the beginning, metadata columns in the middle, amount columns after, etc etc -- this would apply to both base model and sector-wide model

henrystats:
Thank you very much ser @jeff-dude!

1. I'd suggest we keep it as bend_dao if you're good with either of the two. I use bend_dao because that's the name on the docs site, website etc etc. 

2. I agree, I wanted to follow a similar model to nft.events because I believe there are some lending projects that would have marketplaces as well that would be included in nft.events & nft.trades. But since there are some column name changes & new columns, I agree with having it in just one model `.lending`. 

3. Okaiee ser, will update!

4. okaieee ser, will update this as well. 

I'll push an update for these shortly.

jeff-dude (approved):
awesome, thanks again!

PR #2344, opened 2022-12-17
Fix incremental update of curve.fi on ethereum

augustog:
Fixing a smol typo on the curve.fi model on ethereum that caused it to not be updated during the incremental run.


soispoke (approved):
lgtm! will probably merge on its own

PR #2260, opened 2022-12-08
Add ellipsis_finance bnb to dex trades

henrystats:
Brief comments on the purpose of your changes:

**For Dune Engine V2**

I've checked that:

### General checks:
* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each sql file is a select statement and has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [ ] `coin_id` represents the ID of the coin on coinpaprika.com
* [ ] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:
* [ ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:
* [ ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ ] where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #2345, opened 2022-12-17
OP Curve - Fix Metapool Decimals

MSilb7:

Query checking the fix: https://dune.com/queries/1772295

Metapools on Curve seem to just use the curve pool's decimals (18) on the buy side, even if the token bought has < 18 decimals.

This should drop and replace all curve trades & downstream tables (i.e dex.trades)

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:

### DIFF
--- a/models/curvefi/optimism/curvefi_optimism_trades.sql
+++ b/models/curvefi/optimism/curvefi_optimism_trades.sql
@@ -143,14 +143,16 @@ SELECT DISTINCT
         when lower(COALESCE(erc20a.symbol,p_bought.symbol)) > lower(COALESCE(erc20b.symbol,p_sold.symbol)) then concat(COALESCE(erc20b.symbol,p_sold.symbol), '-', COALESCE(erc20a.symbol,p_bought.symbol))
         else concat(COALESCE(erc20a.symbol,p_bought.symbol), '-', COALESCE(erc20b.symbol,p_sold.symbol))
     end as token_pair,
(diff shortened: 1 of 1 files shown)


henrystats:
@augustog I'm actually working on ellipsis (a fork of curve) and just completed a model today that I believe would work for curve contracts. 

I Also think we can use a similar approach for Curve Optimism, Avalanche & Ethereum. 

This is my open pr - https://github.com/duneanalytics/spellbook/pull/2260

One thing though, all the trades are decoded into one table, Stableswap (Metapool, Plainpool & Cryptopool). 

So it's easier to actually use this same model for other curve chains since we'd only need to change the base pool addresses. I also added a check pools seed test to test that it works correctly. 

@MSilb7 if you could take a look ser, lemme know if this approach makes sense and I'd be happy to help update the Curve trades table for all the chains.

MSilb7:
> Do you know why the approach for Curve on OP is different than for other chains? Asking to see if we need to update other models.

Other chains were manually adding pool addresses I believe. I was trying to take an approach here that could catch any created pool. Will look at @henrystats' approach

augustog:
Oh this is an awesome solution!! Yes, we've had complaints about the fact that we need to add pools manually, so whichever approach works I'll suggest we move all chains to it.

dot2dotseurat:
I'm going to merge because I want to resolve data quality issue but pls I will be so happy if you follow up with a PR with a couple seed values on the test.

dot2dotseurat (approved):

PR #2351, opened 2022-12-19
[CoW - HotFix] Deduplicate Order Rewards

bh2smith:
Turns out the entire rewards table has been duplicated... This query prevents that from making its way into our queries (although we will continue to debug the root cause as well).

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
@@ -7,9 +7,9 @@
 
 -- PoC Query here - https://dune.com/queries/1752782
 select
+    distinct order_uid,
     tx_hash,
-    solver as solver_address,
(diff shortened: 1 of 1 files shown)


dsalv (approved):

0xRobin:
CC: @soispoke probably the cause of the cow_trades error we saw this morning on the production refresh.
@bh2smith Seems like the root cause was resolved though? Both distinct and non distinct queries return 59k records now.

bh2smith:
Yea we had dropped the table and it and the view didn't exist for a while. All good now!

0xRobin:
> Yea we had dropped the table and it and the view didn't exist for a while. All good now!

Awesome! This PR can be closed then probably?

bh2smith:
I haven't yet determined the root cause of the duplication so I'd like to have this in here to ensure the issue doesn't leak out to the view. Would really like to slaws this merged if only temporarily.

PR #2353, opened 2022-12-19
rename element base sources

0xRobin:
Rename in `model/element` and `tests/element` due to changes in the decoding.
Only changing ethereum sources (others were not renamed)

ERC721OrdersFeature - > OrdersFeature
ERC1155OrdersFeature -> OrdersFeature


soispoke (approved):
lgtm

PR #2356, opened 2022-12-19
[CoW] Update trader ETH Flow orders

bh2smith:
We forgot to include the sender column in the `eth_flow` orders table. This PR changes the trader column to the sender of the ETH flow orderCreation. Otherwise the trader is shown "incorrectly" as the ETH Flow contract itself.

We also change the `sell_token` to ETH instead of what looks like WETH.

Check out the [POC query](https://dune.com/queries/1779431) on 

```sh
order_uid=0xb7e1f2599dd2268cd98f9fff203ec4d94ac9f5f0be86a2986bf9d6816e78e89f40a50cf069e992aa4536211b23f286ef88752187ffffffff
```

or any order_uid ending with `ffffffff`

 and validate that the `sender` address is NOT either of 
 
 ```
'0xd02de8da0b71e1b59489794f423fabba2adc4d93' 
'0x40a50cf069e992aa4536211b23f286ef88752187'
```

cc @josojo and @gentrexha for internal review.

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_eth_flow_orders.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_eth_flow_orders.sql
@@ -48,18 +48,4 @@ eth_flow_orders as (
     {% endif %}
 )
 
-select
-  block_date,
-  block_time,
-  block_number,
-  tx_hash,
-  valid_to,
-  quote_id,
-  sell_amount,
-  fee,
-  buy_amount,
-  buy_token,
-  receiver,
-  app_hash,
-  order_uid
-from eth_flow_orders
+select * from eth_flow_orders
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
@@ -13,7 +13,7 @@
     )
 }}
 
--- Find the PoC Query here: https://dune.com/queries/1759305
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: models/cow_protocol/ethereum/cow_protocol_ethereum_eth_flow_orders.sql [hunk 1]
```diff
@@ -48,18 +48,4 @@ eth_flow_orders as (
...
-  quote_id,
-  sell_amount,
-  fee,
-  buy_amount,
-  buy_token,
-  receiver,
-  app_hash,
-  order_uid
-from eth_flow_orders
+select * from eth_flow_orders
```
bh2smith, line 51: This is why I accidentally forgot to include sender in the previous version... `select *` will not leave sender behind.

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -147,16 +147,33 @@ uid_to_app_id as (
     where i = j
 ),
 
+eth_flow_senders as (
```
bh2smith, line 150: Since we only need sender and order_id, i think its more practical for us to simply construct the two column table directly here. The addition of sender column above is merely so that we aren't missing a crucial part of the ETH flow data in our view.

josojo (approved):
Great

dot2dotseurat (approved):

PR #2358, opened 2022-12-19
freshness check

dot2dotseurat:

This test checks the "staleness" the day level for dex.trades. [Inspo](https://dune.com/queries/1772161?d=1) from @augustog. 

Key things to check here:
- The original query showed that ~6 of the project/blockchain pairs had no trades in the last day. Which we initially interpreted as staleness on our part. 
- I grabbed the underlying decoded trades table from their sources (or at least what appears to be) for those pairs. In the `sources` CTE, I do a similar calculation as we do from `delays` (age_of_last_record_days). 
- The test fails when either age_of_last_record_days is not zero or if the diff between the sources and delays value age_of_last_record_days is not zero. 

I will do some sleuthing to see if these projects really haven't seen action in the last day or so. If they have we need to escalate to decode.

### DIFF
--- a/.github/workflows/dbt_precommit_hooks.yml
+++ b/.github/workflows/dbt_precommit_hooks.yml
@@ -31,6 +31,7 @@ jobs:
         with:
           format: space-delimited
           token: ${{ secrets.GITHUB_TOKEN }}
+        continue-on-error: true
       - name: Set modfied and added files as arg
         run: |
           echo "FILES=$(echo ${{ steps.abc.outputs.added_modified }})" >> $GITHUB_ENV
--- a/tests/dex/dex_trades_assert_freshness.sql
+++ b/tests/dex/dex_trades_assert_freshness.sql
@@ -1,3 +1,45 @@
+-- Add sources for decoded projects where trades may not happen daily
+-- Project, Blockchain, Table Schema, Table Name, Time Column
+{% set trade_sources = [
+    {'project': 'fraxswap',
+     'blockchain': 'avalanche_c',
+     'schema': 'fraxswap_avalanche_c',
+     'table_name': 'FraxswapPair_evt_Swap',
+     'time_column': 'evt_block_time'},
+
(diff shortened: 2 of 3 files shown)


augustog:
> The test fails when either age_of_last_record_days is not zero or if the diff between the sources and delays value age_of_last_record_days is not zero.

this is an excellent idea

.github/workflows/dbt_precommit_hooks.yml
```diff
@@ -31,6 +31,7 @@ jobs:
         with:
           format: space-delimited
           token: ${{ secrets.GITHUB_TOKEN }}
+        continue-on-error: true
```
augustog, line 34: 👍 should be a warning, not break the pipeline

augustog (approved):
looks good to me!

only minor comment is we may want to parametrise the per-project CTE somehow (rn it's taking only the special cases we know, maybe would be good to have a general approach to this and capture all dexes? is there any way we can grab the source tables automagically from the sources.yml files?)

tests/dex/dex_trades_assert_freshness.sql
```diff
@@ -0,0 +1,63 @@
+with delays as (
+    SELECT
+        project
+        , blockchain
+        , datediff(now(), max(block_time)) age_of_last_record_days
+    from {{ ref('dex_trades') }}
+    group by 1,2
+)
+
+, sources as
```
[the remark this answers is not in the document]
dot2dotseurat: thats a good ideas because it's very easy to put match the wrong project/blockchain to table.

.github/workflows/dbt_precommit_hooks.yml
dot2dotseurat: I definitely thought this would just not set the FILES arg on a failure. I'm sort of shocked.

dot2dotseurat:
@augustog newly parameterized version. 

I'll add some discussion on the height task.

PR #2359, opened 2022-12-19
dex.trades volume test

dot2dotseurat:
This test hardcodes Uniswap on Ethereum as the most active dex and when volume is surpassed on other dexs' the test will fail. I like this test despite the inclusion of a hard code because should pay attention when something like that happens for real and when it happens due to a data error. 

This test will fail until we fix Curve. It does not fail on V1 through a manual [test](https://dune.com/queries/1780756?d=4). 

My goal is to add additional curve tests and resolve that issue before merging this test.

### DIFF
--- a/models/dex/dex_schema.yml
+++ b/models/dex/dex_schema.yml
@@ -50,6 +50,9 @@ models:
       - &amount_usd
         name: amount_usd
         description: "USD value of the trade at time of execution"
+        tests:
+          - dbt_utils.accepted_range:
+              max_value: 1000000
       - &token_bought_address
         name: token_bought_address
         description: "Contract address of the token bought"
--- a/models/dex/dex_schema.yml
+++ b/models/dex/dex_schema.yml
@@ -52,7 +52,7 @@ models:
         description: "USD value of the trade at time of execution"
         tests:
           - dbt_utils.accepted_range:
-              max_value: 5000000
+              max_value: 100000000
       - &token_bought_address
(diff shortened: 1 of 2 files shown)


soispoke:
Does this mean the test will fail if another DEX legitimately has more volume than Uniswap in the future? Doesn't seem right

augustog:
as a short term measure it should work, but I do agree with @soispoke that maybe we should have a more general check here, ideally so we can replicate it for other sectors.

a few thoughts I had:
- check that the normalised values are within a certain range (uniswap breaks this test on a normal day so not good) https://dune.com/queries/1784284
- simple check on order of magnitude https://dune.com/queries/1784363

maybe the second could be good enough?

I'm not opposed to having a test check for something as fundamental to this market as uniswap being the lead, but it will probably not generalise.

dot2dotseurat:
> Does this mean the test will fail if another DEX legitimately has more volume than Uniswap in the future? Doesn't seem right

Yes but my argument is that when that happens, it will be good to know and either replace uniswap with that project or rewrite the test. This is going on the assumption that they will continue to be the dominant project for a while which I don't think is a crazy assumption.

soispoke (approved):

dot2dotseurat:
@soispoke @augustog  I'll remove the max value test from this since it's still failing and just merge the uniswap volume comparison. 

The top volumes look way too high still and the curve fix is already in so we still need some digging to find out why.

PR #2360, opened 2022-12-20
[Easy] - add fantom signatures to abi spell - Ready For Review

henrystats:
Added fantom signatures to the spell by Ilemi!

### DIFF
--- a/models/abi/signatures.sql
+++ b/models/abi/signatures.sql
@@ -5,14 +5,14 @@
         file_format = 'delta',
         incremental_strategy = 'merge',
         unique_key = ['created_at', 'unique_signature_id'],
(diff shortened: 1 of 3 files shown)


henrystats:
Hola ser @andrewhong5297, I updated the source tables but I couldn't actually find the chain signatures table on dune. 

Is there something I'm doing wrong as the pre-commit hooks test is failing on missing source (chain signatures)

andrewhong5297:
yeah, the pre-commit hooks fail because of the way I reference sources in the for loop I believe. It's a known issue, the commit will still work fine. However for adding new chains we might have to do a full reload since the materialization only checks last 2 days

soispoke (approved):
yeah pre commit fails because of for loop but otherwise lgtm, thanks!

PR #2332, opened 2022-12-15
[Easy] [Labels] Implement trader age label

gentrexha:
Brief comments on the purpose of your changes:

Adding a trader age label for addresses that have completed at least one trade on a dex or dex aggregator.

cc @bh2smith for internal review

**For Dune Engine V2**

I've checked that:

### General checks:
* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [x] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
~~* [ ] if adding a new model, I added a test~~
* [x] the filename is unique and ends with .sql
* [x] each sql file is a select statement and has only one view, table or function defined  
* [x] column names are `lowercase_snake_cased`
~~* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)~~
* [x] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

PR #2364, opened 2022-12-20
[Easy] Fix trader age label name

gentrexha:
I'm updating the trade age label name based on the comment [here](https://github.com/duneanalytics/spellbook/pull/2332#issuecomment-1358980521).

### DIFF
--- a/models/labels/trader_age/ethereum/trader_age_ethereum.sql
+++ b/models/labels/trader_age/ethereum/trader_age_ethereum.sql
@@ -25,15 +25,15 @@ select
   array("ethereum") as blockchain,
   address,
   case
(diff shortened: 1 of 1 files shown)


soispoke:
thanks a lot @gentrexha !

soispoke (approved):

soispoke:
wait, I don't think DEX is in the right place !

gentrexha:
> wait, I don't think DEX is in the right place !
> 
> '5 years old DEX trader' instead of '5 years DEX old trader' makes more sense right?

You're right. My bad :man_facepalming:. It should be fixed now.

PR #2365, opened 2022-12-20
Add Furio token contract to tokens_bnb_bep20

dedalus76:

I have added the Furio token contract to tokens_bnb_bep20.sql so that it will included in dex.trades and dex.prices. The prices.usd data for this Token is incomplete (coinpaprika data does not include the first 3 weeks of prices)

**For Dune Engine V2**

I've checked that:

### General checks:
* [X ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)

I queried the temporary table generated during the dbt run check using dune.com and the added token contract was included

image


soispoke (approved):
lgtm, thanks!

PR #2366, opened 2022-12-20
[WIP] T-22566 - Hide views resulting in OOM in DuneSQL

belen-pruvost:
The following views are currently timing out in DuneSQL:

- opensea.burns
- aave_ethereum.borrow
- aave_ethereum.supply
- aave_v1_ethereum.supply
- aave_v1_ethereum.borrow
- ironbank_ethereum.redeem
- prices.usd_latest

This PR aims to hide them as a short term solution. A long term solution will be applied as part of T-21577.

### DIFF
--- a/models/aave/ethereum/aave_ethereum_borrow.sql
+++ b/models/aave/ethereum/aave_ethereum_borrow.sql
@@ -1,6 +1,6 @@
 {{ config(
       alias='borrow'
-      , post_hook='{{ expose_spells(\'["ethereum"]\',
+      , post_hook='{{ expose_spells_hide_trino(\'["ethereum"]\',
                                   "project",
                                   "aave",
                                   \'["batwayne", "chuxin"]\') }}'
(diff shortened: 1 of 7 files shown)


0xRobin:
hmm.. prices.usd_latest should really be available in DuneSQL.
I'll pull up a quick fix!

0xRobin:
hmm, I just tried the compiled sql from the usd_latest view and it ran fine on Dune SQL?
https://dune.com/queries/1784575?d=11

belen-pruvost:
> hmm, I just tried the compiled sql from the usd_latest view and it ran fine on Dune SQL? https://dune.com/queries/1784575?d=11

Yes, you're right. And this is working on Dune Dev and Prod. However it fails via Query Execution (how I found this) and directly on Trino. So 🤷‍♀️. 

Here is a plan:
- I'll mark this PR as WIP
- I'll raise the concern about different experiences across  App / QES / CLI internally
- I'll test again on the App the views I've hidden on this PR
- I'll ping you folks here once I have some better understanding if anything else needs to be hidden

0xRobin:
Sounds like a plan, you can also try the improved query from my PR (both with and without the WHERE clause).
I don't think the result will be much different though as I expect the execution plan to be equivalent.

belen-pruvost:
Ok - good news, we don't need to hide these: we should just make sure that on the ITs side we don't run all of these concurrently, which I'll address separately. I'll close this PR for now.

PR #2367, opened 2022-12-20
T-22170 - Materialize spells that rely on array indexes

belen-pruvost:
These views rely on array indexes. The array base index varies between Spark and DuneSQL. Therefore, we discussed on the scope of this task that it'd make sense to materialized the views that rely on this. 

Disclaimer: I haven't tested these changes as I'm not really sure on what to look for.


models/curvefi/ethereum/curvefi_ethereum_view_pools.sql
```diff
@@ -1,6 +1,8 @@
 {{ config(
     materialized = 'view',
     alias = 'view_pools',
+    materialized='table',
```
dot2dotseurat, line 4: materialized is used twice which causes a compilation error.

belen-pruvost: Ah shoot, thanks for the pointer - I'll fix this!

models/seaport/ethereum/seaport_ethereum_base_pairs.sql
```diff
@@ -1,6 +1,15 @@
 {{ config(
-     alias = 'base_pairs'
-     )
+    alias = 'base_pairs',
+    partition_by = ['block_time'],
+    materialized = 'incremental',
```
dot2dotseurat, line 4: This shouldn't be incremental. There's no incremental logic in the section below. Can you turn this into a table instead?

belen-pruvost: I pasted this from other models like `seaport_optimism_base_pairs`. I honestly don't know: what would be better? To have this be a table, or to have this be an incremental table (and add the if incremental jinja bits to the model)?

0xRobin: :warning:  This *should* be made incremental, or it will make the seaport.trades model only update once a day.

belen-pruvost: Okay, let me fix that 👍 thanks!

belen-pruvost:
@dot2dotseurat / @jeff-dude: it's been a while for this PR, do you think it can still make it to main? 😄

0xRobin (approved):
I've resolved the merge conflict with the seaport base pairs.
This was already transformed to a materialized incremental table on `main` :)

soispoke:
merging this then 👍

PR #2368, opened 2022-12-20
improve prices.usd_latest performance

0xRobin:
Triggered by PR: https://github.com/duneanalytics/spellbook/pull/2366
This PR reduces the `prices.usd_latest` query to only take into account tokens that had an update in the last week.

This should improve the query performance on Spark and DuneSQL (although it doesn't look to bad [initially](https://dune.com/queries/1784575?d=11))

### DIFF
--- a/models/prices/prices_usd_latest.sql
+++ b/models/prices/prices_usd_latest.sql
@@ -4,24 +4,17 @@
         post_hook='{{ expose_spells(\'["ethereum", "solana", "arbitrum", "gnosis", "optimism", "bnb", "avalanche_c"]\',
                                     "sector",
                                     "prices",
-                                    \'["hildobby"]\') }}'
+                                    \'["hildobby", "0xRob"]\') }}'
         )
 }}
 
-SELECT pu.blockchain
+SELECT
+  pu.blockchain
 , pu.contract_address
 , pu.decimals
-, pu.minute
-, pu.price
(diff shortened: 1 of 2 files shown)


models/prices/prices_usd_latest.sql
```diff
@@ -4,24 +4,17 @@
...
-    GROUP BY blockchain, contract_address
-    ) latest
-LEFT JOIN {{ source('prices', 'usd') }} pu ON pu.blockchain=latest.blockchain
-    AND pu.contract_address=latest.contract_address
-    AND pu.minute=latest.latest
-
+, max(pu.minute) as minute
+, max_by(pu.price, pu.minute) as price
+FROM {{ source('prices', 'usd') }} pu
+WHERE minute > now() - interval 7 day
```
belen-pruvost, line 19: The issue we would have with this for now is DuneSQL compatibility

belen-pruvost: Exactly, this would only work if this was materialized (which I'm guessing we'd like to avoid considering the nature of the info it holds)

0xRobin:
While not necessary for PR-2366, we can still look to merge this as it's a nice example of the use of the max_by function.
Which really improves the readability of these kind of queries.

soispoke (approved):
didn't know about the max_by function, very cool

PR #2379, opened 2022-12-21
DBT project cleanup

0xRobin:
This PR brings the dbt_project.yml back under 1k lines :broom: :smile:

1. removes the `enabled:true` config on seeds (it's true by default so redundant)
2. sets `materialize:view` on the project level for models (there were 2 models which weren't in the project.yml)
3. defines the test_data seed schema at project level (there were ~40 seeds not listed in the project.yml, and thus under wrong/none schema)
4. fix the sudoswap seed so the test is passing


0xRobin:
I think it would also be an improvement to migrate the seed column definitions to seed specific yml files instead of having them all together in 1 file which can get really cluttered.

dot2dotseurat:
nice work. I think this will might result in a lot of merge conflicts initially but if you're down to help wizards navigate them, I'm down to merge this in.

dot2dotseurat (approved):

0xRobin:
@dot2dotseurat yeah I think merging this in and handling the conflicts on open PRs during the "no-merge" freeze should be good. :+1:

PR #2386, opened 2022-12-21
chore: add RSR pool to curvefi_ethereum_trades

larrythecucumber321:

Adding RSR Curve V2 pool to curvefi_ethereum_trades. Screenshot of working query below:

image

Note that the other token in this pool, FRAXBP (the Frax Base Pool comprising USDC & FRAX) at `0x3175Df0976dFA876431C2E9eE6Bc45b65d3473CC` has not been verified yet, but does not pose a problem for me, since the `amount_usd` values in the results are accurate.

**For Dune Engine V2**

I've checked that:

### General checks:


0xRobin (approved):
I've added the token to tokens.erc20.

Issue #1901, opened 2022-11-06
Invalid Label Storing


bh2smith:
When calling (for address = '0xf13f7bf69a5e57ea3367222c65dd3380096d3fbf')
```
select * from labels.all where address = '{{Address}}'
```
as set to default address for in [this query](https://dune.com/queries/1531811?d=1)

we see two rows that are "essentially identical" except that one has null labels and the other has non-null labels:

![Screenshot 2022-11-06 at 1 06 51 AM](https://user-images.githubusercontent.com/11778116/200147949-5103614a-ad39-4834-934b-3994c030384b.png)

These are counter intuitive results for such a table, and one would expect each row to be unique on 

`blockchain, address, category, contributor` (cc @soispoke -- since you are the contributor of the duplicate rows). 

This seems like a bug although the created an updated times are different and it might be a good idea to enforce some validation when updating this columns and only allow individual contributors to occupy one slot per (blockchain, address, category).

PR #2363, opened 2022-12-20
T-22855: fix foundation model

0xRobin:
Fixes bug that caused duplicates in the foundation model.

PR #2396, opened 2022-12-23
Fix failing tests

0xRobin:
1. `dao_addresses.gnosis_aragon`
There were null values causing duplicates. Originated as artifacts of an incremental run. Changed to `INNER JOIN` to never return null values.
2. typo in native token test
3. collision in native mints unique id
4. nft trader labels contained null values when not in top 10%. These are now filtered out.
-> this closes https://github.com/duneanalytics/spellbook/issues/1901
5. trigger a full refresh for `wash_trades` as [this change](https://github.com/duneanalytics/spellbook/pull/2363) hasn't rippled through to it. 
6. cow_protocol tx hash labels had a bug which catches duplicates from trades being in both `dex.trades` as in `dex_aggregator.trades`, fixed by a correctly placed `distinct`


0xRobin:
In current state this should mean we only have 2 failing tests remaining.
- `dex_trades_assert_freshness` (expected to fail in current state)
- `source_unique_fantom_transactions_hash`
Which both have root causes outside the spellbook repo.

(The CI might timeout given the full wash_trades refresh..)

henrystats:
hola ser 👋 - I wrote the dao_addresses.gnosis spell and the LEFT JOIN is essential to the data from the tables.

Can you use COALESCE(gw.dao_wallet_address, '') instead? that would solve the duplicates issue. I used that for the ethereum & polygon models so I can submit a seperate pr to fix that so it doesn't break the anymore.

dot2dotseurat:
There's got to be some timing component to this failure because currrently this returns 6 records. 
```
00:52:55  Failure in test dex_trades_assert_freshness (tests/dex/dex_trades_assert_freshness.sql)
00:52:55    Got 23 results, configured to fail if != 0
00:52:55  
00:52:55    compiled SQL at target/compiled/spellbook/tests/dex/dex_trades_assert_freshness.sql
```

0xRobin:
> hola ser wave - I wrote the dao_addresses.gnosis spell and the LEFT JOIN is essential to the data from the tables.

Hi, can you explain why the left join is necessary?

dot2dotseurat:
My general preference is that this should be split into separate PRs since each change has different stakeholders. We should have a PR @bh2smith is tagged on for the cow protocol change and @henrystats for the DAO gnosis spell.

henrystats:
> Hi, can you explain why the left join is necessary?

The dao_addresses table returns dao addresses & their wallet addresses from the following creator tools (Aragon, syndicate, zodiac, colony & dao-haus). 

Aragon daos have an address which is the Dao's contract address and then they have a seperate address (could be multiple) which is the Dao's wallet address. 

Not all DAOs have wallet addresses hence why that column is null for some Aragon DAOs but each DAO has it's own unique address which is important.

bh2smith:
I'm cool with this as is, but others who contributed to respective project last might like to be tagged. Also, the labels are all @gentrexha so he may be interested to take a look.

henrystats:
submitted a pr fix - https://github.com/duneanalytics/spellbook/pull/2398 cc @0xRobin

0xRobin:
> ot all DAOs have wallet addresses hence why that column is null for some Aragon DAOs
Ok fair, I've reverted to a LEFT JOIN there, a full refresh of this table should unblock the failing test cause it doesn't produce any duplicates. :ok_hand:

dot2dotseurat (approved):

soispoke (approved):

0xRobin:
@dot2dotseurat @soispoke I believe this is clear to merge? :pray:

soispoke:
@0xRobin yes, let's try this :)

PR #2398, opened 2022-12-23
[Easy] - fix dao-addresses - Ready for review 

henrystats:
Smol fix to fix dao_addresses failing tests


0xRobin:
This will just replace the null values with empty strings, so it's not necessary. 
The failing test will be fixed with a full refresh of the model :ok_hand: 
I'll close this PR (and reverted to a left join on the original one)

henrystats:
okie no worries, ty ser!

PR #2399, opened 2022-12-23
dex freshness convert to hours

dot2dotseurat:
This changes the check from "day" to hours. I thought there'd be a cleaner way to do hour intervals but this is the best I came up with after reading the INTERVAL docs. 

As a separate PR, I'm going to go through each of the sources on the underlying tables and make sure they have source freshness checks. This PR essentially silences some errors if the decoded tables are behind. I think we want this but only if we also keep freshness checks on sources.


dot2dotseurat:
@0xRobin can I have a review on this? the pre-commit is failing on expected error due to looping through variables of sources instead of referencing the source directly.

0xRobin (approved):

PR #2401, opened 2022-12-23
script to find sources missing checks

dot2dotseurat:
We want to ensure full coverage on source tables freshness checks for important spells. The script validate_source_freshness_checks.py should help us 1) identify the gaps and then 2) add a check to ensure they stay covered. 

How it works:
This class Sources recurs over all node dependencies to retrieve all sources for the arg model and then we search the manifest for freshness checks. As a step in CI (forthcoming?), this will return an error if there are any sources without freshness checks for specific important models. 

What else is this hellishly long PR?
I've somewhat indiscriminately added the default freshness checks to all sources missing freshness checks on nft_trades and dex_trades. These are going to need adjusting but I figured better to overreact than underreact. For sources that we don't expect additional data from, we can use an arbitrarily large value as the warn/error threshold. 

How can I check if a model has missing freshness checks on its sources? (after compiling)
Run `python validate_source_freshness_checks.py --model {model_slug}`
eg `python validate_source_freshness_checks.py --model  nft_trades`

What next?
We can add dbt jobs like `dbt source freshness --select +nft_trades` and subscribe the ingest and decode teams when we think the warnings are tuning correctly.


models/zeroex/optimism/zeroex_api_optimism_sources.yml
```diff
@@ -17,9 +17,6 @@ sources:
     # for native orders on v3
   - name: zeroex_v3_optimism
     description: "Optimism decoded tables related to 0x v3 Exchange Contract"
-    freshness:
```
dot2dotseurat, line 20: this model is missing and I've raised a height task wth decode

dot2dotseurat:
This failure `20:58:33  1 of 23 FAIL 6 dex_trades_assert_freshness ..................................... [FAIL 6 in 15.57s]` is waiting on this [test](https://github.com/duneanalytics/spellbook/pull/2399) change to be merged.

models/zeroex/optimism/zeroex_api_optimism_sources.yml
dot2dotseurat: spooky. sounds like this table never existed from our records.

jeff-dude (approved):
the script & output make sense. question is: 
- post merge, are we setting up a job in prod specifically to run source freshness on nft/dex?
- then we will go through each project which returns errors, determine proper timeframe, make adjustments, then merge in again?

as you mentioned, adding to CI will help future models prior to initial merge.

dot2dotseurat:
1) Yes, I'd like to add freshness checks specifically to the dex and nft trades jobs. 
2) Yeah, I think iteratively fixing that is the right way to go.

PR #2403, opened 2022-12-26
Allow for local compile of incremental models. 

0xRobin:
This PR:

Enhances the default `is_incremental` logic to allow an override with a command line parameter.
The default value is set as False.
Usage:
`dbt compile --vars '{force-incremental: True}'`

The base logic was taken from here: [is_incremental.sql](https://github.com/dbt-labs/dbt-core/blob/1.2.latest/core/dbt/include/global_project/macros/materializations/models/incremental/is_incremental.sql)

### DIFF
--- a/macros/dune/is_incremental.sql
+++ b/macros/dune/is_incremental.sql
@@ -1,0 +1,12 @@
+{% macro is_incremental() %}
+    {#-- do not run introspective queries in parsing #}
+    {% if not execute %}
+        {{ return(False) }}
+    {% else %}
+        {% set relation = adapter.get_relation(this.database, this.schema, this.table) %}
+        {{ return(relation is not none
+                  and relation.type == 'table'
+                  and model.config.materialized == 'incremental'
+                  and not should_full_refresh()) }}
+    {% endif %}
+{% endmacro %}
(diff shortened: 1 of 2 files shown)


0xRobin:
I'm not sure why but the pre-commit doesn't pick up the configs in `macros/macros_schema.yml`

0xRobin:
@dot2dotseurat @jeff-dude @soispoke 
What do you think? :)

dot2dotseurat:
Would we get the same thing from the proposed [change](https://github.com/dbt-labs/dbt-core/issues/6526) in mentioned in this [issue](https://github.com/databricks/dbt-databricks/issues/243)?

0xRobin:
@dot2dotseurat I don't think so, for me they don't really seem related.
This PR just allows us to locally get the compiled sql for an incremantal model (as if the is_incremental would return true)
It's really helpful when developing complex models and testing them out on Dune.

dot2dotseurat (approved):

0xRobin:
Thanks!
@cryptofreedman we should probably add a note of this in the docs somewhere.
I would add it next to "make sure to compile your models and test them out on Dune.com" but it seems we don't really suggest that anywhere currently. :thinking: 

Not sure where the best place would be

cryptofreedman:
@0xRobin Noted, will do, ty!

0xRobin:
> Let me know if this will solve your use case, thanks!

While certainly useful I don't think your current setup would be able to replace our workaround here.
We're looking for the compiled sql of an incremental model (as if `is_incremental()` would return true)

0xRobin:
@dot2dotseurat the `--no-allow-introspection` option that would be introduced could be a workaround for [this other issue](https://github.com/databricks/dbt-databricks/issues/243) we're having with long compile times?

Issue #2404, opened 2022-12-26
Seaport matching logic causing duplicates (T-23078)


0xRobin:
The current seaport logic produces duplicates on polygon.
Current understanding: This happens when multiple ERC1155s offers on the same token_id are accepted in 1 trade.
Example trade: 
https://polygonscan.com/tx/0x1878f6836e3d165fd57ab76292659d337a5be24713ac8ecc73ac7e0a03bf815a
We can see these records in the base_pairs
https://dune.com/queries/1807354?d=1
![image](https://user-images.githubusercontent.com/83790096/209544007-90eaf6ec-4adf-4990-969e-5f411fc9b2fb.png)
We should match offer 1 -> cons 2, offer 2 -> cons 3,..
But instead we are matching offer 1  -> cons (2,3,4,5), offer 2 -> cons (2,3,4,5), resulting in x4 duplicates.
Matching code is here: https://github.com/duneanalytics/spellbook/blob/6abab39c370afd0da3a646584b87a0061eaef1fc/models/seaport/polygon/seaport_polygon_trades.sql#L135-L146

## Possible solutions
1. We could apply a distinct at the top level, comes with a heavy computation cost though..
2. We could deduplicate the offers (we need all the distinct considerations as they contain the correct buyer fields), we would lose the sub_idx of the offer and replace it with the sub_idx of the consideration.
3. Find a solution that only ever matches 1 offer with 1 consideration, we could do a row_number() partitioned on the join keys on both tables and match on the row_number.. This also sounds computationally expensive though.

0xRobin:
@sohwak, do you have an insight how we should best resolve this?
cc: @jeff-dude @soispoke @dot2dotseurat

0xRobin:
Looks like we also have incorrect seaport data on ethereum, not sure if the root cause of the issue is the same.
```
select * from seaport_ethereum.trades
where tx_hash ='0xa4c28ec1ffe901ddc3c4b2b6e389e80ea6e56861452d54559af5dfc90787679e'
```
![image](https://user-images.githubusercontent.com/83790096/209547740-00aad280-f64e-4519-895c-d084746141fd.png)

0xRobin:
Ok so ethereum data is another root issue. :confused: 
This query reduces the trades to the problematic join, I tried it on other chains but for now only Polygon has duplicate records spawning from this. https://dune.com/queries/1807232

0xRobin:
[...]

PR #2406, opened 2022-12-27
Fix seaport polygon duplicates

0xRobin:
Provides a fix for the polygon duplicates discussed here: https://github.com/duneanalytics/spellbook/issues/2404

This isn't fixing the root cause but eliminates the duplicates.
It should be reverted/made obsolete when a solid seaport rewrite is done.


0xRobin:
@augustog this is a patch for the polygon duplicates, if we can merge this we can remove the exclude and unblock polygon trades on the hourly incremental job.

augustog (approved):
As a temporary fix it lgtm!

- Adds a test for the bug it solves ✅ 
- Extends seed test ✅ 
- Tests pass ✅

PR #2409, opened 2022-12-29
Remove hardcoded fields from nft_trades

belen-pruvost:
Found this hardcoded field that I don't think belongs in `nft.trades`


0xRobin (approved):
good catch, that's indeed a relic from testing that should be removed.

belen-pruvost:
I'll wait until the other fix runs on DBT before merging this, just in case 🤞

PR #2416, opened 2023-01-02
Add Quix Trades to `nft.trades`

chuxinh:
- Cast `amount_raw` values to decimal(38, 0) to be consistent with that in `nft.trades` (let me know if it's better to do that in individual tables instead and if that'd affect incremental build since data type change
- add `quix_optimism_trades` to `nft.trades`. There is no `unique_trade_id` in those event tables. I remember I was told to use `unique_key` in config instead. Let me know whichever works

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


chuxinh:
Also curious what's the reason for breaking out platform fee and creator fee from `nft.trades`? 
i realised that the quix trades one also does not have those despite fees exist in separate event tables

0xRobin:
Had a quix look: 

1. To expose the fee data, we should make a `quix_optimism_fees` model and include it in the main `nft.fees` model.
2. We definitely want to get rid of the `unique_trade_id` field in the future, for now though it's still in place. :(
This means we should add it to all the different event abstractions (a simple concat of all the unique_key fields will do the trick)
3. as a rule of thumb we want to cast the fields in the materialized tables, not in the views on top. This means doing the cast in the base event tables. There are other fields we'd like to see cast (like `number_of_items`, I'm hoping @belen-pruvost can provide a list.)

belen-pruvost:
Thanks for summoning me @0xRobin 😁 yes, both `number_of_items` and `amount_raw` should be explicitly casted to decimal(38,0) on `quix_v1_optimism_events`, `quix_v2_optimism_events`, `quix_v3_optimism_events`, `quix_v4_optimism_events`, `quix_v5_optimism_events` and `quix_seaport_optimism_trades`.

`number_of_items` already has the right type on `quix_optimism_trades` but for readibility we'd rather cast it explicitly on the underlying models.

chuxinh:
Thanks! Have already addressed the comments above. 
Note that I created an extra cte as final in v1 - v5 so it seemed as if there were lots of changes. It's just for convenience that i can select * and a the `unique_trade_id` :p

0xRobin (approved):
@chuxinh I pushed some final tweaks to get everything integrated. 
This looks good now, thanks for the contribution!

jeff-dude (approved):
thank you @chuxinh & @0xRobin 🤝

PR #2426, opened 2023-01-04
T-23065 - optimism transfers: update value filter for string datatype

jeff-dude:
fyi @chuxinh @MSilb7 
`optimism.traces` column `value` has been (or will be today) updated to string datatype, similar to other blockchains. to be safe, i looked into spells which read from it and filter on `value`. this appears to be the only use case. we've had minor issues when filtering on `value` which is string without applying single quotes. looking to get ahead of any potential issues.

the spell already writes out `value` as string (due to databricks infer schema, without implicitly casting a type). the `value_decimal` should write out as double still, which it is now as well.

also added a minor change to remove usage of `::` as future state architecture may not like it.

edit: also removed `/eth/` in directory structure, as it's not necessary


jeff-dude:
@duneanalytics/team-abstract this is a quick change ready to review

dot2dotseurat (approved):

MSilb7:
Hi, sorry just seeing this. Thanks for the edit, but wondering why value is going to be a string vs a numeric?

jeff-dude:
> Hi, sorry just seeing this. Thanks for the edit, but wondering why value is going to be a string vs a numeric?

our ingest team has made the value string across all blockchains. the source field in solidity is `uint256` which isn't applicable in spark, leading to downstream overflow errors and such. at this time, string is the most compatible to retrieve the values properly. this may change with dunesql, as we have more customization on datatypes, but i'm not 100% sure the timeline on that

PR #2430, opened 2023-01-05
T-22136: Add prices for all supported tokens

0xRobin:
This adds all tokens that:

- are present in tokens_ethereum.erc20
- have an active price feed in coinpaprika
(for now only eth mainnet)

relevant query: https://dune.com/queries/1842392

This should help solve some of the null values in nft.fees due to missing price information.


0xRobin:
Closing this as it includes inactive coins, will re-open a new PR

0xRobin:
re-opened to check new price checker logic from https://github.com/duneanalytics/spellbook/pull/2431

0xRobin:
I think all of the errors are from tokens that were already in the original file (they got changed cause I needed to adapt the quotes)
Once the new price-checker is merged I'll adapt this PR to comment out the inactive tokens.

0xRobin:
This looks ready for a quick review.
Also tagging @aalan3 cause this will add ~600 tokens to the price feeds, want to make sure that would not cause any troubles.

aalan3:
600 is a lot and will likely cause problems. In general we shouldn't add this many at the same time. Could you instead add the ones that are actually needed?

0xRobin:
I'm afraid they are all needed. :smiling_face_with_tear: 
They have all been traded in the last month on a DEX or have been used to buy NFTs.

If we currently aren't able to support this, I guess we could add some arbitrary limit that tokens need to have a certain amount of trades or so before being considered..

0xRobin:
Ok @aalan3 looks like the tokens are really heavily skewed on the dex side. 
I'll split it in 2 PR with just the important tokens for nft trades and the most important ones for dex.trades

What would you say is the max number of token additions that's feasible?

aalan3:
I'd say let's start by adding the first 60 and follow up with rest of the 150, that's a great idea to segment like that.

0xRobin:
Closing this as it's now spitted between these PRs:

https://github.com/duneanalytics/spellbook/pull/2437
https://github.com/duneanalytics/spellbook/pull/2438
https://github.com/duneanalytics/spellbook/pull/2439

PR #2431, opened 2023-01-05
use batch api calls for price checker

0xRobin:
Use the `/coins` and `/contracts/<chain-slug>` endpoints from coinpaprika to prevent calling for each token id.

### DIFF
--- a/scripts/check_tokens.py
+++ b/scripts/check_tokens.py
@@ -30,7 +30,7 @@ for new_line in filtered_lines:
     try:
         checker.validate_token(new_line)
         # Sleep to (hopefully) avoid limits
-        time.sleep(2)
+        # time.sleep(2) (not needed anymore since we fetch everything in batch at the beginning)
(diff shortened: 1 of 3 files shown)


0xRobin:
Ok so this adapts the CI price checker so it's able to handle a lot more entries in 1 PR.
Previous setup example run:
https://github.com/duneanalytics/spellbook/actions/runs/3845190393/jobs/6549043390
hit rate limit on the coinpaprika API and would have taken an hour to finish given the sleeps between calls

The new setup only makes a couple calls to the API in the beginning to retrieve all the necessary information.
This allows it to run +1,000 checks in only a couple seconds.
New setup example run (on same diff):
https://github.com/duneanalytics/spellbook/actions/runs/3847318130/jobs/6553627051
@Hosuke @dot2dotseurat

dot2dotseurat (approved):
sick! great work.

PR #2439, opened 2023-01-06
Add missing prices for dex.trades

0xRobin:
Adds prices from tokens used in dex trades.
For now limited to tokens that had + 30k trades.

Includes tweaks from https://github.com/duneanalytics/spellbook/pull/2437 for the token checker.


jeff-dude:
test shows four inactive ones here

PR #2447, opened 2023-01-08
Feat/balancer vebal votes abstraction

Stefenon:

Migrate `balancer_vebal_votes` abstraction from Dune V1.

cc: @markusbkoch @mendesfabio @hedgehog-jacek 

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


Stefenon:
I tried turning this query into incremental, but I don't think it works. Do you have any thoughts on that?

Hosuke:
> I tried turning this query into incremental, but I don't think it works. Do you have any thoughts on that?

I tried to make the spell into incremental in the last commit:
- not incremental: https://dune.com/queries/1888672
- incremental: https://dune.com/queries/1888779

Entries will be updated based on unique_keys defined in the config block.
Please have a review.

Stefenon:
> > I tried turning this query into incremental, but I don't think it works. Do you have any thoughts on that?
> 
> I tried to make the spell into incremental in the last commit:
> 
> * not incremental: https://dune.com/queries/1888672
> * incremental: https://dune.com/queries/1888779
> 
> Entries will be updated based on unique_keys defined in the config block. Please have a review.

Hey @Hosuke, I believe this incremental spell doesn't output the expected results, because the query relies on previous data from the table to create its new rows. This happens because even if providers did not vote in a specific round, they should still be present on it if they voted previously.

Therefore, when we build the spell as incremental, there will be information missing from providers that only voted in rounds previous to the ones from the incremental filter.

Perhaps if this spell is built as a materialized table, it wouldn't be heavy to query. Since voting rounds happen on a weekly basis, not having the most updated version of the spell every time we query it should not be a problem.

Hosuke:
> Hey @Hosuke, I believe this incremental spell doesn't output the expected results, because the query relies on previous data from the table to create its new rows. This happens because even if providers did not vote in a specific round, they should still be present on it if they voted previously.
> 
> Therefore, when we build the spell as incremental, there will be information missing from providers that only voted in rounds previous to the ones from the incremental filter.

The incremental view, will update incrementally, so it preserve the old/original records, and append/update new records to the view.
Incremental view records = not incremental records(first time refresh) + incremental records update daily

Stefenon:
> > Hey @Hosuke, I believe this incremental spell doesn't output the expected results, because the query relies on previous data from the table to create its new rows. This happens because even if providers did not vote in a specific round, they should still be present on it if they voted previously.
> > Therefore, when we build the spell as incremental, there will be information missing from providers that only voted in rounds previous to the ones from the incremental filter.
> 
> The incremental view, will update incrementally, so it preserve the old/original records, and append/update new records to the view. Incremental view records = not incremental records(first time refresh) + incremental records update daily

I believe the problem is in the query from row 86 to 126. We rely on the data from previous rounds to produce the data from the next ones, which is why I believe an incremental strategy wouldn't work except if we can query the spell inside itself to get the data from previous rounds.

Hosuke:
> I believe the problem is in the query from row 86 to 126. We rely on the data from previous rounds to produce the data from the next ones, which is why I believe an incremental strategy wouldn't work except if we can query the spell inside itself to get the data from previous rounds.

Noted.

Hosuke (approved):
LGTM.✅
Thank you @Stefenon 

Due to https://github.com/duneanalytics/spellbook/pull/2447#issuecomment-1400155301 this PR is currently not modified to incremental view. The full-refresh time is within 2 minutes, based on [dbt slim ci](https://github.com/duneanalytics/spellbook/actions/runs/4030407963/jobs/6929046025#step:9:13).

[Sample votes](https://dune.com/queries/1936727).

PR #2451, opened 2023-01-10
Create DAO readme.md

soispoke:

After doing the same for the `bridge` spells, I think we should add `readme` files to describe high-level spells. I wrote this one for DAO spells (dao.votes and dao.proposals).


0xRobin (approved):
This is great, I should do the same for `nft.trades` and include some contributor guidelines for those implementing a lower level model. :+1:

PR #2453, opened 2023-01-10
update Pipfile.lock

0xRobin:
Ran `pipenv update` which updates the packages in `pipfile.lock`.


0xRobin:
@dot2dotseurat @jeff-dude 
I think it's a good idea to periodically update the lockfile so it's in line with what people would get if they made a clean clone and did a `pipenv install`..
We could also change the guidelines to use `pipenv sync` instead of install. This takes the exact versions of the lockfile instead of locally resolving the version from the Pipfile requirements. This is however not really "standard practice" from what I can tell online, not sure why not.

dot2dotseurat (approved):

PR #2207, opened 2022-12-04
[DRAFT] OP Refactor Tokens & Housekeeping

MSilb7:
Brief comments on the purpose of your changes:

- Add functionality to automatically pull tokens from deterministic creators (i.e. Aave, L2StandardTokenFactory)
- Other housekeeping things (moving ovm1 folder under optimism)
- Small one-off static updates

**For Dune Engine V2**

I've checked that:

### General checks:
* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each sql file is a select statement and has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [ ] `coin_id` represents the ID of the coin on coinpaprika.com
* [ ] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:
* [ ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:
* [ ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ ] where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #2455, opened 2023-01-10
fix and cleanup optimism spell directory & schema naming standards

jeff-dude:
cc @MSilb7 

i noticed in #2207 merged PR, we overlooked a schema name for `l2_token_factory` spell. when looking to add schema name, this led me down a mini rabbit hole to reorg some of the optimism spells. now that spell will have a schema name (`ovm_optimism`) as intended, plus a few other things below.

in general, we don't typically have a blockchain at the root of spellbook models, but rather within a project or sector directory. i noticed there was an optimism directory at the root (`/models/optimism/`). across some models within directories, there were inconsistencies in schema naming and approach to it. i've tried to standardize it here.

can you review the new directory alignment and schema naming for each?
- added `ovm` at the root, moved some spells within, added default schema naming in dbt_project
- moved the bridge spell under `bridge/optimism/` directory, added to dbt_project for schema names
- consolidated source/schema yaml files for ease of use where applicable


MSilb7:
hey, awesome thanks for going through and standardizing. Seems good by me. Should any table names in the Dune UI be expected to be modified (I haven't done anything with the tables from PR #2207 yet, so should be good there).

So basic takeaways:
- Don't do an 'optimism' top-level folder (think of chain names like reserved keywords)
- For any generic sector things, try to do a top-level folder before 'optimism' (i.e. bridges)
- We'll use 'ovm' to mean internal Optimism system stuff
cc @chuxinh

jeff-dude:
> hey, awesome thanks for going through and standardizing. Seems good by me. Should any table names in the Dune UI be expected to be modified (I haven't done anything with the tables from PR #2207 yet, so should be good there).
> 
> So basic takeaways:
> 
> * Don't do an 'optimism' top-level folder (think of chain names like reserved keywords)
> * For any generic sector things, try to do a top-level folder before 'optimism' (i.e. bridges)
> * We'll use 'ovm' to mean internal Optimism system stuff
>   cc @chuxinh

i think your takeaways are accurate. blockchains are typically always nested within another root level directory. helps set the foundation for cross-chain spells.

as for this specific PR, schema/table name changes would be:
- `bridge_optimism. standard_bridge_flows`
- `ovm_optimism. cross_domain_messenger_messages`
- `ovm_optimism.l2_token_factory`

i don't believe anything else would be affected.

there's a chance in the future, we reorg the root level of directories and dune data explorer UI to match, if it continues to grow at this pace. i think there is some sentiment of difficult navigation at times. tbd on that! if that does happen, the goal would be to keep schema/table/view names to avoid breaking queries, just reorg how we view it.

soispoke (approved):

PR #2460, opened 2023-01-11
Fix duplicates caused by aggregator hash markers

0xRobin:
Some transactions by nftnerds.ai are causing duplicate matches in the aggregator_markers joins.
This fixes that by doing an anti join with itself on the markers.

This basically guarantees that a join on
`RICHT(tx.data, marker_siz) = marker_hash`
will only ever produce 1 result. 

~~This PR also includes better integration of the marker table in other nft marketplace models.~~
Better integrations into markeplace models will be done once this is merged
cc: @hildobby

### DIFF
--- a/models/nft/ethereum/nft_ethereum_aggregators_markers.sql
+++ b/models/nft/ethereum/nft_ethereum_aggregators_markers.sql
@@ -1,11 +1,11 @@
 {{ config(
         alias ='aggregators_markers',
-		materialized = 'view',
+		materialized = 'table',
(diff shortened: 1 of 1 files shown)


soispoke (approved):
lgtm thanks!

jeff-dude (approved):

0xRobin:
@jeff-dude can you also review the related PRs listed above? :pray: 
I just resynced them to `main` so changes should be minimal. 
Some of them time out after 60minutes in the CI though.

PR #2462, opened 2023-01-11
Integrate aggregator markers in Blur

0xRobin:
Integrates changes from https://github.com/duneanalytics/spellbook/pull/2460


jeff-dude:
just going to comment on one, but should apply universally:
i wonder if the reason for timeout is due to using a left join on that other model, where the model is a v large set of data (assumption, since it reads from base table for all of time). we almost always have to inner join those, otherwise we get the poor performance.

its odd that all of these timeout, when prior to this join being added, all ran within the 60-minute window. i think it's worth digging deeper for performance sake.

0xRobin:
Hmm
Logically it has to be a left join, we can't really get around that..
Let me check the runtimes before the join with a separate PR

0xRobin:
Looks like performance wasn't really inflated with the added join. 
The model itself ran within ~40min the job failed due to other models being included and also taking up some time.
I've retriggered the CI to hopefully get a clean run

0xRobin:
Ok checks ran fine now, I assume they were just slow yesterday because of many concurrent checks running.. :thinking:

jeff-dude (approved):

jeff-dude:
> Ok checks ran fine now **in 8min**, I assume they were just slow yesterday because of many concurrent checks running.. 🤔

yeah, with them all running at the same time, i bet they just clogged up the test server

PR #2468, opened 2023-01-12
Fix and test the aggregator markers

0xRobin:
There was a small oversight in https://github.com/duneanalytics/spellbook/pull/2460.

Adding tests to catch the issue


0xRobin:
test failed as expected, pushing fix

jeff-dude (approved):

PR #2469, opened 2023-01-12
[Easy] - reload ellipsis-finance - Ready for Review

henrystats:
In the recently merged ellipsis finance pr, there was a particular pool that wasn't decoded with the right ABI (it was decoded via factory and the original submission didn't have the TokenExchangeUnderlying event). 

This has now been fixed and the trades are showing in the Stableswap.tokennexchangeunderlying event, and they are also being fed into the dex.trades table but it's only fetched during incremental runs. 

I've added back the seed I removed to test that the trade from before is in the ellipsis_finance trades and also added a ; to the ellipsis_finance bnb trades file so it gets reloaded and includes the old trades. 

cc @jeff-dude


jeff-dude (approved):
nice -- glad to see ya pick up on the little tricks along the way 🤝

henrystats:
Thank you ser 😅😁 for all your help.

PR #2483, opened 2023-01-14
[QuickFix] Fix type consistency in quix_optimism_events

Hosuke:
image

Column 30 in `quix_optimism_events` is `platform_fee_percentage`, which is hardcoded as '2.5' in sub-spell.

I also cast the following fields as double in the quix_optimism_events:
- platform_fee_amount_raw
- platform_fee_amount
- platform_fee_amount_usd
- platform_fee_percentage
- royalty_fee_percentage

PR #2487, opened 2023-01-16
[quick fix] Address DuneSQL compatibility regression

augustog:
Quix models are currently breaking nft.* on DuneSQL due to a type incompatibility issue. #2483 addressed this but missed a few fields that caused the error to persist.

This PR casts the remaining numeric fields to double in the quix events view.


soispoke (approved):
lgtm

PR #2498, opened 2023-01-17
[Prices] Remove Bogus Token Price

bh2smith:
This token is causing problems with Dex Trade valuation and clearly has a very bad price feed (in more than just the coin paprika market).

From [this transaction](https://etherscan.io/tx/0x918eacff2b6c1fdbb14920473974dd471301f9f305c010baa3085b9ed59c33a6) we can see that several trillion of this token were swapped for only 1.5K USDC. This implies that even Etherscan has incorrect valuation of the token.

Here we see our own bogus trade valuation: https://dune.com/queries/1901839?d=1

I believe the problem here to be related to "market depth". The prices are likely showing "spot prices" but don't account for the depth of the pool showing these prices. When such a large order comes in, the pool is only able to return very low amounts. The only solution I can think of at the moment is to cut this off at the source. Since the coin paprika page does not appear to display any active markets for this token. Maybe it does not make sense to keep it around... 

[Coin Paprika page for this token](https://coinpaprika.com/coin/erc20-erc20/) shows the value at ~0.007 cents at the time of this trade which clearly was not realized


bh2smith:
Tests appear to be failing because there is a diff in prices file... not sure if this make sense.

augustog (approved):
This looks good, not sure why the test failed...

josojo (approved):
I support this PR

jeff-dude (approved):
thank you

PR #2511, opened 2023-01-19
X2Y2 token_id fix

0xRobin:
This fixes `token_id` problems in the x2y2 model when the id is to large to be handled by `bytea2numeric_v2()`


0xRobin:
checking both tests fail like expected.

0xRobin:
closing due to https://github.com/duneanalytics/spellbook/pull/2647

PR #2521, opened 2023-01-20
[bug] fix airswap model not updating on incremental runs

augustog:
Airswap model is falling behind by days, but works perfectly when running a full_refresh and raises no errors while running an incremental refresh.

The culprit is an incremental filter set to = instead of >=.

This PR fixes it.

### DIFF
--- a/models/airswap/ethereum/airswap_ethereum_trades.sql
+++ b/models/airswap/ethereum/airswap_ethereum_trades.sql
@@ -115,7 +115,7 @@ INNER JOIN {{ source('ethereum', 'transactions') }} tx
     AND tx.block_time >= '{{project_start_date}}'
     {% endif %}
     {% if is_incremental() %}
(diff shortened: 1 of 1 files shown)


Hosuke (approved):
LGTM.✅
Thanks for the fix

jeff-dude (approved):
doh! good catch. i wonder the best approach to test this via automation. will add to doc.

augustog:
I'm looking into sql parser libraries to see if we can express these patterns somehow... not making much progress there so if anyone has any ideas please ping :D

an alternative would be to test # of rows added after an incremental run?

jeff-dude:
> I'm looking into sql parser libraries to see if we can express these patterns somehow... not making much progress there so if anyone has any ideas please ping :D
> 
> an alternative would be to test # of rows added after an incremental run?

i think quick approach would be to query count of what was incremetnally loaded, yes, but we haven't exactly set that up in CI testing yet either (ability to query and spit out results)

PR #2526, opened 2023-01-21
add tetuswap to dex trades - wip 

henrystats:
I've also included tetuswap fantom to the dex_assert_freshness test as a source that might not be updated daily. 

The last trade was October last year and it looks like the pools have no tvl any longer so possibly that folks don't trade there anymore but I'll check.

I plan to add the polygon model as well so I'd check to see why the last time the fantom table was updated was October (but I already checked ftm scan and the pools are inactive).


jeff-dude:
can anyone update a status on this PR?

PR #2547, opened 2023-01-24
bytea2numeric_v3

0xRobin:
This introduces a `bytea2numeric_v3` function that allows for hex strings up to 64 chars/bytes. 

todo:

- [x] provide test seed covering hex strings of all lengths


0xRobin:
Hmm.. the macro needs to be run to create/register the function in order for the test to run.
I can't do this however in the dbt project as this always needs a DB connection.
Not sure how to proceed with this.

@dot2dotseurat @soispoke @jeff-dude

macros/public/bytea2numeric_v3.sql
```diff
@@ -0,0 +1,66 @@
...
+        ,substring('26959946667150639794667015087019630673637144422540572481103610249216',1,17)::decimal(38) as base_56_4
+        ,substring('26959946667150639794667015087019630673637144422540572481103610249216',18,17)::decimal(38) as base_56_3
+        ,substring('26959946667150639794667015087019630673637144422540572481103610249216',35,17)::decimal(38) as base_56_2
+        ,substring('26959946667150639794667015087019630673637144422540572481103610249216',52,17)::decimal(38) as base_56_1
+    )
+
+    --compose all factors together in the correct decimal(38)s
+    , decimals as (
+    select
+         hex_input_64
```
augustog, line 33: where should this column come from? it's not defined in `base_factors`

macros/public/bytea2numeric_v3.sql
```diff
@@ -0,0 +1,66 @@
...
+            + (hex_4 * base_42_1) + right(hex_4 * base_42_2,17)::decimal(38)*pow_10_17
+            + (hex_5 * base_56_1) + right(hex_5 * base_56_2,17)::decimal(38)*pow_10_17
+         )::decimal(38) as dec_1
+        from base_factors
+    )
+
+    --handle carry overs from dec1 -> dec2 and dec2 -> dec3
+    , carries as (
+        select
+        ,dec_3 + left(lpad(dec_2,35,'0'),1)::decimal(38) as dec_3
```
augustog, line 51: smol parse error here (`select ,dec_3...`)

macros/public/bytea2numeric_v3.sql
```diff
@@ -0,0 +1,66 @@
...
+        ,right(dec_2,34) as dec_2
+        ,dec_1
+        from (
+            select
+                ,dec_3
```
augustog, line 56: smol parse error here (select ,dec_3...)

augustog (changes requested):
I manually created the UDF in databricks and fixed the three lines that I commented here to have it run.

Thanks for taking this on ser 🙏🙏🙏

macros/public/bytea2numeric_v3.sql
0xRobin: Thanks, I removed this column, it's not needed and was a relic from exploration.

0xRobin:
Tests are passing :heavy_check_mark:.
This means we have a working implementation for hex -> decimal string conversion that works for up to 64 chars. :partying_face:

augustog (approved):
woopwoop!!
so happy to finally have this working <3

PR #2552, opened 2023-01-25
[Easy] [Labels] Include dex aggregator trades in arb trader label

gentrexha:
Include dex aggregator trades in the arb trader label which has been considering only dex_trades up until now.

cc @soispoke which has been responsible for labels recently


soispoke:
thanks @gentrexha ! lgtm 👍

soispoke (approved):

PR #2555, opened 2023-01-25
[Easy] make trader age spell crosschain - Ready for review 

henrystats:
@gentrexha I'd like to make a small change to the trader age spell you built. 

I removed the **blockchain/ethereum** filter so there are more blockchains included for this spell as the dex.trades table is a lot more robust now with multiple dexes from multiple chains and the aggregator trades table will also soon have more models from other chains (I've a couple of pr's up).

Please let me know what you think!

PR #2557, opened 2023-01-25
[Easy] make trader frequencies spell cross chain - Ready for review 

henrystats:
@gentrexha this is similar to pr: https://github.com/duneanalytics/spellbook/pull/2555

This pr makes the trader diversity spell crosschain 

erm not sure why the spell Is failing in gh. The compiled query works on dune.com


henrystats:
@augustog, @gentrexha originally worked on this spell and we discussed on a previous pr (similar to this one) and we opted to have these split by blockchains instead of combining all into one. 

I also think it's better splitted since many times one would want t use this for a specific blockchain and not cross-chain.

augustog:
I think we may have sth lost in translation here...
this PR is building a a cross-chain spell on a single file (as opposed to having spell/ethereum/...sql spell/optimism/...sql), which I think, for this use case, is correct (we may want to partition by blockchain and block_date to facilitate single chain usage, but I think this way minimises duplication).

hence, I think this is ready to merge :)

henrystats:
ah I see, awesome!

soispoke (approved):
lgtm!

PR #2564, opened 2023-01-26
Add curve fantom to dex trades - Ready for review

henrystats:
Followed the same design as used in ellipsis_finance. A lot more cleaner! 

I also wrote a quick test query here to check that all the pools in token exchange & underlying show up.

https://dune.com/queries/1933132

Only a few didn't and max number of transactions from those pools are 15. 

This is now ready for review 🫡 -


Hosuke (approved):
LGTM.✅
Thank you @henrystats 

[Sample curve fantom trades](https://dune.com/queries/2072986?d=1)

PR #2565, opened 2023-01-26
Upload manifest

dot2dotseurat:

Workflow that will upload a new manifest after a pr is merged. 

next step: is to download the manifest in the slim ci instead of use the manifest from dbt cloud api.


.github/workflows/commit-new-manifest.yml
```diff
@@ -0,0 +1,41 @@
+name: Update Manifest
+
+on:
+  pull_request:
+    types: [closed]
```
jeff-dude, line 5: just double checking: this means on merge to main? not just any closed PR?

dot2dotseurat:
@jeff-dude I'm honestly only like 70% sure on that. That's what the docs lead me to believe but it certainly *looks* like it would also upload when a PR is closed not merged. It is OK in that failure mode because it's uploading from main anyway. So it would be an unnecessary upload but not an incorrect one.

jeff-dude (approved):
makes sense after your comment & the "worst case" being another upload from main

PR #2567, opened 2023-01-26
Fix errors in and enhance maker_ethereum.accounting

web3lyt:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


web3lyt:
> * we need to be careful on the partition column. spark doesn't like a ton of small files (which each partition will be its own parquet file), so rather than timestamp, we'd like want to group transactions into a larger timeframe -- day? week? month? depends on data volume in each group, but we'd want relatively large chunks of data in each partition. we can continue that conversation on the new PR
> * combo of columns works just fine! then we can uncomment the incremental filters and also add a few more config options potentially
> 
> looking forward to continuing this on the new PR

@jeff-dude Good to know that about partitioning! ts is definitely not suitable then!
Is it possible to only have ts in the output table but still partition by date or month? Like can we put logic into where we set the partition_by parameter? Something like `partition_by = LEFT(['ts'],7),`. I know this is a longshot but wondering if we could do this.

If not, we can just go back to code. That's the best alternative. I'll change it back as soon as I get a response to the above.

Do you have any idea what is causing the errors above in this PR?

Sounds good! Will submit this PR and get it working as is (once I fix partition column) and then work on making it incremental in the next PR. Thanks so much!

jeff-dude:
> > * we need to be careful on the partition column. spark doesn't like a ton of small files (which each partition will be its own parquet file), so rather than timestamp, we'd like want to group transactions into a larger timeframe -- day? week? month? depends on data volume in each group, but we'd want relatively large chunks of data in each partition. we can continue that conversation on the new PR
> > * combo of columns works just fine! then we can uncomment the incremental filters and also add a few more config options potentially
> > 
> > looking forward to continuing this on the new PR
> 
> @jeff-dude Good to know that about partitioning! ts is definitely not suitable then! Is it possible to only have ts in the output table but still partition by date or month? Like can we put logic into where we set the partition_by parameter? Something like `partition_by = LEFT(['ts'],7),`. I know this is a longshot but wondering if we could do this.
> 
> If not, we can just go back to code. That's the best alternative. I'll change it back as soon as I get a response to the above.
> 
> Do you have any idea what is causing the errors above in this PR?
> 
> Sounds good! Will submit this PR and get it working as is (once I fix partition column) and then work on making it incremental in the next PR. Thanks so much!

i don't think you can partition that way. storage isn't a major concern, so adding a column specifically to partition is fine. my suggestion would be to just add a `block_date` or `block_month` column and just build it based on timestamp. otherwise, if `code` has a relatively even distribution of data volume across them, that's fine too!

i'll look at the errors now

models/maker/ethereum/maker_ethereum_sources.yml
```diff
@@ -27,4 +27,20 @@ sources:
...
+        identifier: VOW_call_file
+      - name: dssvesttransferrable_call_create
+        identifier: DssVestTransferrable_call_create
+      - name: dssvesttransferrable_call_yank
+        identifier: DssVestTransferrable_call_yank
+      - name: dssvesttransferrable_evt_vest
+        identifier: DssVestTransferrable_evt_Vest
+      - name: mkr_evt_transfer
+        identifier: mkr_evt_Transfer
+  - name: prices
```
jeff-dude, line 39: this is the error here. prices as a source is already established in another source file, located [here](https://github.com/duneanalytics/spellbook/blob/main/models/prices/prices_sources.yml) -- no need to readd 👍

jeff-dude: in case you haven't used our CI testing before, you can see that [here](https://github.com/duneanalytics/spellbook/pull/2567/checks#step:7:11)

jeff-dude: i'd also highly recommend you run `dbt compile` on your local machine after updating models there, prior to pushing here, to ensure it completes successfully. even when compile completes, you can take the output code in target directory, which is plain sql, and run on dune to ensure it's as you'd like to see it

models/maker/ethereum/maker_ethereum_accounting.sql
```diff
@@ -924,256 +934,479 @@ WITH dao_wallet AS (
...
+        , 1 AS vested
+    FROM {{ source('maker_ethereum', 'dssvesttransferrable_evt_vest') }}
+), pause_proxy_mkr_trxns_raw AS
+(
+    SELECT evt_block_time       ts
+        , evt_tx_hash hash
+        , value         AS      expense --positive expense which is a reduction in equity
+        , `to`          AS      address
+    FROM {{ source('maker_ethereum', 'mkr_evt_transfer') }}
+    WHERE `from` = '0xbe8e3e3618f7474f8cb1d074a26affef007e98fb'
```
jeff-dude, line 1019: minor note -- backticks should no longer needed on columns

models/maker/ethereum/maker_ethereum_accounting.sql
```diff
@@ -924,256 +934,479 @@ WITH dao_wallet AS (
...
+            , 39999 AS code
+            , 0 AS value
+            , token
+            , 'Currency Translation to Presentation Token' AS descriptor
+            , NULL AS ilk
+        FROM m2m_levels
+    ) unioned
+    USING (code)
+    LEFT JOIN token_prices
+    ON unioned.ts::DATE = token_prices.ts::DATE
```
jeff-dude, line 1348: we are migrating away from `::` usage for datatypes, please use `cast(column as type)` instead through the script(s)

jeff-dude:
i'm sure you've seen, but [this test](https://github.com/duneanalytics/spellbook/blob/main/tests/maker/ethereum/maker_ethereum_accounting_test.sql) is what's failing in the CI tests

web3lyt:
@jeff-dude is this good to go now? Or what else is needed on my side?

Thanks so much for your help!

jeff-dude:
> @jeff-dude is this good to go now? Or what else is needed on my side?
> 
> Thanks so much for your help!

i'm just returning from some time off -- let me take a look at where you stand now 🤝

jeff-dude (approved):
looks good, thanks again for the improvements! since it's a full-refresh table, it'll update once a day.

PR #2587, opened 2023-01-30
[updated] Feat/balancer vebal votes abstraction

jeff-dude:
appears there was an access issue when editing PR #2447 -- closing there and opening here


jeff-dude:
for reference, the final commit helped with a few items:
- update model naming standards to include blockchain name, similar to others in directory
- add `table` materialization in config block, this should give better downstream performance vs. a view -- currently, prod will refresh this table once per day

augustog (approved):
lgtm!

PR #2609, opened 2023-02-02
create nft floor price spell - Ready for review

henrystats:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


models/nft/ethereum/nft_ethereum_floor_price.sql
```diff
@@ -0,0 +1,126 @@
+{{ config(
+    alias = 'floor_price',
```
augustog, line 2: one good point of feedback from @0xRobin: this is a good base for a collection stats table, so I would rename the spell to `collection_stats`

augustog (commented):
It's overall looking good! My only concern would be probably adding a few (unit) tests to ensure it behaves as expected.

I made a smol comment re naming, otherwise lgtm

models/nft/ethereum/nft_ethereum_floor_price.sql
henrystats: ty ser, I'll add tests later tonight and update the naming.

models/nft/ethereum/nft_ethereum_floor_price.sql
```diff
@@ -0,0 +1,126 @@
...
+WITH 
+
+src_data as (
+    SELECT 
+        *,
+        date_trunc('day', block_time) as block_date
+    FROM 
+    {{ ref('nft_trades') }} 
+    WHERE blockchain = 'ethereum'
+    AND trade_type = 'Single Item Trade'
```
0xRobin, line 24: nft.trades is already setup so each row contains only one `token_id` traded.
It might thus be possible to drop this conditional.

One row might still contain a trade with multiple ERC1155s though, if you don't want to deal with that I suggest using `number_of_items = 1` as filter.

models/nft/ethereum/nft_ethereum_floor_price.sql
```diff
@@ -0,0 +1,126 @@
...
+prof_data as (
+    SELECT 
+        src.block_date, 
+        src.nft_contract_address,
+        percentile_cont(.05) WITHIN GROUP 
+            (ORDER BY src.amount_original/(
+                CASE 
+                    WHEN src.currency_symbol IN ('ETH', 'WETH') THEN 1 
+                    ELSE prices.price
+                END
```
0xRobin, line 77: From my understanding this price calculation doesn't hold up.
You're doing:

For ETH/WETH trades
`ETH amount / 1`

For other ERC20 trades
`ERC20 amount / ETH price`

models/nft/ethereum/nft_ethereum_floor_price.sql
```diff
@@ -0,0 +1,126 @@
...
+)
+
+SELECT 
+    d.day as block_date, 
+    d.nft_contract_address, 
+    COALESCE(prof.currency_volume, 0) as currency_volume,
+    COALESCE(prof.trades, 0) as trades, 
+    COALESCE(prof.fifth_percentile, 0) as fifth_percentile,
+    COALESCE(prof.currency_min, 0) as currency_min, 
+    COALESCE(prof.currency_max, 0) as currency_max
```
0xRobin, line 120: given you're not grouping on different currencies used, I think it's good to adapt the naming here to include ETH or USD here instead of  the generic currency so it's clear in what currency the columns are expressed.

0xRobin (commented):
I gave this a quick scan and left some comments.

models/nft/ethereum/nft_ethereum_floor_price.sql
henrystats: agreed sir, I'll update now.

models/nft/ethereum/nft_ethereum_floor_price.sql
henrystats: okaiee ser, I'll update thank you very much!

models/nft/ethereum/nft_ethereum_floor_price.sql
henrystats: updated.

models/nft/ethereum/nft_ethereum_floor_price.sql
henrystats: fixed

models/nft/ethereum/nft_ethereum_floor_price.sql
henrystats: updated

models/nft/ethereum/nft_ethereum_floor_price.sql
henrystats: tests added and name updated

henrystats:
@augustog, @0xRobin I've updated the naming & also added some seeds to test some of the results. 


I wrote a simple & separate query for the seed results - https://dune.com/queries/1978740?d=1

I'm now facing some rounding error with the tests/seeds not sure how to write it in way that'd make it work. The seed results are correct just not rounded properly which is weird.

0xRobin:
@henrystats I'm sorry can I suggest a final renaming? 
Just want to make sure it's as future proof and clear as possible so we don't need to rename something later on when there's already a bunch of queries depending on it.

currency_volume_eth -> volume_eth
fifth_percentile_eth -> price_p5_eth
currency_min_eth -> price_min_eth
currency_max_eth -> price_max_eth

The rounding problems can get a bit annoying, I'm sorry about that. 
Unfortunately I've not found why they seem to be inconsistent sometimes. 
I'd say try some combinations and if it doesn't work drop a row from the seed if necessary.

henrystats:
@0xRobin yes sure sir, I'll update now.

I've tried multiple combinations and at least one of it is always failing :(. Lemme try another and see if it works.

henrystats:
column names updated, tests still failing due to the rounding errors.

jeff-dude:
> column names updated, tests still failing due to the rounding errors.

i haven't dove deep into the rounding errors from seeds, but i'm wondering if it makes sense to change the test approach here from seeds to just a query in the `tests/nft/ethereum/` directory? if you have a CTE for `nft.trades` logic with subset of data, CTE for new model logic with same subset, then join together and perform calculation there? ensure it returns no rows?

maybe seeds are more important for decoded projects -> sector spells, where a simpler query could be used here?

henrystats:
Hola ser @jeff-dude, sorry I don't really understand...

jeff-dude:
okay @henrystats -- of the three linked PRs, i edited this one first. here are some notes:
- incremental models are tricky when using day sequence as driving table, then left joining in `nft.trades` data. if you look at latest code, i'm only using incremental filter on days CTE, otherwise the others pull all history in order to aggregate properly
- removed any seed references and used a test specific to this model. my point above was intended to mean that any spells which are not sector-wide (i.e. `dex.trades`, `nft.trades`, `perpetual.trades`, etc) can use more specific tests for their model, rather than the seed approach. i think this makes it easier to build and test. since you read from `nft.trades`, we can feel confident from prior seed tests that data is correct and can focus on the specific test case for this model

does this look to accomplish what you need from this spell?

jeff-dude (approved):

henrystats:
Yes ser @jeff-dude, thank you very much for your help ser!

0xRobin (approved):

seeds/nft/nft_collection_stats_seed.csv
```diff
@@ -0,0 +1,4 @@
+block_date,nft_contract_address,volume_eth,price_p5_eth,price_max_eth,price_min_eth
+2023-01-31,0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d,732.3399999999999,66.211,310,66.16
+2023-01-19,0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d,1249.3517,65.45,136.9,65.25
+2023-01-22,0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d,1594.1301999999998,67.85109,194.2,67.5
```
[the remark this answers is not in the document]
jeff-dude: i see less of a need to force a seed file here, as the spell is built specifically for one team -- even though it's under the `nft` schema which adds a bit of confusion. as these spells mature (or if they do), then i think forcing the seeds and fixing this is a priority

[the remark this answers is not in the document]
0xRobin: @jeff-dude not sure why my comment was included in my "approval". I assume it was a comment from last week that stayed in the drafts.. :thinking:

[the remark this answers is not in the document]
jeff-dude: weird!

PR #2611, opened 2023-02-03
create nft wallet pnl - Ready for review 

henrystats:
This spells create a nft.wallet_pnl model that returns the pnl of nft traders on ethereum.

The spell pulls data from nft.trades and calculates pnl by subtracting the buy volume from the sell volume for the same nft contract addresses.


henrystats:
@Hosuke file format delta is not supported for insert + overwrite incremental strategy. :(

https://docs.getdbt.com/reference/resource-configs/spark-configs

Hosuke:
> @Hosuke file format delta is not supported for insert + overwrite incremental strategy. :(
> 
> https://docs.getdbt.com/reference/resource-configs/spark-configs

If no `partition_by` is specified, the entire table will be overwrited with new data.
In this case, we may remove the incremental filters in this model/spell.

models/nft/ethereum/nft_ethereum_wallet_pnl.sql
```diff
@@ -0,0 +1,149 @@
...
+    LEFT JOIN 
+    trades t 
+        ON prev.wallet = t.wallet 
+        AND prev.nft_contract_address = t.nft_contract_address
+    GROUP BY 1, 2, 3
+    {% endif %}
+)
+
+SELECT 
+    ROW_NUMBER() OVER (PARTITION BY nft_contract_address ORDER BY pnl DESC) as index,
```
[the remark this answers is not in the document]
henrystats: yess, so this spell returns the pnl of the wallets for a specific nft_contract_address so yeah it can serve and is enough as the unique key. 

Also, just curious, wouldn't merge cause duplicates here? if we're using nft_contract_address & wallet as the unique keys then what happens seeing as the pnl & rank would possible change?

henrystats:
ah Interesting.

henrystats:
@Hosuke I just tested the test_schema on dune.com and it returns the correct results... using the merge incremental strategy without the partition_by. 

Should I proceed and remove the incremental filters? this is amazing!! and also just set alias & materialised for this.

Hosuke:
> @Hosuke I just tested the test_schema on dune.com and it returns the correct results... using the merge incremental strategy without the partition_by.
> 
> Should I proceed and remove the incremental filters? this is amazing!! and also just set alias & materialised for this.

Yes, we can simply leave it as `view` without incremental as we need full-refreshed table every time the source is updated.

henrystats:
hmm, but would this act as some sort of incremental table though? the idea is that every time the spell is refreshed, it doesn't have to go back and read the entire nft.trades table. 

Since we're {{This}} and I can use the filter (SELECT MAX(updated_at)) to only check for recent results then add up the pnl, would that still work? would it  only check for the trades after that period?

Hosuke:
> hmm, but would this act as some sort of incremental table though? the idea is that every time the spell is refreshed, it doesn't have to go back and read the entire nft.trades table.
> 
> Since we're {{This}} and I can use the filter (SELECT MAX(updated_at)) to only check for recent results then add up the pnl, would that still work? would it only check for the trades after that period?

Then you can review my current update, which is an incremental view. (And probably no dupes produced. Records will be updated based on the unique_key.)

henrystats:
ohhh amazinggg! looks good to me then! I checked as well and no dupes. 

Much appreciated ser!

models/nft/ethereum/nft_ethereum_wallet_pnl.sql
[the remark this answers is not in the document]
augustog: hmmm, the index would probably be problematic. I think we can leave that out of the materialisation tbh. Seems like it will add too much complexity, looking at how the query is set up.

models/nft/ethereum/nft_ethereum_wallet_pnl.sql
```diff
@@ -0,0 +1,149 @@
...
+        COALESCE(SUM(CASE 
+            WHEN t.trade_type = 'Sells'
+            THEN t.eth_amount
+            ELSE 0 
+        END), 0) + prev.eth_received as eth_received,
+        COALESCE(SUM(t.eth_amount), 0) +  prev.pnl as pnl, 
+        COALESCE(SUM(t.trades), 0) + prev.trades as trades 
+    FROM 
+    {{this}} prev 
+    LEFT JOIN
```
augustog, line 137: I see this will mean we will be working with the entire history of the dataset every time we do an incremental update, do you need this for anything other than calculating the rank at the end?

If we can do this as an inner join instead I think it would be best, so we're not running a merge with all previous records.

I may be thin on the performance optimisation details here so if you guys know better please correct me.

augustog (commented):
The logic looks good and the implementation is spot on, thanks @henrystats and @Hosuke !!

I left two inline questions, we may be able to live without the rank at the end if it adds too much complexity to the query.

augustog:
@Hosuke thank you so much for jumping in and helping here! and thanks @henrystats for the contribution!

I see the error is likely unrelated to this PR, I will double check but that should not get in the way of merging this PR.

models/nft/ethereum/nft_ethereum_wallet_pnl.sql
henrystats: hmm, I'm happy to remove the index. But the reason we're using a left join is to add up the partial pnl. At the time I wrote this incremental I had not written for the top sales & top minters so I think it's better to use a UNION ALL here?

So instead of left join, the incremental query runs a UNION ALL of SELECT * FROM {{this}} and then unions that with all the new entries after the max(last_updated)?

henrystats: Also, it doesn't actually look through the entire history of the dataset, well that's what I think I have written but it's my first time playing with incremental like this. 

But the only time it goes through the dataset entire history is the first run? and in the incremental run, it just takes the previous results of the table ({{this}}) and then adds the pnl from the previous run to the results after (max(last_updated))

henrystats: bump ser @augustog 👀

augustog: Sorry for the slow response! 

1. my suggestion to use inner join is indeed wrong, sorry for the noise there! If we used inner join here we would never add new wallets to the dataset
2. my point was, however, that if we select from {{this}} without any filters, we'll pull all records the spell ever created, change a few (the ones with new trades => different pnl) and then merge into the existing dataset, having to drop most of the records (as they will be the same). This is what I meant with "scanning the whole history of the dataset" (sorry if it was not very clear).

My advise would be to turn this into a RIGHT JOIN. This would mean we pull all new trades since last run, and all the PnLs that match those trades by wallet and collection) I'm not sure whether this would improve performance of the join itself too much, but at least we won't be generating a copy of the dataset for every update.

Does this make sense?

henrystats: An alternative would be to use a simple query that sum's the pnl of traders that's grouped and partitioned by block_date. 

And then do the SUM(pnl) when the spell is queried on dune.com. I'm not sure which is better though but just something I thought of.

jeff-dude: i think the issue here is if you read from {{this}} object and left join new incremental data, it'll only ever contain the wallets from the initial run. it won't add in any new wallets, as the left join keeps only from {{this}} model and corresponding matches from incremental.

my suggestion would be either keep it simple and make this just a table with no incremental logic -- CI test shows it runs pretty quickly and will just full refresh each time -- or potentially switch from left join to full outer join, to get wallets from both sides, then the logic to add across both tables should work.

this is a bit difficult to test on live prod data sources, it'd be good if we could replicate data to play out these scenarios.

henrystats: hmm, ohh you're right ser. 

What do you think about the **UNION ALL** approach? Would that be simpler / faster?

The issue is that there's no reason to have this on spellbook if it's going to have to go through the entire nft.trades table during every run. That way it can just be queried from dune.com? Whilst the incremental run would only sum up the previous data + new data.

henrystats: bump ser @jeff-dude

jeff-dude (approved):

PR #2612, opened 2023-02-03
create nft top sales - Ready for review

henrystats:
This spell adds an incremental model that returns the top 50 sales for all nft's on ethereum and pulls data from the nft.trades table.


henrystats:
@jeff-dude, wanted to tag you to this one incase you haven't seen it. It's very straight forward and the incremental query isn't complex.

jeff-dude:
> @jeff-dude, wanted to tag you to this one incase you haven't seen it. It's very straight forward and the incremental query isn't complex.

oops -- i did not see this one! sorry, will look soon

henrystats:
ty ser

jeff-dude:
i think this one needs a redesign, if i'm understanding correctly. this is how i've understood it:
- at any given time, the table will have 50 records for top-50 sales all time
- table should be unique on `nft_contract_address` and `nft_token_id`
- the `seller`, `price` & `tx_hash` will change any time a new sale happens, which may or may not be higher than the last sale
- even if those three fields change, they may drop out of the top 50 altogether, so it may not matter

in the incremental logic, since we're pulling all data from {{this}}, then we may as well read it from source itself each time, to avoid confusing logic.

i can try to play around with the thoughts above, in the meantime, feel free to tell me where i'm wrong, as surely i am :)

henrystats:
Hola ser @jeff-dude 👋

So I believe the table should be unique on just the nft_contract_address seeing as all the other columns can change. The top sale is for the **nft_contract_address** not for the token_id. 

I guess I'd need to update the unique key columns but I used to think that it need to be the same as what we had in dbtutilsunique combinations. And the unique test will fail unless we add the other columns (row number especially). 

When you say source, do you mean nft.trades table itself? 

I was using {{this}} to avoid reading the entire table seeing as the only possible thing is that a new trade has an higher sale volume than a trade in the top 50 so it'll get added and the lowest next rank gets displaced). 

And I thought it was okay  because we could also use the normal incremental filter (1 week) without worries about duplicates due to the UNION.

jeff-dude:
> Hola ser @jeff-dude 👋
> 
> So I believe the table should be unique on just the nft_contract_address seeing as all the other columns can change. The top sale is for the **nft_contract_address** not for the token_id.
> 
> I guess I'd need to update the unique key columns but I used to think that it need to be the same as what we had in dbtutilsunique combinations. And the unique test will fail unless we add the other columns (row number especially).
> 
> When you say source, do you mean nft.trades table itself?
> 
> I was using {{this}} to avoid reading the entire table seeing as the only possible thing is that a new trade has an higher sale volume than a trade in the top 50 so it'll get added and the lowest next rank gets displaced).
> 
> And I thought it was okay because we could also use the normal incremental filter (1 week) without worries about duplicates due to the UNION.

got it. i think the issue with current setup is that rows will never be removed. i'm reading this as: future sales can replace existing top sales. if it's a brand new nft, it'll have a new address, therefore will insert into target. i don't think the target (i.e. the spell) will stay at 50 records at all times. it'll grow over time.

i think one solution may be to have `rn` as the only value in unique key? that way we always recalculate all history -- keeping your union logic -- but then update the spell based on new rn values only.

does that make sense?

henrystats:
Alrighty ser, from what I think I understand, yes that'll work. I'll update the unique key now.

henrystats:
Thank you very much ser

jeff-dude:
seems the incremental portion is not performing. i checked the compute resources and there isn't a blocker there. we may need to revisit how to design the incremental portion

jeff-dude:
i have some ideas, let me play around locally

models/nft/ethereum/nft_ethereum_top_sales.sql
```diff
@@ -0,0 +1,82 @@
...
+{% if not is_incremental() %}
+WITH src AS
+(
+    SELECT
+        nft_contract_address
+        , token_id as nft_token_id
+        , seller
+        , amount_original as price
+        , tx_hash
+        , ROW_NUMBER() OVER (ORDER BY amount_original DESC) as rn
```
jeff-dude, line 23: alright ser @henrystats -- i fixed the logic here to make this work in a timely manner. there's a big if/assumption here. based on our convo in comments, if this is true:
the output should be top-50 sales of all time, at any given time. the output will have 50 rows at all times. updates will only happen if a rank has shifted.

with that in mind, we don't need to partition by `nft_contract_address` in the row number function. if we do, then we will obtain top-50 sales of every single nft contract address, therefore output would be count of nft contracts * 50 -- leading to an exponentially higher number. this is why it was taking much longer previously.

is this as expected?

jeff-dude: if we do need top 50 for each address, then we will need to just take out all "is incremental" logic and do a full read each time, change from merge to delete/insert, ideally that'd be quicker. easiest thing is to just take out incremental altogether if this scenario is what is needed, so it'd just replace the table each time.

henrystats: Thank you very much ser @jeff-dude. 

The top 50 is for the top sales for all the nft_contract_address, it was why I felt an incremental logic would help since the table would only look at trades in the past 7 days and check for trades higher than the lowest value in the previous top 50. 

If we have to look through the entire table, then would it be better to just query this from dune.com? 

cc @0xRobin (you pointed originally that without incremental then it'd be better to have these on dune.com so I'm curious what you think ser. 

Also tagging @0xBoxer for your thoughts on this.

jeff-dude: the simple approach is materialize as a table. it'll overwrite each time. the initial build on incremental wasn't taking long anyway. we can see how this runs.

jeff-dude:
```
select
  nft_contract_address,
  count(1)
from
  test_schema.git_7c4244a7_nft_ethereum_top_sales
group by
  nft_contract_address
order by
  2 desc
```

henrystats: alrighty ser, thank you very much!

jeff-dude: not sure if it was related to this spell, but dbt won't let us do insert-overwrite strategy for incremental on delta table format. it also won't work over sql endpoints, which is what our infrastructure uses currently. this is the only way, as we don't have unique columns to join on for a merge, it'll create cross joins that explode the row counts.

jeff-dude (approved):

PR #2615, opened 2023-02-03
replace dbt cloud api with s3

dot2dotseurat:

Removes DBT Cloud API dependency and replaces it with s3 download. Integration test files are just being used to trigger the CI as "modified" models.


0xRobin:
technically I think we'd want to use 
```
on:
  push:
    branches:
      - 'main'
```
here: https://github.com/duneanalytics/spellbook/blob/14fc92d7ce110f73286ebc5f383ef64dcde7f802/.github/workflows/commit-new-manifest.yml#L3-L5

The difference is really minor, but the current setup will also be triggered on:
1. closing PRs (not merging)
2. merging/closing PRs to branches other than main
Which we don't really need

dot2dotseurat:
@0xRobin I'll update the commit manifest workflow in a follow-up pr. I did realize that it would run additional uploads when a PR was closed but not merged but I figured that would be fine because the manifest wouldn't have changed. But it's confusing and unnecessary so I might as well fix it!

PR #2618, opened 2023-02-03
create nft minters - Ready for review

henrystats:
This pr adds an incremental model that returns the list of the top minters of nft's on ethereum.


soispoke:
we should turn this into labels @henrystats :)

models/nft/ethereum/nft_ethereum_top_minters.sql
```diff
@@ -0,0 +1,78 @@
...
+        SUM(eth_spent) as eth_spent, 
+        SUM(no_minted) as no_minted, 
+        last_updated
+    FROM 
+    src_data
+    GROUP BY 1, 2, 5
+)
+
+SELECT 
+    ROW_NUMBER() OVER (PARTITION BY nft_contract_address ORDER BY no_minted DESC) as rank_,
```
augustog, line 75: To reduce performance overhead, we can select and materialise the top 100 minters per collection. This would allow us to keep the union above from becoming too big with very small wallets. I would leave the number as a parameter, in case we want to change it later.

models/nft/ethereum/nft_ethereum_top_minters.sql
```diff
@@ -0,0 +1,78 @@
...
+        buyer as minter, 
+        SUM(amount_original) as eth_spent, 
+        COUNT(*) as no_minted,
+        NOW() - Interval '5 Minutes' as last_updated
+    FROM 
+    {{ ref('nft_mints') }}
+    WHERE blockchain = 'ethereum'
+    AND currency_symbol IN ('WETH', 'ETH')
+    AND amount_original IS NOT NULL
+    AND block_time < NOW() - Interval '5 Minutes'
```
augustog, line 29: we should use the same materialisation interval as other spells, at the moment we don't expect to have less than 1 hour freshness. Sorry if this got lost in translation! 

I'd suggest using:
- `max(block_time)` as the value for `last_updated`
- `block_time > (select max(last_updated) from {{this}} )` as the incremental filter

I think we can reasonably tolerate discrepancies caused not using the usual 7 day window here, but if someone disagrees please flag.

augustog (commented):
Thanks @henrystats for the contribution!! I left a few inline comments, pls review and let me know what you think.

models/nft/ethereum/nft_ethereum_top_minters.sql
henrystats: hmmm, the UNION ALL is summing the number of mints & eth spent from the previous results when the spell was last run.

If we limit the previous results to top 100 minters then would mean if **small** wallets mint huge number of tokens then the number of tokens minted & eth spent would be inaccurate..

models/nft/ethereum/nft_ethereum_top_minters.sql
henrystats: looks good to me, I'll update!

henrystats:
Thanks very much ser @augustog, replying nowww!

models/nft/ethereum/nft_ethereum_top_minters.sql
henrystats: bump ser @augustog

augustog: this is indeed correct ser, going again through the logic it makes sense as it is. My bad!

jeff-dude:
after a quick view of this spell & #2611 -- i wonder if we should emphasize @soispoke point above to make these labels. we have query labels that follow this type of format in a way. then with those labels, they are materialized as tables and run full refresh (as mentioned as potential solution in other linked PR).

are there particular use cases for these spells as to why they wouldn't fit into labels?

models/nft/ethereum/nft_ethereum_top_minters.sql
jeff-dude: fyi, using the `block_time > (select max(last_updated) from {{this}} )` type logic has caused us issues in the past performance wise, which is why we switched to hardcoded seven day period. i'd love to have that revisited in the future, especially as duneSQL is more prevalant and may allow for this logic

henrystats:
Oh I was thinking we would just write a query in the labels model and it didn't have to be included in this one.

models/nft/ethereum/nft_ethereum_top_minters.sql
henrystats: 👀🥲

jeff-dude:
> Oh I was thinking we would just write a query in the labels model and it didn't have to be included in this one.

sorry for the confusion @henrystats -- i'm realizing the request behind this PR, #2609 & #2611 now. let me work through these again with the added context.

henrystats:
Thank you very very much ser @jeff-dude

jeff-dude (approved):

henrystats:
Hola ser @jeff-dude, thanks very much ser.

I wanted to ask a question as well. Since we're only specifying a AND block_time > NOW() - interval '1 Week' filter, wouldn't this cause for some mints to get calculated multiple times? when we're trying to get the eth_spent or count the number of NFTs they minted?

jeff-dude:
changed back to normal incremental logic, the new way was taking longer than i'd like

henrystats:
Apologies for the late reply ser, our major ISP was down for a long period today and I'm just able to reply on GitHub. 

I took a brief look and I'm not sure if the incremental logic returns the correct results. This table is supposed to aggregate the number of NFTs  & eth spent to mint those NFTs by a specific wallet address grouped by nft_contract_address.

jeff-dude:
> Apologies for the late reply ser, our major ISP was down for a long period today and I'm just able to reply on GitHub.
> 
> I took a brief look and I'm not sure if the incremental logic returns the correct results. This table is supposed to aggregate the number of NFTs & eth spent to mint those NFTs by a specific wallet address grouped by nft_contract_address.

can you submit a fix PR with your thoughts? then i can review/iterate?

henrystats:
okaie ser will do that!

PR #2619, opened 2023-02-03
Create nf ttop holders (non incremental)

henrystats:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:

### DIFF
--- a/models/nft/ethereum/nft_ethereum_schema.yml
+++ b/models/nft/ethereum/nft_ethereum_schema.yml
@@ -168,3 +168,31 @@ models:
         description: "Name of protocol that routed this trade using the aggregator, is NULL if it's the same as aggregator_name"
       - name: hash_marker_size
         description: "Number of characters the hash_marker is made of, can be useful on joins"
+  
+  - name: nft_ethereum_top_holders
+    meta:
+      blockchain: ethereum
+      sector: nft
+      contributors: Henrystats
+    config:
+      tags: ['nft', 'ethereum', 'top Holder']
+    description: >
+        Top 50 NFT Holders
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - nft_contract_address
+            - wallet_address
+    columns:
+      - name: nft_contract_address
+        description: "nft contract address"
+      - name: wallet_address
+        description: "Wallet Address that Holds the nft"
+      - name: balance
+        description: "Number of NFTs held"
+      - name: supply_share
+        description: "Share of NFT Total Supply"
+      - name: supply
+        description: "NFT Total Supply"
+      - name: rn
+        description: "row number"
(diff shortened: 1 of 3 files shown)


augustog:
thanks @henrystats for creating this!

Two points:
- given the cost of running this (hits balances which is a pretty heavy spell) I'd ideally want to have it materialised somehow. I'd imagine we can leverage `{{this}}` macro and leverage any of the rolling balance spells. Do you think this would be feasible?
- there's a test failing on uniqueness, added a comment inline

models/nft/ethereum/nft_ethereum_schema.yml
```diff
@@ -168,3 +168,32 @@ models:
...
+      sector: nft
+      contributors: Henrystats
+    config:
+      tags: ['nft', 'ethereum', 'top Holder']
+    description: >
+        Top 50 NFT Holders
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - nft_contract_address
```
augustog, line 184: this is the test that is failing, although I cannot imagine why it fails. My first impression was that you're calculating `rn` separately for each token standard and then unioning, so it could be causing duplicates... but I can't imagine how the same nft_contract_address would exist in both token standards?

I'm pretty puzzled by why this is failing.

henrystats: yeah it's odd. I'll be taking a look as I plan to rewrite this as an incremental version. Although the balances table is actually returning inaccurate data and the query looks correct.

I mean the balances_erc72_latest that's on dune already. I'm trying to see if I can look for a small address with minimal transactions so I can see where the error is.

the total supply is accurate but the balances isn't.

henrystats:
Thanks you very much ser @augustog, I was waiting to finalise the other spells where I used {{this}}before re writing the incremental version of this spell. It's very similar to wallet_pnl so I wanted to finalise that before rewriting this as an incremental file. 

There's also something wrong with the balances table.

henrystats:
@jeff-dude this is still WIP as it depends on the fix for erc721 & erc1155 balances tables (still wip too)

henrystats:
Still not sure what's causing the unique tests to fail but the [test schema](https://dune.com/queries/2039597) returns the correct results when I test on dune.com

Although before the fix for the erc721 table was merged, there were some erc721 tokens that were in the list of nft's that failed the tests so I wonder if it's down to the erc1155 table not being accurate.

So we'll see if it fixes once that's merged.

henrystats:
uhh weird, this query is returning duplicated entries for the entire top 50 holders 

https://dune.com/queries/2047884?d=1 with accurate balances & supply

henrystats:
looks like something is up with my logic, will update.

PR #2624, opened 2023-02-04
[EASY] Add missing tokens

MykytaOksak:
This PR adds Connext USDC and WETH LP tokens across all 6 networks currently presented in the protocol


jeff-dude (approved):
thank you for the adds

PR #2640, opened 2023-02-07
Add models for Safe singletons across all networks

tschubotz:

I've added models for singleton addresses used with Safe (formerly Gnosis Safe). I intend to use them in a second step to create/update models to fetch all Safes across networks. Having the singletons dynamically available like this will make other queries more clean. 

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


Hosuke (approved):
LGTM.✅
Thank you @tschubotz

models/safe/arbitrum/safe_arbitrum_singletons.sql
```diff
@@ -0,0 +1,21 @@
+{{ 
+    config(
+        materialized='incremental',
```
jeff-dude, line 3: i'm debating in my head if all of these should be incremental tables. even if they do remain incremental, i wonder if we make the incremental strategy append-only vs. merge since there is only one column in the table, nothing to update. i think there are a few options, let me know if you think any are feasible:
1 - make all of these views, keep it simple (remove incremental logic and properties)
2 - keep incremental, make append-only, and to ensure only new addresses are added, can perform a `LEFT ANTI JOIN` against the model using jinja syntax {{this}} to ensure new addresses coming in last week don't already exist
3 - make all tables, not incremental, and just full refresh each time. based on CI test via GH action attached to PR, these run pretty quick anyway for full history (less than 10 seconds or so)

tschubotz: > 1 - make all of these views, keep it simple (remove incremental logic and properties)

One of the reasons for this abstraction was, that we wanted to have new singletons dynamically show up. (I was planning on use this new abstraction in the safes_list spell, once merged). Hence I would like to keep the incremental logic here, if possible.

>2 - keep incremental, make append-only, and to ensure only new addresses are added, can perform a LEFT ANTI JOIN against the model using jinja syntax {{this}} to ensure new addresses coming in last week don't already exist
> 3 - make all tables, not incremental, and just full refresh each time. based on CI test via GH action attached to PR, these run pretty quick anyway for full history (less than 10 seconds or so)

Whether option 2 or 3, I actually don't know since I don't know about the performance implications. Judging from my outside perspective, option 3 seems to be the more simple one and also feasible given the short query run time. (If that would help, refreshing this view only once a week would also be sufficient actually, if that would further help.)

jeff-dude: let me try to push changes to make option 3 happen. it'll fit into our existing prod job to run 1x/day, which should be acceptable given your timeframe of up to 1x/week

jeff-dude: @tschubotz how does it look now? all tests pass, the tables would be full refresh 1x/day if merged

tschubotz: @jeff-dude Thanks a lot, looks good to me. Full refresh 1x/day is perfect 👌

jeff-dude (approved):

PR #2644, opened 2023-02-07
[draft] adding detailed documentation to spells

0xBoxer:
A first attempt at aligning around creating proper documentations for spells.

- added in line documentation for arbitrum_v3_trades
- added readme.md documentation for both `uniswap_v3_arbitrum_trades` and `uniswap_arbitrum_trades`

Some of my thoughts around this:

1. I think I might have gone overboard with the in-line documenation in the .sql files, think a bit less would be reasonable.

2. Adding hyperlinks to other models in the readme is useful for having a more structured approach to learning and discovering spellbook content, albeit those dependencies up- and downstream can already be looked at natively in the usual spellbook documentation.

I realized we can pack multiple readme docs into one md file using the jinja template for docs, that really helps to not litter the repo with readme files.

Thinking that going forward we can probably do one for sources and one for models.

I didn't extend this to column descriptions and source descriptions yet, if we would include those as well we could build up a very relevant knowledge graph over time.

Compile the docs locally and check it out!

### DIFF
--- a/models/labels/sandwich_attackers/README.md
+++ b/models/labels/sandwich_attackers/README.md
@@ -13,8 +13,8 @@ The logic to identify a sandwich attacker is as follows:
 7. (index of t1 >= index of t2 + 2) or (index of t2 >= index of t1 + 2)
 8. Exclude uniswap v2 router address since this gets included as a false positive for some reason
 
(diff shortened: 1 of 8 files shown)


jeff-dude:
going to close this for now, but you'll have your branch for reference when this becomes a priority again

PR #2654, opened 2023-02-08
Remove AWS auth setup from upload workflow

dot2dotseurat:
The container should already have the awscli installed and permissions are already granted. Removing these steps will allow the workflow to continue and not fail on AWS creds set up when PRs from forked branches are merged.

### DIFF
--- a/.github/workflows/commit-new-manifest.yml
+++ b/.github/workflows/commit-new-manifest.yml
@@ -1,8 +1,9 @@
 name: Update Manifest
 
 on:
-  pull_request:
-    types: [closed]
+  push:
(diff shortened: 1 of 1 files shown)


a-monteiro (approved):

.github/workflows/commit-new-manifest.yml
```diff
@@ -22,15 +22,6 @@ jobs:
       with:
         ref: main
 
-    - id: install-aws-cli
```
a-monteiro, line 25: I can't comment on that line but you can remove lines 14-16

PR #2676, opened 2023-02-13
[not easy] - fix balances_erc721 - Ready for review

henrystats:
This is ready for review, not sure what is causing the pre-commit hooks to fail though as I added the description for the amount column. 

I've tested the balances against etherscan using BAYC nft contract address to test. 

https://dune.com/queries/1997732 

https://etherscan.io/token/0xbc4ca0eda7647a8ab7c2061c2e118a18a936f13d#balances


0xBoxer:
Wouldn't a rebase have been the better call here?

Just pushing all the commits from main into this branch makes it really hard to isolate the changes on this branch now.

https://www.atlassian.com/git/tutorials/merging-vs-rebasing

henrystats:
This is now ready for review ser @jeff-dude

models/balances/ethereum/erc721/balances_ethereum_erc721_hour.sql
```diff
@@ -17,23 +17,41 @@ with
...
+    collection
+FROM 
+(
+SELECT 
     'ethereum' as blockchain,
     d.hour,
     b.wallet_address,
     b.token_address,
     b.tokenId,
+    SUM(amount) as amount,
```
0xBoxer, line 49: I'm wondering if we need a sum() here, I think we could just get by with defining
``where amount = 1``

henrystats: hmm true lemme look into that. There was originally a **Where amount = 1** filter in the hourly/daily balance table (where we did the lead window function). 

Somehow it affected the results of the query causing wallets that don't own a tokenid to get counted multiple times 😅. Maybe the amount filter should just be below. 

I tried mirroring getting balances for an erc20 token (hence the SUM(amount)).

0xBoxer:
@henrystats got this model to work, should be good for merging.

I've reviewed the code and defined pretty strict logical tests, the results are pretty bulletproof from my POV.

0xBoxer (approved):

PR #2678, opened 2023-02-13
 fix erc1155 balances table  - Ready for review

henrystats:
fix balances table for erc1155 model.


jeff-dude:
adding WIP based on title, tag when ready

henrystats:
will update this so it's similar to the erc721 one @jeff-dude

henrystats:
@jeff-dude, @0xBoxer this is ready for review sers. I only made small changes to the partitioning to return the accurate balances and made some small schema changes. 


This query - https://dune.com/queries/2039208?d=1 returns the token holders list for this nft - `0xf36c296e87dd2d7adcef251a542561d4bbe07714`. 

The previous balances table was incorrect as can be seen in this query - https://dune.com/queries/2039384 for the same NFT. 

I cross checked with etherscan as well (the number of holders tallied). Actually etherscan was 2210 & the spell was 2211 (although I wonder if it's because of `0x000000000000000000000000000000000000dead` and if etherscan doesn't count it in holders as it has a balance of 1 NFT (that can also be tracked on etherscan as a transfer in). 

The assert sample failing test is because the samples are incorrect, I double checked on etherscan and also wrote queries here to show. 

current spell version (correct) - https://dune.com/queries/2039420?d=1

previous spell version with samples (incorrect) - https://dune.com/queries/2039412

This can also be double checked on etherscan - https://etherscan.io/token/0xf1f3ca6268f330fda08418db12171c3173ee39c9?a=0x09a5943a6d10919571ee2c9f63380aea747eca97

I believe this is good to go seeing as there are only small changes made to the spells. 

cc @0xBoxer

jeff-dude (approved):
i'm good with the code itself, however, i'll let @0xBoxer be the final approval on logic as i'm not too familiar with the data here and issues in current state

jeff-dude:
this is confusing to me:
> The assert sample failing test is because the samples are incorrect, I double checked on etherscan and also wrote queries here to show.

we need to fix the test prior to merge, does that mean removing the incorrect samples? or fixing them?

0xBoxer:
logic looks good to me and ran some individual tests against the git generated table, seems to add up.

No idea what's happening with the balance assert test, I would guess that was as wrong as this script 😆 

Would be great if you can define a new seed test @henrystats.

0xBoxer:
You could use the collection page from etherscan to get a snapshot of the holder count for example, I haven't looked into other possibilites yet.

https://etherscan.io/nft/0x33fd426905f149f8376e227d0c9d3340aad17af1/10

henrystats:
ohh yes @jeff-dude, @0xBoxer, the original test samples were incorrect results. I feel so dumb lol. I could have just changed them to the correct one. 

will update now and the tests should pass.

henrystats:
tests are now passing, this is ready for merging ser, @jeff-dude

PR #2685, opened 2023-02-14
Add More Ethereum NFT Contracts

hildobby:

Source: https://dune.com/queries/1854190

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


hildobby:
Screenshot 2023-02-15 at 00 23 52

I'm not sure what to do here 🤷‍♂️

pepe-the-frog-thinking

dot2dotseurat:
Yeah, it's too big to be a seed or a query from dbt. I can be a community dataset upload and then we can document it as a source in dbt. @dsalv and @aalan3 what do you think?

0xRobin:
> I can be a community dataset upload

Would that work if it still needs frequent updates in the future?

jeff-dude:
maybe this comes back to our other idea in the past around GH repo for large storage? of course that'd take some setup/config, but this has been a lingering issue for a bit with various ideas floating around. the community dataset system may be too early for files edited so often, as rob is mentioning?

dot2dotseurat:
github large file storage doesn't help us here because dbt is the bottleneck. I also don't think it's not reasonable to expect manual edits to files that are 94,417 lines. 

One possible solution: If people want to manually add contracts, we could have turn this model into one that unions those onto a community data uploaded table. 
So large file -> community data table
Individual rows appended to dbt model which uses community data table as a source. 

The other q I have is, where did the 90k rows come from? Can we just connect directly to that source and get this ingest there and forget about manual edits?

hildobby:
> The other q I have is, where did the 90k rows come from? Can we just connect directly to that source and get this ingest there and forget about manual edits?

@dot2dotseurat It came from reservoir's `reservoir.collections` table and other than a few name rewrites (due to some having single quotes, double or both), the rows added in this PR are just the downloaded dataset from this query I made: https://dune.com/queries/1854190

dot2dotseurat:
@hildobby ok, in that case, we should just use your query as a spell rather than inputting the values. then we don't have to update it when reservoir adds new contracts. 

Does this redesign make sense to people?
tokens_ethereum_nft -> wizard_curated_tokens_ethereum_nft (basically what this spell is now; people manually adding contracts. but we remove the expose spell macro.)
hildobby's query -> reservoir_tokens_ethereum_nft (new spell that we don't add an expose spell macro too) 
wizard_curated_tokens_ethereum_nft union  reservoir_tokens_ethereum_nft -> tokens_ethereum_nft 

Then wizards can keep manually adding token contracts but we also get all the contracts from reservoir.collections?

hildobby:
How about changing the currently existing spell to `tokens_ethereum.nft_wizards_curated` and change it so that `tokens_ethereum.nft` is basically:

```
SELECT *
FROM tokens_ethereum.nft_wizards_curated

UNION ALL

SELECT *
FROM reservoir.collections
```

That way, it doesn't change for dune users but everyone wins?

dot2dotseurat:
Yeah that works as long as the formatting in your query wasn't needed to make the two union-able.

hildobby:
Alright I got to work, how is this PR looking now @dot2dotseurat?

I'm also assuming that `incremental_strategy = 'merge'` is equivalent to a `UNION`. Wouldn't there be some kind of `incremental_strategy = 'append'` which would be equivalent to a `UNION ALL`? This would probably make the query more efficient 🤔

dot2dotseurat:
I'm wondering if this needs to be incremental at all or if this can just be a view without a big performance hit. Maybe the anti join makes that not work. Do you mind if I  mess around with this tomorrow? I want to see if we can avoid double storing the reservoir data.

hildobby:
Yeah ofc, feel free to play around with it and see what works best!

dot2dotseurat:
@hildobby I played around with this and  a view is a bit too sluggish but I think a table would work better than incremental here. A table materialized in about 2 seconds so incremental feels like overkill. 

I did a quick check to see if UNION would be more efficient than an ANTI JOIN and UNION ALL but looks like ANTI JOIN is performing better. Nice! 

I edited the alias on the wizard-curated table because I was getting an error about duplicated table names. We are injecting the git_sha into the alias in the CI so our tests missed that duplication.

dot2dotseurat (approved):

PR #2686, opened 2023-02-14
fix nft-minters 

henrystats:
@jeff-dude, I reverted this to the original incremental logic I submitted. I will try and work through the pr to see the changes you made but one of the main issues with this is the issue you highlighted with using `WHERE block_time > SELECT MAX(last_updated) from {{this}]`.

But this looks like the best way to avoid duplicates in the aggregated values and also not have to go through the entire nft.trades every time the spell is ran.

An alternative would be to aggregate this by block_date and then do the sum total through dune.com instead of at the spell level. 

Please lmk what you think ser @0xBoxer.

This is the same issue I'm having with what I plan to update wallet_pnl to, so I'd wait for this to be resolved before starting to work on that.

### DIFF
--- a/models/nft/ethereum/nft_ethereum_top_minters.sql
+++ b/models/nft/ethereum/nft_ethereum_top_minters.sql
@@ -11,29 +11,65 @@
     )
 }}
 
-WITH src_data as
-(
+WITH 
+
+src_data as (
+    {% if not is_incremental() %}
     SELECT 
-        src.nft_contract_address,
-        src.buyer as minter,
-        SUM(src.amount_original) as eth_spent,
-        COUNT(1) as no_minted,
-        MAX(src.block_time) as last_updated
-    FROM
-        {{ ref('nft_mints') }} src
-    WHERE
-        1 = 1
-        {% if is_incremental() %}
-        AND block_time >= date_trunc("day", now() - interval '1 week')
-        {% endif %}
-        AND blockchain = 'ethereum'
-        AND currency_symbol IN ('WETH', 'ETH')
-        AND amount_original IS NOT NULL
-    GROUP BY 1, 2
+        nft_contract_address,
+        buyer as minter, 
+        SUM(amount_original) as eth_spent, 
+        COUNT(*) as no_minted,
+        (SELECT MAX(block_time) FROM {{ ref('nft_mints') }}) as last_updated
+    FROM 
+    {{ ref('nft_mints') }}
+    WHERE blockchain = 'ethereum'
+    AND currency_symbol IN ('WETH', 'ETH')
+    AND amount_original IS NOT NULL
+    GROUP BY 1, 2, 5
+    {% endif %}
+    -- incremental run
+    {% if is_incremental() %}
+    SELECT 
+        nft_contract_address,
+        buyer as minter, 
+        SUM(amount_original) as eth_spent, 
+        COUNT(*) as no_minted,
+        (SELECT MAX(block_time) FROM {{ ref('nft_mints') }} WHERE block_time > (SELECT MAX(last_updated) FROM {{this}})) as last_updated -- speedup
+    FROM 
+    {{ ref('nft_mints') }}
+    WHERE block_time > (SELECT MAX(last_updated) FROM {{this}})
+    AND blockchain = 'ethereum'
+    AND currency_symbol IN ('WETH', 'ETH')
+    AND amount_original IS NOT NULL
+    GROUP BY 1, 2, 5 
+
+    UNION ALL 
+
+    SELECT 
+        nft_contract_address,
+        minter,
+        eth_spent,
+        no_minted,
+        (SELECT MAX(block_time) FROM {{ ref('nft_mints') }} WHERE block_time > (SELECT MAX(last_updated) FROM {{this}})) as last_updated -- speedup
+    FROM 
+    {{this}}
+    {% endif %}
+), 
+
+combined as (
+    SELECT 
+        nft_contract_address, 
+        minter, 
+        SUM(eth_spent) as eth_spent, 
+        SUM(no_minted) as no_minted, 
+        last_updated
+    FROM 
+    src_data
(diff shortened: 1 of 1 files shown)


jeff-dude:
> @jeff-dude, I reverted this to the original incremental logic I submitted. I will try and work through the pr to see the changes you made but one of the main issues with this is the issue you highlighted with using `WHERE block_time > SELECT MAX(last_updated) from {{this}]`.
> 
> But this looks like the best way to avoid duplicates in the aggregated values and also not have to go through the entire nft.trades every time the spell is ran.
> 
> An alternative would be to aggregate this by block_date and then do the sum total through dune.com instead of at the spell level.
> 
> Please lmk what you think ser @0xBoxer.
> 
> This is the same issue I'm having with what I plan to update wallet_pnl to, so I'd wait for this to be resolved before starting to work on that.

when we aggregate the data, it's not as big a concern to read the full `nft.trades` each time. of course, it's good to cut down where we can, but spark is optimized to run the group by/aggregate queries. can you write out in words (or visuals) what you expect the output to be, that way i can help further? are you saying the combo of each distinct `minter`/`nft_contract_address` should never be overwritten, as in the first time it's ever read should be the final result? and if the data comes through again (since we read back one week each time), ignore it? if so, can we use left anti join on `{{this}}` to ensure we don't rewrite/overwrite data that already got written?

henrystats:
Hmm, so the issue with using that left anti join is that we're joining with nft_contract_address & wallet_address so there won't be any way to filter for transactions that already got added.

But here's what the table is supposed to do:

* Return an aggregated sum of eth_spent to mint an nft, the number of NFTs the address has minted. 
* So for example, whenever the spell runs, I'd be able to get how many BAYC `0xabc` has minted and the number of eth they spent to mint those tokens.
* The idea with the incremental logic is that, on the first run, the spell returns the current number of BAYC NFTs minted by `0xabc` and the eth spent to mint them. But after every incremental run, the previous table (already aggregated) is added to any new mints. So If from previous run, wallets mints 5 more BAYC & spends 50 more ETH, these will be added to it's previous sum (hence the use of {{this}} so we don't have to go back through the whole table).


I'm also suggesting using a UNION ALL (instead of a join) because of new wallets so they don't get missed out.

jeff-dude:
> @jeff-dude Hmm, so the issue with using that left anti join is that we're joining with nft_contract_address & wallet_address so there won't be any way to filter for transactions that already got added.
> 
> But here's what the table is supposed to do:
> 
> * Return an aggregated sum of eth_spent to mint an nft, the number of NFTs the address has minted.
> * So for example, whenever the spell runs, I'd be able to get how many BAYC `0xabc` has minted and the number of eth they spent to mint those tokens.
> * The idea with the incremental logic is that, on the first run, the spell returns the current number of BAYC NFTs minted by `0xabc` and the eth spent to mint them. But after every incremental run, the previous table (already aggregated) is added to any new mints. So If from previous run, wallets mints 5 more BAYC & spends 50 more ETH, these will be added to it's previous sum (hence the use of {{this}} so we don't have to go back through the whole table).
> 
> I'm also suggesting using a UNION ALL (instead of a join) because of new wallets so they don't get missed out.

okay, so each address/wallet can mint x number of BAYC at any time and we want to ensure they get added to any previous rows. it's possible that a mint could last a few days, a few weeks, months, etc -- which means the historical run can catch say 70%, but incremental will continue to add until it potentially hits 100% of mints. if that sounds right, i can look again and see what you're thinking and/or propose an approach

jeff-dude:
@henrystats i think this latest commit is the safest approach. it uses our normal incremental timeframe, filters the `nft.mints` table down to those specific wallet/nft_contract_addresses, then does normal aggregates on the entire table

henrystats:
Hola ser 👋 - Apologies for my late reply. 

But in reply to your original messages then that is indeed what we're trying to achieve with this spell. I just took a look at the current approach. 

And yess this is perfectttt! If I understand correctly, the main function of the incremental logic here is that whenever it is run incrementally, it only looks at the minters that make a new interaction with an nft_contract_address within the past week then it scans the nft.trades table just for those wallets.

Since we're partioning on minter & nft_contract_address, any minter not picked up in that 1 week window is left unchanged, and the ones picked up will have their values updated. 

Thank you very much ser!

jeff-dude:
> Hola ser @jeff-dude 👋 - Apologies for my late reply.
> 
> But in reply to your original messages then that is indeed what we're trying to achieve with this spell. I just took a look at the current approach.
> 
> And yess this is perfectttt! If I understand correctly, the main function of the incremental logic here is that whenever it is run incrementally, it only looks at the minters that make a new interaction with an nft_contract_address within the past week then it scans the nft.trades table just for those wallets.
> 
> Since we're partioning on minter & nft_contract_address, any minter not picked up in that 1 week window is left unchanged, and the ones picked up will have their values updated.
> 
> Thank you very much ser!

yep, that's right, the only caveat being partitioning isn't the reason -- it's the unique key. partitions are simply a storage mechanism for more efficient reads. the unique key is what will dictate net new inserts and/or updates to old records. the majority of rows in the target (i.e. the spell) will remain untouched, as they won't have activity within the last week.

jeff-dude:
it ran much quicker locally, i imagine the server here is overloaded again. i'm not too worried on the runtimes.

jeff-dude:
happy to merge if you're all set

jeff-dude (approved):

henrystats:
> yep, that's right, the only caveat being partitioning isn't the reason -- it's the unique key. partitions are simply a storage mechanism for more efficient reads. the unique key is what will dictate net new inserts and/or updates to old records. the majority of rows in the target (i.e. the spell) will remain untouched, as they won't have activity within the last week.

ahh I see, thank you ser!

henrystats:
> happy to merge if you're all set

yes ser, looks good to me. I'll open a similar pr to update nft_wallet_pnl as well as it's very similar logic needed.

PR #2700, opened 2023-02-15
add solana fungible tokens

andrewhong5297:
Brief comments on the purpose of your changes: adding some of the top/active fungible tokens from Solana. We should find a way to somehow source token pricing data for Solana into `prices.usd` table too.

**For Dune Engine V2**

I've checked that:

### General checks:
* [x ] if adding a new model, I added a test
* [x ] the filename is unique and ends with .sql
* [x ] each sql file is a select statement and has only one view, table or function defined  
* [ x] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ x] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [ x] `coin_id` represents the ID of the coin on coinpaprika.com
* [ x] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:
* [x ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:
* [x ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [x ] where block_time >= date_trunc("day", now() - interval '1 week')
* [x ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [x ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')


andrewhong5297:
not sure how to expose this in Dune SQL, but that's a seperate problem. This will help with the release of solana decoding tables. I still need to figure out a way to add NFT collections for Solana as well in a `tokens.nonfungible` solana model. It will also have to be a new one, as it likely won't fit in the traditional schema.

jeff-dude:
pushed some minor formatting / naming standards
- table will now be `tokens_solana.fungible` rather than `tokens_solana_fungible.fungible`
- added trim() function since it's manual addition, helps avoid human error
- renamed schema file to be more generic, in case other models get added in directory (i.e. non-fungible)

jeff-dude:
other blockchains require lower address values -- should solana remain as-is, no functions to lowercase?

andrewhong5297:
> other blockchains require lower address values -- should solana remain as-is, no functions to lowercase?

Thanks for edits! We store solana programs/addresses in our raw/decoded tables as mixed case, so probably makes sense to keep these mixed case too so nothing needs to be changed on joins.

jeff-dude (approved):

PR #2711, opened 2023-02-16
Hide DuneSQL not compatible spells

belen-pruvost:
Hiding some spells that are not DuneSQL compatible. 

They are tracked so at some point we can work together to fix them and bring them back.


hildobby:
Hey @belen-pruvost, I'm just curious here, what would make a spell incompatible with DuneSQL? That way I can ensure to not make those mistakes :)

jeff-dude (approved):

belen-pruvost:
> Hey @belen-pruvost, I'm just curious here, what would make a spell incompatible with DuneSQL? That way I can ensure to not make those mistakes :)

Hey @hildobby 🙌 thank you for asking! We are going to be launching a GH check in a few weeks to help wizards understand what's compatible and what isn't, whenever they open a new PR to Spellbook. 

In the meantime, [here](https://dune.com/docs/reference/dune-v2/query-engine/#syntax-comparison) is a summary on how the syntax differs. Bear in mind that spells are still built using Spark SQL, so they need to be Spark SQL compatible... but as DuneSQL compatible as possible. Most of the time that means making sure there's explicit casting. 

On the spells hidden by this PR, these are some of the issues:
- `balances_ethereum.genesis_balances` has numeric literals which are too big for DuneSQL. We have numeric types like int256 and uint256 available where these numbers would fit, but they don't exist on Spark so we are stuck until we migrate how spells are built.
- `nomad_ethereum.view_bridge_transactions` is doing comparisons on fields of different types, which DuneSQL can't handle (e.g. `p1.minute >= '2022-01-01'`). These one may not be tricky to fix, and I could do it eventually. For the sake of time + priorities, I decided to hide it for now.
- `perpetual.trades` is doing a union on columns with different types. The solution to this is usually casting them explicitly to match the same type. Of course it implies making sure nothing breaks downstream.
- `balancer_v2_ethereum.liquidity`, `balancer_v2_optimism.liquidity`, `balancer_v2_polygon.liquidity` and `balancer_v2_arbitrum.liquidity` are all using `LEAD(DAY, 1, NOW())` which is not compatible in terms of syntax (see [Trino docs](https://trino.io/docs/current/functions/window.html))
- `cow_protocol_gnosis.solvers` is doing comparisons on fields of different types
- `cow_protocol_ethereum.price_improvement` and `cow_protocol_ethereum.referrals` rely on a field from `cow_protocol_ethereum.app_data` called `slippage_bips` defined to be numeric, which is not a compatible type with DuneSQL. The solution would be to cast it to decimal and make sure it does not generate issues downstream.

We have other spells which have also been hidden for DuneSQL. You can find them by searching which ones are using the `expose_spells_hide_trino` macro.

PR #2713, opened 2023-02-16
Explicit casting on tessera models for DuneSQL compatibility

belen-pruvost:


belen-pruvost:
> i gotta ask our favorite question 🙂 did we check downstream models which read from this one, to ensure data type compatability?

Hey Jeff! The resulting type of the column should not change here as we were already getting a double (see screenshots), so I did not test the downstream models as nothing would really change. This is just an intermediate explicit casting for the divisions to work.
Screenshot 2023-02-17 at 08 00 53

Screenshot 2023-02-17 at 08 01 58

PR #2717, opened 2023-02-16
[labels] downstream impact fixes from redesign

jeff-dude:
looking for fixes on downstream models which read from labels. issues seem to be narrowed down to filtering the blockchain column


andrewhong5297:
good catch!

andrewhong5297 (approved):

PR #2726, opened 2023-02-17
enhance the PR CI test GH action

jeff-dude:
the goal:
- don't run CI test on `main` after merge, as we only merge if passing on feature branch -- this will also save compute resources for other PRs
- don't run CI test when syncing feature branch with main -- most commonly done via the new button on PRs to keep up-to-date with main branch prior to merge. we should have a passing CI test previously, then if another PR is merged, we're asked to updated to main again, therefore kicking off CI test again, taking up compute resources for other PRs

cc @duneanalytics/team-abstract

### DIFF
--- a/.github/workflows/dbt_slim_ci.yml
+++ b/.github/workflows/dbt_slim_ci.yml
@@ -4,15 +4,14 @@ name: dbt slim ci (in beta)
 on:
   workflow_dispatch:
   pull_request:
+    branches-ignore:    
+      - 'main'
     paths-ignore:
       - 'deprecated-dune-v1-abstractions/**'
       - 'scripts/**'
       - 'Pipfile'
       - '.gitignore'
       - '.github/**'
-  push:
-    branches:
-      - 'main'
 
 concurrency:
   group:  ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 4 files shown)


dot2dotseurat:
I think we want the first case but not the second.

tests can fail because the branch is not up to date with main and that’s why we added the button for syncing to main.

jeff-dude:
> I think we want the first case but not the second.
> 
> tests can fail because the branch is not up to date with main and that’s why we added the button for syncing to main.

second case can get finicky quick. for example:
- review PR1, tests pass, approve PR, no other PRs have been merged into main so the button isn't there, don't merge PR1 quite yet
- review PR2, tests pass, approve PR, no other PRs have been merged into main so the button isn't there, don't merge PR2 quite yet
- merge PR1
- revisit PR2, now the button pops up, click to sync with main from changes on PR1 (even if unrelated completely), CI tests run again after sync with main
- wait for success and then merge PR2

while i outline that, i'm still okay removing the second case to be safe. the workarounds:
- don't sync with main, bypass and merge, avoid the tests again
- just be patient and let it ride

dot2dotseurat:
Yeah, I think bypassing is the better solution imho.

jeff-dude:
> Yeah, I think bypassing is the better solution imho.

that's fine with me. happy to merge if you approve it 🙏

dot2dotseurat (approved):

PR #2727, opened 2023-02-17
fix wallet pnl - Ready for review 

henrystats:
rewriting this so it's similar to nft_top_minters. 

cc @jeff-dude, this is ready for review. I checked on dune.com with the test schema as well to make sure the results are what is expected.

### DIFF
--- a/models/nft/ethereum/nft_ethereum_wallet_pnl.sql
+++ b/models/nft/ethereum/nft_ethereum_wallet_pnl.sql
@@ -11,11 +11,8 @@
     )
 }}
 
-WITH 
-
 {% if is_incremental() %}
-
-weekly_unique_wallet_address as 
+WITH weekly_unique_wallet_address as 
 (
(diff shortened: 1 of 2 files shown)


models/nft/ethereum/nft_ethereum_wallet_pnl.sql
```diff
@@ -11,65 +11,98 @@
...
 }}
 
-WITH trades as
+WITH 
+
+{% if is_incremental() %}
+
+weekly_unique_wallet_address as 
+(
+    SELECT DISTINCT
```
jeff-dude, line 20: am i crazy or are these two queries being union'ed exactly the same?

henrystats: haha no you're not ser. Just fixed, should have been **seller**

jeff-dude (approved):

PR #2559, opened 2023-01-25
[Easy] make trader platform spell crosschain -  Ready for Review

henrystats:
@gentrexha this is similar to pr: https://github.com/duneanalytics/spellbook/pull/2555

This pr makes the trader diversity spell crosschain

PR #2728, opened 2023-02-17
[Easy] make trader platform labels crosschain - Ready for review

henrystats:
previous pr was merged https://github.com/duneanalytics/spellbook/pull/2559 but got dropped in the relaunch. 

Resubmitting to make crosschain again.


models/labels/dex/Persona/trader_platforms/labels_trader_platforms.sql
```diff
@@ -1,8 +1,48 @@
...
+                    "labels",
+                    \'["gentrexha", "Henrystats"]\') }}'
     )
 }}
 
-SELECT * FROM {{ ref('labels_trader_platforms_ethereum') }}
\ No newline at end of file
+with trader_platforms as (
+    select taker           as address,
+           MIN(block_time) as first_trade,
```
Hosuke, line 13: It seems `first_trade` column is not used in the later select part.
We may remove the `order by first_trade`, though it might be a useful feature.

henrystats: I think it was intentionally used but will defer to @gentrexha on that.

gentrexha: The `order by first_trade` keeps the order inside of the label on which platform they've used first.

Hosuke: I see, so the platforms' name will be ranked in order.

Hosuke (approved):
LGTM.✅
Thank you @henrystats @gentrexha 

[Sample trader platform labels](https://dune.com/queries/2042331?d=1)

PR #2734, opened 2023-02-17
[Quick] update tofu arbitrum events unique trade id - Ready for review

henrystats:
@jeff-dude in tofu_arbitrum_events, I used the unique keys you mentioned we were moving to instead of a combination of block_date & unique_trade_id. 

So I left the unique_trade_id column as '' since it was still a column in the nft_trades_table. 

I saw earlier that Hildobby submitted a pr to make nft_wash_trades cross-chain and the spell relies on the unique_trade_id column in nft.trades so doing a quick fix to update it from '' to the usual concat.

I didn't update the unique keys from the previous one since that's what we'd use going forward.


henrystats:
uhm weird, this is similar to the nft_transfers_fantom error.

jeff-dude:
> uhm weird, this is similar to the nft_transfers_fantom error.

it's looking for the seed file. since this PR doesn't include the seed, it doesn't build it in the PR, then looks to find in prod but it hasn't been built yet. make a dummy change to the seed(s) to force it here

henrystats:
woahhhhh. Okaiee looking through now ser thank you!

henrystats:
ran now, thank you ser @jeff-dude

0xRobin (approved):

PR #2735, opened 2023-02-18
Batch fix old generic test in dex.trades

Hosuke:
Use `check_dex_seed` test to replace old generic test for the following projects:

   - curvefi_ethereum_trades
   - dodo_ethereum_trades
   - bancor_ethereum_trades
   - mstable_ethereum_trades

Also remove the cast() in the curvefi rollup spell.


jeff-dude:
thank you for taking this on 🙏 seems like a few failures we need to look into. some could be false positive results, especially the overall dex trade assert freshness test. we should look closer at the seed issues and unique combo failure

Hosuke:
> thank you for taking this on 🙏 seems like a few failures we need to look into. some could be false positive results, especially the overall dex trade assert freshness test. we should look closer at the seed issues and unique combo failure

Yes, there are some seeds for curvefi ethereum I still need to check/fix.

Hosuke:
Curvefi ethereum duped records:
https://dune.com/queries/2025966?d=1

andrewhong5297:
> > Curvefi ethereum duped records: https://dune.com/queries/2025966?d=1
> 
> > 
> 
> > sample Factory V1 Plain dupe details: https://dune.com/queries/2027787?d=1&blockchain=ethereum&category=decoded_project&namespace=curvefi
> 
> > 
> 
> > We can see that all dupes are related to CVX-pxCVX exchange.
> 
> > 
> 
> > There are two `pxCVX/CVX` pools with the same `pool_address` in curvefi_ethereum.view_pools https://dune.com/queries/2027816?d=1&blockchain=ethereum&category=decoded_project&namespace=curvefi
> 
> > 
> 
> > The root cause is there are two records in `ethereum.contracts` for the same address. https://dune.com/queries/2027842
> 
> 
> 
> thank you for the analysis.
> 
> 
> 
> @andrewhong5297 raising awareness here on curve ethereum issue that hosuke has proposed a fix for. you can see here: https://github.com/duneanalytics/spellbook/pull/2735/files#diff-1fed5c5f9cee8d203eeca71baf0445beaa7ea16f5f0419bc22f2a668b0d574ea
> 
> 
> 
> it's not exactly the original intention of this PR, but came up in testing :)

Ah yeah good find! Should prob use labels instead since i think those are unique. Thanks Hosuke!

jeff-dude (approved):

PR #2747, opened 2023-02-21
Add NFTEarth NFT to Optimism 

chuxinh:
Add NFTEarth NFT events and trades to Optimism and `nft.trades` etc.

It's a fork of seaport contract but some of the logic deviates from the one here: https://github.com/duneanalytics/spellbook/tree/main/models/seaport/optimism

Specifically
- logic to determine `is_platform_fee` or `is_creator_fee`
- adding transfer of erc1155 to fill in missing seller 
- logic for `eth_erc_idx` 

Let me know if there's any concern to the logic above, specifically if that can be applied flexibly to future trades for this contract.

Some test `tx_hash` 
```
0x983c231e80249610f35c81037850305d0e9b11db12e646320b0e0c16c3485d84
0x49b404f371734ead979e4949882d2de9ba51be3403758314d3627ad859bda606
0xffa3f07b9f9cdabe7ec226c5dfc9c613f812bfa4f98874b2d49ff2039b8a70f3
0xceaacb2199a3e007fc181a430becc749d78cf91beb3dc084755705b9acab5a7c
0x139c44b3f1c1bf2c2dc624a111f0b3de984a9229543bbaf562776a1ca12c15db
0x9763544da8232f02c1dfd5da76a633e8ad5842f99f1e7aaeab6279180988c89
```

Also a quick change to `quix` schema

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


0xRobin:
Hey @chuxinh, I'm trying to understand the differences between the fork and the original Seaport deploy.
Do we have a view on the actual code diff between the contracts?

Ideally we make it very clear in the model where and why different logic is used. Do you think it's possible to keep the original base_pairs logic and put any modifications in the events model (not saying that's a must, just asking if it would be feasible given the contract differences)?

My main concern here is that the current seaport logic will probably change in the future because the current implementation is not accurate enough. We'd want to be able to port any modifications to this model.

chuxinh:
> Hey @chuxinh, I'm trying to understand the differences between the fork and the original Seaport deploy. Do we have a view on the actual code diff between the contracts?
> 
> Ideally we make it very clear in the model where and why different logic is used. Do you think it's possible to keep the original base_pairs logic and put any modifications in the events model (not saying that's a must, just asking if it would be feasible given the contract differences)?
> 
> My main concern here is that the current seaport logic will probably change in the future because the current implementation is not accurate enough. We'd want to be able to port any modifications to this model.

Thanks for looking into it @0xRobin. I don't have a view on the difference in code (probably can ask and reach out to their team), but when I directly fork the current seaport logic to NFTEarth (e.g. base pair), the results are not aligned with the actual trade info onchain. Maybe for your understanding as well, is it better for me to use the original base pair on some of the tx_hash above to demonstrate the difference and see how to move from there?

chuxinh:
Quickly pulled through the data for this PR version of base_pair and the original seaport for NFTEarth data here:
- This PR: https://dune.com/queries/2040719?d=1
- Original: https://dune.com/queries/2040744

You can search for some of the tx hashes above and see issues related to how the order of `platform_fee` and `creator_fee` using `eth_erc_idx` does not work for those transactions in the original version. As well as the creator fee idx. 

Would definitely like a better solution if you have any suggestion!

0xRobin:
@chuxinh ok thanks for the extra context.
Seems like some of the ordering assumptions in the original seaport/opensea abstractions don't hold here. 
My guess is the seaport contract they deployed is identical, but they just setup the orders differently (which is fine).

Ideally we should have in the future:
- a seaport.base_trades model purged from any marketplace specific assumptions
- an opensea.trades model built on top of the base logic with opensea specific logic to determine payment/royalty/fees
- an NFTEarth.trades model built on top of the base logic with NFTEarth specific logic to determine payment/royalty/fees

@hildobby is working towards this setup, so I'd say we can merge this now in the current state. 
Keeping in mind we should probably adapt it whenever we have a better seaport/opensea setup.

0xRobin (approved):

jeff-dude (approved):

PR #2748, opened 2023-02-21
[tx_hash label] [Easy] Added new columns and fixed query in tx_hash label

gentrexha:
I've added some of the new label columns to the tx_hash spell. In retrospect, it would've been a good idea to discuss this label as part of the [Relaunching Labels discussion](https://github.com/duneanalytics/spellbook/discussions/2634) and discussing the possibility of this being a separate category of labels.

PR #2749, opened 2023-02-21
[tx_hash label] [Easy] Add bluechip_investment tx_hash label

gentrexha:

Implement the blue-chip investment tx hash label

**For Dune Engine V2**

I've checked that:

### General checks:


models/cow_protocol/tx_hash_labels/bluechip_investment/cow_protocol_tx_hash_labels_bluechip_investment.sql
```diff
@@ -0,0 +1,8 @@
+{{
+    config(
+        alias='tx_hash_labels_bluechip_investment',
+        post_hook='{{ expose_spells_hide_trino(\'["ethereum"]\', "sector", "tx_hash_labels", \'["gentrexha"]\') }}'
+    )
+}}
+
+SELECT * FROM {{ ref('cow_protocol_tx_hash_labels_bluechip_investment_ethereum') }}
```
bh2smith, line 8: nit:
```suggestion
SELECT * FROM {{ ref('cow_protocol_tx_hash_labels_bluechip_investment_ethereum') }}

```

bh2smith (approved):
LGTM!

models/cow_protocol/tx_hash_labels/bluechip_investment/cow_protocol_tx_hash_labels_bluechip_investment.sql
gentrexha: I can't seem to spot the difference between the two. Can you elaborate on that?

gentrexha:
This will be fixed once #2748 is merged.

models/cow_protocol/tx_hash_labels/bluechip_investment/cow_protocol_tx_hash_labels_bluechip_investment.sql
bh2smith: New line at EOF!

models/cow_protocol/tx_hash_labels/bluechip_investment/ethereum/cow_protocol_tx_hash_labels_bluechip_investment_ethereum.sql
```diff
@@ -0,0 +1,49 @@
...
+with
+ bluechips as (
+    select
+        bluechip_address
+    from
+        (
+            VALUES  ('0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2'),
+                    ('0x2260fac5e5542a773aa44fbcfedf7c193bc2c599'),
+                    ('0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee')
+        ) t(bluechip_address)
```
bh2smith, line 16: Could you add comment strings identifying the token symbols here?

bh2smith (approved):

jeff-dude:
@gentrexha i think we'll need to wait for prod to refresh with #2748 being merged. should be within the hour.

jeff-dude:
we're past the merge issues, but now have failure on unique test [here](https://github.com/duneanalytics/spellbook/blob/main/models/cow_protocol/tx_hash_labels/cow_protocol_tx_hash_labels_all_schema.yml#L6) -- please do investigate and fix. it'd also be better to move these unique test to the base spells which contain the logic, rather than overall view, to catch it sooner.

gentrexha:
> we're past the merge issues, but now have failure on unique test [here](https://github.com/duneanalytics/spellbook/blob/main/models/cow_protocol/tx_hash_labels/cow_protocol_tx_hash_labels_all_schema.yml#L6) -- please do investigate and fix. it'd also be better to move these unique test to the base spells which contain the logic, rather than overall view, to catch it sooner.

Thanks for taking a look into these.

I've removed the test due to it not being valid anymore because we are okay with the same tx_hash showing across multiple different projects/trades with CoWs. cc @bh2smith The thing we discussed where the same tx_hash being shown twice started happening.

PR #2782, opened 2023-02-22
T-25470 - Fix ironbank_ethereum

belen-pruvost:

### DIFF
--- a/models/ironbank/ethereum/ironbank_ethereum_borrow.sql
+++ b/models/ironbank/ethereum/ironbank_ethereum_borrow.sql
@@ -11,7 +11,7 @@ b.evt_block_number AS block_number,
 b.evt_block_time AS block_time,
 b.evt_tx_hash AS tx_hash,
 b.evt_index AS `index`,
-b.contract_address,
+CAST(b.contract_address AS VARCHAR(100)) AS contract_address,
 b.borrower,
 i.symbol,
 i.underlying_symbol,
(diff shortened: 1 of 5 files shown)


belen-pruvost:
I would also like to fix `ironbank_ethereum.redeem_underlying` but I can't find its definition on spellbook, how can I proceed? 🤔

dot2dotseurat:
@belen-pruvost can you tell where the table comes from based on the metadata? I also don't see it in spellbook.

belen-pruvost:
> @belen-pruvost can you tell where the table comes from based on the metadata? I also don't see it in spellbook.

@dot2dotseurat It seems to come from spellbook based on the table properties but I don't see it 🤷‍♀️ see [here](https://dbc-7ee55e77-eabe.cloud.databricks.com/sql/explore/data/hive_metastore/ironbank_ethereum/redeem_underlying?o=4190327660257548)
Screenshot 2023-02-22 at 15 46 40

dot2dotseurat:
hmmm. I wonder if it was renamed or removed. Let me see if there's a good way to search history on github.

belen-pruvost:
@dot2dotseurat found it! It redeem was called redeem_underlying before this change: https://github.com/duneanalytics/spellbook/commit/36e981f60559a1d5c250e1c6eef283f64dfba491#diff-e2ae353ea70cf7252d068e50fbc981a331841e341e563aaf001d10d3daf5c789, shouldn't the old alias view be removed?

dot2dotseurat:
Yes, I agree, let's remove the old alias.

jeff-dude:
this is high on my list of automation opportunities for new engineer(s) to help. find and remove any spells in the metastore/lake which no longer exist in main branch. we've had to manually remove for customers a few times (those who notice/appear in explorer)

PR #2784, opened 2023-02-22
T-25470 - Fix ironbank_optimism

belen-pruvost:


belen-pruvost:
I would also like to fix ironbank_optimism.redeem_underlying but I can't find its definition on spellbook, how can I proceed? 🤔

dot2dotseurat (approved):

PR #2786, opened 2023-02-22
Make dex_aggregator.trades DuneSQL compatible

belen-pruvost:
This PR aims to fix dev_aggregator.trades to be compatible with DuneSQL and it is unrelated to the rest of the PRs in the batch which were linked to data type changes.

Here the columns `atoms_sold` is changed from double to decimal(38,0),  and `atoms_bought` is changed from string to decimal(38,0) so that they can be correctly unioned on `dex_aggregator.trades`.

Full disclaimer: I did not fully test this, I need a refresher on how to do this (it's been a while). What I did do: I checked if other models were referring to these modified columns [here](https://dune.com/spellbook#!/model/model.spellbook.cow_protocol_ethereum_trades) and [here](https://dune.com/spellbook#!/model/model.spellbook.cow_protocol_gnosis_trades) and I didn't find any.

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
@@ -184,9 +184,9 @@ valued_trades as (
                  else concat(buy_token, '-', sell_token)
                end as token_pair,
            units_sold,
-           atoms_sold,
+           CAST(atoms_sold AS DECIMAL(38,0)) AS atoms_sold,
            units_bought,
-           atoms_bought,
+           CAST(atoms_bought AS DECIMAL(38,0)) AS atoms_bought,
            (CASE
                 WHEN sell_price IS NOT NULL THEN
                     -- Choose the larger of two prices when both not null.
--- a/models/cow_protocol/gnosis/cow_protocol_gnosis_trades.sql
+++ b/models/cow_protocol/gnosis/cow_protocol_gnosis_trades.sql
@@ -167,9 +167,9 @@ valued_trades as (
                  else concat(buy_token, '-', sell_token)
                end as token_pair,
            units_sold,
-           atoms_sold,
+           CAST(atoms_sold AS DECIMAL(38,0)) AS atoms_sold,
            units_bought,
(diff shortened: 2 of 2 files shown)


jeff-dude:
@belen-pruvost this shouldn't affect `dex_aggregator.trades` downstream, as the column is specific to cow team usage on upstream models.

however, if we run:
`dbt ls --resource-type model --output name --select cow_protocol_ethereum_trades+`

we can see some other models to check, which read from the model changed here (same can be done for gnosis, but we'll keep it clean here):
> cow_protocol_batches
> cow_protocol_ethereum_batches
> cow_protocol_ethereum_price_improvement
> cow_protocol_ethereum_referrals
> cow_protocol_ethereum_trades
> cow_protocol_trades
> cow_protocol_tx_hash_labels_all
> cow_protocol_tx_hash_labels_stable_to_stable
> cow_protocol_tx_hash_labels_stable_to_stable_ethereum
> dex_aggregator_trades
> labels_all
> labels_arbitrage_traders
> labels_arbitrage_traders_ethereum
> labels_average_trade_values
> labels_average_trade_values_ethereum
> labels_dex
> labels_dex_aggregator_traders
> labels_dex_aggregator_traders_ethereum
> labels_trader_age
> labels_trader_dex_diversity
> labels_trader_dex_diversity_ethereum
> labels_trader_frequencies
> labels_trader_platforms
> labels_trader_platforms_ethereum
> labels_trader_portfolios
> labels_trader_portfolios_ethereum


most of those can be safely ignored, but i'd want cow team to confirm a few aren't affected, as it appears `atom` fields are used in calculations in models such as `cow_protocol_ethereum_price_improvement`

maybe @bh2smith can help approve?

belen-pruvost:
Makes sense - thanks for checking @jeff-dude ! 🙏

bh2smith:
I'm the past this was a string and somehow arithmetic magically worked. I think it should be fine to go through with it as there is no place I would expect there to be a use case of this field that is not with integer as the intention.

jeff-dude (approved):
thank you both 🤝

PR #2813, opened 2023-02-24
Try to fix trove_arbitrum_trades

Hosuke:
Not sure the exact cause for the issue:
https://dune.com/queries/2042429?d=11&category=abstraction
https://discord.com/channels/757637422384283659/999683200563564655/1078530354194960535
image

The model works fine for spark sql though.

### DIFF
--- a/models/trove/arbitrum/trove_arbitrum_trades.sql
+++ b/models/trove/arbitrum/trove_arbitrum_trades.sql
@@ -27,14 +27,30 @@ with marketplace as (
         evt_tx_hash as tx_hash,
         evt_block_number as block_number
     from (
-        select evt_block_time, tokenId, quantity, seller, pricePerItem,
-            paymentToken, nftAddress, evt_tx_hash, evt_block_number,
-            contract_address, bidder as buyer
+        select evt_block_time,
+               tokenId,
+               quantity,
+               seller,
+               pricePerItem,
+               paymentToken,
+               nftAddress,
+               evt_tx_hash,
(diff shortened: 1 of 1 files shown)


models/trove/arbitrum/trove_arbitrum_trades.sql
```diff
@@ -63,12 +79,14 @@ select
     project_contract_address,
     cast(null as varchar(5)) as aggregator_name,
     cast(null as varchar(5)) as aggregator_address,
-    tx_hash,
+    mp.tx_hash,
     mp.block_number,
     tx.from as tx_from,
```
belen-pruvost, line 84: Hey @Hosuke! `from` is a reserved word on DuneSQL so it needs to be escaped when it's used for selecting a column. Changing this to be tx.`from` should do the trick 🤞

belen-pruvost (changes requested):

models/trove/arbitrum/trove_arbitrum_trades.sql
Hosuke: like this?
```suggestion
     tx.`from` as tx_from,
```

belen-pruvost: Exactly 👍

belen-pruvost (approved):
This looks good to me! I'll tag the team that owns Spellbook to give it the final thumbs up + merge

0xRobin (approved):
thank you!

PR #2814, opened 2023-02-24
Added aavegotchi polygon nft trades

springzh:

This PR is for adding Aavegotchi marketplace on Polygon to nft.trades.

Notes:
1. There looks like be a 2% sent to 0x385eeac5cb85a38a9a07a70c73e0a3271cfb54a7 and 1% sent to Aavegotchi: DAO at 0xb208f8bb431f580cc4b216826affb128cd1431ab . I use 0x385eeac5cb85a38a9a07a70c73e0a3271cfb54a7 as the fees.

2. Mint events added without calculating the amount for each token due complexity. There can be multiple tokens minted in single transaction. https://polygonscan.com/tx/0x14006c44d86c0a48cbc823b95b09d77cc40bc116cce62691c2de5243a379be5a . So I handled it the same way as here: https://github.com/duneanalytics/spellbook/blob/main/models/nftb/bnb/nftb_bnb_events.sql#L37 . Which is set the mint to be 0.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


models/aavegotchi/polygon/aavegotchi_polygon_events.sql
```diff
@@ -0,0 +1,190 @@
...
+    2 * amount_raw / 100 AS platform_fee_amount_raw,
+    2 * amount_raw / power(10, coalesce(p.decimals, gp.decimals)) / 100 AS platform_fee_amount,
+    2 * amount_raw / power(10, coalesce(p.decimals, gp.decimals)) * coalesce(p.price, gp.price) / 100 AS platform_fee_amount_usd,
+    CAST(2 AS DOUBLE) AS platform_fee_percentage, -- Treasury 0xd4151c984e6cf33e04ffaaf06c3374b2926ecc64 receive 2%
+    0 AS royalty_fee_amount_raw,
+    0 AS royalty_fee_amount,
+    0 AS royalty_fee_amount_usd,
+    0 AS royalty_fee_percentage,
+    NULL::double AS royalty_fee_receive_address,
+    NULL::string AS royalty_fee_currency_symbol,
```
Hosuke, line 180: We can rewrite `::` to `CAST()`

models/aavegotchi/polygon/aavegotchi_polygon_events.sql
```diff
@@ -0,0 +1,190 @@
...
+        `to` AS buyer,
+        NULL AS seller,
+        contract_address AS nft_contract_address,
+        token_id,
+        amount AS number_of_items,
+        token_standard,
+        '0x385eeac5cb85a38a9a07a70c73e0a3271cfb54a7' AS currency_contract, -- All sale are in GHST
+        CAST(0 as DECIMAL(38,0)) AS amount_raw, -- It's hard to get the mint price. So handle it similar as in nftb_bnb_events
+        NULL AS category,
+        NULL AS executed_time
```
Hosuke, line 43: We can cast null to their expected types

models/aavegotchi/polygon/aavegotchi_polygon_events.sql
springzh: Updated.

models/aavegotchi/polygon/aavegotchi_polygon_events.sql
springzh: Updated.

0xRobin (approved):
Hey @springzh, I've mostly simplified this during my review. 
1. I removed the mints since we could not get the prices and they were sourced from nft transfers which can be easily done by a user in a query. (This also reduces the complexity of the query a lot) 
If you have a hard requirements for the mints table, we should include the logic as a separate model so it doesn't impact the trades model complexity.
2. I've removed the backwards price fill, I think we prefer to have no historical price if we don't have the accurate price as it might otherwise incorrectly inflate USD volume.

Also as a side note, we've stepped away from listing all contributors in the top level models (nft.events etc).

models/aavegotchi/polygon/aavegotchi_polygon_fees.sql
```diff
@@ -0,0 +1,43 @@
...
+CAST(platform_fee_percentage AS DOUBLE) AS platform_fee_percentage,
+CAST(royalty_fee_amount_raw AS DOUBLE) AS royalty_fee_amount_raw,
+royalty_fee_amount,
+royalty_fee_amount_usd,
+CAST(royalty_fee_percentage AS DOUBLE) AS royalty_fee_percentage,
+royalty_fee_receive_address,
+royalty_fee_currency_symbol,
+token_standard,
+trade_type,
+CAST(number_of_items AS DECIMAL(38,0)) AS number_of_items,
```
jeff-dude, line 28: i notice this is casted as int in `events` then decimal here (which is correct type). please move this to `events` and override the int casting

models/aavegotchi/polygon/aavegotchi_polygon_schema.yml
```diff
@@ -0,0 +1,213 @@
+version: 2
+
+models:
+  - name: aavegotchi_polygon_events
```
jeff-dude, line 4: where is the unique test here? for incremental models, we'd like to see an unique test run on the unique key columns of the model. this can be either at column level (if single column is unique) or with dbt unique combo of columns (found in many other model schemas)

jeff-dude (changes requested):

models/aavegotchi/polygon/aavegotchi_polygon_schema.yml
springzh: Added to the unique_trade_id field.

models/aavegotchi/polygon/aavegotchi_polygon_fees.sql
springzh: Updated.

models/aavegotchi/polygon/aavegotchi_polygon_fees.sql
```diff
@@ -0,0 +1,43 @@
...
+                                    \'["springzh"]\') }}')
+}}
+
+SELECT blockchain,
+project,
+version,
+block_time,
+token_id,
+collection,
+CAST(platform_fee_amount_raw AS DOUBLE) AS platform_fee_amount_raw,
```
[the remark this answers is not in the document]
springzh: Updated.

jeff-dude (approved):

PR #2824, opened 2023-02-24
T-25678 prices file redesign: segment by blockchain

jeff-dude:


Hosuke (approved):
LGTM.✅
I added description and accepted_values tests in the last commits.

jeff-dude:
> LGTM.✅ I added description and accepted_values tests in the last commits.

nice addition, much appreciated!

0xRobin (commented):
I assume we'll also need to adapt the token checker CI.

dot2dotseurat:
I agree with @0xRobin, we should update the token checker. Right now, it references this [prices tokens file](https://github.com/duneanalytics/spellbook/blob/f320ea73cdc8811c076d8fb4871b9741a8c9377b/.github/workflows/prices_check.yml#L33) to generate a diff file. You could potentially just check each file but it would be a little brittle in that each new chain would require adding a new step to the GH Action.

jeff-dude:
good catch. i can revisit that soon, won't have time today. if anyone wants to take a look at it, feel free, otherwise i'll look asap

Hosuke:
I tried to make git diff scan all .sql files under models/prices/ directory, but not sure if it will work as expected.

jeff-dude:
> I tried to make git diff scan all .sql files under models/prices/ directory, but not sure if it will work as expected.

hmm, the gh action output shows that it ran across id values from multiple different blockchains. i can't tell if that's only because the final `prices.tokens` is also edited in this PR, which unions them back together, and that is why it's picking them all up.

in future PRs, that file won't be edited (most the time), so we can't rely on that

.github/workflows/prices_check.yml
```diff
@@ -30,7 +30,7 @@ jobs:
       - name: Get prices diff
         shell: bash
         run: |
-          git diff ${{ github.event.pull_request.base.sha}}..${{ github.event.pull_request.head.sha }} models/prices/prices_tokens.sql | grep '^\+ ' > scripts/new_lines.txt
+          git diff ${{ github.event.pull_request.base.sha}}..${{ github.event.pull_request.head.sha }} models/prices/**/*.sql | grep '^\+ ' > scripts/new_lines.txt
```
Hosuke, line 33: I was thinking this line recursively scan all .sql files in the directory, and there is a pipe `|` operation after git diff, so in the `grep` step the diff from all files get merged into one flow/place, and write into a single file `new_lines.txt` by `>` operation.

jeff-dude:
@Hosuke do you mind helping go through all the open PRs which add prices and format to the new standard, as used here?

Hosuke:
> @Hosuke do you mind helping go through all the open PRs which add prices and format to the new standard, as used here?

Will do soon, but probably tomorrow.

jeff-dude:
> > @Hosuke do you mind helping go through all the open PRs which add prices and format to the new standard, as used here?
> 
> Will do soon, but probably tomorrow.

no rush! thank ya

PR #2829, opened 2023-02-27
T-25470 - Fix trove.trades for DuneSQL

belen-pruvost:
No changes on final column types, just intermediate explicit casting for the queries to work on DuneSQL


0xRobin:
@belen-pruvost, these models were not supposed to be views.
I've got a PR coming which makes them incremental models, do we still need these DuneSQL adaptations if the models are materialized?

belen-pruvost:
> @belen-pruvost, these models were not supposed to be views. I've got a PR coming which makes them incremental models, do we still need these DuneSQL adaptations if the models are materialized?

Hey @0xRobin! No, if these are being materialized then we don't need this fix 🙌 I'll close this PR then :)

PR #2840, opened 2023-02-27
T-23515 update dbt_slim_ci.yml with dunesql check step

jmsgu:
Add step to run dunesql check script in the `dbt slim ci (in beta)` github workflow.


jeff-dude:
👀 it has arrived, huh? seems like all the heavy lifting is on the server, then a simple step here?

.github/workflows/dbt_slim_ci.yml
```diff
@@ -59,3 +59,8 @@ jobs:
       - name: dbt test incremental model(s) if applicable
         if: env.INC_MODEL_COUNT > 0
         run: "dbt test --select state:modified,config.materialized:incremental --defer --state ."
+
+      - name: Run DuneSQL Check
+        if: env.INC_MODEL_COUNT > 0
```
jeff-dude, line 64: i don't think this is needed here. i'd imagine you want to run on all spells, regardless of materialization type? tables, views and incremental

.github/workflows/dbt_slim_ci.yml
```diff
@@ -59,3 +59,8 @@ jobs:
...
+        run: "/runner/dunesql_check.py --schema test_schema --pr_schema git_$GIT_SHA"
+        continue-on-error: true
```
jeff-dude, line 66: assume this means it won't fail the CI test, if it fails the dunesql check? that's probably best in the early stages?

jeff-dude: sounds good 👍

jeff-dude (approved):
i'm good with it. we can let @dot2dotseurat give a look as well

jeff-dude:
fyi @0xRobin @Hosuke for future PRs, don't want this to cause confusion. this will be the dunesql compatibility automated test.

.github/workflows/dbt_slim_ci.yml
dot2dotseurat: oh wow I love this solution. 

so much cleaner to use the schema instead of the manifest and I appreciate the continue on error as a conservative way of adding this.

dot2dotseurat (approved):

0xRobin:
> fyi @0xRobin @Hosuke for future PRs, don't want this to cause confusion. this will be the dunesql compatibility automated test.

So we won't consider this test failiures blocking for PRs for now, right?

PR #2846, opened 2023-02-28
dbt_project cleanup: seeds

0xRobin:
This PR cleans up the `dbt_project.yml` file by removing the seed definitions.

Following the dbt docs [[link](https://docs.getdbt.com/reference/resource-configs/column_types#recommendation)] we should only specify column types for seeds where the automatic inference does not work as expected.
I've thus removed all `column_types` definitions and have reinstated those where the inferred type caused the tests to fail.
This mostly happens for big numbers (like token_id's) that are inferred as decimal and loose some precision, so we need to cast and compare them as strings. 

**to-do's**
- [x] fix failing tests
- [x] check and update readme/guidelines

While all the tests are passing, it might still be the case that the new inferred type for a column doesn't match the current column type in production. For those cases we need to manually drop (and purge the data from s3) the seeds to unblock the seed job. If possible we could just drop all seed data and let the job repopulate the tables, that way we don't need to handle specific cases.


0xRobin:
When merged I'll go through the open PRs to adapt them if needed.

dot2dotseurat (approved):

jeff-dude:
> Is there still something blocking this?

not a blocker per se, but we could get failures as you mentioned since dbt seems to have a bugs for "drop and create" seeds. we likely have to go into prod data lake and manually remove all seeds and then rerun.

jeff-dude:
i'll look to do this once we catchup a bit on the regular PRs

jeff-dude:
what i think we need to do:
- write a query to drop all tables in the `test_data` schema (validate only seeds use this schema)
- ensure dropped in both metastore/data lake
- merge this
- run `dbt seed` command to run all seeds

jeff-dude:
> what i think we need to do:
> 
> * write a query to drop all tables in the `test_data` schema (validate only seeds use this schema)
> * ensure dropped in both metastore/data lake
> * merge this
> * run `dbt seed` command to run all seeds

these steps are done & i think we're all rebuilt 👍

PR #2858, opened 2023-03-01
CI: test if all models have a custom schema set.

0xRobin:
CI: test if all models have a custom schema set.
Works as follows:

1. In `dbt_project` we assign a `no_schema` to the top level, this applies to all models.
2. Models are expected to override this schema in their own specific config
3. we use `dbt ls --select config.schema:no_schema` to find any models that forget to do this.

This test will help us to clean up the `dbt_project` file more without the risk that models will forget to define a custom schema. 

As an example here's the test during the migration steps of the `abi` models

example test failure: https://github.com/duneanalytics/spellbook/actions/runs/4304576455/jobs/7505801192
image

example test passing after adding a custom schema:
https://github.com/duneanalytics/spellbook/actions/runs/4304603500/jobs/7505864079
image

### DIFF
--- a/.github/workflows/dbt_precommit_hooks.yml
+++ b/.github/workflows/dbt_precommit_hooks.yml
@@ -36,6 +36,10 @@ jobs:
         run: "dbt deps"
       - name: dbt compile
         run: "dbt compile"
+      - name: check schemas
+        run: |
+          test=$(dbt --quiet --no-print ls --select config.schema:no_schema --output path)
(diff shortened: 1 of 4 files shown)


0xRobin:
The schema change happening in the dbt slim ci job prevents the test from working correctly. 
Moving it to the pre-commit hooks job which does a normal compile.

jeff-dude (approved):
will merge as it's pretty low risk too, thanks for the addition

PR #2873, opened 2023-03-02
[Easy] [tx_hash label] Add fixes for tx_hash labels

gentrexha:

Added a new unique tx_hash key for tx_hash labels since tx_hash only doesn't suffice.

**For Dune Engine V2**

I've checked that:

### General checks:


gentrexha:
cc @jeff-dude @Hosuke 

I have a presentation about these today and would appreciate it if we could merge the fix before that so I can show the correct data. Thanks!

jeff-dude (approved):

PR #2917, opened 2023-03-08
fix cow trace address bug for dunesql

jeff-dude:
fyi @gentrexha @bh2smith

### DIFF
--- a/models/cow_protocol/cow_protocol_trades.sql
+++ b/models/cow_protocol/cow_protocol_trades.sql
@@ -31,7 +31,7 @@ FROM
             tx_hash,
             trader AS tx_from,
             receiver AS tx_to,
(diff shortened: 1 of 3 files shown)


belen-pruvost (approved):
Great! So now... the function is only applied on the materialized tables, therefore making any views built on top compatible, right?

Thank you for this! 🙌

belen-pruvost:
> > Great! So now... the function is only applied on the materialized tables, therefore making any views built on top compatible, right?
> > Thank you for this! 🙌
> 
> yes. my fault for not catching this. in a final push to get 1inch integrated in this spell and have them all compatible 🙏

Don't worry - Grimoire should hopefully make this easier for the future 🙏 🙌

PR #2922, opened 2023-03-08
remove outlier token

dot2dotseurat:


jeff-dude (approved):
leaving this for future reference, if need be. outlier values on 2022/03/18:
image

PR #2923, opened 2023-03-08
patch sha length

dot2dotseurat:


jeff-dude (approved):
could this have implications elsewhere? some random thoughts:
- job to delete gh test tables every 24 hours
- grimoire for dunesql
- ...?

dot2dotseurat:
@jeff-dude no, its a good call out but this only affects how we write the manifest files to the manifest.

PR #2931, opened 2023-03-09
T 26068 abstraction category

dot2dotseurat:
adding another post hook which  add category abstraction to all spells regardless if they are exposed in the data explorer

my recommendation is we do *not* run the full ci on this because it will require running very single spell.


jeff-dude (approved):

jeff-dude:
we'll be good to merge soon 👍

PR #2948, opened 2023-03-13
Make opensea_v3_ethereum.trades DuneSQL compatible

belen-pruvost:
No resulting column types are changed, only explicit casting to varchar is applied to perform in filters and to concat values.


augustog (approved):
thanks for the fix @belen-pruvost !

PR #2883, opened 2023-03-03
Add spell for ERC20 token transfers on Arbitrum

tschubotz:
Brief comments on the purpose of your changes:

- This PR is attempting to add a spell for ERC20 token transfers on Arbitrum.
- I got inspired by the erc20 transferes spell on Ethereum (https://github.com/duneanalytics/spellbook/blob/main/models/transfers/ethereum/erc20/transfers_ethereum_erc20.sql) and created all files accordingly (i.e. most importantly **without** incremental logic. Please let me know in case you prefer something else. 

**For Dune Engine V2**

I've checked that:

### General checks:
* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [x] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [x] if adding a new model, I added a test
* [x] the filename is unique and ends with .sql
* [x] each sql file is a select statement and has only one view, table or function defined  
* [x] column names are `lowercase_snake_cased`
* [x] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [x] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Join logic:
* [x] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:

Not used, see comment at the start of this PR.

PR #2949, opened 2023-03-13
Add spell for ERC20 transfers on OP

tschubotz:

- This PR adds a spell for ERC20 token transfers on Optimism.
- It is inspired/copied from #2883 which in turn followed [transfers_ethereum_erc20.sql](https://github.com/duneanalytics/spellbook/blob/main/models/transfers/ethereum/erc20/transfers_ethereum_erc20.sql)
- **no** incremental logic used since the 2 links above don't use it either.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

doesn't apply.

### Join logic:

### Incremental logic:

not used. see beginning of PR decription.

cc @MSilb7

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -414,6 +414,8 @@ models:
         +schema: transfers_bnb
       arbitrum:
         +schema: transfers_arbitrum
+      optimism:
+        +schema: transfers_optimism
 
     sudoswap:
       +schema: sudoswap
--- a/models/transfers/optimism/erc20/transfers_optimism_erc20.sql
+++ b/models/transfers/optimism/erc20/transfers_optimism_erc20.sql
@@ -1,0 +1,65 @@
+{{ config(materialized='view', alias='erc20',
+        post_hook='{{ expose_spells(\'["optimism"]\',
+                                    "sector",
+                                    "transfers",
+                                    \'["soispoke", "dot2dotseurat", "tschubotz"]\') }}') }}
+
+with
+    sent_transfers as (
+        select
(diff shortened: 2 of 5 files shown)


MSilb7:
What's teh value prop of this table vs existing transfer tables? Is it that the addresses are unified in one field?

tschubotz:
> What's the value prop of this table vs existing transfer tables? Is it that the addresses are unified in one field? (i.e. why would I query this table)

For me it's 2 things:
- It simplifies balance queries over time.
- It adds WETH deposits and withdrawals. Doing that manually in every query caused performance issues in the past for me, on some networks. 

> If so, then should the `value` field be a double or some other numeric value rather than a varchar?

- I wouldn't mind, but considering there's already [transfers_ethereum_erc20.sql](https://github.com/duneanalytics/spellbook/blob/main/models/transfers/ethereum/erc20/transfers_ethereum_erc20.sql) which uses, I went with `varchar` to stay consistent.

jeff-dude:
> > What's the value prop of this table vs existing transfer tables? Is it that the addresses are unified in one field? (i.e. why would I query this table)
> 
> For me it's 2 things:
> 
> * It simplifies balance queries over time.
> * It adds WETH deposits and withdrawals. Doing that manually in every query caused performance issues in the past for me, on some networks.
> 
> > If so, then should the `value` field be a double or some other numeric value rather than a varchar?
> 
> * I wouldn't mind, but considering there's already [transfers_ethereum_erc20.sql](https://github.com/duneanalytics/spellbook/blob/main/models/transfers/ethereum/erc20/transfers_ethereum_erc20.sql) which uses, I went with `varchar` to stay consistent.

once spellbook is running on dunesql (i.e. trino), we can revisit many spells where varchar is used on `value` type fields and leverage the custom data type for unit256

PR #2950, opened 2023-03-13
Add spell for erc20 transfers on Polygon

tschubotz:

- This PR adds a spell for ERC20 token transfers on Polygon.
- This PR is similar to #2949 
- It is inspired/copied from #2883 which in turn followed [transfers_ethereum_erc20.sql](https://github.com/duneanalytics/spellbook/blob/main/models/transfers/ethereum/erc20/transfers_ethereum_erc20.sql)
- **no** incremental logic used since the 2 links above don't use it either.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

doesn't apply.

### Join logic:

### Incremental logic:

not used. see beginning of PR decription.

cc @MSilb7

### DIFF
--- a/models/transfers/polygon/erc20/transfers_polygon_erc20.sql
+++ b/models/transfers/polygon/erc20/transfers_polygon_erc20.sql
@@ -1,0 +1,65 @@
+{{ config(materialized='view', alias='erc20',
+        post_hook='{{ expose_spells(\'["polygon"]\',
+                                    "sector",
+                                    "transfers",
+                                    \'["soispoke", "dot2dotseurat", "tschubotz"]\') }}') }}
+
+with
+    sent_transfers as (
+        select
+            CAST('send' AS VARCHAR(4)) || CAST('-' AS VARCHAR(1)) || CAST(evt_tx_hash AS VARCHAR(100)) || CAST('-' AS VARCHAR(1)) || CAST(evt_index AS VARCHAR(100)) || CAST('-' AS VARCHAR(1)) || CAST(`to` AS VARCHAR(100)) as unique_transfer_id,
+            `to` as wallet_address,
+            contract_address as token_address,
+            evt_block_time,
+            value as amount_raw
+        from
+            {{ source('erc20_polygon', 'evt_transfer') }}
+    )
+
+    ,
+    received_transfers as (
+        select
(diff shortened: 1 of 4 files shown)


jeff-dude:
> Take one step ahead, I may redesign the `transfers_*.erc20` spells to get rid of `unique_transfer_id`. I want to introduce a new field `transfer_type` which can be one of 4 values in [`send`, `receive`, `deposit`, `withdrawn`] and together with `evt_tx_hash`, `evt_index`, `wallet_address` to form unique_key.
> 
> I will try to implement my design in the next commit to this PR.
> 
> https://dune.com/queries/2217266?d=1
> 
> cc. @dot2dotseurat @jeff-dude for review.

just noticed this comment, relates to #2953 where i commented to look into dunesql compatibility. i will revisit and review here, then we can determine how we are managing all these spells which are the same across different blockchains.

jeff-dude:
related PRs: #2953 , #2951 , #2949 , #2952

models/transfers/polygon/erc20/transfers_polygon_erc20.sql
```diff
@@ -0,0 +1,109 @@
...
+        {{ source('erc20_polygon', 'evt_transfer') }}
+    {% if is_incremental() %}
+    where evt_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
+),
+received_transfers as (
+    select 'receive'                          as transfer_type,
+           evt_tx_hash,
+           evt_index,
+           "from"                             as wallet_address,
```
Hosuke, line 32: I am not sure for now we need to use dune sql syntax for spellbook, as the dbt-test still uses spark sql syntax to check, and `"from"` may cause dbt-test failure at the moment.

tschubotz: I used it since @jeff-dude asked to make the query future proof (i.e. compatible with Dune SQL) already [here](https://github.com/duneanalytics/spellbook/pull/2953#issuecomment-1476527976).

jeff-dude: yes -- sorry @Hosuke for not tagging, but we'll try to push views to be compatible on both engines where possible (not always at this time).

jeff-dude (approved):
noticed @Hosuke pushed this one to incremental table vs. a view, where other blockchains were a view. not to push more work, but you could explore doing this on all spells now if you'd like. it's preference, if the view isn't performant. polygon is high data volume, so this may make most sense, but others could be done if this is much more performant on the querying side

Hosuke:
> noticed @Hosuke pushed this one to incremental table vs. a view, where other blockchains were a view. not to push more work, but you could explore doing this on all spells now if you'd like. it's preference, if the view isn't performant. polygon is high data volume, so this may make most sense, but others could be done if this is much more performant on the querying side

I will modify other similar PRs to this one as well.

PR #2751, opened 2023-02-21
OP Token Distributions v1

MSilb7:
Brief comments on the purpose of your changes:

!Keeping this up as a draft to test as I go, and make sure changes to labels and things are implemented.~

Should be good to go now! The contains various address mappings tables and token transfer tagging. The stated project wallets are already in labels, but many other distribution wallets are speculative/guesses for now (we will have a formal reporting process getting started which will help).

Also, once https://github.com/duneanalytics/spellbook/pull/2731 gets merged, we can add transfers to these wallets to known incentive distributions.

Goal: Migrate the data that powers this in to a spell so that queries run faster & it's less of a mess & others can contribute.

Open Q:
- How can we ensure that the "Distributions" model updates whenever address lists are changed?

**For Dune Engine V2**

I've checked that:

### General checks:
* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each sql file is a select statement and has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [ ] `coin_id` represents the ID of the coin on coinpaprika.com
* [ ] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:
* [ ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:
* [ ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ ] where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #2975, opened 2023-03-17
Fix labels project wallets

masquot:
Fix for following bug:
the labels_project_wallets file tries to get the column project_name from addresses_optimism_grants_funding, but this column was removed earlier this week in this PR:
https://github.com/duneanalytics/spellbook/pull/2751/files#diff-ea031d3129a72af68fb2fc6a316d555fefb2dd75f7597ecb01dcf3d2c637735d

solution suggested by @MSilb7

### DIFF
--- a/models/addresses/optimism/addresses_optimism_grants_funding.sql
+++ b/models/addresses/optimism/addresses_optimism_grants_funding.sql
@@ -131,8 +131,9 @@ FROM (VALUES
 SELECT
    category
    , entries.address
-   , coalesce(cast(wallets.project_name as varchar(100)), cast(entries.proposal_name as varchar(100))) as project_name
+   , cast(coalesce(pnm.project_name, entries.proposal_name) as varchar(100)) as project_name
    , cast(entries.proposal_name as varchar(100)) as proposal_name
    , cast(entries.funding_source as varchar(100)) as funding_source
 FROM entries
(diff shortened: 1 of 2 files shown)


dot2dotseurat (approved):

dot2dotseurat:
Oh no. I would have expected this to be caught by CI. I need to look more deeply into why this wasn’t caught.

jeff-dude:
> Oh no. I would have expected this to be caught by CI. I need to look more deeply into why this wasn’t caught.

can you clarify this comment? you expected `labels_project_wallets` file to have failed in CI in previously merged PR? it's downstream from the merged PR files and we don't currently run that in CI (could open up much longer runtimes)

jeff-dude:
> > > Oh no. I would have expected this to be caught by CI. I need to look more deeply into why this wasn’t caught.
> > 
> > 
> > can you clarify this comment? you expected `labels_project_wallets` file to have failed in CI in previously merged PR? it's downstream from the merged PR files and we don't currently run that in CI (could open up much longer runtimes)
> 
> Sorry, I realized that after my comment. But regardless, our CI _should_ be able to prevent this type of error but I do understand why it didn't hear.
> 
> The same way there is a `fast-fail` option, I wish we have a `fast-pass` option for downstream models. e.g. something that checks if there is a syntax error due to a missing column but does not run the entire query.

agreed, we should create a height task to track this potential enhancement in the future

PR #2981, opened 2023-03-20
Fix missing 0x prefix on taker address

augustog:
We found exactly 3 trades on dex.trades that had a misformated taker address (starting with '00' instead of '0x', which breaks some explicit type casts allowing DSQL to treat these as varbinary (=faster and happier).

I applied this fix which seems fairly naive, I understand `topic2` supports binary values larger than an address, so the raw field has many more leading zeroes. Assuming this is true, I just changed the string manipulation to get a correct address hex string.

If any of this is wrong pls flag. cc  @MSilb7 @Hosuke (fyi @jeff-dude)

### DIFF
--- a/models/curvefi/optimism/curvefi_optimism_trades.sql
+++ b/models/curvefi/optimism/curvefi_optimism_trades.sql
@@ -88,7 +88,7 @@ SELECT
             'basic' as pool_type,
             block_time,
             block_number,
-            substring(topic2,25,40) AS taker,
+            concat('0x',substring(topic2,27,40)) AS taker,
             '' AS maker,
             conv(substring(data,3+64*3,64),16,10) as token_bought_amount_raw, --2nd bought
             conv(substring(data,3+64*1,64),16,10) as token_sold_amount_raw, --1st sold


Hosuke (approved):
Seems legit for the `taker` addresses.
https://dune.com/queries/2234208?d=1

augustog:
dumb question: is it expected that the BasicPoolSwap section produces only 3 rows?

dot2dotseurat:
Are we sure this works? Could we write a test? I think we could do something like a query that confirms all taker fields are valid hexadecimals.

dot2dotseurat (changes requested):
smol test?

dot2dotseurat:
I think the dbt equivalent of this would do it.

```
select taker
from dex.trades
where taker not like '0x%'
``` 

```
select taker
from dex.trades
where len(taker) != 42
```

jeff-dude:
please see [this line](https://github.com/duneanalytics/spellbook/blob/main/models/curvefi/ethereum/curvefi_ethereum_trades.sql#L20) from @andrewhong5297 curve query on ethereum -- makes me feel more confident this is the correct approach

augustog:
added a generic test and instantiated it at dex.trades level, open TODO's (I'd say post merge but call me out if not):
- move test from dex.trades to upstream models
- implement test for non-evm addresses

also I made a smol mistake as seen in the failing CI, will fix asap

models/dex/dex_schema.yml
```diff
@@ -53,27 +53,39 @@ models:
       - &token_bought_address
         name: token_bought_address
         description: "Contract address of the token bought"
+        tests:
+          - assert_evm_address
```
dot2dotseurat, line 57: this is gonna be such a useful generic test.

0xRobin: oh yeah this is great!!!

tests/generic/assert_evm_address.sql
```diff
@@ -0,0 +1,14 @@
+{% test assert_evm_address(model, column_name) %}
+
+    with unit_test as
+    (
+        select
+            rlike({{ column_name }}, '0[x][0-9a-f]{40}') as assert_evm_address
```
0xRobin, line 6: Can we include `column_name` as a separate column in the select here? 
That will make it easier to track down any failures.

tests/generic/assert_evm_address.sql
```diff
@@ -0,0 +1,14 @@
...
+        from {{ model }} m
+        where blockchain in ('ethereum', 'optimism', 'arbitrum', 'avalanche_c', 'polygon', 'bnb', 'gnosis', 'fantom') -- TODO: turn into generic macro (all_evm_chains fails atm)
```
0xRobin, line 8: We can remove this "where" clause and make use of the generic `where` config for tests to filter any model. 
docs for more info: https://docs.getdbt.com/reference/resource-configs/where#examples


```
tests:
    - assert_evm_address:
        where: "blockchain in ( 'ethereum, 'optimism, ...)"
```

0xRobin: That should make this generic test applicable to any models! :partying_face:

augustog:
okay, yeah, agree with jeff this is growing in scope. I'll create a new PR for the test and remove it from this one.

augustog:
Moved generic test to #2995

augustog:
As soon as the CI runs I'll run the address test on the output and post it here.

tests/generic/assert_evm_address.sql
augustog: oh interesting!

I applied the filter at the definition level because the test looks specifically for evm address format, and will fail on valid e.g. solana addresses.

were you thinking  of that scenario? or scenarios where we want to use this in a single chain model (e.g. `blur/ethereum`) when we apply it upstream?

btw moved the test to #2995

augustog:
https://dune.com/queries/2241358
checks out so I'd say ready to merge

jeff-dude (approved):

PR #2985, opened 2023-03-20
fix blur: remove distinct

0xRobin:
A top level `distinct` was introduced some time back to fix duplicates.

This PR introduces a sustainable fix. 
- Removed join with nft transfers to pull in the token standard, use tokens.nft as a source instead
- Removed transfer joins to fix buyer/seller on aggregator trades, instead we fall back to the transaction initiator when the aggregator address pops up in buyer/seller field.

Big bonus that this change greatly reduces the model complexity meaning less future bugs.


0xRobin:
Duplication issue is with 1 NFT transfer emitting multiple events.
Example tx:
https://etherscan.io/tx/0x84b38ba4573a92496acffee3e2cfab8d1778c99ecd21acc7972d657f9fe37205

We should provide a cleaned up nft transfers table that ensures compliance with the ERC standards.

0xRobin:
defined the underlying issue here: 
https://github.com/duneanalytics/spellbook/issues/2992

jeff-dude (approved):

PR #2989, opened 2023-03-21
looksrare: remove distinct

0xRobin:
Tracking down the duplicate issues with looksrare.

- Removed transfer join to deduce token standard, leverage `tokens.nft` instead.
- Removed transfer joins to fix buyer/seller on aggregator trades, instead we fall back to the transaction initiator when the aggregator address pops up in buyer/seller field. 

Big bonus that this change greatly reduces the model complexity meaning less future bugs.


jeff-dude (approved):
thank you for continued cleanup -- love seeing this 🙌 

one note that i won't consider a blocker, just looked weird at first glance -- the incremental run took longer than full historical initial run. could easily be inaccurate long term, the code looks set up fine

PR #2994, opened 2023-03-21
T-26293 - Improve aave_ethereum.supply performance

belen-pruvost:

PR #2998, opened 2023-03-21
Revert "T-26293 - Improve aave_ethereum.supply performance"

augustog:
Reverts duneanalytics/spellbook#2994 as it seems to have caused duplicates. https://dune.com/queries/2241088


jeff-dude:
let's close this as we explore materializing either approach of union/union all

PR #3005, opened 2023-03-22
[CoW] Add batch_rewards view

bh2smith:

Parsing/unpacking the data field from the raw community source table `cowswap.raw_batch_rewards`. Very similar to PR #2303 which parses `raw_order_rewards`.

PoC Query: https://dune.com/queries/2247554

cc 
- @dsalv -- as participant in construction of community source table.
and
- @gentrexha for internal review

I am a bit worried about all this type casting in a view and hope it does not affect query performance too much.

**For Dune Engine V2**

I've checked that:

### General checks:

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_batch_rewards.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_batch_rewards.sql
@@ -1,0 +1,23 @@
+{{ config(alias='batch_rewards',
+        post_hook='{{ expose_spells(\'["ethereum"]\',
+                                    "project",
+                                    "cow_protocol",
+                                    \'["bh2smith"]\') }}'
+)}}
+
+-- PoC Query here - https://dune.com/queries/2247554
+select
+    block_deadline,
+    block_number, -- Null here means the settlement did not occur.
+    from_hex(solver) as winning_solver,
+    from_hex(tx_hash) as tx_hash,
+    -- Unpacking the data
+    data.winning_score,
+    data.reference_score,
+    data.surplus,
+    data.fee,
+    data.execution_cost,
+    data.uncapped_payment_eth,
+    data.capped_payment,
+    transform(data.participating_solvers, x -> from_hex(x)) as participating_solvers
+from {{ source('cowswap', 'raw_batch_rewards') }}
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
@@ -359,7 +359,7 @@ models:
       - &uncapped_payment_eth
         name: uncapped_payment_eth
         description: the (uncapped) payment (in ETH) to be awarded to the winning solver for the batch.
-      - &uncapped_payment_eth
+      - &capped_payment
         name: capped_payment
         description: the payment (in ETH) to be awarded to the winning solver for the batch.
       - &participating_solvers
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
@@ -352,9 +352,7 @@ models:
       - &surplus
         name: surplus
         description: Total surplus achieved (in ETH) for the batch. Zero when tx_hash is Null.
-      - &fee
-        name: fee
-        description: total fee (in ETH) collected for the batch. Zero when tx_hash is Null.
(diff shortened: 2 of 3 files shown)


models/cow_protocol/ethereum/cow_protocol_ethereum_batch_rewards.sql
```diff
@@ -0,0 +1,23 @@
...
+                                    "cow_protocol",
+                                    \'["bh2smith"]\') }}'
+)}}
+
+-- PoC Query here - https://dune.com/queries/2247554
+select
+    block_deadline,
+    block_number, -- Null here means the settlement did not occur.
+    from_hex(solver) as winning_solver,
+    from_hex(tx_hash) as tx_hash,
```
bh2smith, line 13: We are getting an error here! 

```
10:00:13    [UNRESOLVED_ROUTINE] Cannot resolve function `from_hex` on search path [`system`.`builtin`, `system`.`session`, `hive_metastore`.`default`].; line 12 pos 4
```

Seems to not exist in spark.

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -320,4 +320,48 @@ models:
...
+      - &surplus
+        name: surplus
+        description: Total surplus achieved (in ETH) for the batch. Zero when tx_hash is Null.
+      - *fee
+      - &execution_cost
+        name: execution_cost
+        description: execution cost of the batch (can also be acquired from `ethereum.transactions`)
+      - &uncapped_payment_eth
+        name: uncapped_payment_eth
+        description: the (uncapped) payment (in ETH) to be awarded to the winning solver for the batch.
```
gentrexha, line 361: Nit:

```suggestion
        description: The (uncapped) payment (in ETH) to be awarded to the winning solver for the batch.
```

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -320,4 +320,48 @@ models:
...
+      - &capped_payment
+        name: capped_payment
+        description: the payment (in ETH) to be awarded to the winning solver for the batch.
+      - &participating_solvers
+        name: participating_solvers
+        description: an array of solvers who submitted a valid score/solution for the batch. Used to compute participation/secondary rewards
```
gentrexha, line 367: Nit:

```suggestion
        description: An array of solvers who submitted a valid score/solution for the batch. Used to compute participation/secondary rewards
```

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -320,4 +320,48 @@ models:
...
+      In particular, how much COW token solvers are entitled for each settled order.
+      Furthermore, since the introduction of limit orders into the protocol, 
+      this table also contains the surplus fee.
+      The specification for Batch Reward data can be found at 
+      [Snapshot CIP-20](https://snapshot.org/#/cow.eth/proposal/0x2d3f9bd1ea72dca84b03e97dda3efc1f4a42a772c54bd2037e8b62e7d09a491f)
+    columns:
+      - *block_number
+      - &block_deadline
+        name: block_deadline
+        description: latest block for which winning solver must submit their solution. After this block, the payment is ZERO.
```
gentrexha, line 341: Nit:

```suggestion
        description: Latest block for which winning solver must submit their solution. After this block, the payment is ZERO.
```

gentrexha (approved):

models/cow_protocol/ethereum/cow_protocol_ethereum_batch_rewards.sql
jeff-dude: and you plan to use the view in dunesql i assume. the CI test isn't adjusted to handle that quite yet. i'm comfortable pushing this through as it's a view, doesn't roll into other spells outside of cow and low risk

bh2smith: Should I write it as a spark query instead?

jeff-dude: if it's written in spark but still a view, then queried via dunesql, we'll still have issues. can this be materialized as a table, instead of a view? how often is the new community source updated? if daily or less, a table could be simple approach -- backend runs on spark, queryable on dunesql too since functions aren't used anymore on front-end

PR #3007, opened 2023-03-22
Add stealcam to `nft.trades`

hildobby:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:

### DIFF
--- a/models/stealcam/arbitrum/stealcam_arbitrum_events.sql
+++ b/models/stealcam/arbitrum/stealcam_arbitrum_events.sql
@@ -1,0 +1,106 @@
+{{ config(
+    alias = 'events',
+    partition_by = ['block_date'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['block_date', 'unique_trade_id'],
+    post_hook='{{ expose_spells(\'["arbitrum"]\',
+                                "project",
+                                "stealcam",
+                                \'["hildobby"]\') }}')
+}}
+
+
+{% set project_start_date = '2023-03-10' %}
+
+SELECT 'arbitrum' AS blockchain
+, 'Stealcam' AS project
+, 'v1' AS version
+, sc.evt_block_time AS block_time
+, date_trunc('day', sc.evt_block_time) AS block_date
+, sc.evt_block_number AS block_number
+, 'Single Item Trade' AS trade_type
+, 'Buy' AS trade_category
+, CASE WHEN sc.value=0 THEN 'Mint' ELSE 'Trade' END AS evt_type
+, sc.from AS seller
+, sc.to AS buyer
+, sc.contract_address AS nft_contract_address
+, 'Stealcam' AS collection
+, sc.id AS token_id
+, 'erc721' AS token_standard
+, 1 AS number_of_items
+, '0x82af49447d8a07e3bd95bd0d56f35241523fbab1' AS currency_contract
+, 'ETH' AS currency_symbol
+, sc.value AS amount_raw
+, sc.value/POWER(10, 18) AS amount_original
+, pu.price*sc.value/POWER(10, 18) AS amount_usd
+, sc.contract_address AS project_contract_address
+, NULL AS aggregator_name
+, NULL AS aggregator_address
+, sc.evt_tx_hash AS tx_hash
+, at.from AS tx_from
+, at.to AS tx_to
+, COALESCE(sc.value-(roy.value+not_fee.value), 0) AS platform_fee_amount_raw
+, COALESCE((sc.value-(roy.value+not_fee.value))/POWER(10, 18), 0) AS platform_fee_amount
+, COALESCE(pu.price*(sc.value-(roy.value+not_fee.value))/POWER(10, 18), 0) AS platform_fee_amount_usd
+, COALESCE(100*(sc.value-(roy.value+not_fee.value))/sc.value, 0) AS platform_fee_percentage
+, CASE WHEN sc.value-(roy.value+not_fee.value) > 0 THEN 'ETH' ELSE NULL END AS royalty_fee_currency_symbol
+, COALESCE(roy.value, 0) AS royalty_fee_amount_raw
+, COALESCE(roy.value/POWER(10, 18), 0) AS royalty_fee_amount
+, COALESCE(roy.value/POWER(10, 18)/pu.price, 0) AS royalty_fee_amount_usd
+, COALESCE(100*roy.value/sc.value, 0) AS royalty_fee_percentage
+, m._creator AS royalty_fee_receive_address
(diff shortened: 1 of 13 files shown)


models/stealcam/arbitrum/stealcam_arbitrum_events.sql
```diff
@@ -0,0 +1,106 @@
...
+FROM {{ source('stealcam_arbitrum', 'Stealcam_evt_Stolen') }} sc
+INNER JOIN {{ source('arbitrum', 'transactions') }} at ON at.block_number=sc.evt_block_number
+    AND at.hash=sc.evt_tx_hash
+    {% if is_incremental() %}
+    AND at.block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
+    {% if not is_incremental() %}
+    AND at.block_time >= '{{project_start_date}}'
+    {% endif %}
+LEFT JOIN prices.usd_forward_fill pu ON pu.blockchain = 'ethereum'
```
jeff-dude, line 64: please apply jinja syntax

models/stealcam/arbitrum/stealcam_arbitrum_fees.sql
```diff
@@ -0,0 +1,42 @@
...
+                                    \'["hildobby"]\') }}')
+}}
+
+SELECT blockchain,
+project,
+version,
+block_time,
+token_id,
+collection,
+CAST(platform_fee_amount_raw AS DOUBLE) AS platform_fee_amount_raw,
```
jeff-dude, line 16: can we push the cast statements to the base incremental model, rather than rolled up views?

models/stealcam/arbitrum/stealcam_arbitrum_trades.sql
```diff
@@ -0,0 +1,38 @@
...
+SELECT blockchain,
+project,
+version,
+block_time,
+token_id,
+collection,
+amount_usd,
+token_standard,
+trade_type,
+CAST(number_of_items AS DECIMAL(38,0)) AS number_of_items,
```
jeff-dude, line 19: same idea here

jeff-dude (changes requested):
in addition to below comments, i hate to ask but gotta: did we verify this doesn't break the nft sector spells on dunesql?

hildobby:
> in addition to below comments, i hate to ask but gotta: did we verify this doesn't break the nft sector spells on dunesql?

how can this be checked?

models/stealcam/arbitrum/stealcam_arbitrum_fees.sql
jeff-dude: yeah, no problem, an old habit we're trying to clean up moving forward :)

jeff-dude:
> > in addition to below comments, i hate to ask but gotta: did we verify this doesn't break the nft sector spells on dunesql?
> 
> how can this be checked?

maybe more of a question for @0xRobin to check -- but until our automation checks are released, an approach may be grab the view output from the CI test (with github prefix) and run on dune app under dune sql

0xRobin:
> > > in addition to below comments, i hate to ask but gotta: did we verify this doesn't break the nft sector spells on dunesql?
> > 
> > 
> > how can this be checked?
> 
> maybe more of a question for @0xRobin to check -- but until our automation checks are released, an approach may be grab the view output from the CI test (with github prefix) and run on dune app under dune sql

Yeah I've not found a solid way to check this unfortunately. 
From my understanding these models/views build by the CI are only available in spark and not in dunesql?

jeff-dude:
> > > > in addition to below comments, i hate to ask but gotta: did we verify this doesn't break the nft sector spells on dunesql?
> > > 
> > > 
> > > how can this be checked?
> > 
> > 
> > maybe more of a question for @0xRobin to check -- but until our automation checks are released, an approach may be grab the view output from the CI test (with github prefix) and run on dune app under dune sql
> 
> Yeah I've not found a solid way to check this unfortunately. From my understanding these models/views build by the CI are only available in spark and not in dunesql?

good point. @belen-pruvost do you know why we can run the below on spark but not dunesql? is it something with the metastore not recognizing the `test_schema` schema?
```
select
  *
from
  test_schema.git_7d0cc25a_stealcam_arbitrum_trades
limit
  10
  -- select * from test_schema.git_7d0cc25a_nft_trades limit 10
```

belen-pruvost:
> good point. @belen-pruvost do you know why we can run the below on spark but not dunesql? is it something with the metastore not recognizing the `test_schema` schema?

This is a good question for @jmsgu 😄 

You should be able to query the test schemas, though you may run into errors like `Error: Error getting snapshot for <table>`, which is what the spell import job in Grimoire will solve. Is this the error you were seeing?

belen-pruvost:
> > `Error: Error getting snapshot for <table>`, which is what the spell import job in Grimoire will solve. Is this the error you were seeing?
> 
> Yes exactly.

Makes sense 👍 essentially new tables created by dbt have a dbfs (databricks file system) location, which DuneSQL can't access. Therefore grimoire would migrate tables like these before attempting to query them on DuneSQL

jeff-dude:
> > > `Error: Error getting snapshot for <table>`, which is what the spell import job in Grimoire will solve. Is this the error you were seeing?
> > 
> > 
> > Yes exactly.
> 
> Makes sense 👍 essentially new tables created by dbt have a dbfs (databricks file system) location, which DuneSQL can't access. Therefore grimoire would migrate tables like these before attempting to query them on DuneSQL

unfortunately, i think we have to wait for the new CI test. it's a security risk to allow the CI test to write to s3 directly rather than dbfs, which would allow dunesql to query the test tables (confirmed with meghan my thoughts)

jeff-dude:
@belen-pruvost should we just merge tomorrow and then retroactively check dunesql?

hildobby:
So we will soon be able to see test tables in DuneSQL to make sure they work fine there, right @belen-pruvost?

belen-pruvost:
> So we will soon be able to see test tables in DuneSQL to make sure they work fine there, right @belen-pruvost?

Yes, they won't appear on the data explorer but they'll be queryable. Also, the CI check that moves those tables to be queryable with DuneSQL will do a select on each affected model to make sure it's compatible with DuneSQL.

hildobby:
> > So we will soon be able to see test tables in DuneSQL to make sure they work fine there, right @belen-pruvost?
> 
> Yes, they won't appear on the data explorer but they'll be queryable. Also, the CI check that moves those tables to be queryable with DuneSQL will do a select on each affected model to make sure it's compatible with DuneSQL.

Damn that's even better if I don't have to manually check, CI tests should definitely fail if incompatible with DuneSQL 🙌

jeff-dude:
fyi @0xRobin to keep an eye on this one for dunesql :) we'll be around to help if needed

PR #3034, opened 2023-03-24
Fix creator earnings and platform fee on `opensea_ethereum.trades` v4 

sohwak:

**For Dune Engine V2**

Modifications:

- Instead of using the existing `seaport_ethereum.base_pairs`, `OrderFulfilled` raw table was used directly. While `base_pairs` could have been useful because it bundles eth/erc20/erc721/erc1155 transfers, the additional columns that contain derived data were inaccurate and hard-coded, making them unusable.
- The fee wallet addresses are used to determine `is_platform_fee`. OpenSea changed its fee policy in 2023, so there may be cases where platform fees are not incurred. Although Seaport can determine what tokens are being sent, it cannot always determine why they are being sent. Therefore, the best option was to pre-define the addresses and give them meaning. The ERC20 sent to that address was found to be fairly accurate. If other marketplaces that use Seaport contract to obtain fees, they will be included and this will be applied to the Seaport spell tables.
- The maximum number of destination addresses for creator earning through OpenSea was changed from 4 to 5 columns. Although it's not a good idea to list the value of the same attribute in multiple columns, it is affordable with the current Seaport spell.
- Transactions from two kinds of functions (`matchAdvancedOrders`, `matchdOrders`) can result in 2 rows, but it can be assumed to be a single trade by connecting with the `OrdersMatched`'s `order_hash`.

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


sohwak:
To Dune team: 
I would like to check the result table first, and I will let you know when the merge is needed.

models/opensea/ethereum/opensea_v4_ethereum_trades.sql
```diff
@@ -8,62 +15,525 @@
...
+          ,posexplode(orderhashes) as (om_order_id, om_order_hash)
+      from {{ source('seaport_ethereum','Seaport_evt_OrdersMatched') }}
+     where contract_address = '{{c_seaport_contract_address}}'
+)
+,iv_platform_fee_wallet (wallet_address, wallet_name) as (
+    values   ('0x5b3256965e7c3cf26e11fcaf296dfc8807c01073','opensea')
+            ,('0x8de9c5a032463c561423387a9648c5c7bcc5bc90','opensea')
+            ,('0x34ba0f2379bf9b81d09f7259892e26a8b0885095','opensea')
+            ,('0x0000a26b00c1f0df003000390027140000faa719','opensea')
+)
```
mikeshultz, line 77: One thing to keep in mind here is that there are other platforms that are using Seaport.  Blur, Gem, and Origin come to mind.  I'm not sure if they're all doing platform fees and I don't have them all available but would you be interested in them when I come across them?

sohwak: If some transfer destination is in that addresses, then it would be considered as a `platform_fee`. If not in this list, it would be considered as `royalty_fee` (or creator_earning). As far as I know, there are ens.vision and alphasharks that receive fees.
FYI, this is a correction for opensea_ethereum.trades and nft.trades. Does not affect seaport_ethereum.trades directly. But I'm trying to fix it in this way sooner or later.

sohwak:
I checked the gh-generated tables and most of the values are more accurate than the previous. 
@augustog @jeff-dude Could you please check this?

0xRobin:
@sohwak can you include a seed test for the opensea trades v4 that also includes the royalty and platform fee columns?

models/opensea/ethereum/opensea_v4_ethereum_trades.sql
mikeshultz: oh, so it is.  The issue I reported was in `seaport_ethereum_base_pairs`.  That looks like the dep that handles the `is_creator_fee` logic this depends on.  Not sure where it makes the most sense, but it would be cool if it wasn't only supporting of opensea platform fees.  Your call, thanks for working on this.

0xRobin: Since seaport has no notion of creator/marketplace fees, only payments, we should aim to move any creator fee logic out of the seaport models.
That should be handled by the marketplace specific models which build on top of the seaport ones. Not something we want to do in this PR, but definitely something we should do in the future to improve further.

mikeshultz: I get where you're coming from.  However, a bit of a snag on that is that there's no clean way to separate all the different marketplaces out on Seaport.  Just recently it's become the norm to append a domain hash to the txdata, but it's not enforced and isn't on historical transactions.  

That said, I'm not offering up any grand ideas here, it's just to point out that this spellbook isn't filtering out other marketplaces and may count other platform fees on Seaport as creator royalties.

sohwak:
Yep, sure thing.

models/opensea/ethereum/opensea_v4_ethereum_trades.sql
sohwak: I'm not exactly sure I understand the concept of Seaport, but I would like to tell you what I think.

- Seaport was developed by OpenSea, but it was developed as an open source platform that anyone can use. That's why the schema name is Seaport not OpenSea, and it's actually freely used by several marketplaces. 

- However, as 0xRob mentioned, there is no tagging like the purpose of the token being transferred (creator earnings/platform fees). I initially thought it was an order, but I was wrong. I guess it's because it's a flexible platform for various trades.

- This PR is for `opensea.trades`. I wanted to modify Seaport to be universally usable if possible and just filter out OpenSea trades, but that proved to be difficult, so I ended up specifying specific addresses like this PR. I don't think this logic is perfect either, I just think it's the best I could do.

- It's not clear how to distinguish between each marketplace in Seaport. I thought maybe `zone_address` was a separate column for this purpose, but now other marketplaces are using the same value as OpenSea, so it's quite difficult to find transactions from OpenSea in Seaport. For reference, the domain hash is used as a value to determine where the transaction came from. Currently, OpenSea trades are determined by `fee_receive_address` and `zone_address`.

sohwak:
Hi @0xRobin, I added a test case by referencing `seaport` . I tried to add it to the `tests/opensea/ethereum` but it failed a lot, so I ended up adding it to the `tests/generic/ethereum`.

jeff-dude (approved):
from a general spellbook perspective, this looks good to me. added one more test to validate unique key.

jeff-dude:
@jmsgu adding context for the dunesql label here -- it appears to be running successfully (as i type, there's a fresh CI test running, but previous run was `true`). my question would be: this spell feeds into `nft.trades` downstream, while we know we check if the modified model in the PR is compatible, do we check downstream objects? it's viable that this spell is fine, but it breaks `nft.trades`.

checking downstream could open up a lot more required logic (i.e. only check most important spells?) and cut down on the noise of other downstream spells. thinking out loud here, but may be something to hardcode in the check:
- does this feed into nft sector spells? (trades, events, mints, burns, fees) -- for this PR specifically, yes
- does this feed into dex/dex_aggreagator sector spells? (trades) -- for this PR, no
- ...potentially a few other sector-level checks tbd

0xRobin (approved):

sohwak:
While working on various unit table operations that make up `nft.trades`, I am not sure if there will be any issues when they are combined. As a minimum defense, I am working with an exclusive dataset to prevent overlaps with other rows in `nft.trades`. (In this case, it is uniquely identified by `blockchain=ethereum`, `project=opensea`, and `version=v4`.)

I am not sure about the issues that might arise from other `UNION ALL` operations. To my knowledge, a spell constructs tables using SparkSQL syntax and the completed content is used in DuneSQL. Errors in column names or numbers are likely to occur in SparkSQL. The data type seems to be the most frequent error so far, but I am not sure how to prevent it in advance. In general, `UNION ALL` in a typical database convert data types implicitly, but in DuneSQL, errors often occur. I am trying to maintain the same data types as the existing script as much as possible.

I hope I have understood your message correctly and provided a satisfactory response.

jeff-dude:
> > @jmsgu adding context for the dunesql label here -- it appears to be running successfully (as i type, there's a fresh CI test running, but previous run was `true`). my question would be: this spell feeds into `nft.trades` downstream, while we know we check if the modified model in the PR is compatible, do we check downstream objects? it's viable that this spell is fine, but it breaks `nft.trades`.
> > checking downstream could open up a lot more required logic (i.e. only check most important spells?) and cut down on the noise of other downstream spells. thinking out loud here, but may be something to hardcode in the check:
> > 
> > * does this feed into nft sector spells? (trades, events, mints, burns, fees) -- for this PR specifically, yes
> > * does this feed into dex/dex_aggreagator sector spells? (trades) -- for this PR, no
> > * ...potentially a few other sector-level checks tbd
> 
> @jeff-dude the dunesql check can only cover the models that are handled by dbt steps prior. for this PR, the check did not cover `nft.trades`. is it possible to have dbt look further downstream to detect the `nft.trades` dependency, perhaps run the spell sector-level checks?

can you create a height subtask off your main one for this use-case specifically? we can discuss details there

PR #3039, opened 2023-03-24
fix balancer liquidity v2

mendesfabio:
Adds `token_amount_raw` and `token_amount` columns to the `liquidity` views and fixes an issue when joining with token weights for estimating liquidity.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


Hosuke:
We may need to fix the duneSQL compatibility in this PR.
https://github.com/duneanalytics/spellbook/actions/runs/4596713919/jobs/8118522392?pr=3039#step:14:18

mendesfabio:
it's not clear to me how to make it compatible with both engines -- do you have any examples? also, do I have to fix it on my side or will you guys do it?

models/balancer/arbitrum/balancer_v2_arbitrum_liquidity.sql
```diff
@@ -20,10 +20,11 @@ WITH pool_labels AS (
         SELECT
             date_trunc('day', minute) AS day,
             contract_address AS token,
+            decimals,
             AVG(price) AS price
         FROM {{ source('prices', 'usd') }}
         WHERE blockchain = "arbitrum"
-        GROUP BY 1, 2
+        GROUP BY 1, 2, 3
```
Hosuke, line 27: We may need to rewrite the group by in full column names.

Hosuke: Substring function may also varies.

Hosuke (commented):
We may try to run the compiled query on dune sql first.
Current codes fits the spark sql syntax, but for dune sql there are some built-in functions need to be replaced..

models/balancer/arbitrum/balancer_v2_arbitrum_liquidity.sql
mendesfabio: should I do this change? how do I make it compatible with both Spark and Dune SQL if I use built-in functions for e.g. generating a time series that is different in each engine?

mendesfabio: are PRs only merged if the spell is compatible with both engines from now onwards? that's not clear to me too

Hosuke (approved):
LGTM.✅
This PR introduces `token_balance_raw` and `token_balance` for balancer_v2_*_liquidity.
Thank you @mendesfabio

jeff-dude:
hi @mendesfabio sorry for the confusion on the separate engines in spellbook right now. the issue we have is that the backend ETL pipelines still run on spark, yet the spells created ideally can run on dunesql. views in particular are the most difficult, as it has to succeed on spark and dunesql (trino) syntax. if we don't make it spark compatible, then the ETL process will fail. if we don't make it trino compatible, then querying the view on the app will fail (hence the error in the CI test attached to PR).

for these particular views, the current error message in the gh action CI test is yelling at us for using the `explode` function, as it's not available on trino.
```
    calendar AS (
        SELECT explode(sequence(to_date('2021-04-21'), CURRENT_DATE, interval 1 day)) AS day
    ),
```

we either need to look into materializing these spells into tables, which makes it easier to run on spark/query just the output on dunesql (rather than run the underlying view code each time), or find a function that is compatible for both engines for `explode`

if we explore materializing, how often do you need these updated? if only 1x/day, it can be a standard table. if multiple times a day, it'd need to be an incremental table.

if you can find a function to replace `explode`, then it can remain a view

markusbkoch:
cc @Stefenon @thetroyharris FYI, see Jeff's comment

mendesfabio:
once a day is good enough for now, can you materialize this table @jeff-dude? thanks in advance!

jeff-dude (approved):
okay, got them materialized as tables. they'll refresh 1x/day

PR #3055, opened 2023-03-27
MagicEden solana: fix unique trade id

0xRobin:
fixes null unique trade id

### DIFF
--- a/models/magiceden/solana/magiceden_solana_mints.sql
+++ b/models/magiceden/solana/magiceden_solana_mints.sql
@@ -1,5 +1,6 @@
 {{ config(
         alias ='mints'
+
         )
 }}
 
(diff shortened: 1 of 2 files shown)


jeff-dude:
about to push a batch of merges, are we thinking this should be good after it's CI run? i can wait for this, as a large refresh in prod will already be happening

0xRobin:
@jeff-dude tests have failed, so no need to further wait on this.

jeff-dude (approved):
thanks for the fix 🤝 

was thinking about changing `unique_key` to just `id` rather than `unique_trade_id`, but we can save that for the larger overhaul

also going to save this merge for the morning EST tomorrow, so we're all available

jeff-dude:
context for dunesql label:
- our gh action as a whole times out at 60 minutes. this model takes almost all that time to run/test (solana volume is large). i wonder if this makes us rethink including in existing CI gh action or if we should extend the time 🤔

PR #3057, opened 2023-03-27
[EAST]Update Datatype for Key

chuxinh:
- As per @jeff-dude enforcing datatype of `key` to string so that it's displayed as expected on both DuneSql and Sparksql: https://dune.com/queries/2274231?d=1


jeff-dude (approved):
thank you 🤝

PR #3068, opened 2023-03-28
[EASY]Add `seaport_optimism_trades` to `opensea_trades`

chuxinh:
Intention is to add `seaport_optimism_trades` to `opensea_trades` so that it will be eventually included in `nft.trades` as well as `nft.wash_trades`.

Note that it's currently missing `unique_trade_id` in  `seaport_optimism_trades`. Let me know if it's better to add it back to the original incremental build instead.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


0xRobin (changes requested):
Hi @chuxinh
Definitely a good idea to get the optimism trades into `nft.trades`.
There are however a couple more steps we need to put in place for this.

1. write trades, events, fees models in opensea/optimism
These can pull from the seaport optimism model but should filter on the correct zone address (and fee recipient) to only include opensea trades. I'll tag @sohwak here to see if he has more input on this. 
2. Include those models in the generic opensea events/trades/fees models.

I do think it's better to include the `unique_trade_id` in the base model instead of doing it in intermediary views. Although we will eventually step away from using `unique_trade_id`, for now consistency over all models is more important.

sohwak:
Thank you for letting me know this. I will check this, too.

MSilb7:
Hi, yeah on this - maybe an ideal solution is to have some Zone <> Marketplace mapping table that `seaport.trades` can join to before being loaded in to `nft.trades`

As I understand Seaport, multiple marketplaces plug in to it, so it would be super helpful to have some canonical mapping of what Zones mean what.

Pseudo-code:
```
Select [nft stuff],  COALESCE(z.marketplace,'Unknown Marketplace') AS marketplace
FROM seaport.trades t
LEFT JOIN seaport.zones z
 on t.zone = z.zone
```

MSilb7:
> > Hi, yeah on this - maybe an ideal solution is to have some Zone <> Marketplace mapping table that `seaport.trades` can join to before being loaded in to `nft.trades`
> 
> Unfortunately the seaport events alone aren't enough to extract the correct royalty and marketplace fees. So we'll still need a model for each marketplace, I don't think a generic model is trivial. And from what I understand from @sohwak there have been other marketplaces that share the same zone address with OpenSea but use their own fee_recipient? So zone address alone is insufficient to determine the correct marketplace. 🤔

Ok, so configuration could be joining on zone + fee recipient (or just fee recipient).

It is possible / sensible to just take care of `trades` and worry about other tables later? The importance of this is for wash_trades to cover all marketplaces, rather than "all except Opensea"

From an older version, I have:
```
when 
fee_receive_address in 
                ('0x5b3256965e7c3cf26e11fcaf296dfc8807c01073',
                '0x8de9c5a032463c561423387a9648c5c7bcc5bc90',
                '0x0000a26b00c1f0df003000390027140000faa719')
then 'OpenSea'
```
As an interim solution, maybe we just filter `seaport.trades` on this?

0xRobin:
@MSilb7 that sounds like a decent interim solution.
We will still need the other (events/fees) models, but I can add those during the final review of the PR. Extra columns that are needed for those models can be set to null in the beginning if it's not trivial to get the correct information. It's important we try to not have any discrepanties between trades/events where some trades only show up in 1 of the tables.

sohwak:
Hi all. I just caught up the comments. Sorry that I have not updated the opensea schema for so long.

Initial thought was:

- Other NFT markets seem to create `opensea.events` first, and then make view `opensea.trades`, `opensea.fees`, and `opensea.mints` by filtering trade type and getting related columns 
- In my opinion, opensea should have `opensea.trades` for what is traded on the platform and `opensea.mints` for what is produced with the OpenStore or on SeaDrop, so I was thinking of `opensea.events` = `opensea.mints`(1st market) + `opensea.trades`(2nd market)

- When thought about multi-chain
  - `opensea.trades` = `opensea_solana.trades` + `opensea_ethereum.trades` + `opensea_optimism.trades`
  - `opensea.events` = `opensea_solana.events` + `opensea_ethereum.events` + `opensea_optimism.events`

- I think it would be nice to create a view layer explicitly called `opensea_optimism.trades` even though `seaport_optimism.trades` is same to `opensea_optimism.trades` (I should have done that beforehand, I couldn't)

- The problem I have faced in Ethereum is that there are many project using Seaport, and some of them also use same `zone_address` value to OpenSea, so it's hard to distinguish trades that are purely from OpenSea. Currently, we're referring to both `zone_address` and `fee_receive_address` for now.

- I don't know the situation of other chains other than Ethereum because I haven't had time yet. Polygon seems to be used in a way that we didn't expect as well, and I don't know about Optimism now. 

- I have a little more time than before. I'd like to take the time to help with the multichain side as well.

chuxinh:
> * In my opinion, opensea should have `opensea.trades` for what is traded on the platform and `opensea.mints` for what is produced with the OpenStore or on SeaDrop, so I was thinking of `opensea.events` = `opensea.mints`(1st market) + `opensea.trades`(2nd market)
> *

Thanks @sohwak 
AFAIK, right now there's still no native mint on Optimism for OpenSea. So `opensea_optimism.events` should just be `opensea.trades` with the zone + fee receiver address filter on seaport? 

i think that'd serve the purpose of this PR and team from OpenSea and help figure out the rest like fees etc? 

Let me know if this understanding is correct

sohwak:
I think that should work. 
We have `seaport_optimism.trades`, so I think we can use that to create `opensea_optimism.trades` by filtering with the `zone` + `fee receiver address`, and then make the view `opensea_optimism.events`. Probably you have no problem if you just align columns.

chuxinh:
> I think that should work. We have `seaport_optimism.trades`, so I think we can use that to create `opensea_optimism.trades` by filtering with the `zone` + `fee receiver address`, and then make the view `opensea_optimism.events`. Probably you have no problem if you just align columns.

Nice. That sounds good to me. Do you have the list of addresses that we can filter on?

sohwak:
Oh, I don't have now. I'll ask the corresponding team later and list it.

chuxinh:
> Oh, I don't have now. I'll ask the corresponding team later and list it.

@sohwak let me know if you've heard back from the team. 🙏

sohwak:
Sorry I didn't realize you were waiting for an answer from me on this PR, I'll let you know as soon as I know.

MSilb7:
Also - seems like `seaport_optimism.trades` is > 1 day behind

PR #3083, opened 2023-03-30
DO NOT MERGE: test add valid marketplace

0xRobin:
Test the DuneSQL CI step when adding a valid new marketplace.

The code was copy/pasted from `archipelago_ethereum.events` so we know it should pass the tests


0xRobin:
Looks like we have a clash between the DuneSQL check and the DuneSQL type overrides.. :smiling_face_with_tear:

jeff-dude:
closing this, but can always refer back for context

PR #3093, opened 2023-03-31
T-26472 materialize nft/perpetual sector

jeff-dude:


couralex6 (approved):

jeff-dude:
timing out on copying data over in dunesql check. @jmsgu will look to filter in the CTAS statement to make this timeout occur less frequently

dot2dotseurat:
@jeff-dude looks like this is ready barring the dune sql check?
should we coordinate with query (@belen-pruvost) to see if they need to add any types overrides for this table or just remove the timeout constraint for this pr to get the grimoire test running.

jeff-dude:
> @jeff-dude looks like this is ready barring the dune sql check? should we coordinate with query (@belen-pruvost) to see if they need to add any types overrides for this table or just remove the timeout constraint for this pr to get the grimoire test running.

we should be set on the overrides, as there's been a lot of iteration on these as views already. i can double check quickly.

as for the dunesql check, i'm comfortable moving on with the timeout, but we can let it run without one just to be sure since i won't merge until a bit later anyway

dot2dotseurat:
@jeff-dude mind it I merge?

dot2dotseurat (approved):

jeff-dude:
> @jeff-dude mind it I merge?

let's go for it. i just merged a batch of belen's fixes for dunesql, so might as well put this one in too

jeff-dude:
forgot to remove the timeout change :) i can follow up

PR #3106, opened 2023-04-03
Optimism: bridged ERC20 prices

0xRobin:
This PR automatically includes bridged optimism ERC20s into the price feeds if we have a price feed for the corresponding ETH erc20.


0xRobin:
cc: @MSilb7 @chuxinh

0xRobin:
> If an L1 <> L2 mapping exists, check if we have a price feed for the L1 token. If so, map this price feed to the L2 token? (with checks that this token isn't already in the curated L2 list).

exactly!

jeff-dude (approved):

PR #3115, opened 2023-04-04
T-26744 - Make opensea.inorganic_volume_filter_wallet_funders DuneSQL compatible

belen-pruvost:


augustog (approved):
All tests pass, changes make sense
Thanks @belen-pruvost  :)

PR #3116, opened 2023-04-04
T-26744 - Make transfers_bnb.bep20 spells DuneSQL compatible

belen-pruvost:


belen-pruvost:
@jmsgu  Looks like grimoire is failing with `Query exceeds cluster capacity, please optimize your query` when testing the models - should the resources for the cluster be increased to avoid this, or what could be a good strategy to not block PRs?

belen-pruvost:
@jeff-dude even though Grimoire is check is failing, I tested the new models directly on prod ([here](https://dune.com/queries/2328768) and [here](https://dune.com/queries/2328774)) and they are working, so this should hopefully be okay for review 🙏

PR #3117, opened 2023-04-04
T-26744 - Make transfers_ethereum token spells DuneSQL compatible

belen-pruvost:


belen-pruvost:
Querying the changed models on Prod works successfully ([here](https://dune.com/queries/2328849), [here](https://dune.com/queries/2328854) and [here](https://dune.com/queries/2328859)) so I believe this PR is good for review.

augustog (approved):
Manual testing on dune.com passes ✅ 
Syntax works in Spark & DuneSQL ✅

PR #3118, opened 2023-04-04
T-26744 - Hide spells that can't be made compatible with DuneSQL

belen-pruvost:


belen-pruvost:
I believe the CI check failure is unrelated to the changes proposed by this PR

augustog:
indeed, they should be unrelated. Will retry the run just to be sure.

jeff-dude:
is this still valid? i noticed a few merged PRs edited some of these files to make compatible?

belen-pruvost:
> is this still valid? i noticed a few merged PRs edited some of these files to make compatible?

I think it still is, but let me confirm today now that the other spell PRs were merged. I'll keep you posted.

belen-pruvost:
I confirm, these are still incompatible and therefore I believe hiding them for now is the way

belen-pruvost:
It makes sense for the DuneSQL check to fail considering these models are incompatible

jeff-dude:
fyi @Stefenon -- similar to comment [here](https://github.com/duneanalytics/spellbook/pull/3039#issuecomment-1505632145), since these are views and not dunesql compatible, they are being hidden in the data explorer. the example in that link on explode function isn't necessarily the case universally, but wanted to show context from that comment on why it's happening. these will still work in spark sql, but if we keep as views, they will need edited to be queryable on dunesql.

jeff-dude (approved):

PR #3121, opened 2023-04-04
T-22455 - Show balances_ethereum.erc721_latest

belen-pruvost:
This spell can now be queried on DuneSQL - after the cluster resources adjustments + spill to disk logic that have been rolled out.

The reason why this was hidden was because querying it was resulting in `Query exceeded per-node memory` - but it's no longer happening (see [query](https://dune.com/queries/2329212))

### DIFF
--- a/models/balances/ethereum/erc721/balances_ethereum_erc721_latest.sql
+++ b/models/balances/ethereum/erc721/balances_ethereum_erc721_latest.sql
@@ -1,6 +1,6 @@
 {{ config(
         alias='erc721_latest',
-        post_hook='{{ expose_spells_hide_trino(\'["ethereum"]\',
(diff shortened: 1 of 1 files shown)


augustog:
The spell does indeed work in prod. Do we have a sense of what's causing the slim ci check to fail ?

augustog (approved):
DSQL compatibility check is failing but check on dune.com passes. We should be okay bringing this back to the UI.

:)

belen-pruvost:
> The spell does indeed work in prod. Do we have a sense of what's causing the slim ci check to fail ?

Yes - the Trino cluster that Grimoire uses to run the compatibility queries is not powerful enough so the checks fail with `Query exceeds cluster capacity, please optimize your query`. @jmsgu is already on the loop about these so I believe he's working on increasing resources for them.

PR #3122, opened 2023-04-04
T-22455 - Fix balances_ethereum.genesis_balances to be DuneSQL compatible

belen-pruvost:


augustog:
All tests pass
No data inconsistencies introduced https://dune.com/queries/2332628

PR #3123, opened 2023-04-04
T-22455 - Fix cryptopunks_ethereum.all_events to be DuneSQL compatible

belen-pruvost:


augustog (approved):
Tests pass
Syntax works for both engines

PR #3128, opened 2023-04-05
Lido ethereum accounting

gregshestakovlido:
Creating spell for Lido accounting. Lido would like to make all expenses open, but the simple querying query computed too long, so we would like to make a materialized table out of this query

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


tests/lido/ethereum/lido_ethereum_accounting_test.sql
```diff
@@ -0,0 +1,17 @@
+with  tests as (
+select  'Check Lido deposits' as test_name, 
+case when amount_staked = 224000000000000000000 then true else false end as success
```
Hosuke, line 3: Hi @gregshestakovlido, the test fails for `Check Lido deposits`.
https://github.com/duneanalytics/spellbook/actions/runs/4696020625/jobs/8325724396?pr=3128#step:10:19

The amount_staked seems to be `32000000000000000000`
https://dune.com/queries/2367842?d=1

Please verify which result is correct.

gregshestakovlido: Thank you, fixed

models/lido/ethereum/lido_ethereum_accounting_deposits.sql
```diff
@@ -0,0 +1,21 @@
...
+}}
+--https://dune.com/queries/2011901
+--ref{{'lido_ethereum_accounting_deposits'}} 
+
+	SELECT 
+        block_time AS period,
+        amount_staked*POWER(10,18) AS amount_staked,
+        LOWER('0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2') AS token, --ETH
+        tx_hash
+    FROM {{ ref('staking_ethereum_deposits') }}
```
Hosuke, line 20: Hi @gregshestakovlido,
since `staking_ethereum_deposits` is an incremental view,
the result may not be static(results may be changed unexpectedly).
If the accounting is linked to certain timeframe, we may hardcoded the time filter in `WHERE` condition.

tests/lido/ethereum/lido_ethereum_accounting_test.sql
```diff
@@ -0,0 +1,17 @@
+with  tests as (
+select  'Check Lido deposits' as test_name, 
+case when sum(amount_staked) = 224000000000000000000 then true else false end as success
```
Hosuke, line 3: And the result of `sum()` may change as well, and this test may not be working properly.

Hosuke (commented):
Hi @gregshestakovlido,
please be careful with the `source` and `ref` used in the spells.
If the accounting is linked to certain timeframe, we may need to add `where` condition.

models/lido/ethereum/lido_ethereum_accounting_deposits.sql
gregshestakovlido: Got it, not sure we could link it to certain timeframe, but will think of how we could avoid this, will return in 1-2 days, thank you

Hosuke: The raw table updates regularly as well.
If there is any new deposit between these two addresses, this table will update as well.
And `sum(amount_staked)` will increase over time.

This will result in failure in test, as the test may be run periodically: 
https://github.com/duneanalytics/spellbook/pull/3128/files#diff-47eec97d268bb1c56cfd27f6f77e57dc3e1641b1ab6952f6e08ab38ea6be00ceR3

gregshestakovlido: Sorry, I suppose I didn’t get the mistake correctly. In test, we check certain transaction hash, and I thought that within 1 transaction info couldn't be changed (so we can use it for test), but, apparently, am wrong here and data regarding certain transaction could be changed?

Hosuke: Sorry I missed the tx_hash part in the where condition.

Hosuke (approved):
LGTM.✅
Thank you @gregshestakovlido 

Sorry I missed the specified `tx_hash` in the test.

You can still use `ref('staking_ethereum_deposits')` as source.
The `table` materialization will update daily(roughly one time per day for now).

jeff-dude (approved):
thanks all! looking forward to hosting this on spellbook

PR #3129, opened 2023-04-05
Iniial restructuring nft

0xRobin:
WIP


models/_sector/nft/trades/MIGRATION.md
```diff
@@ -0,0 +1,156 @@
+# Sector: NFT
+
+We're migrating the NFT sector spells to a new structure which will make implementing and maintaining marketplace abstractions a lot easier and more robust.
+
+## `nft.trades`
+
+- partition_by: `[blockchain, block_date]`
```
jeff-dude, line 7: may be worthwhile for us to do deeper analysis on the partition columns, for example:
- run group by these fields, grab counts, see if any combo have much too small volume of data for a partition (could be different on trino vs. spark)
- potentially leverage internal analytics tools to see usage on `nft.trades` to find most common group by/join conditions/etc that help partitions, though these fields are most likely candidates (@augustog can help us here)

models/_sector/nft/trades/MIGRATION.md
```diff
@@ -0,0 +1,156 @@
+# Sector: NFT
+
+We're migrating the NFT sector spells to a new structure which will make implementing and maintaining marketplace abstractions a lot easier and more robust.
+
+## `nft.trades`
+
+- partition_by: `[blockchain, block_date]` 
+- unique_key: `[blockchain, project, project_version, block_number,tx_hash, sub_tx_trade_id]`
```
jeff-dude, line 8: ahhh universal unique key, i love it

models/_sector/nft/trades/MIGRATION.md
```diff
@@ -0,0 +1,156 @@
...
+
+| Column Name             | Descriptions                                                       |
```
jeff-dude, line 10: probably not in this markdown file for migration, but we should likely add a readme at the root directory of `_sector/nft/trades` which reuses some of this info, then compliment it with more:
- null / not null column, rather than "if applicable" in description
- accepted values column
- ...maybe more

even though these will be applied in schema file(s), i like the idea of having it laid out in readme for quick pointers

models/_sector/nft/trades/MIGRATION.md
```diff
@@ -0,0 +1,156 @@
...
+| nft_amount               | The amount of NFTs (only >1 for ERC1155)                                   |
+| price_raw                | The raw full amount (ERC20) paid for the NFT, not adjusted for decimals    |
+| currency_contract        | The address of the contract for the currency used in the NFT trade         |
+| project_contract_address | The address of the contract for the project associated with the NFT trade  |
+| platform_fee_amount_raw  | The raw amount (ERC20) paid in marketplace fees, not adjusted for decimals |
+| platform_fee_address     | The address to which the platform fee was sent to (if applicable)          |
+| royalty_fee_amount_raw   | The raw amount (ERC20) paid in royalties, not adjusted for decimals        |
+| royalty_fee_address      | The address to which the royalty fee was sent to (if applicable)           |
+
+Enrichment of these base tables is done in 1 model for each chain.
```
jeff-dude, line 79: so excited for this

models/_sector/nft/trades/MIGRATION.md
```diff
@@ -0,0 +1,156 @@
...
+
+
+### Migration
+
+#### User changelog
+
+The following changes affect the users:
+
+##### models:
+- [deprecated] `nft.events`, use `nft.trades` instead
```
jeff-dude, line 100: curious to get the team's thoughts on deprecating these, as it could break a lot of downstream queries/dashboards without proper precautions & communication
cc @augustog @0xBoxer @andrewhong5297

jeff-dude: with that said, i love the idea of it, but want to make sure we're raising awareness early

jeff-dude: applies to both `nft.events` and `nft.fees`, want to make sure the following line isn't missed

models/_sector/nft/trades/MIGRATION.md
```diff
@@ -0,0 +1,156 @@
...
+- [deprecated] `nft.fees`, use `nft.trades` instead
+
+
+##### columns:
```
jeff-dude, line 104: similar idea here to raise awareness on column changes
cc @augustog @0xBoxer @andrewhong5297

models/_sector/nft/trades/ethereum/platforms/_sources.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+
+sources:
+  - name: archipelago_ethereum
+  - name: superrare_ethereum
+#  sources are already defined in the "older" directories and can't be defined twice. In the end we should migrate them to here.
```
jeff-dude, line 6: noted

models/_sector/nft/trades/ethereum/platforms/_schema.yml
```diff
@@ -0,0 +1,92 @@
...
+     sector: nft
+     project: archipelago
+     contributors: 0xRob
+   config:
+     tags: ['ethereum', 'nft', 'trades', 'archipelago']
+   description: "Archipelago base trades"
+   tests:
+     - dbt_utils.unique_combination_of_columns:
+         combination_of_columns: ['block_number','tx_hash','sub_tx_trade_id']
+     - dbt_utils.equal_rowcount:
```
jeff-dude, line 16: this is great, specifically comparing to source

jeff-dude: but i notice the others can't necessarily apply the same test. will that always be the case due to complexity or will they eventually add similar tests?

models/_sector/nft/trades/schema.yml
jeff-dude, line 1: minor, but any reason this schema file doesn't use your new approach of preceding `_`?

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
```diff
@@ -0,0 +1,78 @@
+{{ config(
+    schema = 'archipelago_ethereum',
+    alias ='base_trades',
```
jeff-dude, line 3: do you have a strong opinion on using `base_` on table name and file name? is it to overly express that this is base layer or can we remove and just assume it's base layer?

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
```diff
@@ -0,0 +1,78 @@
...
+    )
+}}
+{% set project_start_date='2022-6-20' %}
+
+WITH
+trade_events as (
+    SELECT * FROM {{ source('archipelago_ethereum','ArchipelagoMarket_evt_Trade') }}
+    {% if is_incremental() %}
+    WHERE evt_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
```
jeff-dude, line 18: minor nit:
i could see this going one of two ways:
- remove the non-incremental filter altogether, as this reads from decoded table, so it'll always be at project start date anyway -- although i've found it useful to have to easily apply small data timeframe filters for dev/debugging, i.e. just update the project start date to now() - 1 week or something
- if we keep, change from `endif`, `if` to just `else` for simplicity -- again, minor, but something i've found more readable when building new models

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
```diff
@@ -0,0 +1,78 @@
...
+    {% if not is_incremental() %}
+    WHERE evt_block_time >= '{{project_start_date}}'
+    {% endif %}
+),
+token_events as (
+    SELECT * FROM {{ source('archipelago_ethereum','ArchipelagoMarket_evt_TokenTrade') }}
+    {% if is_incremental() %}
+    WHERE evt_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
```
jeff-dude, line 27: same idea here

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
```diff
@@ -0,0 +1,78 @@
...
+        , sum(amount) filter (
+            where recipient not in ('0xa76456bb6abc50fb38e17c042026bc27a95c3314','0x1fc12c9f68a6b0633ba5897a40a8e61ed9274dc9')
+            ) as royalty_amount
+        , sum(amount) filter (
+            where recipient in ('0xa76456bb6abc50fb38e17c042026bc27a95c3314','0x1fc12c9f68a6b0633ba5897a40a8e61ed9274dc9')
+            ) as platform_amount
+    FROM {{ source('archipelago_ethereum','ArchipelagoMarket_evt_RoyaltyPayment') }}
+    {% if is_incremental() %}
+    WHERE evt_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
```
jeff-dude, line 45: same idea

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
```diff
@@ -0,0 +1,78 @@
...
+    GROUP BY evt_block_number, tradeId
+)
+
+SELECT
+    date_trunc('day',trade.evt_block_time) as block_date
+    ,trade.evt_block_number as block_number
+    ,trade.evt_tx_hash as tx_hash
+    ,trade.contract_address as project_contract_address
+    ,CAST(null as varchar(5)) as trade_category
+    ,'secondary' as trade_type
```
jeff-dude, line 58: apply accepted value(s) test in schema file?

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
```diff
@@ -0,0 +1,78 @@
...
+    ,trade.buyer
+    ,trade.seller
+    ,tok.tokenAddress as nft_contract_address
+    ,tok.tokenId as nft_token_id
+    ,1 as nft_amount
```
jeff-dude, line 63: apply accepted value(s) test in schema file? i believe markdown file said only `1` across the board?

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
jeff-dude: since i'm already commenting a lot, assume all comments in `archipelago` apply to other platform base trades too

models/_sector/nft/trades/ethereum/platforms/foundation_ethereum_base_trades.sql
```diff
@@ -0,0 +1,127 @@
...
+  'ethereum' as blockchain
+, 'foundation' as project
+, 'v1' as version
+, date_trunc('day', t.block_time) AS block_date
+, t.block_time
+, t.block_number
+, t.nft_token_id
+, 1 AS nft_amount
+, t.trade_category
+, cast( null as string) as trade_type
```
jeff-dude, line 114: there's logic for `trade_type` in CTE, any reason it's forced null?

models/_sector/nft/trades/ethereum/platforms/foundation_ethereum_base_trades.sql
```diff
@@ -0,0 +1,127 @@
...
+    FROM {{ source('foundation_ethereum','market_evt_PrivateSaleFinalized') }} f
+     {% if is_incremental() %} -- this filter will only be applied on an incremental run
+     WHERE f.evt_block_time >= date_trunc("day", now() - interval '1 week')
+     {% else %}
+     WHERE f.evt_block_time >= '{{project_start_date}}'
+     {% endif %}
+    )
+
+SELECT
+  'ethereum' as blockchain
```
jeff-dude, line 105: i notice this model contains blockchain/project/version, yet archipelago doesn't? are columns intended to be universal across all base layers?

models/_sector/nft/trades/ethereum/platforms/foundation_ethereum_base_trades.sql
```diff
@@ -0,0 +1,127 @@
...
+, 'foundation' as project
+, 'v1' as version
```
jeff-dude, line 107: a minor thing that i've wondered awhile: `nft.trades` uses `v` prefix on versions, yet `dex.trades` does not -- should we standardize and remove the `v`?

models/_sector/nft/trades/ethereum/platforms/superrare_ethereum_base_trades.sql
```diff
@@ -0,0 +1,227 @@
...
+
+SELECT
+    'ethereum' as blockchain,
+    'superrare' as project,
+    'v1' as version,
+    cast(date_trunc('day', a.block_time) AS date) AS block_date,
+    a.block_time,
+    a.block_number,
+    a.nft_token_id,
+    CAST(1 AS DECIMAL(38,0)) as nft_amount,
```
jeff-dude, line 187: seems this is casted here, but not other base models

models/_sector/nft/trades/ethereum/platforms/superrare_ethereum_base_trades.sql
```diff
@@ -0,0 +1,227 @@
...
+    case
+        when a.seller = coalesce(minter.to, minter_superrare.to)
+        then cast(0 as decimal(38))
+        else cast(ROUND(0.10 * (a.price_raw)) as decimal(38))  -- fixed 10% royalty fee on secondary sales
+    end as royalty_fee_amount_raw,
+    cast(NULL as varchar(1)) as royalty_fee_address,
+    cast(NULL as varchar(1)) as platform_fee_address,
+    sub_tx_trade_id
+from all_superrare_sales a
+left join {{ source('erc721_ethereum','evt_transfer') }} minter on minter.contract_address = a.contract_address
```
jeff-dude, line 215: do these have to be left join, rather than inner?

models/_sector/nft/trades/ethereum/platforms/superrare_ethereum_base_trades.sql
```diff
@@ -0,0 +1,227 @@
...
+    and minter.tokenId = a.nft_token_id
+    and minter.from = '0x0000000000000000000000000000000000000000'
+    {% if is_incremental() %}
+    and minter.evt_block_time >= date_trunc("day", now() - interval '1 week')
+    {% endif %}
```
jeff-dude, line 220: maybe add else filter for project start date?

models/_sector/nft/trades/nft_ethereum_trades_beta.sql
```diff
@@ -0,0 +1,31 @@
...
+
+-- macros/models/sector/nft
+{{
+    enrich_trades(
+        blockchain='ethereum',
+        models=base_models,
+        transactions_model=source('ethereum','transactions'),
+        tokens_nft_model=ref('tokens_ethereum_nft'),
+        tokens_erc20_model=ref('tokens_ethereum_erc20'),
+        prices_model=ref('prices_usd_forward_fill'),
```
jeff-dude, line 27: is using `forward_fill` a new approach?

models/_sector/nft/trades/nft_ethereum_trades_beta.sql
```diff
@@ -0,0 +1,31 @@
...
+] %}
+
+
+-- macros/models/sector/nft
+{{
+    enrich_trades(
+        blockchain='ethereum',
+        models=base_models,
+        transactions_model=source('ethereum','transactions'),
+        tokens_nft_model=ref('tokens_ethereum_nft'),
```
jeff-dude, line 25: i'm good with this, but curious your thoughts on leveraging the chain-specific spell vs. the broader spell and adding blockchain to join condition?

models/prices/ethereum/prices_ethereum_tokens.sql
```diff
@@ -15,7 +15,7 @@ SELECT
 FROM
 (
     VALUES
-
+    ("eth-ethereum","ethereum","ETH","{{ var('ETH_ERC20_ADDRESS') }}",18),
```
jeff-dude, line 18: this already exists in the `prices_native_tokens` spell

jeff-dude: or if it doesn't should likely?

models/_sector/nft/trades/ethereum/platforms/_schema.yml
```diff
@@ -0,0 +1,92 @@
+version: 2
+
+models:
+ - name: archipelago_ethereum_base_trades
```
jeff-dude, line 4: if we leave out column definitions in schema file(s), then we lose some flexibility to populate the docs page. i'm not sure how much usage there is there, but is something to consider.
maybe @0xBoxer can help understand the usage on spellbook docs page and if we remove columns here, then we could be stripping out details provided there

macros/models/_sector/nft/enrich_trades.sql
```diff
@@ -0,0 +1,107 @@
...
+    base.platform_fee_amount_raw/pow(10,coalesce(erc20.decimals,18))*p.price as platform_fee_amount_usd,
+    base.royalty_fee_amount_raw/pow(10,coalesce(erc20.decimals,18))*p.price as royalty_fee_amount_usd,
+    agg.contract_address as aggregator_address,
+    {% if aggregator_markers != null %}
+    coalesce(agg_mark.aggregator_name, agg.name) as aggregator_name
+    {% else %}
+    agg.name as aggregator_name
+    {% endif %}
+FROM base_union base
+INNER JOIN {{ transactions_model }} tx
```
jeff-dude, line 84: can we make this flexible enough to filter on `block_time` as well? since the model calling this is incremental?

macros/models/_sector/nft/enrich_trades.sql
```diff
@@ -0,0 +1,107 @@
...
+    base.nft_contract_address,
+    base.nft_token_id,
+    base.nft_amount,
+    base.price_raw,
+    base.currency_contract,
+    base.platform_fee_amount_raw,
+    base.royalty_fee_amount_raw,
+    base.platform_fee_address,
+    base.royalty_fee_address,
+    tx.block_time,
```
jeff-dude, line 65: any reason we use `block_time` from base transactions source? rather than decoded source(s)?

macros/models/_sector/nft/enrich_trades.sql
```diff
@@ -0,0 +1,107 @@
...
+ON tx.block_number = base.block_number
+    AND tx.hash = base.tx_hash
+LEFT JOIN {{ tokens_nft_model }} nft
+ON nft.contract_address = base.nft_contract_address
+LEFT JOIN {{ tokens_erc20_model }} erc20
+ON erc20.contract_address = base.currency_contract
+LEFT JOIN {{ prices_model }} p
+ON p.blockchain = base.blockchain
+    AND p.contract_address = base.currency_contract
+    AND p.minute = date_trunc('minute',tx.block_time)
```
jeff-dude, line 94: same idea here for `tx.block_time`

jeff-dude (commented):
this is so f'in great -- thank you for taking the time to help revamp 🙏  i'm beyond excited to get this moving. i've left (probably too much) comments below, looking forward to moving it forward

models/_sector/nft/trades/MIGRATION.md
0xRobin: agreed. :heavy_check_mark: 
This markdown is mostly just a work document for now, but I think once we've finalized the migration we distill it to a good overview and put it in a readme.

models/_sector/nft/trades/MIGRATION.md
0xRobin: I agree we need to explore the possibilities here further.
For now I went with blockchain as most of the queries and dashboards I've seen are still blockchain specific so this setup should allow for easy/efficient filtering. The thought behind this is also that we don't want blockchains with a lot of (spam) NFT traffic to influence compute times for queries running only on mainnet for example.

models/_sector/nft/trades/MIGRATION.md
0xRobin: Yes, it's a discussion we need to have.
What https://github.com/duneanalytics/spellbook/pull/3253 already does is reduce those models to a view on top of `nft.events`. So we can keep those "compatibility views" around while we decide if we want to actually deprecate them in favor of simplicity or keep them in favor of backwards compatibility.

It would be cool if we could figure out a way to automatically migrate user queries to handle these things like model and column renamings so they don't see any impact.

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
0xRobin: I agree "if incremental" -> "else" -> "endif" is the preferred way. :+1:

models/_sector/nft/trades/nft_ethereum_trades_beta.sql
0xRobin: I'd say it's preferred over the regular `prices.usd` as it enables more recent trades to also have a `amount_usd` value.
It won't be 100% accurate but will be on following runs once price.usd has catched up. 

I know Hildobby has been using it in his marketplace models so I also did not want to remove that here.

models/_sector/nft/trades/schema.yml
0xRobin: no reason, I've been mostly doing it just whenever I get annoyed that the schema and source files are put in between model files. 
If we really want to adopt this as the general approach we can do a full pass over the repo and rename all schema and source files.

models/prices/ethereum/prices_ethereum_tokens.sql
0xRobin: Yeah those aren't in the format we need here as they don't have a `chain` and `contract_address` value.

The goal is that we can just use `var(ETH_ERC20_ADDRESS)` in models and don't need any other custom logic to handle the native token case for prices or erc20 matching..

But I agree that having
1. a native ETH row in `prices_ethereum_tokens` with our own `ETH_ERC20_ADDRESS`
2. a native ETH row in `prices_native_tokens` with no address
3. a native ETH row in `tokens_native` with the WETH symbol and address.

Is not what we'd want, I'll try to think of a better setup that works for everything. Will need to look into the current uses of 2) and 3) first.

0xRobin: https://dune.height.app/T-24057?task=T-27753

models/_sector/nft/trades/ethereum/platforms/superrare_ethereum_base_trades.sql
0xRobin: These have to be left joins as superare has non-erc721 tokens.

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
0xRobin: No strong opinion on `base_` but I do think having both `archipelago_ethereum_trades` and `nft_ethereum_trades` when they don't follow the same schema is confusing. Hence the introduction of `base_trades`.
Very open to other naming suggestions.

models/_sector/nft/trades/MIGRATION.md
0xRobin: Task so we don't forget and circle back on this. 
https://dune.height.app/T-24057?task=T-27749

models/_sector/nft/trades/MIGRATION.md
0xRobin: Let's continue here: https://dune.height.app/T-24057?task=T-27752

models/_sector/nft/trades/ethereum/platforms/_schema.yml
0xRobin: definitely still the plan to add all columns in the schema file with proper descriptions and tests.

Big bonus is that we only have to specify the tests once and they'll be automatically applied to all platform models.

macros/models/_sector/nft/enrich_trades.sql
0xRobin: It's not part of the base schema currently.
But now that I think of it, it might be better included there..

macros/models/_sector/nft/enrich_trades.sql
0xRobin: I wonder if we can just include incremental blocks in the macro and if they are handled well then.

If not we can incrementally filter them before they are passed to the macro (so the `transactions_model` is then already filtered in a CTE in `nft_ethereum_trades_beta). This is not yet the case to be clear.

models/_sector/nft/trades/MIGRATION.md
0xRobin: https://dune.height.app/T-24057?task=T-27752

models/_sector/nft/trades/ethereum/platforms/foundation_ethereum_base_trades.sql
0xRobin: Always a fan of standardization, but I think I might prefer `v` prefixed in this case as it forces the column to be interpreted as a string.

models/_sector/nft/trades/nft_ethereum_trades_beta.sql
0xRobin: I think it allows for a bit more flexibility when using the macro (esp during the initial development now)

0xRobin:
Thanks for the thorough review and all the comments @jeff-dude!
I've gone through them all and setup some subtasks for issues/discussions I don't intend to handle in this PR.
(all under https://dune.height.app/T-24057)

You still had some good remarks that I've left unresolved, I'll implement them this week and we can try to move this forward.

models/_sector/nft/trades/schema.yml
jeff-dude: fair

models/_sector/nft/trades/ethereum/platforms/archipelago_ethereum_base_trades.sql
jeff-dude: i think i'm good with `base_` prefix. not sure many users are using these base tables anyway, ultimately just using `nft.trades`

models/_sector/nft/trades/ethereum/platforms/foundation_ethereum_base_trades.sql
jeff-dude: fair, as some `dex.trades` values are actual strings

macros/models/_sector/nft/enrich_trades.sql
0xRobin: :point_up: this worked

jeff-dude (approved):
lfg 🚀

jeff-dude:
fyi @Hosuke -- rob took on a cleanup of the `nft` sector spells & we will be doing the same for `dex` in the near future. i wanted to tag you here to take some time to review and get comfortable in the structure. i can take on the initial build, similar to here, but then we'll need to tackle all underlying base models within `dex` at some point. ideally with some help from wizards, as the model count is high!

Hosuke:
I will try to submit a mock PR for `dex`, maybe I can understand codes in this PR better through mocking.

PR #3144, opened 2023-04-07
add token balances and sol balances daily on Solana

andrewhong5297:
Brief comments on the purpose of your changes: `solana.account_activity` is too large to work with, so getting daily balances should shrink the table and make it much more usable. We can then get a `latest_balance` calculated from this table later on.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


andrewhong5297:
@jeff-dude should be an easy merge when you get the chance - need it for Solana decoding launch next week 🙏 

this table can be updated just once a day (preferably late in the day UTC), so that it doesn't hammer the clusters lol.

jeff-dude:
> @jeff-dude should be an easy merge when you get the chance - need it for Solana decoding launch next week 🙏
> 
> this table can be updated just once a day (preferably late in the day UTC), so that it doesn't hammer the clusters lol.

in the current setup, since it's an incremental spell, it would run every hour with the rest. i'd say we could explore leveraging `solana` tags to put anything on solana into it's own job, but with our current orchestration tool, we're restricted to one compute cluster anyway, so it doesn't really help us at this time.

we have a few solana spells that run hourly. they aren't horrible once past the initial build

models/solana_utils/solana_utils_daily_balances.sql
```diff
@@ -0,0 +1,35 @@
...
+            WHERE tx_success 
+            {% if is_incremental() %}
+            AND block_time >= date_trunc("day", now() - interval '1 week')
+            {% endif %}
+      )
+
+--is there a way to then only keep the most recent "updated_at" for each address?
+SELECT 
+      *
+      , now() as updated_at
```
jeff-dude, line 33: to your commented question, this `updated_at` value will change for each unique key assigned above when it comes through. since we read a week back each run, it'll update a rolling 7-day window of the unique keys every hour. so since `day` is part of the unique key, this value will change more than just `address`

jeff-dude:
> > @jeff-dude should be an easy merge when you get the chance - need it for Solana decoding launch next week 🙏
> > this table can be updated just once a day (preferably late in the day UTC), so that it doesn't hammer the clusters lol.
> 
> in the current setup, since it's an incremental spell, it would run every hour with the rest. i'd say we could explore leveraging `solana` tags to put anything on solana into it's own job, but with our current orchestration tool, we're restricted to one compute cluster anyway, so it doesn't really help us at this time.
> 
> we have a few solana spells that run hourly. they aren't horrible once past the initial build

you can see [here](https://github.com/duneanalytics/spellbook/actions/runs/4642221886/jobs/8215900505?pr=3144#step:12:14) that it's not horrible on an incremental run

models/solana_utils/solana_utils_daily_balances.sql
andrewhong5297: got it - so it won't cause duplicate issues it'll just update this "updated_at" field. Thanks!

jeff-dude (approved):

PR #3159, opened 2023-04-12
OP - Add all ERC20 tokens to erc20.tokens

MSilb7:

~Note: This should only run after https://github.com/duneanalytics/spellbook/pull/3158~

Make tokens_optimism.erc20 match the structure of tokens.nft , where all tokens are identified via transfers, but only curated/known tokens have symbols.

This structure also lets up pull in "algorithmically-generated" tokens (i.e. aave atoken factory) - which we can expand to other protocols (given no contract read solution).

*Q: Is it possible from here to run contract reads on unmapped tokens to fill in symbols and decimals? This would remove all admin work of curating token lists. This is the ideal solution imo.*

For generating the broader tokens.erc20 table, only known tokens are shown.

---

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


MSilb7:
Made an edit for the Aave and Granary tokens file. The error seems to be from not recognizing the jinja.

MSilb7:
bumping - the error seems to be from the test not recognizing dynamic jinja (this part was unchanged).

But also relevant if this is how erc20 tables should be represented (to match nft version). I think it's helpful for tokens table to include all known tokens, then a few are curated or generated. cc @jeff-dude

models/tokens/optimism/tokens_optimism_erc20_generated.sql
```diff
@@ -0,0 +1,56 @@
+{{ config( alias='erc20_generated', tags=['static'],
```
jeff-dude, line 1: seems this one shouldn't be tagged static?

models/yearn/optimism/yearn_optimism_vaults.sql
```diff
@@ -0,0 +1,29 @@
+{{ config(
+        schema = 'yearn'
+        , alias = 'vaults'
+        , materialized = 'table'
```
jeff-dude, line 4: seems this needs to be incremental

models/tokens/optimism/tokens_optimism_erc20_generated.sql
MSilb7: ah yeah copy and paste error probably

models/yearn/optimism/yearn_optimism_vaults.sql
MSilb7: good catch!

MSilb7:
> 

Yeah agree, it will be a weird thing whenever these Aave tables are hit.

Made the table info edits!

jeff-dude (approved):
should be all good

PR #3188, opened 2023-04-17
Work on `opensea.trades` on Polygon 

sohwak:

**For Dune Engine V2**

Modifications:

- Worked on tables related to `opensea.events` on polygon
- Split `opensea.events` by contract tables and unify them in the same way as other nft.trade projects.
  - v2 : opensea data based on existing zeroex
  - v3 : opensea data from Seaport v1.1
  - v4 : opensea data of Seaport v1.4

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:

### DIFF
--- a/models/opensea/polygon/opensea_polygon_events.sql
+++ b/models/opensea/polygon/opensea_polygon_events.sql
@@ -1,150 +1,187 @@
 {{ config(
-    schema = 'opensea_polygon',
-    alias = 'events',
-    partition_by = ['block_date'],
-    materialized = 'incremental',
-    file_format = 'delta',
-    incremental_strategy = 'merge',
-    unique_key = ['block_time', 'unique_trade_id'],
-    post_hook='{{ expose_spells(\'["polygon"]\',
-                              "project",
-                              "opensea",
-                              \'["springzh"]\') }}'
-    )
+         schema = 'opensea_polygon',        
+         alias ='events'
+         )
 }}
 
-{% set nft_start_date='2021-06-28' %}
-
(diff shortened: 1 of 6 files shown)


sohwak:
If this PR (https://github.com/duneanalytics/spellbook/pull/3186) merged in advance, it would help this table a bit.

0xRobin (commented):
@sohwak Thanks for providing some structural equality on the opensea models for different chains! 
I left some small comments on the union view model.

Please also provide a small seed test and a uniqueness test on the `unique_key` columns for each of the sub models.

sohwak:
Hi @0xRobin, I have modified as your suggestions. But, I've held off on adding an empty null column to the underlying table of the union table (opensea_polygon_events) for now.

Usually as the contract changes (v2, v3, v4), it's common to at least keep the old columns (for backwards compatibility), while new columns are added in later versions. I don't think it's a good idea to add the old columns every time. So, I put the later version on top when using `union all`, so that the datatype is determined by the top subquery columns. I know trinosql has problems with implicit datatype conversion, but I thought `NULL` would be fine.

I'd like to hear your thought on this.

0xRobin:
@sohwak I see what you're saying. 
Given these columns are not "core" nft.trades columns but rather opensea specific, it's fine to keep them as is. 
The DuneSQL check in the CI is passing so we should be confident the current setup is ok. :+1:

0xRobin (approved):
:heavy_check_mark:

jeff-dude (approved):

PR #3156, opened 2023-04-12
[Prices.USD] Add 311 missing, but available token prices

bh2smith:
Running the [missing_prices script](https://github.com/cowprotocol/data-misc/blob/main/src/missing_prices.py) displayed the following logs (similar to our missing_tokens script)

```
Getting Coin Paprika token list
Excluded address for 4937 entries out of 53440
Loaded 2927 coins from Coin Paprika
2023-04-12 10:55:37,283 INFO dune_client.client waiting for query execution 01GXTATWX2BNBJ9DA6ADT5JZ5X to complete: ExecutionState.EXECUTING
...
2023-04-12 10:56:08,307 INFO dune_client.client waiting for query execution 01GXTATWX2BNBJ9DA6ADT5JZ5X to complete: ExecutionState.EXECUTING
Fetched 2170 traded tokens from Dune without prices
Found 311 matches
Results written to missing-prices.txt
```

and created this file (which has been pasted here).

[missing-prices.txt](https://github.com/duneanalytics/spellbook/files/11209282/missing-prices.txt)

PR #3193, opened 2023-04-18
remove CAW from prices - permanently

0xRobin:
The CAW token has a price feed that includes wrong data on 27/10/2022, this inflates the usd volumes of dex trades by billions.
This PR removes the token and makes sure it doesn't come back.

It was previously already commented out here:
https://github.com/duneanalytics/spellbook/blob/712907bf4636e6722df718e2766c2dab9f0a6311/models/prices/ethereum/prices_ethereum_tokens.sql#L441
but was re-added in a more [recent PR](https://github.com/duneanalytics/spellbook/pull/3156) by @bh2smith's automatic token scripts. 

Dune queries which show the wrong prices:
https://dune.com/queries/2380231/3903027
https://dune.com/queries/2380260

We can see here that the issue has not yet been resolved by coinpaprika:
https://coinpaprika.com/coin/caw-a-hunters-dream/

cc: @aalan3

### DIFF
--- a/models/prices/ethereum/prices_ethereum_tokens.sql
+++ b/models/prices/ethereum/prices_ethereum_tokens.sql
@@ -1640,4 +1640,5 @@ FROM
 ) as temp (token_id, blockchain, symbol, contract_address, decimals)
 where contract_address not in (
     '0x841fb148863454a3b3570f515414759be9091465' -- SHIH has bad price feed.
+    ,'0xf3b9569f82b18aef890de263b84189bd33ebe452' -- CAW has bad price feed.
 )


bh2smith (approved):
LGMT, Sorry about that! Too bad we have to exclude the whole thing for just one day of bad data.

0xRobin:
Yeah, complete exclusion seems to be our only option for now.. 
Not sure if we're in contact with coinpaprika support to push for a resolution on their side @aalan3?

aalan3:
@0xRobin Not really no but I believe anyone can reach out to them about these things (since coins can be submitted by anyone).

bh2smith:
Also, as an update, it would be good to rebuild the tables that depend on this 

the `cow_protocol_ethereum.trades` table is still showing the bogus valuation for SHIH token removed last week:

https://dune.com/queries/2381418

jeff-dude:
> Also, as an update, it would be good to rebuild the tables that depend on this
> 
> the `cow_protocol_ethereum.trades` table is still showing the bogus valuation for SHIH token removed last week:
> 
> https://dune.com/queries/2381418

interesting, there was a full refresh yesterday. we'll have to see why that is happening. `prices.usd` doesn't return any rows for that symbol

Issue #3030, opened 2023-03-23
[BUG] oneinch_unoswap_v5_ethereum_trades spell contains outlier tx's which explode the amount values


jeff-dude:
### Description

a handful of transactions coming from the source `oneinch_ethereum.AggregationRouterV5_call_unoswap` in the `oneinch_unoswap_v5_ethereum_trades` spell incorrectly join to the `ethereum.traces` base table. the single join condition in question is [here](https://github.com/duneanalytics/spellbook/blob/main/models/oneinch/ethereum/unoswap/oneinch_unoswap_v5_ethereum_trades.sql#L120). the tx's in question:
[0x4f98ac5d5778203a0df3848c85494a179eae35befa64bb6fc360f03851385191](https://etherscan.io/tx/0x4f98ac5d5778203a0df3848c85494a179eae35befa64bb6fc360f03851385191)
[0xce87a97efbf1c6c0491a72997d5239029ced77c9ef7413db66cc30b4da63fe86](https://etherscan.io/tx/0xce87a97efbf1c6c0491a72997d5239029ced77c9ef7413db66cc30b4da63fe86)
[0x62c833c1ab66d17c42aeb1407755c00894f9af8691da2c2ca0f14392e3a6334c](https://etherscan.io/tx/0x62c833c1ab66d17c42aeb1407755c00894f9af8691da2c2ca0f14392e3a6334c)
[0x774ad4c15a6f776e71641fe4e9af3abd5bb80f7511c77548d130c2ee124ba80a](https://etherscan.io/tx/0x774ad4c15a6f776e71641fe4e9af3abd5bb80f7511c77548d130c2ee124ba80a)

**_please note:_** this is also an issue on v1 engine, so it's not an issue with migration to v2 engine, but rather the logic itself. the "quick fix" on v1 engine was to just force the `usd_amount` to `0` to remove outlier values that blow up dashboard visuals.
### Current behavior

all transactions from `oneinch_ethereum.AggregationRouterV5_call_unoswap` outside of the linked above work fine within the context of the current code join conditions. in the initial build of this spell, some of the tx's were manually filtered out as they seemed to be outliers. then new ones appeared every so often and are not filtered out, therefore showing high `amount_usd` for a few transactions.

[...]

PR #3197, opened 2023-04-18
add oneinch to dex aggregator

jeff-dude:
there is still an outstanding issue for oneinch #3030 

we don't have a solution quite yet, but this issue also persists in the original code from v1 engine. we can move forward knowing that the bug needs resolved and short-term fixes may be necessary to filter out tx's which blow up `amount_usd`

~~edit: looking to add test for `amount_usd`, but will require thought on how to run in prod~~
will follow up with new PR to add tests


augustog (approved):
Lgtm!

bh2smith:
I think that putting left commas on the erc20 token list did not belong here and should be reverted. If we want to go left commas its fine, but not without warning and in a totally irrelevant PR!

jeff-dude:
> I think that putting left commas on the erc20 token list did not belong here and should be reverted. If we want to go left commas its fine, but not without warning and in a totally irrelevant PR!

you're correct, that was oversight on my part. we'll work together to get this back in a solid state.

jeff-dude:
fyi @gentrexha -- i do believe you wanted a heads up once this data was available

gentrexha:
> fyi @gentrexha -- i do believe you wanted a heads up once this data was available

Yes, I did. Thanks!

PR #3198, opened 2023-04-19
Stealcam model fix

0xRobin:
Fixes the stealcam duplicates by doing some paper math and not relying on traces which break on edge cases. 

royalty and platform fees are now calculated based on the rules layed out in:
https://www.stealcam.com/faq


jeff-dude (approved):
thank you 🙏

PR #3212, opened 2023-04-21
WIP [oneinch] add static contract addresses

grkhr:
**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


augustog:
hey @jeff-dude, trying to confirm the approach here. If I understand correctly (@grkhr correct me if I'm wrong) they are trying to add this static file and use it from DSQL, which means we should use strings here and explicit cast them to varbinary on the overrides, no?

jeff-dude:
> hey @jeff-dude, trying to confirm the approach here. If I understand correctly (@grkhr correct me if I'm wrong) they are trying to add this static file and use it from DSQL, which means we should use strings here and explicit cast them to varbinary on the overrides, no?

yes, for the time being, the approach is to write spells compatible in spark sql (added bonus if compatible on both). once merged, if we want any field to be varbinary specifically, we'll handle that in a separate private repo. more to come on this process to leverage dunesql types, where it's not hidden behind private repo

jeff-dude:
i've pushed a commit to standardize, hopefully assumptions made are fine -- otherwise, feel free to overwrite
- added column alias to get past failure
- removed usage of `metadata` in schema/table naming, keep clean
- added a few tests (unique address, accepted values for blockchain)

jeff-dude:
> i've pushed a commit to standardize, hopefully assumptions made are fine -- otherwise, feel free to overwrite
> 
> * added column alias to get past failure
> * removed usage of `metadata` in schema/table naming, keep clean
> * added a few tests (unique address, accepted values for blockchain)

looks like combo of blockchain/address still isn't unique -- is this intended? if so, we can remove. otherwise, could clean up the file

grkhr:
Removed duplicates. It wasn't intended, I just didn't notice, my bad here. @augustog @jeff-dude 

Everything is great in my opinion. Thanks for the cleanup. 

Can be merged.

jeff-dude (approved):
thank you!

PR #3213, opened 2023-04-21
add accepted range usd test on dex/nft schema

jeff-dude:
tagging mostly for awareness @Hosuke @0xRobin 

we're looking to add a minor test to help be proactive finding irregular `amount_usd` values. majority of issues come down to API data from coinpaprika being incorrect, but tokens bought/sold being flipped can also impact due to decimal sizes in equations.


jeff-dude:
two failures:
1 - dex trades has three rows which exceed 10bil -- todo: track them down and validate
2 - nft mints has high number of failures, this seems irregular since events/trades didn't fail. research to be done there

0xRobin:
I plan to revisit the `nft.mints` so can investigate these irregularities at that moment.

There are more irregular `dex.trades` rows if we lower the threshold (eg. to 1 bil).
See also the recent convo on discord.

Are we planning to fo more frequently test runs and monitoring if we introduce this? (we haven't run tests since March 7th)

jeff-dude:
> This is great. I plan to revisit the `nft.mints` so can investigate these irregularities at that moment.
> 
> There are more irregular `dex.trades` rows if we lower the threshold (eg. to 1 bil). See also the recent convo on discord.
> 
> Are we planning to do more frequently test runs and monitoring if we introduce this? (we haven't run tests since March 7th)

we can explore lowering the threshold too. the idea would be to put test commands in the hourly job. we may surpass the allotted time with this extra step though (in addition to new prices step). i also want to explore splitting the job up across a few (dex, nft, etc), allocating certain thread counts to each.

jeff-dude:
progress here:
- lowered the accepted range threshold down from $10b -> $1b
- manually ran the tests locally, identified tx's which failed, all came down to bad price feeds (no surprise)
- added plenty more tokens to the permanent exclusion list in prices spell(s)

next steps:
- plan how to merge this, with main thoughts around where to run `dbt test` command, if we should split up jobs in production to remain in the hourly refresh rate, how to allocate threads for performance across jobs in prod, etc.

long term:
- in dex/nft redesign efforts, update logic on `amount_usd` to first look for large mcap tokens (i.e. ETH, USDC, etc etc), use those if applicable as the price feeds are much more reliable, then fall back onto a secondary check where both token bought/sold are calculated and the accurate response is used

jeff-dude:
merging now to get bad price feeds removed, usage of tests in prod to come soon.
@Hosuke @0xRobin in case you see these tests fail on PRs CI test:
- could be prod hasn't fully refreshed yet (should be in a few hours)
- could be new bad price feeds, first place to look
- could be `nft.mints`, which is a known issue at this time and should be fixed in redesign

PR #3253, opened 2023-04-28
Big NFT sector cleanup!

0xRobin:
This PR cleans up the NFT trades sub sector spells.

Baiscally it reduces every platform specific models to only the "events" models, removing the "trades", "fees", "burns", "mints" models which should all just be a view on the "events" model. 

We had some inconsistencies with platforms only implementing some of these tables which resulted in platforms being present in events but not in trades or fees,..

New setup:
- Every platform implements just the "events" model
- All platforms are combined in the `nft_events` model
- 'nft_trades', `nft_fees`, `nft_burns` and 'nft_mints' are views on top of `nft_events`

This cleanup will also help a lot during the ongoing NFT restructuring migration.

There's a small impact on users as models like `blur.fees` don't exist anymore after this is merged and they should use `blur.events`or the overall `nft.fees` model.

### **Part 2:**
1. Moved all the "old" (but still in use!) models under the '_sector/nft` directory structure.
2. added explicit schema's to those models
3. removed the `expose_spel` from those marketplace events models
4. split any marketplace union views in just the seperate version models (except opensea, that's a mess)
5. cleaned up dbt_project
6. introduces `<marketplace>.trades` view, which is the only marketplace specific view that will get exposed on the App


0xRobin:
Here's a comparison between the production tables and the tables in the PR to confirm I've not forgotten any specific platform.
https://dune.com/queries/2424413
image

0xRobin:
:heavy_check_mark: CI checks are actually passing. 
Only checks not passing are DuneSQL checks for models that are currently already hidden in Trino.

0xRobin:
:heavy_check_mark:  checks are passing expect dunesql checks on big models due to java heap space.

couralex6 (approved):

jonathandune (approved):

dbt_project.yml
```diff
@@ -326,31 +286,11 @@ models:
...
-        +schema: magiceden_polygon
-      solana:
-        +schema: magiceden_solana
-
     nomad:
       +schema: nomad
       ethereum:
         +schema: nomad_ethereum
 
-    ellipsis_finance:
```
jeff-dude, line 349: this one a mistake?

jeff-dude:
can you fix merge conflicts and let CI run again?

i gave a passthrough of the code, i think i follow the general flow:
- cleanup project file by removing directories which are being moved into the `_sector` directory, continuing cleanup we've discussed
- change `nft.events` to be view on top of all the "old" events base models, which were moved into new directory structure and clearly labeled `old`
- remove old base models from data explorer
- create platform-specific views on top of existing `nft.trades` spell
- nft trades/burns/fees also views, similar to events, kept to sector-level
(follow up: but why is fees/trades on `nft_events` while burns is on `nft_events_old`?)
(also, is mints missing?)
- remove and/or edit all the old directories that fed nft sector
- clean up seeds
- clean up tests

dbt_project.yml
0xRobin: no, it was defined double in the `dbt_project` file.

jeff-dude (approved):
lets push this and only this through, let it ride in prod

0xRobin:
awesome, I'll monitor the next hourly run :+1:

PR #3275, opened 2023-05-03
Fix eth erc20 address

0xRobin:
Follow up PR to https://github.com/duneanalytics/spellbook/pull/3271
This re-enables the ETH ERC20 address functionality for the restructured models without impacting existing ones.

### DIFF
--- a/models/_sector/nft/trades/nft_ethereum_trades_beta.sql
+++ b/models/_sector/nft/trades/nft_ethereum_trades_beta.sql
@@ -24,7 +24,7 @@ WITH cte_prices_patch as (
     {% endif %}
     UNION ALL
     SELECT
-        '{{ var(ETH_ERC20_ADDRESS) }}' as contract_address
+        '{{ var("ETH_ERC20_ADDRESS") }}' as contract_address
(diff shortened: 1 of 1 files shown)


models/_sector/nft/trades/nft_ethereum_trades_beta.sql
```diff
@@ -24,8 +52,11 @@
...
         tokens_erc20_model=ref('tokens_ethereum_erc20'),
-        prices_model=ref('prices_usd_forward_fill'),
+        prices_model='cte_prices_patch',
         aggregators=ref('nft_ethereum_aggregators'),
         aggregator_markers=ref('nft_ethereum_aggregators_markers')
     )
 }}
+)
+
+select * from enriched_trades
```
jeff-dude, line 62: now both the macro & this model have this select -- is that intentional?

0xRobin: yes, it's due to the 2 levels of CTE now present

jeff-dude (approved):

PR #3283, opened 2023-05-05
Fix Balancer Trades

viniabussafi:
Brings changes to the trade queries on all blockchain, reflecting the original dune v1 abstractions https://github.com/duneanalytics/dune-v1-abstractions/blob/main/deprecated-dune-v1-abstractions/ethereum/dex/trades/insert_balancer.sql#L102


models/balancer/arbitrum/balancer_arbitrum_trades.sql
```diff
@@ -30,9 +30,13 @@ with v2 as (
     on s.poolId = p.poolId
     {% if not is_incremental() %}
         where s.evt_block_time >= '{{project_start_date}}'
+        AND CAST( token_bought_address as VARCHAR) != CAST( project_contract_address AS VARCHAR)
```
jeff-dude, line 33: please see error [here](https://github.com/duneanalytics/spellbook/pull/3283/checks#step:9:72)

also, `token_bought_address` is just an alias in this query, you'll need to use the actual name until outside the CTE: `tokenOut` -- same concept for sold

another note -- since this filter should apply on both incremental and full historical runs, i'd suggest moving the where clause out of the incremental block and then making the block_time filter use `and` instead, since `where` will already be established

jeff-dude (changes requested):

models/balancer/arbitrum/balancer_arbitrum_trades.sql
viniabussafi: Thank you for your help, Jeff! Hopefully, everything is okay this time!

Issue #2756, opened 2023-02-22
Bug: Blurs offer accepted shows wrong currency information


0xRobin:
### Description
Blur sells show WETH as currency used. The real ERC20 used in these trades is the [Blur Pool token ](https://etherscan.io/token/0x0000000000a39bb272e79075ade125fd351887ac).

### Current behavior
There's separate logic for the Blur pool address to consider it the same as native ETH.
https://github.com/duneanalytics/spellbook/blob/4f2e482d48ab5af1b1fc0e83066668885883fc17/models/blur/ethereum/blur_ethereum_events.sql#L40-L51

### Expected behavior
We should report the currency contract and symbol correctly, both in the trade currency and the royalty_fee currency.
example: https://etherscan.io/tx/0x2af35739e7d37d05d0f210435e90c35e6779d6a2f28f04f4a657889818d2a9c5
![image](https://user-images.githubusercontent.com/83790096/220593129-6f295958-102c-4943-bcfb-7a7fb1089af3.png)

### Concerned model(s)
```
blur_ethereum.events
```

### Impacted model(s)
```
nft.events
nft.trades
nft.fees
```
~540k rows affected
```sql
select count(*) from blur_ethereum.events
where trade_category = 'Offer Accepted'
```

### Possible solution
Properly integrate the Blur Pool token in the ERC20 tokens and prices models so we don't need custom logic.

ytoast:
Hi, is this issue still relevant? Can I work on this?

0xRobin:
hey @ytoast, I'm currently in the middle of restructuring the Blur model to a new setup for NFT trades. 
I'll probably tackle this issue while I'm at it.

ytoast:
@0xRobin okay! are there any other nft related issues that I can work on at the moment? Would love to contribute!

0xRobin:
Yes, all of them will be affected.
So it's maybe a bit of an unfortunate timing at the moment.

I'd say give it 2-3 more weeks and then try to tackle some of the open NFT issues.
Any contributions will be really appreciated, but I wouldn't like to see any of your effort go to waste by bad timing.

ytoast:
Okay sure, thank you I appreciate it!

PR #3284, opened 2023-05-05
Restructure Blur to new NFT trades setup

0xRobin:
Blur trades restructuring.

Also tackles Issue: https://github.com/duneanalytics/spellbook/issues/2756

* [x] row-count test passing
* [x] seed test passing
* [x] migration validation test passing


0xRobin:
@jeff-dude all tests relevant to this PR are now passing.
Maybe we should also exclude the new tag in the github CI runs so they don't show up in every PR?

PR #3290, opened 2023-05-05
add prod_exclude tags for models to disable in production

jeff-dude:
fyi @0xRobin -- trying out something new for excluding in prod to keep orchestration side clean. there are a few ways to achieve this, but it's a quick and clean one.


0xRobin:
good idea :+1: . 
If we like this setup, it's probably worth it to convert to full yaml selector definitions later which will keep the exclusion list in 1 place instead of scattered through model configs.

Issue #3309, opened 2023-05-08
[BUG] Duplicates in odos trades


ytoast:
### Description

Some context: This is related to https://github.com/duneanalytics/spellbook/pull/2853. This is a known issue but I thought I'd document it here since I spent some time looking at it. I was looking into this as I noticed that the [uniqueness check for odos_avalanche_c_trades](https://github.com/duneanalytics/spellbook/blob/05544a9a4a7c33ffe74026fccbeb16aa34f25b7f/models/odos/avalanche_c/odos_avalanche_c_schema.yml#L15) is failing. Example [here](https://github.com/duneanalytics/spellbook/actions/runs/4915383239/jobs/8777815949?pr=3306).

Main reason:
1. **Odos performs batch swaps**. Each swap transaction here is the equivalent of multiple swap trades in the more "traditional" swap protocols. For example, Odos will allow users to swap token_A to token_B and token_C in a single swap trade. Compare this to something like Uniswap, where swapping token_A to token_B and token_C can only take place in 2 swap trades. This introduces complexity into the current logic used to normalise the swaps table

### Current behavior

There are 2 problems in the current table:
1. incorrect `token_bought_amount` - USDC has the wrong token_bought_amount
2. incorrect `token_sold_amount` - the sum of both rows should be 12.2
https://dune.com/queries/2452828
<img width="700" alt="image" src="https://user-images.githubusercontent.com/52529078/236859020-3464f47a-2688-466c-b8f6-7bda32e5f826.png">

|token_pair|token_bought_amount|token_sold_amount|
|--|--|--|
|USDC-WAVAX| 0.350273 |12.2|
|BTC.b-WAVAX| 0.00350273 |12.2|

### Expected behavior

Source of truth: https://snowtrace.io/tx/0x48bb32dc44006204e851bac988260b779918e1f4bcdddb86102f36ecb5b68846
|token_pair|token_bought_amount|token_sold_amount|
|--|--|--|
|USDC-WAVAX| 105.245739 |6.2242610707|
|BTC.b-WAVAX| 0.00350273 |5.975738929258659482|

### Impacted model(s)

```
odos.avalanche_c.trades
```

### Possible solution

There are a few changes to be made:
1. Fix https://github.com/duneanalytics/spellbook/blob/main/models/odos/avalanche_c/odos_avalanche_c_trades.sql to ensure that accurate `token_bought_amount` and `token_sold_amount` are reflected
[...]

PR #2853, opened 2023-03-01
odos dupes - wip 

henrystats:
Brief comments on the purpose of your changes:

**For Dune Engine V2**

I've checked that:

### General checks:
* [ ] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each sql file is a select statement and has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [ ] `coin_id` represents the ID of the coin on coinpaprika.com
* [ ] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:
* [ ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:
* [ ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ ] where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #3310, opened 2023-05-08
fix odos duplicate issue

ytoast:

ODOS supports batch swaps. This previous logic only retrieves the data relevant to the first token bought and first token sold. This logic will break when the user performs a batch swap. 

Related to https://github.com/duneanalytics/spellbook/issues/3309. 

I have decided to reformat the schema of Odos trades. The main changes are:

* Addition of `tokens_bought` and `tokens_sold` ARRAY<STRUCT> fields. This fields will contain an array of token information. This will allows each row to represent a single trade, while encapsulating details of the trade into those 2 columns (see struct type [here](https://github.com/duneanalytics/spellbook/pull/3310/files#diff-074973819ca7750a2bba1b7659c0582cede986a42d4229145026b3a4d2e829bdR42-R50)). 
* Removal of all `token_bought`/`token_sold` columns as they are now encapsulated in the new `tokens_bought` and `tokens_sold` fields

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


Hosuke:
Thank you @ytoast , hopefully you can find a solution to odos batch swap.
Some works you can recycle here: https://github.com/duneanalytics/spellbook/pull/2853

ytoast:
Tests are failing (`models/odos/odos_trades.sql`). I think it's because I'm changing the underlying schema

Hosuke:
> Tests are failing (`models/odos/odos_trades.sql`). I think it's because I'm changing the underlying schema. @Hosuke, would you know how I can resolve this?

Since odos.trades should be under dex_aggregator.trades meta-table, it should have all required fields/columns in its roll-up spell, and its base models.

If you change the fields in the roll-up spell, you may need to change the output of base models as well, and in the same time odos.trades cannot fit into dex_aggregator.trades.

I suggest that only modify the base models. However it may be hard to solve dupes in odos.trades using current unique_key, as odos.trades may have batch swaps having the same unique_key.

ytoast:
@Hosuke The current fields required in the roll-up spell (dex_aggregator.trades) is insufficient, and will not be able to capture what is happening in an odos transaction. The current schema for the roll-up spell only allows us to capture information from a single-swap transaction. And hence I was proposing to change the schema for the base odos model. This way, the odos model is guaranteed to capture accurate information - with each row representing a transaction, regardless of whether it is a batch swap or a single swap.

But I recognise the concern - which is that the new schema proposed will not be able to join the roll-up spell. I did consider this and decided that it would be better to represent accurate information in the odos spell instead.

Seems like a better solution will be to change the schema for the base models and the rollup spell too.

I can create a separate PR to also add the `tokens_bought` and `tokens_sold` fields (ARRAY<STRUCT>) to the other dexs spells. What do you think?

Hosuke:
> @Hosuke The current fields required in the roll-up spell (dex_aggregator.trades) is insufficient, and will not be able to capture what is happening in an odos transaction. The current schema for the roll-up spell only allows us to capture information from a single-swap transaction. And hence I was proposing to change the schema for the base odos model. This way, the odos model is guaranteed to capture accurate information - with each row representing a transaction, regardless of whether it is a batch swap or a single swap.
> 
> But I recognise the concern - which is that the new schema proposed will not be able to join the roll-up spell. I did consider this and decided that it would be better to represent accurate information in the odos spell instead.
> 
> Seems like a better solution will be to change the schema for the base models and the rollup spell too.
> 
> I can create a separate PR to also add the `tokens_bought` and `tokens_sold` fields (ARRAY) to the other dexs spells. What do you think?

I agree with your idea.

I am thinking for the first step, we exclude the odos.trades from `dex_aggregator.trades` meta-table.

And you can redesign/reimplement odos.trades and test thoroughly, in this PR.

If everything looks good for odos.trades, then we continue to other dex_aggregator.trades spells.

ytoast:
@Hosuke `odos` has alrdy been left out of the `dex_aggregator_trades` spell in a previous commit bcos of the duplicates. I updated the PR to also remove odos from the seed file

models/odos/odos_trades.sql
```diff
@@ -21,16 +21,9 @@ FROM (
...
-        token_pair,
-        token_bought_amount,
-        token_sold_amount,
-        token_bought_amount_raw,
-        token_sold_amount_raw,
         amount_usd,
-        token_bought_address,
-        token_sold_address,
+        tokens_bought,
+        tokens_sold,
```
Hosuke, line 26: Since tokens_bought/tokens_sold is not defined in the base models, you may need to put columns available in base models here to fix.
https://github.com/duneanalytics/spellbook/actions/runs/5169331430/jobs/9311513716?pr=3310#step:9:22

Hosuke (changes requested):

models/odos/odos_trades.sql
ytoast: I did add these columns in the base models. I see that the error here says `ERROR creating incremental model test_schema.git_ab46cf33_dex_aggregator_trades `. Could the error be due to the model trying to build incrementally? Incremental load wont work now because I change the underlying base model. I need to somehow run dbt with a full refresh for this model

PR #3339, opened 2023-05-12
ENH: New table to facilitate matching  `chain_id -> chain_name` in prices tables

cc7768:

Many events on-chain (especially bridge events) track chains by their chain id rather than the identifier used in the `prices.usd` table. This PR adds a new table to the `chain_info` schema called `dune_chains` which is a simple mapping from `chain_id` to `dune_name` where `dune_name` corresponds to the naming conventions in the prices table.

We are using a version of this internally and thought it would be of more general use.

**For Dune Engine V2**

I've checked that:

### General checks:


Hosuke (approved):
LGTM.✅
Thank you @cc7768 

chain_id checked in https://chainlist.org/ except id for `solana`

cc7768:
Cheers! Thanks for the review.

That Solana chain id came up when I googled but I'm not sure it's a thing tbh. Maybe we should remove it unless we have a more reputable source than "I googled it and saw a single result" :sweat_smile: 

Thoughts? I'll open a suggestion with that change and you can accept it if you agree.

models/chain_info/chain_info_dune_chains.sql
```diff
@@ -0,0 +1,23 @@
...
+  VALUES
+    ('ethereum', 1),
+    ('optimism', 10),
+    ('bnb', 56),
+    ('gnosis', 100),
+    ('polygon', 137),
+    ('fantom', 250),
+    ('arbitrum', 42161),
+    ('avalanche_c', 43114),
+    ('solana', 1399811149)
```
cc7768, line 22:
```suggestion
```

Hosuke: We can add back solana id when we have more confident source.

jeff-dude (approved):
works for me -- thank you for the contribution and making general purpose!

PR #3341, opened 2023-05-14
needs to use {{this}} format for incremental

andrewhong5297:

We were getting duplicates with this always running with one day of history, I've changed it to only run if the block_time is in the next day.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


andrewhong5297:
```
Expected a schema version of "https://schemas.getdbt.com/dbt/manifest/v6.json" in manifest.json, but found "https://schemas.getdbt.com/dbt/manifest/v8.json". Are you running with a different version of dbt?
```

hmm, seems like something is wrong with config? cc @jeff-dude

Hosuke (approved):
{{this}} provides a more precise incremental updating option.
Original idea is having an overlapping region when incremental update, to ensure 100% coverage of new incoming records, and dbt uses `unique_key` in the config block to filter out the duplicated/existed records when updating.

models/solana_utils/solana_utils_daily_balances.sql
```diff
@@ -24,7 +24,7 @@ WITH
             FROM {{ source('solana','account_activity') }}
             WHERE tx_success 
             {% if is_incremental() %}
-            AND block_time >= date_trunc("day", now() - interval '1 day')
+            AND block_time > (SELECT max(day) FROM {{this}})
```
0xRobin, line 27: should probably use `>=` here to not drop later incremental updates from the same day.

0xRobin:
CI issues should be resolved now, I've triggered a re-run

jeff-dude:
can you confirm the failing test on duplicates?

andrewhong5297:
> can you confirm the failing test on duplicates?

Looking

andrewhong5297:
the test schema is gone, but I really can't see any reason there would be duplicates given we're using a window function.

andrewhong5297:
@jeff-dude I'm like 100% confident there should be no duplicates - can we merge?

Hosuke:
I think I know the cause of the bug,
there is a tiny time diff in `updated_at`(1st time in full-refresh, 2nd time in incremental refresh).
Perhaps the engine is hard to compare NULL value for `token_mint_address`.
Here is an example:
https://dune.com/queries/2493263?d=1

Hosuke:
I tried to fix the problem by giving a default empty string `''` to NULL `token_mint_address`, and currently empty string is not equal to NULL in some versions of spark.

I actually found out this problem is an opening issue in dbt: https://github.com/dbt-labs/dbt-core/issues/7597

andrewhong5297:
hmm odd. I've run a group by check and got no nulls https://dune.com/queries/2479876?d=1&q=v2+&sidebar=query-explorer

andrewhong5297:
thank you @Hosuke !!

jeff-dude:
> I tried to fix the problem by giving a default empty string `''` to NULL `token_mint_address`, and currently empty string is not equal to NULL in some versions of spark.
> 
> I actually found out this problem is an opening issue in dbt: [dbt-labs/dbt-core#7597](https://github.com/dbt-labs/dbt-core/issues/7597)

great catch! we've seen this elsewhere before too. it's a good thing to keep in mind, that unique keys can't be null or the merge statement fails the join condition, leading to duplicate inserts. i wonder if this means we need to apply `not null` tests to each unique key column 🤔

jeff-dude:
seems like we're getting failures where not null tests were applied (assuming the unique keys)

andrewhong5297:
> seems like we're getting failures where not null tests were applied (assuming the unique keys)

Oh token_mint_address and token_balance can be null, those tests should he removed. Just address and balance should have not null tests.

jeff-dude:
what's the move here?

jeff-dude:
closing PRs marked as `wip` state, prior to the freeze announced and tracked [here](https://github.com/duneanalytics/spellbook/discussions/3558)

PR #3355, opened 2023-05-16
Rename onceinch_lop_v2 to oneinch schema for decoded tables

antonio-mendes:
The 1inch team requested a rename of some decoded tables that were under the wrong namespace `oneinch_lop_v2`. I've done the rename for the tables and this will update the spells to reflect that.

Same as in the Lido case before, I have both the old table and new table existing and will only drop the old ones after this is merged.


antonio-mendes:
No worries, there's no big urgency @jeff-dude ! Thanks again!

PR #3360, opened 2023-05-18
Add Flashloans to NFT Wash Filter

hildobby:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


0xRobin (approved):
@hildobby :heavy_check_mark:  this looks good.

side note:
would you be interested in having these models implemented in a macro? 
(1 macro that's then used for all the different chains, so you'd only need to change the logic in 1 place)
Or do you see the logic diverging a bit between different chains? In that case we'd need to keep the implementation chain specific.

hildobby:
> you be interested in having these models implemented in a macro?
> (1 macro that's then used for all the different chains, so you'd only need to change the logic in 1 place)
> Or do you see the logic diverging a bit between different chains? In that case we'd need to keep the implementation chain specific.

I excluded Solana when making it crosschain since a lot of the fields were missing there for Solana trades and the filter thus didn't work (it was all `is_wash_trade=true` since buyer and seller were both NULLs and thus equal for example). I'm guessing that decoded Solana data changes things though, once Solana marketplaces have their own spell and necessary fileds are not NULL, the implementation should be the same afaik, but my Solana knowledge is limited. I would maybe keep it as is for now unless you view the change as necessary.

What do you think about keeping current structure and adding `is_wash_trade` in `nft.trades`? And have it be NULL for now for non-EVM chains. Imo many would find it practical, and we could also keep `nft.wash_trades` to keep the other per filter columns there for those who want to dig further (and also not overcrowd `nft.trades` with too many new columns)

jeff-dude (approved):

PR #3363, opened 2023-05-18
NFT deprecation 1

0xRobin:
This PR deprecates the older models for marketplaces that are restructured in the new base trades models. 

This means the "source of data" for these marketplaces also switches over the the new models and any future logic changes or bugfuxes should be made into the new models. 
For those marketplaces, this is the last step in the "NFT restructuring" process.

All current Ethereum mainnet marketplaces have been restructured expect `opensea`, `trove` and `liquidifty`


0xRobin:
@jeff-dude the only CI failures are dunesql checks on:
`cryptopunks_ethereum_current_bids`
`cryptopunks_ethereum_floor_price_over_time`
`cryptopunks_ethereum_listings_over_time`

All of which have `expose_spells_hide_trino` post_hooks

0xRobin:
@augustog @couralex6 
This is my final NFT restructuring PR which I'd wanted to include before starting on the trino migrations. 
If we can get a rebase on `main` into trino-spellbook after this is merged that would be perfect.

jeff-dude:
left this out of the batch today, will look to include tomorrow 👍

PR #3368, opened 2023-05-18
cast number of items

dot2dotseurat:

I'm not sure where this issue is introduced but number_of_items is listed in the data explorer as decimal but then seems to be actually varchar based on some query errors. I will validate this make the field actually decimal in the gitsha'd version of the table. 

short term solution as this table is being restructured.


dot2dotseurat:
closing since this issue appears to be upstream

PR #3388, opened 2023-05-22
[CoW] Parse raw_internal_imbalances table

bh2smith:
Currently the types of `cowswap.raw_internal_imbalance` are inferred as strings for large integer fields and hex (like transaction hashes and addresses). We write this spell to cast the types to what one would expect so querying can be made cleaner.

Not sure yet if we can write DuneSQL in spells....

cc @olgafetisova


jeff-dude:
> Not sure yet if we can write DuneSQL in spells....

not quite yet

bh2smith:
> not quite yet

In that case, I would rather wait... any estimate on when? Like days, weeks or months?

jeff-dude:
> > not quite yet
> 
> In that case, I would rather wait... any estimate on when? Like days, weeks or months?

@augustog got a timeframe to share?

jeff-dude:
closing this for now, but we plan to start migration v soon, either next week or the following 🤝

PR #3391, opened 2023-05-22
buffer flow

ppclunghe:
After Lido V2 upgrade we want to make the bunch of changes in accounting model, in this update we add ETH inflow/outflow for protocol buffer
Made this PR instead of 3372

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


models/lido/accounting/ethereum/lido_accounting_ethereum_buffer_inflow.sql
```diff
@@ -0,0 +1,27 @@
+{{ config(
+        alias ='buffer_inflow',
+        partition_by = ['period'],
```
jeff-dude, line 3: if i'm reading the spell correctly, this is partitioning by `evt_block_time`? that is much too granular for partitions. i'd suggest either removing this config option or introducing a new column in the select below that is at minimum day (or week, month, etc) to partition by

models/lido/accounting/ethereum/lido_accounting_ethereum_buffer_outflow.sql
```diff
@@ -0,0 +1,16 @@
+{{ config(
+        alias ='buffer_outflow',
+        partition_by = ['period'],
```
jeff-dude, line 3: same idea here

jeff-dude (changes requested):

models/lido/accounting/ethereum/lido_accounting_ethereum_buffer_inflow.sql
ppclunghe: thank you for review, fixed config with new "day" column

jeff-dude: great! we'll let the test run, then hopefully all set

gregshestakovlido: It seems that test execution takes too long and fails after 90 minutes. @jeff-dude, maybe removing partition will help here or it is something in test flow?

jeff-dude: i don't think it's due to your spells in the PR. for some reason, our gh action ci test is picking up more models than it should and causing longer runtimes, then timeouts. i'm trying to look into that now. i would assume no changes here, just waiting for resolution there and a successful run after

jeff-dude (approved):

models/lido/accounting/ethereum/lido_accounting_ethereum_buffer_inflow.sql
jeff-dude: all good, unrelated issue resolved

PR #3392, opened 2023-05-23
Enable vacuum on spells

a-monteiro:
This enables periodic vacuum of spells.

### DIFF
--- a/macros/expose_spells.sql
+++ b/macros/expose_spells.sql
@@ -7,7 +7,8 @@
         'dune.data_explorer.category'='abstraction',
         'dune.data_explorer.abstraction.type'= '{{ spell_type }}', -- 'project' or 'sector'
         'dune.data_explorer.abstraction.name'= '{{ spell_name }}', -- 'aave' or 'uniswap'
-        'dune.data_explorer.contributors'= '{{ contributors }}'   -- e.g., ["soispoke","jeff_dude"]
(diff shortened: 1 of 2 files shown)


aalan3 (approved):

macros/expose_spells.sql
```diff
@@ -8,6 +8,7 @@
         'dune.data_explorer.abstraction.type'= '{{ spell_type }}', -- 'project' or 'sector'
         'dune.data_explorer.abstraction.name'= '{{ spell_name }}', -- 'aave' or 'uniswap'
         'dune.data_explorer.contributors'= '{{ contributors }}'   -- e.g., ["soispoke","jeff_dude"]
+        'dune.vacuum' = '{"enabled":true}'
```
jeff-dude, line 11: i think you'll need a comma `,` on line 10 to signal this new line

one thing to test locally:
- run dbt compile
- grab the output in `target` directory for the macro, i.e. the alter table sql statement
- try to run in spark for a random model

i imagine this will fail without the comma. i don't think our gh action ci test has great coverage for these types of changes that alter a table, or if it does, you can see it ran all 700+ models that use this macro and timed out. you'll be better off quickly manually testing

jeff-dude (changes requested):

jeff-dude (approved):

PR #3397, opened 2023-05-23
Hashflow cross chain trades

BroderickBonelli:

Adding specific Hashflow Cross Chain Trades models for deeper analysis of cross chain trades and cross chain trade activity. Contains additional data specific to cross chain trades such as destination chain and source chain fields that are not included in other Hashflow trade models upstream of `dex.trades`. These new models will enable easy tracking of token volume across chains, aggregate inflows/outflows by chain, etc. 

**For Dune Engine V2**

I've checked that:

### General checks:
* [X] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [X] `coin_id` represents the ID of the coin on coinpaprika.com
* [X] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:

### Incremental logic:


models/hashflow/hashflow_crosschain_trades.sql
```diff
@@ -0,0 +1,26 @@
...
+    [
+        ref('hashflow_avalanche_c_crosschain_trades')
+        ,ref('hashflow_ethereum_crosschain_trades')
+        ,ref('hashflow_bnb_crosschain_trades')
+    ]
+%}
+
+{% for ref in cross_chain_models %}
+SELECT 
+    *
```
Hosuke, line 21: We may need to rewrite `select *` in full column names.

Hosuke (changes requested):
Hi @BroderickBonelli,
do you think these trades can be added to `dex.trades` meta-table?
If so, we may need to implement all columns dex.trades has.

BroderickBonelli:
> 

Hey @Hosuke , thanks for having a look. These cross chain trade events are already included in the other hashflow trades models that roll into `dex.trades`, they just don't include the cross-chain specific data such as destination chain and source chain for example. I was wary to add the cross-chain specific data to the `dex.trades` table as it will be `NULL` for 95%+ of records within that table, so figured it was best to split it out into a separate model.

models/hashflow/avalanche_c/hashflow_avalanche_c_crosschain_trades.sql
```diff
@@ -0,0 +1,103 @@
+{{
+    config(
+        alias="avalanche_c_cross_chain_trades"
```
Hosuke, line 3: Since the schema is `hashflow_avalanche_c`,
we can just have alias as `cross_chain_trades` here.

The same goes to other base models as well.

models/hashflow/hashflow_crosschain_trades.sql
Hosuke, line 21: Hi @BroderickBonelli,
could you please write full column names in this roll-up spell instead of `*`?

It can avoid mismatch of columns.

Hosuke (changes requested):

models/hashflow/hashflow_crosschain_trades.sql
BroderickBonelli: Hi @Hosuke, no problem, just amended to list out column names.

Hosuke (approved):
LGTM.✅
Thank you @BroderickBonelli 

[Sample hashflow.cross_chain_trades](https://dune.com/queries/2530626)

models/hashflow/avalanche_c/hashflow_avalanche_c_crosschain_trades.sql
```diff
@@ -0,0 +1,103 @@
...
+        SELECT
+            evt_block_time           AS block_time
+            ,trader                  
+            ,quoteTokenAmount        AS token_bought_amount_raw
+            ,baseTokenAmount         AS token_sold_amount_raw
+            ,cast(NULL AS double)    AS amount_usd
+            ,quoteToken              AS token_bought_address
+            ,baseToken               AS token_sold_address
+            ,evt_tx_hash             AS tx_hash
+            ,CASE WHEN dstChainId = 1 OR dstChainId = 101 THEN 'Ethereum'
```
jeff-dude, line 26: @BroderickBonelli i know you guys mentioned creating these spells separate, which is fine, but i think it would still make sense to follow the standard of lowercase for hardcoded values such as blockchain. in case you ever use these spells to join to another spell which uses that approach. can you make these all lowercase?

models/hashflow/avalanche_c/hashflow_avalanche_c_crosschain_trades.sql
```diff
@@ -0,0 +1,103 @@
...
+        SELECT
+            evt_block_time           AS block_time
+            ,trader            
+            ,quoteTokenAmount        AS token_bought_amount_raw
+            ,baseTokenAmount         AS token_sold_amount_raw
+            ,cast(NULL AS double)    AS amount_usd
+            ,quoteToken              AS token_bought_address
+            ,baseToken               AS token_sold_address
+            ,evt_tx_hash             AS tx_hash
+            ,CASE WHEN dstChainId = 1 THEN 'Ethereum'
```
jeff-dude, line 50: same idea here

models/hashflow/avalanche_c/hashflow_avalanche_c_crosschain_trades.sql
```diff
@@ -0,0 +1,103 @@
...
+        , (cross_chain_trades.token_sold_amount_raw / power(10, p_sold.decimals)) * p_sold.price
+        )                                                                   AS amount_usd
+    ,cross_chain_trades.token_bought_address
+    ,cross_chain_trades.token_sold_address
+    ,cross_chain_trades.trader
+    ,cross_chain_trades.tx_hash
+    ,cross_chain_trades.source_chain
+    ,cross_chain_trades.destination_chain
+FROM cross_chain_trades
+LEFT JOIN {{ ref('tokens_erc20') }} erc20a
```
jeff-dude, line 85: you may want to add join conditions on `tokens_erc20` throughout for `blockchain = <current blockchain>`, or you could get accidental duplicate returns

jeff-dude (changes requested):
left a few comments, they would apply to all blockchain specific spells

models/hashflow/avalanche_c/hashflow_avalanche_c_crosschain_trades.sql
BroderickBonelli: Yep no problem, will do

models/hashflow/avalanche_c/hashflow_avalanche_c_crosschain_trades.sql
BroderickBonelli: I tried this initially but realized that it was creating a lot of NULLs and missing data because of the nature of the trade events. All trades from these events are cross chain trades so the blockchains and token addresses are always different/changing depending on what chains the user is swapping to, am unable to specify a single chain to join on here without losing data.

BroderickBonelli: Apologies for the delay, yep I added the join on erc20b

jeff-dude (approved):
thank you!

PR #3406, opened 2023-05-24
update manifest commit gh workflow

jeff-dude:
- remove dagster deploy for the time being
- add concurrency (matches both ci & pre-commit workflow)
- rename workflow for clear understanding of purpose

### DIFF
--- a/.github/workflows/commit_manifest.yml
+++ b/.github/workflows/commit_manifest.yml
@@ -5,7 +5,7 @@ on:
       - "main"
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}
   cancel-in-progress: true
       
 jobs:
--- a/.github/workflows/commit_manifest.yml
+++ b/.github/workflows/commit_manifest.yml
@@ -3,6 +3,10 @@ on:
   push:
     branches:
       - "main"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
       
 jobs:
   commit_manifest:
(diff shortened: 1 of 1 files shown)


couralex6 (approved):

.github/workflows/commit_manifest.yml
```diff
@@ -3,6 +3,10 @@ on:
   push:
     branches:
       - "main"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
```
dot2dotseurat, line 9: I think `cancel-in-progress: true` is correct. But to be explicit, if we merged three PRs in a row only the last merged PR would publish a manifest. I can't think of why we would need any of the previous manifests so this should be ok.

.github/workflows/commit_manifest.yml
```diff
@@ -3,6 +3,10 @@ on:
   push:
     branches:
       - "main"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
```
dot2dotseurat, line 8: I don't know if `github.ref` needed here. I think just `github.workflow` is the correct group. If `ref` is always `main` it shouldn't matter but I'm not 100% thats the case as it might refer to the branch being merged in? I would double check to be certain.

.github/workflows/commit_manifest.yml
```diff
@@ -38,27 +42,4 @@ jobs:
       run: "aws s3 cp previous_manifest.json s3://manifest-spellbook/manifest_$GIT_SHA.json"
 
     - name: upload manifest
-      run: "aws s3 cp manifest.json s3://manifest-spellbook/manifest.json"
-
-  dispatch:
```
dot2dotseurat, line 43: byeeeeeeeee.

.github/workflows/commit_manifest.yml
jeff-dude: yeah, this is what i was wanting to see happen, i think we keep

PR #3408, opened 2023-05-24
remove @ in seed step of ci test

jeff-dude:
@0xRobin it looks like [this pr](https://github.com/duneanalytics/spellbook/commit/3ffc84ac4564d68ca6a909406b4a65caca1dde6b) added `@` to the `state:modified` piece of the seed step. what exactly does that do? i've noticed on other PRs now, all seeds run each time, not just modified. i'm proposing to remove that here.

### DIFF
--- a/.github/workflows/dbt_slim_ci.yml
+++ b/.github/workflows/dbt_slim_ci.yml
@@ -40,7 +40,7 @@ jobs:
         run: "dbt compile"
 
       - name: dbt seed
-        run: "dbt seed --select @state:modified --state ."
+        run: "dbt seed --select state:modified --state ."
 
       - name: dbt run initial model(s)
         run: "dbt -x run --select state:modified --exclude tag:prod_exclude --defer --state ."


0xRobin:
I've added this to resolve seed tests not running on follow up PRs.
We basically want all seeds to run that are necessary for tests that involve any modified model.

scenario without '@'
1. PR with new model + new seed. CI checks runs them both (as both are modified) and tests pass
2. follow up PR with change to model (but no change in seed). CI checks runs only the model and fails on the seed test.

The failure is partly due to us stopping to materialize the seeds in production, because seed modifications where giving us job failures.

The '@' should include all the children of the selected models and their parents.
In our case this should mean modified models, all their children (which includes the tests), and all the parents of the children (seeds are parents of the test models).
The problem is that there no way to do `1@` to only include 1 level of children and their parents, and this results in all the unnecessary seeds being run. 

I'll check if I can find a workaround, but if the seeds are really increasing runtime to much or are hanging we should indeed revert this. (in my initial tests when introducing this, I don't think that was the case)

jeff-dude:
> I've added this to resolve seed tests not running on follow up PRs. We basically want all seeds to run that are necessary for tests that involve any modified model.
> 
> scenario without '@'
> 
> 1. PR with new model + new seed. CI checks runs them both (as both are modified) and tests pass
> 2. follow up PR with change to model (but no change in seed). CI checks runs only the model and fails on the seed test.
> 
> The failure is partly due to us stopping to materialize the seeds in production, because seed modifications where giving us job failures.
> 
> The '@' should include all the children of the selected models and their parents. In our case this should mean modified models, all their children (which includes the tests), and all the parents of the children (seeds are parents of the test models). The problem is that there's no way to do `1@` to only include 1 level of children and their parents, and this results in all the unnecessary seeds being run.
> 
> I'll check if I can find a workaround, but if the seeds are really increasing runtime to much or are hanging we should indeed revert this. (in my initial tests when introducing this, I don't think that was the case)

thanks for the info. you're right on the prod version of seeds lacking. not sure we should explore cleaning that up or not. previously, it was `dbt seed` and `dbt test` which opened up much more in prod, but we could likely have a job to solely run `dbt seed` like 1x/day. this `@` change can introduce large amounts of seeds to be run on some PRs and be ~10 minutes, which can be confusing to users. let's think about this one a bit more

jeff-dude:
side note: where'd you find that in dbt docs or elsewhere? i did a (quick) search and it can be hard to navigate in there sometimes

PR #3411, opened 2023-05-25
Create sector `evms` for crosschain views of EVM raw tables

hildobby:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


hildobby:
Anyone knows why the `gnosis.blocks` base table has different data types? For all other EVM chains' `{chain}.blocks` tables, `difficulty` is a `BIGINT` and `total_difficulty` is a `DECIMAL(38, 0)`. But for `gnosis.blocks` they both are `UINT256` 🤷‍♂️

hildobby:
So only UINT256 would work here, right? What do you think is the best way forward for this to work @dot2dotseurat? Other than waiting for migration to DuneSQL to be complete I mean 😅

Screenshot 2023-06-01 at 12 36 23

couralex6:
> So only UINT256 would work here, right? What do you think is the best way forward for this to work @dot2dotseurat? Other than waiting for migration to DuneSQL to be complete I mean 😅
> 
> Screenshot 2023-06-01 at 12 36 23

Hey @hildobby! 

Yup, `UINT256` will eventually work here once the migration is complete and I'll make sure your spells are correctly translated.

For now, you should be able to cast to `DOUBLE` instead of `DECIMAL`. I just tried and it seems to be working. 

Lmk if that works!

models/evms/evms_blocks.sql
```diff
@@ -0,0 +1,143 @@
+{{ config(
+        alias ='blocks',
+        unique_key=['blockchain', 'number'],
+        post_hook='{{ expose_spells_hide_trino(\'["ethereum", "polygon", "bnb", "avalanche_c", "gnosis", "fantom", "optimism", "arbitrum"]\',
```
jeff-dude, line 4: is this intentionally hidden?

models/evms/evms_blocks.sql
```diff
@@ -0,0 +1,143 @@
...
+                                    "sector",
+                                    "evms",
+                                    \'["hildobby"]\') }}'
+        )
+}}
+
+SELECT 'ethereum' AS blockchain
```
jeff-dude, line 11: any reason we can't use the for loop approach, for cleaner code?

jeff-dude: if it's due to `blockchain` column, there is a way to do a dictionary variable and assign a value for both blockchain and source table in same variable. rob has some examples out there, likely in nft restructure models

models/evms/evms_blocks.sql
jeff-dude: same comment throughout all new spells here

models/evms/evms_blocks.sql
jeff-dude: same idea across all spells here

models/evms/evms_blocks.sql
```diff
@@ -0,0 +1,143 @@
+{{ config(
+        alias ='blocks',
+        unique_key=['blockchain', 'number'],
```
jeff-dude, line 3: if assigning unique keys here, do you want to apply the tests in schema file? for views, it's not vital, but this particular config setting won't do anything as-is (unless its more of a documentation kinda thing)

jeff-dude: again, same applies throughout all spells here

jeff-dude (changes requested):
in addition to comments throughout, are we sure these are performant as views? if you query the cross-chain spells here, then add date filters, does it actually help or will it read entire view and then filter?

models/evms/evms_blocks.sql
hildobby: ah no good catch, will change to unhide them

jeff-dude:
> I unhid the tables from trino, made them incremental materialised tables instead of views, figured out the blockchain field and changed to a loop approach and added tests @jeff-dude

what is the use case for these cross-chain base table spells? these timeouts are even more concerning, with much longer range applied too. this is a lot for spellbook to handle. it may need to be considered more at our base ingest level, or at least segmented out even further into much smaller chunks -- which doesn't seem to apply here, as it's just combining base tables.

jeff-dude:
i mentioned this to the ingest team. here's the option:
- revert back to views, then we can have them live in spellbook
- if the views are inefficient to query downstream, we'll have to find another process outside of spellbook to do this for you. it's not computationally efficient to do so here

hildobby:
> > I unhid the tables from trino, made them incremental materialised tables instead of views, figured out the blockchain field and changed to a loop approach and added tests @jeff-dude
> 
> what is the use case for these cross-chain base table spells? these timeouts are even more concerning, with much longer range applied too. this is a lot for spellbook to handle. it may need to be considered more at our base ingest level, or at least segmented out even further into much smaller chunks -- which doesn't seem to apply here, as it's just combining base tables.

Just any crosschain search really, I would use it a lot and I can imagine others too. Stuff like:
- searching crosschain balance of a wallet
- joining on other crosschain tables (even useful for some spellbook crosschain spells imo)
I have a few some wip queries that im waiting for this on since it would really reduce the number of CTEs I need

I would almost use them more than base tables, so imo it would be awesome to have it materialised somehow 🤷‍♂️

jeff-dude:
> > > I unhid the tables from trino, made them incremental materialised tables instead of views, figured out the blockchain field and changed to a loop approach and added tests @jeff-dude
> > 
> > 
> > what is the use case for these cross-chain base table spells? these timeouts are even more concerning, with much longer range applied too. this is a lot for spellbook to handle. it may need to be considered more at our base ingest level, or at least segmented out even further into much smaller chunks -- which doesn't seem to apply here, as it's just combining base tables.
> 
> Just any crosschain search really, I would use it a lot and I can imagine others too. Stuff like:
> 
> * searching crosschain balance of a wallet
> * joining on other crosschain tables (even useful for some spellbook crosschain spells imo)
>   I have a few some wip queries that im waiting for this on since it would really reduce the number of CTEs I need
> 
> I would almost use them more than base tables, so imo it would be awesome to have it materialised somehow 🤷‍♂️

cc @aalan3 @dsalv @nileslawrence @0xBoxer @andrewhong5297

hildobby:
Turned it back into views & removed tests @jeff-dude

0xBoxer:
I think we'll have a case where these tables are useful sooner or later so we should probably find a solution on how to construct them.

If we continue with this, I think we should use ``chainID`` as an additional identifier to make it easier and more efficient to join these tables on something else. (string vs int, also makes the naming a bit more reliable and in line with how everything else deals with different chains on a infra level)

jeff-dude (approved):
sounds good, let's give views a shot. we can revisit other comments in another forum, if needed in the future

MSilb7:
> 

I think tables like this are 100000% useful (and could even be the default versus a bespoke schema per chain). I'd expect ~50% of my queries now are cross-chain & unioning over and over is annoying.

jeff-dude:
> > 
> 
> I think tables like this are 100000% useful (and could even be the default versus a bespoke schema per chain). I'd expect ~50% of my queries now are cross-chain & unioning over and over is annoying.

i've initiated an internal dialogue about these and potential options

PR #3415, opened 2023-05-25
Fix nft_wash_trades

0xRobin:
The varchar trunctation of `tx_hash` is causing problems with `wash_trades` logic handling the `unique_trade_id`
In short:

`nft_wash_trades` was truncating the `tx_hash` as it is materialized  (and thus interpreted in sparksql)
`nft_trades` was not truncating as the compatibility view (`nft_ethereum_trades_beta_ported`) is not materialized (and thus interpreted in DuneSQL)

example:
nft_wash_trades:
`blur-15898370-0x0f2b1ed2a69f1f0d94d03327cc909de839634ba4a9e11b7cb2609e9db86367d3-158`

nft_trades:
`blur-15898370-0x0f2b1ed2a69f1f0d94d03327cc909de839634ba4-158`

-- 
This PR changes the casting to just varchar so there's no truncation on either `unique_trade_id`.


0xRobin:
@jeff-dude FYI, here's an example of our seed problems
I've already patched the precision error in the sudoswap seed in a previous PR.
But since it's not updated on the production table, this PR still fails on that test and I have to include non-changes to force it into the CI.

0xRobin:
cc: @hildobby this should resolve the issues with `nft_wash_trades`
Let's monitor/check it after this is merged.

hildobby:
ty @0xRobin !!

PR #3424, opened 2023-05-26
NFT enrichment fixes

0xRobin:
1. Fix for buyer/seller
There's a bug in the NFT enrichments where 
`case when base.buyer != coalesce(agg1.contract_address,agg2.contract_address)`
always evals to false when the coalesce returns null.

I've rewritten the logic so the default option is switched in this case. 

2. fix for symbol/decimals
For some tokens, there's no record in erc20.tokens but there is in prices. Introduced a coalesce to prevent null values.


hildobby (approved):
Fix makes a lot of sense, most trades now have `buyer` & `seller` default to `tx.from` because of NULLs right now so this should fix it!

jeff-dude (approved):
make sense 🤝

PR #3425, opened 2023-05-26
(Quick) Fix to DAOs models

augustog:
We are consistently seeing issues with DAO models shortly after full refreshes, as incremental updates kick in. Since we track status of proposals on a rolling basis, an incremental model would require us to update records for as long as a proposal is open.

To simplify our approach to this model, I propose we simply rebuild the whole table each time. Since the tables are relatively small, this should not carry a large performance hit to our overall jobs (checking run times on the CI to verify). We can come back to this issue when the model grows more or when we need to streamline jobs.


jeff-dude (approved):
makes sense

PR #3432, opened 2023-05-26
Rename chain_ids > chain_info_chain_ids hotfix

hildobby:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


jeff-dude (approved):
thank you ser, fixing dbt compile warnings

jeff-dude:
hmm, looks the the anti join isn't dunesql compatible. since these are views, they'd need to be run on dunesql syntax too. i wonder if these were built prior to our gh action test for dunesql compatability. we could also check if they use the `expose_spells_hide_trino` macro in config since they are known to fail there, then we could look into merging regardless of passing that test

Hosuke:
> hmm, looks the the anti join isn't dunesql compatible. since these are views, they'd need to be run on dunesql syntax too. i wonder if these were built prior to our gh action test for dunesql compatability. we could also check if they use the `expose_spells_hide_trino` macro in config since they are known to fail there, then we could look into merging regardless of passing that test

We may rewrite anti join in equivalent left join logics.
eg.
```
select *
from a left join b
  on a.id = b.id
where b.id is NULL
```

hildobby:
> > hmm, looks the the anti join isn't dunesql compatible. since these are views, they'd need to be run on dunesql syntax too. i wonder if these were built prior to our gh action test for dunesql compatability. we could also check if they use the `expose_spells_hide_trino` macro in config since they are known to fail there, then we could look into merging regardless of passing that test
> 
> We may rewrite anti join in equivalent left join logics. eg.
> 
> ```
> select *
> from a left join b
>   on a.id = b.id
> where b.id is NULL
> ```

Issued another commit where I did that, lmk if fix is insufficient

Hosuke:
Hi @hildobby, for the test failure, I cannot identify the exact cause at the moment.
But I am thinking it is not caused by the anti join.

It may be caused by the ` = ` operation in `WHERE` OR `CASE` CLAUSE.
eg.
image

One brute-force way to solve this problem, is to change the roll-up spell into incremental view.

Another appropriate solution, is to locate the column comparison and cast type on either side.
https://github.com/duneanalytics/spellbook/actions/runs/5120014833/jobs/9205990766?pr=3432#step:14:18

MSilb7:
Tested the compiled bridge_flows and ran on dune.com in Spark and it worked with no datatype errors

jeff-dude:
> Tested the compiled bridge_flows and ran on dune.com in Spark and it worked with no datatype errors

is this still the case? i think we've seen this on other PRs too, where dunesql test is failing and i was a bit confused

MSilb7:
> > Tested the compiled bridge_flows and ran on dune.com in Spark and it worked with no datatype errors
> 
> is this still the case? i think we've seen this on other PRs too, where dunesql test is failing and i was a bit confused

we should likely just cast the values anyway, I'll add a comment.

Hosuke:
Wow it fixes.

jeff-dude:
thanks all! i'll look to get this pushed through asap

PR #3398, opened 2023-05-24
add DODOFeeProxy on op & arb & polygon

owen05:
Brief comments on the purpose of your changes:

Update new FeeProxy of DODO on Optimism & Arbitrum & Polygon

**For Dune Engine V2**

I've checked that:

### General checks:
* [x] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [x] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [x] if adding a new model, I added a test
* [x] the filename is unique and ends with .sql
* [x] each sql file is a select statement and has only one view, table or function defined  
* [x] column names are `lowercase_snake_cased`
* [x] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [x] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [x] `coin_id` represents the ID of the coin on coinpaprika.com
* [x] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:
* [x] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:
* [x] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [x] where block_time >= date_trunc("day", now() - interval '1 week')
* [x] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [x] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #3433, opened 2023-05-26
Revert "add DODOFeeProxy on op & arb & polygon (#3398)"

jeff-dude:
fyi @Hosuke this failed in prod. the union into higher levels failed due to incompatible types.
image


jeff-dude:
@owen05 we had to revert this for failures in prod. please work with @Hosuke to fix as needed

Hosuke:
Sorry, I think I will help @owen05 to restructure DODO trades into `dex.trades` and `dex_aggregator.trades` separately in file/spell structure.

owen05:
So we need to keep the string type of trace_address exported to dex.trades and array<int> type of trace_address exported to  aggregator.trades. Is it right?

PR #3442, opened 2023-05-30
exclude failing lido model

0xRobin:
This PR excludes the `lido_liquidity_optimism_kyberswap_pools` model which has been failing on duplicates.

CC: @ppclunghe @gregshestakovlido 
When this is merged the model won't receive any hourly updates anymore, are you able to review the model and adapt so the uniqueness test on `[pool, time]` does not produce any failures?


ppclunghe:
@0xRobin thank you for heads up! Will take a look the model

PR #3455, opened 2023-05-31
time utils

dot2dotseurat:

getting a list of dates or times in dunesql is complex and frankly annoying. my proposal is we have a premade table with date prepopulated so this can be abstracted away from the user. 

this should ultimately live in a section other than spells. possible under "utilies" or some other future data explorer category.  but I don't think we should worry too much about that for now. 

this minutes query was too much when I tested it on dune.com so we might need to drop it but I wanted to test here. 

```
SELECT
    day.day
FROM
    (SELECT
        SEQUENCE(
          CAST(DATE_TRUNC('day', CURRENT_TIMESTAMP - INTERVAL '100' day) as TIMESTAMP),
          CAST(DATE_TRUNC('day', CURRENT_TIMESTAMP) as TIMESTAMP),
          INTERVAL '1' day
        ) AS days
    ) t
CROSS JOIN
    UNNEST(t.days) AS day(day)
    ```

### CODE REVIEW

File: models/time_utils/time_utils_minutes.sql [hunk 1]
```diff
@@ -0,0 +1,7 @@
+{{ config( alias='minutes', tags=['static'], materialized = 'table')}}
+
+select explode(sequence(
+         date_trunc('minute', now() - interval '20 year')
```
augustog, line 4: should we make the range fixed? I can't think any reason why a 40 year window shifting a few years would have any negative consequences on queries analysing crypto data, but feels like a best practice to make it truly static.

wdyt?

dot2dotseurat: I was assuming we'd just run this once but I guess we periodically rerun tables so it makes sense to make this fixed. I can update with a fixed time frame.

jeff-dude:
nice, smart idea!

i wonder if we go ahead and build a few more while we're here. i know i've seen users do weekly time series too.

jeff-dude (approved):

jeff-dude:
fyi @Hosuke @0xRobin 
if we have users looking to build out time series, this goes ahead and pre-populates tables for us that we can use instead

jeff-dude:
in order to debug prod orchestration issues that started friday afternoon, i will revert this PR and start clean. once things look good, we can look to "revert the revert" and add back one at a time (there were a batch of PRs added at same time).

PR #3468, opened 2023-06-03
Adding registrar4 to ENS registration and renewal views

web3lyt:

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


models/ens/ens_view_registrations.sql
```diff
@@ -2,12 +2,15 @@
...
+                                    \'["antonio-mendes","mewwts","steakhouse"]\') }}')}}
+SELECT *, NULL AS premium
 FROM {{source('ethereumnameservice_ethereum', 'ETHRegistrarController_1_evt_NameRegistered')}}
 UNION 
-SELECT *
+SELECT *, NULL AS premium
 FROM {{source('ethereumnameservice_ethereum', 'ETHRegistrarController_2_evt_NameRegistered')}}
 UNION 
-SELECT *
+SELECT *, NULL AS premium
```
Hosuke, line 12: Hi @web3lyt, we can cast NULL to its expected type for `premium` field.

Hosuke (changes requested):
Hi @web3lyt, 
as this PR is currently out-of-date, please run `git merge main` or use other ways to sync with the latest main branch.

Thank you.

models/ens/ens_view_registrations.sql
web3lyt: That doesn't seem necessary. In sparksql, the datatype requirements are lax, and in dunesql, this still makes the field end up as a UINT256. Further, since this view needs to work in both sparksql and dunesql, I can't CAST the datatype to a common one unless I change it to INT in the raw table as well in both versions. It's much messier if I cast, this works better.

web3lyt:
> 

I merged but this PR isn't ready anyway. Was fixing some issues. Reverted back some changes to see why they were causing issue, as it actually should have been working back then.

jeff-dude:
thinking out loud here.

my best guess is since the view is built on spark engine (all of spellbook is, as of now), the fields are still string and not `unit256`, which is only available on dunesql. when using the dune app, an override occurs which shows the field(s) as `uint256` as expected, but that happens due to a backend process. since this dunesql test doesn't run on the dune app, i wonder if the conversion of data type to `unit256` isn't happening, hence the failure on adding varchar/string fields.

@jmsgu can you help confirm this theory or help us see why this is failing?

jeff-dude:
> Catalog setup appears to be working as intended. In the meantime, can we do an explicit type conversion from varchar to uint256 for those columns being added together? Did a quick test on this, and I had to add a cast on the `cost` column for each of the other queries in the chain of unions.

i don't think we can add a cast to unit256 in the spell, as the view is built on spark first, which should fail on that. if our only option is to cast, we will need to either materialize as table or wait until spellbook migration to trino

jeff-dude:
closing PRs marked as `wip` state, prior to the freeze announced and tracked [here](https://github.com/duneanalytics/spellbook/discussions/3558), we can reopen and reformat post-freeze

PR #3479, opened 2023-06-06
Add check-script-has-no-table-name test to pre-commit-hooks

Hosuke:
Try to add a source/ref syntax check in github CI.

Main idea: search regex `^((?!--).)*from\s+[a-z0-9_]+\.+[a-z0-9_]+` in all new added *.sql files to identify invalid source/ref syntax.

i.e.
```
from gnosis_protocol_v2_ethereum.GPv2Settlement_evt_Trade
```


jeff-dude:
didn't we have a check in the pre-commit hook gh action for source/ref usage? can't we extend and/or fix the usage there?

Hosuke:
I found a test here for pre-commit-dbt/dbt-checkpoint: 
https://github.com/dbt-checkpoint/dbt-checkpoint/blob/main/dbt_checkpoint/check_script_has_no_table_name.py

The current version may accidentally identify `array_join` as `join`.
https://github.com/duneanalytics/spellbook/actions/runs/5207812043/jobs/9395725528?pr=3479#step:9:77

Hosuke:
We may need to publish/release our own dbt-pre-commit to use the customized testing scripts.

models/cryptopunks/ethereum/cryptopunks_ethereum_current_bids.sql
```diff
@@ -28,15 +28,15 @@ with combined_events_table as (
 
         select  'Transfer' as event_type
                 , cast(NULL as varchar(5)) as bidder
-                , from as transfer_from
+                , t.from as transfer_from
```
Hosuke, line 31: This line may cause the test error mistakenly.
https://github.com/dbt-checkpoint/dbt-checkpoint/blob/main/dbt_checkpoint/check_script_has_no_table_name.py#L63

Hosuke:
Since there are differences in spark syntax spells and trino syntax spells, and we currently have not published our own customized release yet. We can still use https://github.com/offbi/pre-commit-dbt v1.0.0 for pre-commit-hooks(mostly stable and compatible for now).

Maybe in future when the spellbook is mostly written in trino syntax we can then publish our customized scripts.

jeff-dude:
should we wait till post-migration to proceed, or do you see a sense or urgency to find resolution?

note that we'll have a code freeze on spark spells v soon, as we start the migration

Hosuke:
> should we wait till post-migration to proceed, or do you see a sense or urgency to find resolution?
> 
> note that we'll have a code freeze on spark spells v soon, as we start the migration

Yes we can wait till post-migration, as the ideal test is found, we can add it to the pre-commit-hook anytime.

PR #3497, opened 2023-06-09
OpenSea (Seaport v1.5) fix - Remove duplicate transactions

sohwak:

**For Dune Engine V2**

- fix for duplicate transactions in seaport v1.4 
- from https://github.com/duneanalytics/spellbook/pull/3489

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


jensenity:
ask; would it be better to just dedup the orders matched and maybe just get the min or max or the order id? I checked on those tx hashes, and it seems to have duplicates on etherscan as well.

jeff-dude (approved):
sounds good. we are working through backend server issues, so this will be merged when resolved, but feel free to edit in meantime based on feedback in comments

jeff-dude:
will merge now to ideally get past prod errors, we can revisit long term fixes as needed

sohwak:
Thanks, @jeff-dude.

I was worried that there would be a lot of cases like this, but it seems like there are only a few, but I don't think that excluding transactions will work. This patch should prevent the same type of error as before, but the workaround data may not be accurate.

I can't make an accurate patch in such a short period of time, because I haven't figured out why this is happening and how to interpret it, and I'll have to work through it as more data comes in and interpret it. I'll see how I prioritize patches once I have more of this data.

sohwak:
@jensenity
It's still not clear which actions are causing this to happen. There have been transactions with multiple NFTs in a single transaction before, but this seems different. It looks like a difficult problem to solve in the short term. If possible, I'd like to try to get an accurate interpretation, not just aggregate it.

jensenity:
@sohwak  I will help with the investigation. My team was already doing an analysis on seaport's smart contract but i will share on here once we got the root cause from our investigation 🙏🏻

jensenity:
This is the analysis I got so far,

To find all the same cases, this is the query on dune,
```
select 
    om_tx_hash,
    om_order_id,
    count(*)
from
(select evt_block_time as om_block_time
  ,evt_tx_hash as om_tx_hash
  ,evt_index as om_evt_index
  ,vals.om_order_id
  ,vals.om_order_hash
from seaport_ethereum.Seaport_evt_OrdersMatched,
UNNEST(orderhashes) WITH ORDINALITY AS vals (om_order_id, om_order_hash)
where true
  and contract_address in (0x00000000000001ad428e4906ae43d8f9852d0dd6, 0x00000000000000adc04c56bf30ac9d3c0aaf14dc)
)
group by 1,2
having count(*) > 1
;
```

I found that all the of these 3 transactions had the same pattern,
- When 1 buyer that does a collection offers and 1 seller that accepted those offers on 1 collection with more than 1 token ids. 

After looking into the smart contract, i think i found the codes. Since the seller is from the same address, when the address accept the offer, an array is set to the [matchAdvancedOrders]() which will run the _matchAdvancedOrders() and then _validateOrdersAndPrepareToFulfill returns an output of 4 values in the array. and 2 of them has the same values(which is the order length)
Screen Shot 2023-06-11 at 6 19 33 PM
 
https://github.com/ProjectOpenSea/seaport/blob/8eec593c5be6dafc0870420a7d5da5fb321a40fe/reference/lib/ReferenceOrderCombiner.sol#L218


### Cases:
Using tenderly to debug the events, and found the same order hashes on the array in the OrdersMatched events logs.

Screen Shot 2023-06-11 at 5 52 12 PM

#### Case 1. 
1. https://etherscan.io/tx/0x3ffc50795ecaf51f14a330605d44e41e3aa3515326560037b61edb8990ff80e2
Screen Shot 2023-06-11 at 5 47 10 PM

2. https://etherscan.io/tx/0xff6ab6d78a69bd839ac4fa9e9347367075f3ba2d83216c561010f94291d0118c
Screen Shot 2023-06-11 at 4 01 31 PM

#### Case 2.
3. https://etherscan.io/tx/0x4a6e63a7e0c7ffada153f71d9eef05cea1cec71b0ee56142e6e9dc0dd8e0276f
What if the collection offers is accepted by the same seller and it has more than 2 NFTs. Will it iterate the same amount to the number of nfts accepted. 
Screen Shot 2023-06-11 at 6 52 40 PM
Screen Shot 2023-06-11 at 6 52 15 PM




#### Case 3.
If the collection offer was made accepted by multiple sellers, the transactions are made in different evt_tx_hash, so the order hash and tx hash will be unique. But when it in a same tx hash, the order hash itself will not be unique.

These 2 were bought in a same offer but different sellers. You can see that the tx hashes are different.
https://opensea.io/assets/ethereum/0xd07e72b00431af84ad438ca995fd9a7f0207542d/3979
https://opensea.io/assets/ethereum/0xd07e72b00431af84ad438ca995fd9a7f0207542d/1056

### Any Action items?
1. Since OrderHashes is an array, it doesn't guarantee uniqueness on the order ids. We can see that if the the a multiple NFT offer is accepted by one address, the transaction happens in one tx hash and hence while iterating the tuple from the parameter, the order id will duplicate only once. The cardinality fix seems fine but what do you guys think if the order id has an array or 10,000 of the ids. I personally don't think it's a huge problem. 

I feel like it has to do with this update. 
https://twitter.com/opensea/status/1667218475905892377?s=20

Screen Shot 2023-06-11 at 7 04 27 PM


@sohwak @jeff-dude

jeff-dude:
cc @0xRobin too

PR #3512, opened 2023-06-12
deduplicate nft and dex based on row_number()

0xRobin:
Deduplicate nft.trades and dex.trades based on row_number. 
This is done to prevent upstream marketplace models to introduce duplicates in the sector wide model.


0xRobin:
when merged we should re-enable wash trades, not sure if I should already include that in the PR?

jeff-dude:
> when merged we should re-enable wash trades, not sure if I should already include that in the PR?

wash trades appears to be fine. the job has run a few times successfully with modified added back in. we should be safe to look to merge this in tomorrow, after it's rebased with main and such

jeff-dude (approved):

PR #3515, opened 2023-06-12
Consolidate Uniswap V3 Factory contract names

antonio-mendes:
We've consolidated the Factory Uniswap v3 contract across chains when it comes to the naming of decoded tables. This reflects that on Spells as well.


jeff-dude:
this failure is due to seed files not being created in prod, therefore the CI test can't find them to reference. typically, when spells are first built, the code and seed are in the same PR, so the CI test uses from PR rather than prod. i'm comfortable bypassing this failure and merging, since the initial `dbt run` succeeded and we know the code runs fine in prod

Issue #2802, opened 2023-02-23
Bug: Seaport null currency_address


0xRobin:
### Description
Seaport model reports null `currency_address` for trades that were free.

### Current behavior
Seaport model reports null `currency_address` for trades that were free.
https://github.com/duneanalytics/spellbook/blob/d838417ae0d76c0b16d3298be07c5ac28a50a9d3/models/seaport/ethereum/seaport_ethereum_trades.sql#L159
example tx: https://etherscan.io/tx/0x0001164eddf3ff130b7da0518459353c9f818b13151dcb26f0ee492bf1bf554d

### Expected behavior
While technically correct, this also messes with other columns that rely on erc20 token decimals to be calculated.
It would be better to fill in the native token address and report 0 as the amount for the trade (instead of null).

### Concerned model(s)
```
seaport_ethereum.trades
```

### Impacted model(s)
```
opensea.trades
nft.events
nft.trades
nft.fees
```
~100k rows affected
```sql
select count(*), min(tx_hash) as example from opensea.events
where currency_contract is null and blockchain = 'ethereum'
```

### Possible solution
use `coalesce( max(...), {{ weth_address }})` as a workaround to fill in the WETH contract address

jeff-dude:
based on closed linked PR, maybe time to revisit the fixes here?

0xRobin:
just confirmed the issue is still present.

jeff-dude:
closing issue due to inactivity. if still relevant, and proposed fix is available, feel free to reopen an issue as needed.

PR #3534, opened 2023-06-14
Fix null opensea & seaport null currency_address

jensenity:

Fixes https://github.com/duneanalytics/spellbook/issues/2802 's seaport null currency address if the nft is given for free with weth address

Test Query: https://dune.com/queries/2634765

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


jensenity:
ask; hi @hildobby @sohwak @jeff-dude , I checked on some of the transactions and found out that the currency symbols on seaport v3, v4 are all `ETH` right now. Is there a reason for the symbol to be ETH and not WETH? since the address is changed to WETH's address anyways, wouldn't it more accurate to use WETH?

hildobby:
> ask; hi @hildobby @sohwak @jeff-dude , I checked on some of the transactions and found out that the currency symbols on opensea v3 & v4 + seaport v1, v2 are all `ETH` right now. Is there a reason for the symbol to be ETH and not WETH? since the address is changed to WETH's address anyways, wouldn't it more accurate to use WETH?

Unfortunately the OpenSea models have had a lot of bugs for months and are by far the most inaccurate NFT marketplace models on Dune. I proposed a full rewrite a few months ago (albeit with some fixes needed still) to the OS data team but haven't heard back from them about it since 🤷‍♂️

@0xRobin & I wrote + maintain most other marketplaces but the OS ones are seeing regular PRs from the OS team and current spells' code have me confused more than anything so I'm not sure how to fix them. A full rewrite would imo be the best way forward as I and others have suggested many times. I'm happy to work on the rewrite once the OS team agree it's the best way forward.

Plz do tell me if there are any bugs we haven't caught with other marketplace spells, I'm happy to look into those @jensenity!

sohwak:
@jensenity 
We're changing the structure of opensea_ethereum.events(https://github.com/duneanalytics/spellbook/pull/3567), so I'd like to apply this part later. 
The `token_address = 0x0000....` was caused by a native token, so using `WETH` to get price, but I thought it was correct to represent it as `ETH`.
However, I agree that the problem is the trade being created without a currency should be fixed.

sohwak:
@hildobby 
I understand what you're saying. Opensea data on Dune had considerable discrepancies in February-March 2023, particularly for platform fees and creator earnings, which were caused by some hardcoding in the seaport source I created. You're one of the biggest contributors here, and you've also been working hard to make sure the data is accurate, so I'd like to say that I'm sorry you feel that way.

The reason it didn't get fixed was partly because I didn't find the right way to do it, and partly because we didn't use it that much. I also didn't get to give you feedback on the source review that you recently requested. What I did was, I tried to run the code, and the results were different for the cases that I thought they were, and to give you feedback as to why, I would have to review the code, and reviewing the code takes time, even for me. At the time, I was in the middle of testing a new method for the same error case with Polygon, and I thought it would be better to apply it quickly to solve the problem. But I really apologize for not getting back any response to you.

I'm not trying to monopolize the works on Seaport. I've been writing tables for them much, but I'm not the one who wrote the contract codes, I'm the one who breaks down and interprets the cases for the transactions that happen to that contract. When a new case comes up, the interpretation breaks down, and I have to come up with a cause and solution again, which is also a burden, so it's been out of date for a long time. 

I would like to do this with many other contributors, and I think the biggest problem with not being able to do this is that I haven't been able to share what I've learned and worked with others, but I haven't been able to do. So I'll try to share them like interpretation and the different cases, so that more people can use the data well, and more people can modify it.

0xRobin:
@jensenity thank you for bringing forward a possible solution! 
I'm going to close this PR (for now) since we are starting the freeze and migration to DuneSQL (see the discussions). 
While the main code changes look good, the seed file was also changed drastically which makes me not comfortable enough to get this in before the freeze (later today).

Let's revisit this in ~10 days time.

jensenity:
sure! I was late in checking all your comments! @sohwak I would love to work on the opensea models with you :)

PR #3592, opened 2023-06-22
Add macros to enable offline Trino compiles and speed up Trino compiles

jkylling:
The macros are used to disable listing of relations from the warehouse. Run `dbt compile` with the flag `--vars 'no-relation-listing: "true"'` on the command line to enable this. This speeds up DBT compile times. It must not be used with `dbt run`.

This enables running `dbt compile` without access to a Trino cluster. It is also faster than running with a Trino cluster.

Potentially we could also use this to make the local setup of `dbt` simpler:
* We add a `profiles.yml` file to the root of spellbook:
```
spellbook:
  target: dev
  outputs:
    dev:
      type: trino
      user: trino
      password: trino
      host: trino
      port: 1234
      database: hive
      schema: dbt
      threads: 1
```
* User setup from a fresh repository is then:
  * `pipenv install`
  * `pipenv shell`
  * `dbt deps`
* Dev workflow: `dbt compile` (we use the profile name instead of a vars to enable or disable listings)

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,36 +130,6 @@ pipenv shell
 
 You have now created a virtual environment for this project. You can read more about virtual environments [here](https://realpython.com/pipenv-guide/).
 
-To initiate the dbt project run:
-
-```console
-dbt init
-```
-
-Enter the values as shown below:
-
-```console
-Which database would you like to use?
-[1] databricks
-[2] spark
-
-(Don't see the one you want? https://docs.getdbt.com/docs/available-adapters)
-
-Enter a number: 1
-host (yourorg.databricks.com): .
-http_path (HTTP Path): .
-token (dapiXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX):
-[1] use Unity Catalog
-[2] not use Unity Catalog
-Desired unity catalog option (enter a number): 2
-schema (default schema that dbt will build objects in): wizard
-threads (1 or more) [1]: 2
-```
-
-This will not connect to the database but you have access to some dbt actions.
-**When you are prompted to choose a schema, please enter `wizard` so we know you are an external contributor.**
-Should you make an error during this process (not entering `wizard` being the only one you can make), simply quit the CLI and start over.
-
 To pull the dbt project dependencies run:
 
 ```console
@@ -240,18 +210,6 @@ As a preview, you can do [things](https://docs.getdbt.com/reference/resource-pro
 - Link to other models in your descriptions.
 - Add images / project logos from the repo into descriptions.
 - Use HTML in your description.
-
-### Troubleshooting
-
-If you fail to run `dbt compile`, here are some common error messages:
-
-- `Could not find profile named 'spellbook'` <br> Check `~/.dbt/profiles.yml` and make sure there is a profile named `spellbook`. When you run `dbt init` to initiate a project, a profile gets created. Inside `spellbook` you cannot initiate a project called the same name, so you need to run `dbt init spellbook` outside the project so it creates the profile, or create one with a different name and then manually edit the `profiles.yml` file.
-- ```console
-  Credentials in profile "spellbook", target "dev" invalid: Runtime Error
-   http connection method requires additional dependencies.
-   Install the additional required dependencies with pip install dbt-spark[PyHive]
-  ```
-  You've probably selected the `spark` option instead of the `databricks` option when running `dbt init`. Rerun `dbt init`, overwrite the profile, and select the `databricks` option.
 
 ### DBT Resources:
 
--- a/.gitignore
+++ b/.gitignore
@@ -140,6 +140,7 @@ dmypy.json
 target/
 dbt_packages/
 logs/
+.user.yml
 
 # exclusions
 !target/manifest.json
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -5,7 +5,7 @@ version: "1.0.0"
 config-version: 2
 
 # This setting configures which "profile" dbt uses for this project.
-profile: "spellbook"
+profile: "spellbook-local"
 
 vars:
   DBT_ENV_CUSTOM_ENV_S3_BUCKET: "{{ env_var('DBT_ENV_CUSTOM_ENV_S3_BUCKET', 'local') }}"
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -5,7 +5,7 @@ version: "1.0.0"
 config-version: 2
 
 # This setting configures which "profile" dbt uses for this project.
-profile: "spellbook-local"
+profile: "spellbook"
 
 vars:
   DBT_ENV_CUSTOM_ENV_S3_BUCKET: "{{ env_var('DBT_ENV_CUSTOM_ENV_S3_BUCKET', 'local') }}"
--- a/macros/dune/no-relation-listing.sql
+++ b/macros/dune/no-relation-listing.sql
@@ -1,11 +1,11 @@
 {# Macros for disabling listing of relations from the warehouse.
- Enabled by default for the spellbook-local profile.  Run dbt compile with the flag --vars 'no-relation-listing: "true"'
+ Enabled by default for the spellbook profile.  Run dbt compile with the flag --vars 'no-relation-listing: "true"'
  on the command line to explicitly enable or disable this.
  This speeds up DBT compile times. It must not be used in combination with dbt run.
  #}
 {% macro get_catalog(information_schema, schemas) -%}
   {% do log(target) %}
-  {%- if (var('no-relation-listing', 'false').lower() == 'true') or (target.profile_name == 'spellbook-local') -%}
+  {%- if (var('no-relation-listing', 'false').lower() == 'true') or (target.profile_name == 'spellbook') -%}
     {{ return([]) }}
   {%- else -%}
     {{ return(adapter.dispatch('get_catalog')(information_schema, schemas)) }}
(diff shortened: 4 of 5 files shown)


0xRobin (commented):
This is great!!
We could add:
```yml
vars:
    no-relation-listing: "{{ env_var('DBT_NO_RELATION_LISTING', 'false') }}"
```
to the `dbt_project.yml` file.
This would allow us (wizards) to set it once as an env var and then forget about it.

jeff-dude:
> This is great!! We could add:
> 
> ```yaml
> vars:
>     no-relation-listing: "{{ env_var('DBT_NO_RELATION_LISTING', 'false') }}"
> ```
> 
> to the `dbt_project.yml` file. This would allow us (wizards) to set it once as an env var and then forget about it.

good call, similar to usage [here](https://github.com/duneanalytics/spellbook/blob/main/dbt_project.yml#L11), then we can apply in our local `~/.zshrc` file, such as:
image

jkylling:
I'm wondering if we could take this even further, and enable this for contributors by default.

We could add a `profiles.yml` file to the root of the repository, and have
```
spellbook:
  target: dev
  outputs:
    dev:
      type: trino
      user: trino
      password: trino
      host: trino
      port: 1234
      database: hive
      schema: dbt
      threads: 1
```
and in `dbt_project.yml` keep
```
profile: "spellbook"
```
The macros are modified to have an if statement `{%- if target.profile_name == 'spellbook' -%}`, which enables offline usage. Then contributors will only need to run `dbt compile`, without any extra arguments or configuration setup elsewhere.

Maintainers will need to set up different profiles (e.g., `spellbook-dev`, `spellbook-prod`), and specify this on the command line if they want to execute `dbt run` directly.

jeff-dude:
> I'm wondering if we could take this even further, and enable this for contributors by default.
> 
> We could add a `profiles.yml` file to the root of the repository, and have
> 
> ```
> spellbook:
>   target: dev
>   outputs:
>     dev:
>       type: trino
>       user: trino
>       password: trino
>       host: trino
>       port: 1234
>       database: hive
>       schema: dbt
>       threads: 1
> ```
> 
> and in `dbt_project.yml` keep
> 
> ```
> profile: "spellbook"
> ```
> 
> The macros are modified to have an if statement `{%- if target.profile_name == 'spellbook' -%}`, which enables offline usage. Then contributors will only need to run `dbt compile`, without any extra arguments or configuration setup elsewhere.
> 
> Maintainers will need to set up different profiles (e.g., `spellbook-dev`, `spellbook-prod`), and specify this on the command line if they want to execute `dbt run` directly.

trying to think through those implications. i like it from the contributors perspective, but open questions on maintainers:
- if `main` branch only contains `profile.yml` at the root directory of spellbook as mentioned above, then we as maintainers would keep our profiles locally like we do now under `~/.dbt/profiles.yml` and have two profiles within, dev and prod
- can we somehow automate switching from dev/prod via variables, rather than applying in command line each time?

jkylling:
> 
> trying to think through those implications. i like it from the contributors perspective, but open questions on maintainers:
> 
> * if `main` branch only contains `profile.yml` at the root directory of spellbook as mentioned above, then we as maintainers would keep our profiles locally like we do now under `~/.dbt/profiles.yml` and have two profiles within, dev and prod

Yes, this is correct.

> * can we somehow automate switching from dev/prod via variables, rather than applying in command line each time?

Maybe, what is the workflow you have in mind here? Having to explicitly write `dbt run --profile dev` or `dbt run --profile prod` is a small safeguard against mistakes.

README.md
```diff
@@ -130,36 +130,6 @@ pipenv shell
 
 You have now created a virtual environment for this project. You can read more about virtual environments [here](https://realpython.com/pipenv-guide/).
```
dot2dotseurat, line 131: the dunesql readme draft was updated here if we want to use that version https://github.com/duneanalytics/spellbook-trino/pull/44

dot2dotseurat:
target `prod` will run commands without the user's table prefix i.e. nft.trades versus dbt_jeff_nft.trades. unless we are actually trying to circumvent dbt cloud to run something to expose in the data explorer, we shouldn't use prod in the profiles.yml. I don't think we are restricted to dev and prod as targets. So we could have a local target which uses the macro and dev target for the `dbt_jeff...` prefixed workflow.

jkylling:
> target `prod` will run commands without the user's table prefix i.e. nft.trades versus dbt_jeff_nft.trades. unless we are actually trying to circumvent dbt cloud to run something to expose in the data explorer, we shouldn't use prod in the profiles.yml. I don't think we are restricted to dev and prod as targets. So we could have a local target which uses the macro and dev target for the `dbt_jeff...` prefixed workflow.

Should we have two targets in `profiles.yml`, `dev` and `prod`? E.g.,
```yaml
spellbook-local:
  target: dev
  outputs:
    dev:
      type: trino
      user: trino
      password: trino
      host: trino
      port: 1234
      database: hive
      schema: dbt
      threads: 1
spellbook-local-prod:
  target: prod
  outputs:
    prod:
      type: trino
      user: trino
      password: trino
      host: trino
      port: 1234
      database: hive
      threads: 1
```

dot2dotseurat:
@jkylling what is the purpose of the prod profile for local dev? I'm not following.

jkylling:
> @jkylling what is the purpose of the prod profile for local dev? I'm not following.

Sorry, I think I misunderstood your comment. In `spellbooks/profiles.yml` we should only have a single profile, `spellbook-local`, the one added in this PR. Dune developers will have custom profiles to access dev and prod directly in `~/.dbt/profiles.yml`.

jeff-dude:
i'm trying to go through the flow locally on this branch, but running into issues on the internal dune team flow.

for awareness, i changed the profile name back to `spellbook`, to help simplify the change locally, as most dev's local profile will have that value too. it still works fine when mimicking the workflow for wizards. i can get compile to run as expected when using the `profiles.yml` in the root directory of spellbook.

important note to include in contribution guide: wizards will have to be in the root directory to run compile. we need to explicitly call that out, as it can be easy to navigate directories and forget. otherwise, the profile won't be found.

as for the dune workflow, i've tried two ways:
- set the `DBT_PROFILES_DIR` env variable to my personal `~/.dbt/profiles.yml`
- remove the env variable, run with profile param: `dbt compile --profiles-dir ~jdude/.dbt/profiles.yml`

when i run `dbt debug --config-dir`, it correctly shows my personal profile yml file. however, i repeatedly get the error:
```
Encountered an error while reading the project:  ERROR: Runtime Error
  Could not find profile named 'spellbook'
21:15:30  Encountered an error:
Runtime Error
  Could not run dbt
```

i've confirmed i'm connected to tailscale. i've also used this profile to run on `spellbook-trino` repo successfully. i don't think it's even getting to authentication, it just can't find the `spellbook` profile, yet i'm sure it's set to that locally.

jkylling:
> i'm trying to go through the flow locally on this branch, but running into issues on the internal dune team flow.
> 
> for awareness, i changed the profile name back to `spellbook`, to help simplify the change locally, as most dev's local profile will have that value too. it still works fine when mimicking the workflow for wizards. i can get compile to run as expected when using the `profiles.yml` in the root directory of spellbook.
> 
> important note to include in contribution guide: wizards will have to be in the root directory to run compile. we need to explicitly call that out, as it can be easy to navigate directories and forget. otherwise, the profile won't be found.
> 
> as for the dune workflow, i've tried two ways:
> 
> * set the `DBT_PROFILES_DIR` env variable to my personal `~/.dbt/profiles.yml`
> * remove the env variable, run with profile param: `dbt compile --profiles-dir ~jdude/.dbt/profiles.yml`
> 
> when i run `dbt debug --config-dir`, it correctly shows my personal profile yml file. however, i repeatedly get the error:
> 
> ```
> Encountered an error while reading the project:  ERROR: Runtime Error
>   Could not find profile named 'spellbook'
> 21:15:30  Encountered an error:
> Runtime Error
>   Could not run dbt
> ```
> 
> i've confirmed i'm connected to tailscale. i've also used this profile to run on `spellbook-trino` repo successfully. i don't think it's even getting to authentication, it just can't find the `spellbook` profile, yet i'm sure it's set to that locally.

For the dune dev flow, could you use the `--profile spellbook-trino` flag, and not setting environment variables or the `--profiles-dir` flag (notice, singular form of profile)? This will look in the usual locations, not find spellbook-trino in /profiles.yml, and proceed to find it in ~/.dbt/profiles.yml.

jeff-dude:
image

from what i see, we confirm that the default profile is used. `dbt compile` works fine. `dbt compile --profile spellbook-trino` still can't find the profile (i changed to this in my local profile).

i guess ultimately we can still merge this during the freeze, then find the gaps for dune team local setup after. as long as compile works for the wizards.

jeff-dude (approved):
wizard workflow:
`dbt compile` -- leverages the `profiles.yml` setup in root of spellbook, no connection to trino. note: user must be in root of spellbook on command line to run.

dune workflow:
`dbt compile --profiles-dir ~/.dbt --profile spellbook-trino` -- requires profile setup locally, with correct profile name within

jeff-dude:
> wizard workflow: `dbt compile` -- leverages the `profiles.yml` setup in root of spellbook, no connection to trino. note: user must be in root of spellbook on command line to run.
> 
> dune workflow: `dbt compile --profiles-dir ~/.dbt --profile spellbook-trino` -- requires profile setup locally, with correct profile name within

fyi @Hosuke @0xRobin @couralex6 @a-monteiro @dot2dotseurat 
this will have impact on your local setup when in the spellbook repo

PR #3593, opened 2023-06-22
test lido liquidity model

ppclunghe:
Test the options for lido liquidity model @jeff-dude

### DIFF
--- a/models/lido/liquidity/arbitrum/lido_liquidity_arbitrum_curve_pools.sql
+++ b/models/lido/liquidity/arbitrum/lido_liquidity_arbitrum_curve_pools.sql
@@ -16,7 +16,7 @@
 {% set project_start_date = '2022-10-06' %}
 
 with dates as (
-select explode(sequence(to_date('{{ project_start_date }}'), now(), interval 1 day)) as day
+select explode(sequence(date_trunc("day", now() - interval '1 week'), now(), interval 1 day)) as day
 )
 
 , weth_prices_daily AS (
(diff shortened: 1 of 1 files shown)


models/lido/liquidity/arbitrum/lido_liquidity_arbitrum_curve_pools.sql
```diff
@@ -16,7 +16,7 @@
 {% set project_start_date = '2022-10-06' %}
 
 with dates as (
-select explode(sequence(to_date('{{ project_start_date }}'), now(), interval 1 day)) as day
+select explode(sequence(date_trunc("day", now() - interval '1 week'), now(), interval 1 day)) as day
```
jeff-dude, line 19: you'll need both filters, rather than replace the old. you can add the `if is incremental` blocks here, add the sql statement for each incremental / not incremental.
for example:
```
{% if not is_incremental() %}
select explode(sequence(to_date('{{ project_start_date }}'), now(), interval 1 day)) as day
{% else %}
select explode(sequence(date_trunc("day", now() - interval '1 week'), now(), interval 1 day)) as day
{% endif %}
```

models/lido/liquidity/arbitrum/lido_liquidity_arbitrum_curve_pools.sql
```diff
@@ -27,11 +27,7 @@ select explode(sequence(to_date('{{ project_start_date }}'), now(), interval 1 d
         decimals,
         avg(price) AS price
     FROM {{source('prices','usd')}}
```
jeff-dude, line 29: then i think you're right here, remove incremental filters throughout all other CTEs in order to get full historical aggregations on each day

models/lido/liquidity/arbitrum/lido_liquidity_arbitrum_curve_pools.sql
jeff-dude: you also don't need to default to the one week for incremental runs, it can be just one day back if that works too

ppclunghe: thanks @jeff-dude 
fixed filters
it seems ok 1 week currently

ppclunghe:
@jeff-dude does it looks good to be non-incremental spell?

jeff-dude (approved):

PR #3640, opened 2023-06-29
Add Celo to `evms` tables

hildobby:
# SPELLBOOK FREEZE

From June 22 to June 29, we will be freezing some Spellbook contributions for the migration to DuneSQL. We will not accept contributions to the lineage of the following models:

* dex.trades
* nft.trades
* labels
* token.erc20
* tokens.nft

Run the following command to see the list of affected files:

```
dbt ls --resource-type model --output path --select +dex_trades +labels +nft_trades +tokens_nft +tokens_erc20
```

Don't hesitate to reach out on Discord if you have any questions.

**For Dune Engine V2**

I've checked that:

### General checks:

### Pricing checks:

### Join logic:

### Incremental logic:


jkylling (approved):

models/evms/evms_contracts.sql
jeff-dude, line 1: on dunesql spells, we need to apply two things:
- `tags = ['dunesql'],` in the config block
- `alias = alias('contracts'),` formatting on the alias property

jeff-dude: you should also be duplicating the spells prior to making dunesql changes, renaming to `_legacy.sql` suffix, then applying:
- `alias = alias('blocks', legacy_model=True),` formatting on alias property
- **_do not_** add tag for an engine

notice for `evms_blocks` we have both new version and `_legacy`

jeff-dude: we have a script in testing phase that automates the creation of legacy files and updates alias property for us, we will look to include in repo as it matures

jeff-dude (changes requested):

models/evms/evms_contracts.sql
hildobby: 1. were those tables that were created just a few days ago used on spark though? If not, it might be pointless to have the legacy sql files, no?

2. I didn't add the tag for the engine in `evms_blocks.sql`'s config, should I remove it? or just not add it to other files in the evms sector?

3. for how long will the legacy files be necessary?

jeff-dude: 1 - which tables? we only did one spell in the initial merge, the `evms.blocks` spell, which had the legacy created with the new translated one. in general though, if they are net new spells, no need for legacy. we're simply looking to have a backstop for existing spells from spark just in case
2 - no, the tag in blocks should stay, that is an example for you to follow on all other `evms` schema spells. all non-legacy files, i.e. dunesql compatible, should have the tag added
3 - likely until spark is fully deprecated -- hopefully not long! we will try to release that script to help this part

hildobby: Alright I think I applied all the changes to the original files and duplicated them into their legacy equivalent, lmk if I missed anything!

hildobby:
Hitting an error about varchar <> varbinary incompatibility on joins, could we have all hashes in raw tables be of varbinary datatype? I'm guessing this is caused by the new Celo tables but maybe other chains too 🤔

jeff-dude:
> Hitting an error about varchar <> varbinary incompatibility on joins, could we have all hashes in raw tables be of varbinary datatype? I'm guessing this is caused by the new Celo tables but maybe other chains too 🤔

did you confirm in the data explorer that the troubled column is showing the different data types?

MSilb7:
> Hitting an error about varchar <> varbinary incompatibility on joins, could we have all hashes in raw tables be of varbinary datatype? I'm guessing this is caused by the new Celo tables but maybe other chains too 🤔

to handle for this in queries, I've done `from_hex( cast(column_name as varchar) )` so we're shielded from datatype changes

jeff-dude:
> > Hitting an error about varchar <> varbinary incompatibility on joins, could we have all hashes in raw tables be of varbinary datatype? I'm guessing this is caused by the new Celo tables but maybe other chains too 🤔
> 
> to handle for this in queries, I've done `from_hex( cast(column_name as varchar) )` so we're shielded from datatype changes

is this feasible or should we be exploring the root cause further?

hildobby:
> > > Hitting an error about varchar <> varbinary incompatibility on joins, could we have all hashes in raw tables be of varbinary datatype? I'm guessing this is caused by the new Celo tables but maybe other chains too 🤔
> > 
> > 
> > to handle for this in queries, I've done `from_hex( cast(column_name as varchar) )` so we're shielded from datatype changes
> 
> is this feasible or should we be exploring the root cause further?

I would argue that fixing at the root and having uniformity amongst base tables would be the ideal route instead of building up unnecessary tech dept. You can also tell the discrepancy across raw tables that I had to account for in the `evms` sector (ie `tx_from` and `tx_to` are in `celo.creation_traces` but not in the other `creation_traces` tables, would be cool to add them)

I really can't tell where the error stems from exactly though, those seem to be defined with the same datatypes here, what could be the issue?
Screenshot 2023-07-03 at 17 41 32
Screenshot 2023-07-03 at 17 41 19



It would also be worth addressing the inconsistency of definied raw table names, some are lowercased while some arent, might be worth making all lowercase?

jkylling:
The type issues might be caused by the way the cluster running CI for Spellbook is set up. I'll look into it.

jkylling:
Looks like it is an issue with the underlying types of the base tables, even on dune.com. For instance, `arbitrum.contracts.code` is of type varchar, while `ethereum.contracts.code` is of type varbinary. We will get it fixed.

hildobby:
> Looks like it is an issue with the underlying types of the base tables, even on dune.com. For instance, `arbitrum.contracts.code` is of type varchar, while `ethereum.contracts.code` is of type varbinary. We will get it fixed.

Amazing ty! Do you think it would be possible to add `tx_from` & `tx_to` into the `creation_traces` tables? It's currently in the celo one but not the other chains'. Those two columns would also be an amazing addition to the traces tables, but I'm not sure if that's feasible.

alokeutpal (approved):

models/evms/evms_traces.sql
```diff
@@ -17,6 +18,7 @@
      , ('fantom', source('fantom', 'traces'))
      , ('optimism', source('optimism', 'traces'))
      , ('arbitrum', source('arbitrum', 'traces'))
+     , ('celo', source('celo', 'traces'))
 ] %}
 
 SELECT *
```
jkylling, line 24: Beware that celo.traces contains the additional tx_from and tx_to columns, so this union must be made with explicit column names. It's a good practice to write out all the column names explicitly, as then the spell won't break if additional columns are introduced in some of the upstream table.

alokeutpal (approved):

jkylling:
> > Looks like it is an issue with the underlying types of the base tables, even on dune.com. For instance, `arbitrum.contracts.code` is of type varchar, while `ethereum.contracts.code` is of type varbinary. We will get it fixed.
> 
> Amazing ty! Do you think it would be possible to add `tx_from` & `tx_to` into the `creation_traces` tables? It's currently in the celo one but not the other chains'. Those two columns would also be an amazing addition to the traces tables, but I'm not sure if that's feasible.

Yes, we are working on adding tx_from and tx_to to all base tables, such that it won't be necessary to join with the transactions table all the time. This unfortunately takes some time :florians-favourite-emoji: Celo is a new chain, which is why it already contains tx_from and tx_to.

models/evms/evms_creation_traces_legacy.sql
```diff
@@ -0,0 +1,54 @@
...
+        , ct.block_time
+        , ct.block_number
+        , ct.tx_hash
+        , ct.address
+        , ct.from
+        , ct.code
+        , txs.from AS tx_from
+        , txs.to AS tx_to
+        FROM {{ creation_traces_model[1] }} ct
+        INNER JOIN {{ creation_traces_model[2] }} txs USING (block_number, tx_hash)
```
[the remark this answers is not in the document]
hildobby: Removed it that join for now, I would love for these tables to eventually be materialised at the ingestion level rather than a view on the longer term

models/evms/evms_traces.sql
hildobby: Good catch, I was too lazy to define all the columns in each model when creating these tables, just defined them in all tables

MSilb7:
**Another issue impacting evms.** Some chains seem to have a different amount of columns for logs, traces, and creation_traces which is causing UNION errors (this PR should try to address this as well)

image

MSilb7:
^ Reading the comments, I think adding tx_from and tx_to to all tables should resolve this. Is there an ETA @jkylling  ?

jkylling:
> ^ Reading the comments, I think adding tx_from and tx_to to all tables should resolve this. Is there an ETA @jkylling ?
> 
> It seems like adding them to Arbitrum, but no other chains broke this. We need all EVM chains to have them added before this will work.

Sorry, no definite ETA for this, but it should hopefully be available within the next quarter. The workaround is to explicitly enumerate the columns which are part of the union. Not enumerating the columns is kind of an anti-pattern, as then you don't know what the schema of the model is.

hildobby:
> > ^ Reading the comments, I think adding tx_from and tx_to to all tables should resolve this. Is there an ETA @jkylling ?
> > It seems like adding them to Arbitrum, but no other chains broke this. We need all EVM chains to have them added before this will work.
> 
> Sorry, no definite ETA for this, but it should hopefully be available within the next quarter. The workaround is to explicitly enumerate the columns which are part of the union. Not enumerating the columns is kind of an anti-pattern, as then you don't know what the schema of the model is.

All columns are enumerated since my commits from last week, I have no idea what the current error means (seems to be unrelated tables)

hildobby:
Fixed all issues in this PR, let me know if I should change anything else @jeff-dude @jkylling.

With those models currently broken with the addition of tx_from/to to some base tables, it would be nice to get this merged asap since it fixes those issues :)

jeff-dude (approved):
great stuff ser -- thank you!

PR #3645, opened 2023-07-01
Run Spark CI before Trino

couralex6:
Changing the order of `dbt_slim_ci` matrix, letting the Spark job run before the DuneSQL job. Since most contributions will be to spark spells for the near term, those contributions won't need to wait for dune job for execution. When I have more time, i'll update the ci to only run on job corresponding to the query engine those files target.

### DIFF
--- a/.github/workflows/dbt_slim_ci.yml
+++ b/.github/workflows/dbt_slim_ci.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: [ self-hosted, linux, spellbook-trino ]
     strategy:
       matrix:
-        engine: [ 'dunesql', 'spark' ]
+        engine: [ 'spark', 'dunesql' ]
       max-parallel: 1
     timeout-minutes: 90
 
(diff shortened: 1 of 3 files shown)


.github/workflows/dbt_slim_ci.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: [ self-hosted, linux, spellbook-trino ]
     strategy:
       matrix:
-        engine: [ 'dunesql', 'spark' ]
```
a-monteiro, line 22: Do we need to revert the order?

macros/dune/schema.sql
```diff
@@ -3,8 +3,8 @@
 {% endmacro %}
 
 {% macro default__create_schema(relation) -%}
-  {% set s3_bucket = 'prod-spellbook-trino-118330671040' %}
+  {% set s3_bucket = var('DBT_ENV_CUSTOM_ENV_S3_BUCKET', 'local') %}
   {%- call statement('create_schema') -%}
-   CREATE SCHEMA {{ relation }} WITH (location = 's3a://{{s3_bucket}}/hive/')
```
a-monteiro, line 8: Will this be executed for spark?

PR #3642, opened 2023-06-30
[ERC20] [Easy] Add missing tokens

gentrexha:
Generated with

```
docker run \                                                   
    -e SPELLBOOK_PATH=$SPELLBOOK_PATH \
    -e INFURA_KEY=$INFURA_KEY \
    -e DUNE_API_KEY=$DUNE_API_KEY \
    -v $SPELLBOOK_PATH:$SPELLBOOK_PATH \
    ghcr.io/cowprotocol/data-misc-missing-tokens:main
```

PR #3649, opened 2023-07-03
Add updated missing tokens

gentrexha:
Generated with

```
docker run \                                                   
    -e SPELLBOOK_PATH=$SPELLBOOK_PATH \
    -e INFURA_KEY=$INFURA_KEY \
    -e DUNE_API_KEY=$DUNE_API_KEY \
    -v $SPELLBOOK_PATH:$SPELLBOOK_PATH \
    ghcr.io/cowprotocol/data-misc-missing-tokens:main
```

Updated based on comment [here](https://github.com/duneanalytics/spellbook/pull/3642#issuecomment-1614737942).


bh2smith (changes requested):
Should we be updating both legacy and this table now? or is `_legacy` just a cheeky temp fix during migration?

I noticed that our trades table is pointing to legacy now as well (see here https://github.com/duneanalytics/spellbook/blob/main/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql#L91-L93)

Can we already redirect our table to read from the migrated tokens @jeff-dude or do we have to wait till the whole spell is migrated to dunesql?

In the meantime, I think we will need to update both the _legacy and the new tables...

jeff-dude:
> Should we be updating both legacy and this table now? or is `_legacy` just a cheeky temp fix during migration?
> 
> I noticed that our trades table is pointing to legacy now as well (see here https://github.com/duneanalytics/spellbook/blob/main/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql#L91-L93)
> 
> Can we already redirect our table to read from the migrated tokens @jeff-dude or do we have to wait till the whole spell is migrated to dunesql?
> 
> In the meantime, I think we will need to update both the _legacy and the new tables...

we will release an updated spellbook contribution guide v shortly as we come out of the freeze. short answer: yes, add to both for now (through migration period). any downstream spells which read from them (i.e. in `ref()` usage), we will update to legacy, then next step when migrating downstream spells, such as cow trades, we will remove legacy ref call.

i believe cow trades moved to `dex_aggregator` lineage, so it's not in our scope during the freeze, so you're more than welcome to go ahead and start migrating if you'd like. otherwise, it'll come up in next set of priorities for us asap.

bh2smith:
Awesome thanks @jeff-dude! I may try to take a stab at it soon...

gentrexha:
Closed in favour of https://github.com/duneanalytics/spellbook/pull/3673

PR #3662, opened 2023-07-05
Only check first degree children and parents for DuneSQL and Spark overlap

jkylling:
It's sufficient to check that the first layer of parents or children are non-overlapping. This makes the error messages smaller and easier to inspect.

### DIFF
--- a/scripts/dunesql_spark_empty_intersection.sh
+++ b/scripts/dunesql_spark_empty_intersection.sh
@@ -2,7 +2,7 @@
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


jeff-dude (commented):
you could also simplify the output to just model name, if that helps
here is how i have been running list command:
`dbt ls --resource-type model --output name --select tag:dunesql`

jeff-dude:
pushing this now as other PRs are failing on this step

we can revisit in follow-up PR for further cleanup and/or enhancement

PR #3667, opened 2023-07-06
Migrate Uniswap Trades

couralex6:
Migrating the following models:
```
$ dbt ls --resource-type model --output path --select +uniswap_trades | grep -v -e "legacy"
models/uniswap/arbitrum/uniswap_arbitrum_trades.sql
models/uniswap/bnb/uniswap_bnb_trades.sql
models/uniswap/ethereum/uniswap_ethereum_trades.sql
models/uniswap/optimism/uniswap_optimism_ovm1_pool_mapping.sql
models/uniswap/optimism/uniswap_optimism_pools.sql
models/uniswap/optimism/uniswap_optimism_trades.sql
models/uniswap/polygon/uniswap_polygon_trades.sql
models/uniswap/uniswap_trades.sql
models/uniswap/ethereum/uniswap_v1_ethereum_trades.sql
models/uniswap/ethereum/uniswap_v2_ethereum_trades.sql
models/uniswap/arbitrum/uniswap_v3_arbitrum_trades.sql
models/uniswap/bnb/uniswap_v3_bnb_trades.sql
models/uniswap/ethereum/uniswap_v3_ethereum_trades.sql
models/uniswap/optimism/uniswap_v3_optimism_trades.sql
models/uniswap/polygon/uniswap_v3_polygon_trades.sql
```


models/aave/ethereum/aave_v2_ethereum_interest_rates.sql
```diff
@@ -1,18 +1,18 @@
...
+select 
+  a.reserve, 
+  t.symbol,
+  date_trunc('hour',a.evt_block_time) as hour, 
+  avg(CAST(a.liquidityRate AS DOUBLE)) / 1e27 as deposit_apy, 
+  avg(CAST(a.stableBorrowRate AS DOUBLE)) / 1e27 as stable_borrow_apy, 
+  avg(CAST(a.variableBorrowRate AS DOUBLE)) / 1e27 as variable_borrow_apy
+from {{ source('aave_v2_ethereum', 'LendingPool_evt_ReserveDataUpdated') }} a
+left join {{ ref('tokens_ethereum_erc20_legacy') }} t
+on CAST(a.reserve AS VARCHAR(100)) = t.contract_address
```
couralex6, line 16: This most likely should not be cast to VARCHAR. I'm surprised this doesn't throw an error

models/aave/ethereum/aave_v2_ethereum_interest_rates.sql
```diff
@@ -1,18 +1,18 @@
...
+group by 1,2,3
```
0xRobin, line 17: These are non-changes that we shouldn't commit.. (they also trigger CI tests on these models)
I believe the migration script overwrites line-endings.
Will open a PR with a fix for that.

0xRobin:
@couralex6 can you try re-doing the process with the script modifications in [this PR](https://github.com/duneanalytics/spellbook/pull/3669)

0xRobin:
@couralex6 nevermind on the rerun, I've resolved it by cherry-picking the revert of those unwanted newline changes. :heavy_check_mark:

0xRobin:
this PR will also need the changes from https://github.com/duneanalytics/spellbook/pull/3676 to be able to run the uniswap seed tests.

jkylling:
I've ran the data quality script on the uniswap models in this PR. Most checksum match, except for the optimism models, and
* `token_bought_amount_raw`, because we get a double when doing `abs(string number)` in Spark. We also do a comparison of a string value and '0', which is potentially dangerous.
* `token_sold_amount_raw`, because we get a double when doing `abs(string number)` in Spark
* `amount_usd`, because we use doubles
* `maker`, probably because of different null handling
* `trace_address`, different type for trace address

The optimism models will require some investigation, but this should also be innocent.

0xRobin:
a bit of a special case here since the seed is shared between multiple models (both DuneSQL and Spark).
I think we replicate the seed file with a `_legacy.csv` suffix and reference the legacy seed in the `check_dex_seed_legacy` generic test.

0xRobin (changes requested):

jeff-dude:
> This now looks like the same problem @andrewhong5297 was facing, where duplicates occur after an incremental update when there is a null field in the `unique_key` combination.. I don't yet fully understand how this happens though, en why this behavior is different than on dbt-databricks.
> 
> In this case the `trace_address` is often hardcoded null (which makes sense if the source is an event table and not a call table). We also can't replace the nulls with `[ ]` as that would imply it's the top level function call. 🤔 cc: @jeff-dude @jkylling, ideas?

we did see a similar thing on spark, which is how i knew to check that on andrew's PR. maybe it was also `null` values, which is why we did empty string `''` instead? i can't remember the exact setup that caused it on spark, but i know for sure we did.

anyway, if `trace_address` is the only problem, i recommend we remove it universally in `dex.trades`, as it's never used in the full lineage and we planned to remove once we did the redesign post-migration. since we're migrating in batches of just upstream, i think we can safely remove on each as we go along

0xRobin:
ok, makes more sense that the same issue is also on spark. I probably just never ran into it. 
I now understand why this happens. The merge check is done with the standard equality operator `=` which is not null-safe. Meaning `(null = null) = false`, so the row equality check will always fail, resulting in the insertion of the same row again instead of a merge. 

We could look into using the null safe `<=>` equality operator in the future, but for now it's probably best to enforce a not-null policy on the `unique_key` columns.

+1 for removing the `trace_address` from the dex sector spells as a solution.

0xRobin:
I'll dump some more refs here that I've found as I've got no better place to put them.
null-safe equality in Spark (`<=>`)
https://docs.databricks.com/sql/language-manual/sql-ref-null-semantics.html#comparison-operators
null-safe equality in Trino (`IS NOT DISTINCT FROM`) https://trino.io/docs/current/functions/comparison.html#is-distinct-from-and-is-not-distinct-from

place where I think we could modify the merge behavior: https://github.com/starburstdata/dbt-trino/blob/f3e55e75843934c0d469f17d916f8e9ac03b6d25/dbt/include/trino/macros/materializations/incremental.sql#L162-L174

models/uniswap/arbitrum/uniswap_arbitrum_trades_legacy.sql
```diff
@@ -6,7 +6,7 @@
 }}
 
 {% set uniswap_models = [
-'uniswap_v3_arbitrum_trades'
+'uniswap_v3_arbitrum_trades_legacy'
```
jeff-dude, line 9: i noticed elsewhere that this setup of jinja code made us miss applying `_legacy` in refs downstream

no big deal -- just pointing it out and we'll need to monitor on migrations moving forward

jeff-dude: good catch!

jeff-dude (approved):
lfg 🚀 

approving with caveat for open question i dm'ed

couralex6:
This one is good to go @jeff-dude @0xRobin

0xRobin (approved):
let's get this in! :+1:

PR #3668, opened 2023-07-06
feat: new pool, fix token name, periphery address

vhawk19:
## Changes [timeswap]
- New arbitrum pool added
- Fix token name for the aura rdnt weth pool
- Update periphery addresses for the unshETH pool


jeff-dude (approved):
thank you!

PR #3674, opened 2023-07-06
dos2unix line endings

0xRobin:
We have a couple files with different line endings.
We should default to using unix (LF) instead of DOS (CRLF) line endings.

The changes in this PR are the result of running:
`find . -type f -print0 | xargs -0 dos2unix`
in the root directory. 

Windows users should set:
`git config --global core.autocrlf true` to handle conversions automatically


jkylling (approved):
Could you update the README with the git config suggestion?

jeff-dude (approved):

PR #3676, opened 2023-07-06
Seed and generic test support for DuneSQL

0xRobin:
1. Copied generic spark tests to a `_legacy` version
2. Changed all current test refs to use the `_legacy` version
3. Included DuneSQL versions of the `check_seed` generic macro and tests
4. added support for `varbinary` in seed columns for DuneSQL seeds

### DIFF
--- a/macros/test-helpers/check_seed_macro.sql
+++ b/macros/test-helpers/check_seed_macro.sql
@@ -54,8 +54,9 @@
     matching_count_test as (
         select
             'matched records count' as test_description,
-            count(model_{{seed_matching_columns[0]}}) as `result_model`,
-            1 as `expected_seed`,
+            cast(count(model_{{seed_matching_columns[0]}}) as varchar) as result_model,
+            cast(1 as varchar) as expected_seed,
+            (count(model_{{seed_matching_columns[0]}}) = 1) as equality_check,
             {%- for column_name in seed_matching_columns %}
             seed_{{column_name}} as {{column_name}}{% if not loop.last %},{% endif %}
             {% endfor -%}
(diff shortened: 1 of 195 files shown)


0xRobin:
After this is merged and you migrate a model that uses seed tests you should do the following steps to allow the seed test to run in DuneSQL.

1. add a `_schema.yml` file for your seed, add a `dunesql` tag and type any `varbinary` and `uint256` columns
```
version: 2

seeds:
  - name: your_seed
    config:
      tags: dunesql
      column_types:
        tx_hash: varbinary
        amount: uint256
```
2. Check that any tests reference your new migrated model and not the `_legacy` version.
3. if using generic tests, change to the non-legacy version in the model schema file.

macros/test-helpers/check_seed_macro.sql
```diff
@@ -54,8 +54,9 @@
     matching_count_test as (
         select
             'matched records count' as test_description,
-            count(model_{{seed_matching_columns[0]}}) as `result_model`,
-            1 as `expected_seed`,
+            cast(count(model_{{seed_matching_columns[0]}}) as varchar) as result_model,
+            cast(1 as varchar) as expected_seed,
```
jkylling, line 58: Why  the casts here?

jkylling (approved):

macros/test-helpers/check_seed_macro.sql
jkylling: Ah, makes sense. Could we add a comment on this? E.g., `cast all columns to varchar to unify column types`

PR #3664, opened 2023-07-05
migrate foundation_ethereum_base_trades

0xRobin:
example PR of migrating a standalone NFT model.

1. made a copy of the spark spell to `foundation_ethereum_base_trades_legacy.sql`
2. adapted all refs to use the `_legacy` spark model
3. modified the main spell to by DuneSQL compatible

checks:

- [x] dbt tests are successful
- [ ] migration data validation is successful

PR #3682, opened 2023-07-06
small infra fixes

0xRobin:
extracted the fixes needed to get seed tests working to a separate PR so we can merge this into `main`. 

That would allow us to keep the migration PR(s) clean (ex: https://github.com/duneanalytics/spellbook/pull/3664)


0xRobin:
Q: does the CI hang because the change the the source(..) macro makes all the models "modified"?

If so we'll need to manually merge this and do a prod dbt cloud run without modified I think?
cc: @jeff-dude

jeff-dude:
> Q: does the CI hang because the change to the source(..) macro makes all the spark models "modified"?
> 
> If so we'll need to manually merge this and do a prod dbt cloud run without modified I think? cc: @jeff-dude

we've removed the `state:modified` step a few times during merges, when it involves macros and such. i'd say we probably will continue to need to do so

macros/dune/ref.sql
```diff
@@ -0,0 +1,9 @@
+{% macro ref(model_name) %}
```
jkylling, line 1: Why is this one needed?

0xRobin: it's so that wizards who compile locally (without the db connection and with the dummy profile in the repo) don't have the database names in the compiled SQL. This allows the SQL to be run on dune.com without modifications.

jkylling: Aha, where does this show up? I thought we had fixed this in https://github.com/duneanalytics/spellbook/pull/3592

jkylling: Makes sense!

jkylling (approved):

0xRobin:
canceled the CI's as I reverted the `--exclude state:modified.macros` change so we don't commit that to `main`.

jeff-dude (approved):

PR #3685, opened 2023-07-06
Migrate Traces Gas

MSilb7:
Migrate Traces Gas to DuneSQL cc @hildobby 

[NOTE: This is lower priority than other spells - the best path forward may be to add this logic to ingestion?]


models/gas/arbitrum/gas_arbitrum_fees_traces.sql
```diff
@@ -13,55 +14,57 @@ WITH traces AS (
...
           , trace_address AS trace
-          , gas_used AS gas_used_original
-          , gas_used AS gas_used_trace
+          , cast(gas_used as double) AS gas_used_original
+          , cast(gas_used as double) AS gas_used_trace
           , block_time
           , block_number
           , input
           , type AS trace_type
+          , call_type
```
0xRobin, line 39: Our migration validation checks won't like schema changes (both type changes and new/removed columns).
It's probably preferred to keep all the migration changes in 1 PR do any modifications/additions in a follow up PR.

0xRobin (commented):
If all these models share the same logic I can help move the implementation to a macro so you have 1 place to modify the code instead of going between 7 files.

models/gas/arbitrum/gas_arbitrum_fees_traces.sql
MSilb7: cool i'll remove for now

MSilb7:
seems like this is failing for runtime now

MSilb7:
> If all these models share the same logic I can help move the implementation to a macro so you have 1 place to modify the code instead of going between 7 files.

yeah there's some nuances with L2s, but most are consistent

jeff-dude:
closing this for now, we can revisit after more active PRs are resolved

PR #3692, opened 2023-07-07
Change sources macro to run CI based on smaller sources

jkylling:


.github/workflows/dbt_slim_ci.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         run: "dbt compile $PROFILE $COMPILE_TAG"
 
       - name: dbt seed
-        run: "dbt seed $PROFILE --select state:modified$TAG --exclude $EXCLUDE tag:prod_exclude --state ."
+        run: "dbt seed $PROFILE --select +state:modified${TAG} --exclude $EXCLUDE tag:prod_exclude --state ."
```
0xRobin, line 70: you'll probably need `@` instead of `+` here to achieve what you're trying to do.
(seeds are not a parent from a model, but they are often parents from a test which is a child of a model..) 
We've done this before but removed it as it would bloat all PRs with almost all seeds.

jkylling: It's to fix this CI failure: https://github.com/duneanalytics/spellbook/actions/runs/5485534751/jobs/9994453915
Do you know what is up here?

macros/dune/source.sql
0xRobin, line 1: will this not break a lot of tests?
They often need data from arbitrary points in time (or all of history) that are not in the list of test_dates.

jkylling: Why do they need data from arbitrary points in time?

.github/workflows/dbt_slim_ci.yml
0xRobin: yeah the seeds are currently not materialized in prod.. :(
but I've not seen the previous problems we had with running those, so might be that we can just re-enable that schedule now that we're on dbt 1.4 instead of 1.2 :thinking:

macros/dune/source.sql
0xRobin: - seed tests checking the correctness of all columns for a specific set of transactions
- the uniqueness tests need to ensure the model isn't producing any duplicates on any point in time (not just on the arbitrary dates, as an edge case can very easily happen on another date..)

jeff-dude:
proposing we close this until we have a proper larger discussion on this topic. i definitely think it's worth keeping top of mind, but not sure we'll proceed in immediate short term?

trying to clean up the PR queue to track immediate migration items quicker

PR #3695, opened 2023-07-07
fix missing alias macro on nft parent model

jeff-dude:
fyi @0xRobin 

post-merge, commit manifest workflow failed: https://github.com/duneanalytics/spellbook/actions/runs/5487611575/jobs/9999293874

hopefully this is the only one missing and the logs didn't only show the first instance 🙏 

i tested a `dbt compile` locally before and after the change, and both worked, but that's with my trino profile set. the failure is on spark. this may be tricky to catch up front.


0xRobin:
ah good catch!
I guess this would have been caught by the spark CI on the PR?

jeff-dude:
yeah, v weird, it's happened back-to-back days now

it did fix everything though, commit manifest is back and working

PR #3698, opened 2023-07-07
Fix create_schema macro

jkylling:

### DIFF
--- a/macros/dune/schema.sql
+++ b/macros/dune/schema.sql
@@ -2,7 +2,7 @@
   {{ adapter.dispatch('create_schema', 'dbt')(relation) }}
 {% endmacro %}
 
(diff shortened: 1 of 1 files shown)


macros/dune/schema.sql
```diff
@@ -2,7 +2,7 @@
   {{ adapter.dispatch('create_schema', 'dbt')(relation) }}
 {% endmacro %}
 
-{% macro default__create_schema(relation) -%}
+{% macro trino__create_schema(relation) -%}
```
jeff-dude, line 5: where did you see downstream impact from this prior to change here?

jkylling: This breaks databricks if we ever need to create a new schema (e.g., for local development).

PR #3699, opened 2023-07-07
Migrate Chainlink PriceFeeds to DuneSQL, add Chainlink OCR

anon-r-7:
- Migrates Chainlink Price Feeds to DuneSQL
- Refactor Price Feeds for clarity and namespacing to `price_feeds_<>`
- Add OCR metrics for all supported evm chains

### DIFF
--- a/models/chainlink/bnb/chainlink_bnb_ocr_fulfilled_transactions.sql
+++ b/models/chainlink/bnb/chainlink_bnb_ocr_fulfilled_transactions.sql
@@ -1,0 +1,53 @@
+{{
+  config(
+    tags=['dunesql'],
+    alias='ocr_fulfilled_transactions',
+    materialized='view',
+    post_hook='{{ expose_spells(\'["bnb"]\',
+                                "sector",
+                                "chainlink",
+                                \'["linkpool_ryan"]\') }}'
+  )
+}}
+
+{% set incremental_interval = '7' %}
+
+WITH
+  bnb_usd AS (
+    SELECT
+      minute as block_time,
+      price as usd_amount
+    FROM
+      {{ source('prices', 'usd') }} price
+    WHERE
+      symbol = 'BNB'
+      {% if is_incremental() %}
+        AND minute >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
+      {% endif %}      
+  ),
+  ocr_fulfilled_transactions AS (
+    SELECT
+      MAX(tx.block_time) as block_time,
+      tx."from" as "node_address",
+      MAX((cast((gas_used) as double) / 1e18) * gas_price) as token_amount,
+      MAX(bnb_usd.usd_amount) as usd_amount
+    FROM
+      {{ source('bnb', 'transactions') }} tx
+      RIGHT JOIN {{ ref('chainlink_bnb_ocr_gas_transmission_logs') }} ocr_gas_transmission_logs ON ocr_gas_transmission_logs.tx_hash = tx.hash
+      LEFT JOIN bnb_usd ON date_trunc('minute', tx.block_time) = bnb_usd.block_time
+    {% if is_incremental() %}
+      WHERE tx.block_time >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
+    {% endif %}      
+    GROUP BY
+      tx.hash,
+      tx.index,
+      tx."from"
+  )
+SELECT
+ 'bnb' as blockchain,
+ block_time,
+ node_address,
+ token_amount,
+ usd_amount
+FROM
+  ocr_fulfilled_transactions
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -771,10 +771,18 @@ models:
 
     chainlink:
       +schema: chainlink
-      ethereum:
-        +schema: chainlink_ethereum
+      arbitrum:
+        +schema: chainlink_arbitrum
+      avalanche_c:
+        +schema: chainlink_avalanche_c
       bnb:
         +schema: chainlink_bnb
+      ethereum:
+        +schema: chainlink_ethereum
+      fantom:
+        +schema: chainlink_fantom
+      gnosis:
+        +schema: chainlink_gnosis
       optimism:
         +schema: chainlink_optimism
       polygon:
--- a/models/chainlink/README.md
+++ b/models/chainlink/README.md
@@ -1,4 +1,16 @@
-## OCR Metrics Example
+# Price Feeds
+
+## Optimism 
+
+To add a new price feed, add a row to the `chainlink_optimism_price_feeds_oracle_addresses.sql` table with the following information:
+- **feed_name**: Sourced from [Chainlink Docs](https://docs.chain.link/docs/optimism-price-feeds/)
+- **decimals**: Sourced from [Chainlink Docs](https://docs.chain.link/docs/optimism-price-feeds/) (Note: This refers to the price feed decimals, not the underlying token's decimals)
+- **proxy_address**: Sourced from [Chainlink Docs](https://docs.chain.link/docs/optimism-price-feeds/)
+- **aggregator_address**: Sourced from opening the proxy address on Etherscan ([example](https://optimistic.etherscan.io/address/0x338ed6787f463394D24813b297401B9F05a8C9d1#readContract)), clicking 'Contract' -> 'Read Contract' and getting the address from the 'aggregator' field
+
+*Open Research Area: Is there a way for us to deterministically build the oracle -> address -> feed name -> token links purely by reading on-chain events vs manually entering data from Chainlink docs?*
+
+## OCR Metrics Usage Example
 
 Aggregate OCR data monthly with margin and  aggregation, along with margin by node operator
 
@@ -10,7 +22,7 @@ WITH
       operator_name,
       SUM(token_amount) as reward_token,
       SUM(usd_amount) as reward_usd
-    FROM test_schema.git_dunesql_1294d1da_chainlink_ethereum_ocr_reward_daily
+    FROM chainlink_ethereum_ocr_reward_daily
     WHERE operator_name = 'LinkPool'
       AND date_start >= cast('2023-01-01' as date)
     GROUP BY 1, 2
@@ -21,7 +33,7 @@ WITH
       operator_name,
       SUM(total_token_amount) as gas_token,
       SUM(total_usd_amount) as gas_usd
-    FROM test_schema.git_dunesql_1294d1da_chainlink_ethereum_ocr_gas_daily
+    FROM chainlink_ethereum_ocr_gas_daily
     GROUP BY 1, 2
   ),
   request as (
@@ -31,7 +43,7 @@ WITH
       SUM(fulfilled_requests) as fulfilled_requests,
       SUM(reverted_requests) as reverted_requests,
       SUM(total_requests) as total_requests
-    FROM test_schema.git_dunesql_1294d1da_chainlink_ethereum_ocr_request_daily
+    FROM chainlink_ethereum_ocr_request_daily
     GROUP BY 1, 2
   )
 SELECT 
--- a/models/chainlink/arbitrum/chainlink_arbitrum_ocr_fulfilled_transactions.sql
+++ b/models/chainlink/arbitrum/chainlink_arbitrum_ocr_fulfilled_transactions.sql
@@ -1,0 +1,55 @@
+{{
+  config(
+    tags=['dunesql'],
+    alias='ocr_fulfilled_transactions',
+    materialized='view',
+    post_hook='{{ expose_spells(\'["arbitrum"]\',
+                                "sector",
+                                "chainlink",
+                                \'["linkpool_ryan"]\') }}'
+  )
+}}
+
+{% set incremental_interval = '7' %}
+
+WITH
+  arbitrum_usd AS (
+    SELECT
+      minute as block_time,
+      price as usd_amount
+    FROM
+      {{ source('prices', 'usd') }} price
+    WHERE
+      symbol = 'ETH'
+      {% if is_incremental() %}
+        AND minute >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
+      {% endif %}      
+  ),
+  ocr_fulfilled_transactions AS (
+    SELECT
+      MAX(tx.block_time) as block_time,
+      tx."from" as "node_address",
+      MAX(
+        (cast((gas_used) as double) / 1e18) * effective_gas_price
+      ) as token_amount,
+      MAX(arbitrum_usd.usd_amount) as usd_amount
+    FROM
+      {{ source('arbitrum', 'transactions') }} tx
+      RIGHT JOIN {{ ref('chainlink_arbitrum_ocr_gas_transmission_logs') }} ocr_gas_transmission_logs ON ocr_gas_transmission_logs.tx_hash = tx.hash
+      LEFT JOIN arbitrum_usd ON date_trunc('minute', tx.block_time) = arbitrum_usd.block_time
+    {% if is_incremental() %}
+      WHERE tx.block_time >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
+    {% endif %}      
+    GROUP BY
+      tx.hash,
+      tx.index,
+      tx."from"
+  )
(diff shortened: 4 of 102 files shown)


alokeutpal (approved):

jeff-dude:
thank you for initiating the effort for this migration 🙌 

anything in particular we can help with? the CI test error you're hitting typically happens when you alter part of the lineage for migration, but not the entire lineage.

for example:
- migrate upstream chainlink spell(s)
- any downstream spell needs altered to make the `ref()` use `_legacy` in the referenced model name, until that downstream spell is also migrated to dunesql (ensuring no cross-engine lineage) -- this is likely the step missing, without me looking v closely at the modified files

jeff-dude:
> Thanks @jeff-dude for the reach out! For next steps I was just waiting on the price.usd migration that you'd mentioned last week could happen this week. After that, I'll replicate the new OCR tables from ethereum to all eight supported EVM chains. Any update on the prices.usd?

i had some open questions that gave me pause to merge, but i think i'm all set now. i will make a push to get done early next week

jeff-dude:
when the step for common models fails, it typically is one of two things:
- missing new alias macro on the alias property in config block (example in readme, also in some merged PRs)
- not adding `_legacy` to any `{{ ref( ) }}` in downstream models that are still spark, and need to read from spark, rather than new dunesql spell (assuming you are editing most upstream spells in lineage)

jeff-dude:
also please ensure you add the dunesql tag in models which are migrating (again, readme and merged PRs have examples)

jeff-dude:
> This CI is still failing. The issue has to do with using timestamp in a date_trunc on an incremental view. For example: `cast(date_trunc('hour', tx.block_time) as timestamp)`.
> 
> I've tried all suggestions provided in discord, reviewed dozens of PRs, tried every suggestion in the [Date and Time Docs](https://dune.com/docs/query/DuneSQL-reference/Functions-and-operators/datetime/).
> 
> I've made 24 attempts to get this to work in the CI over the last two weeks, and all have failed despite the same compiled queries working in the UI. At this point, I'm assuming this is a bug that requires fixed. I've created a [bug issue](https://github.com/duneanalytics/spellbook/issues/3799).
> 
> Please advise.

thank you for being proactive and continuing to try solutions. our query team just extended dunesql to allow writing timestamp in various formats as of this morning (EST). can you try one more time, using the format:
`TIMESTAMP <column>/hardcoded date/function`
to write your column as timestamp

please note, you still will not be able to partition by a timestamp column, so ensure all partitions are casted as date, as we intended partitions to never be smaller than a day

maybe focus on one model to try this first. if it still fails, we can tag the query team here to see why

models/chainlink/ethereum/chainlink_ethereum_ocr_reverted_transactions.sql
```diff
@@ -0,0 +1,58 @@
+{{
+  config(
+    tags=['dunesql'],
+    alias='ocr_reverted_transactions',
+    partition_by = ['block_time'],
```
jkylling, line 5: It's not possible to partition by `block_time`. This will create one partition per block_time, and likely an enormous amount of partitions.

You likely want to add an additional column `block_date`, defined as `cast(block_time as date) as block_date`, and partition by `block_date` instead.

models/chainlink/ethereum/chainlink_ethereum_ocr_fulfilled_transactions.sql
```diff
@@ -38,7 +38,7 @@ WITH
     FROM
       {{ source('ethereum', 'transactions') }} tx
       RIGHT JOIN {{ ref('chainlink_ethereum_ocr_gas_transmission_logs') }} ocr_gas_transmission_logs ON ocr_gas_transmission_logs.tx_hash = tx.hash
-      LEFT JOIN ethereum_usd ON date_trunc('minute', cast(tx.block_time as timestamp)) = ethereum_usd.block_time
+      LEFT JOIN ethereum_usd ON date_trunc('minute', timestamp tx.block_time) = ethereum_usd.block_time
```
jkylling, line 41:
```suggestion
      LEFT JOIN ethereum_usd ON date_trunc('minute', tx.block_time) = ethereum_usd.block_time
```

models/chainlink/ethereum/chainlink_ethereum_ocr_reverted_transactions.sql
```diff
@@ -37,7 +37,7 @@ WITH
       MAX(ethereum_usd.usd_amount) as usd_amount
     FROM
       {{ source('ethereum', 'transactions') }} tx
-      LEFT JOIN ethereum_usd ON date_trunc('minute', cast(tx.block_time as timestamp)) = ethereum_usd.block_time
+      LEFT JOIN ethereum_usd ON date_trunc('minute', timestamp tx.block_time) = ethereum_usd.block_time
```
jkylling, line 40:
```suggestion
      LEFT JOIN ethereum_usd ON date_trunc('minute',  tx.block_time) = ethereum_usd.block_time
```

models/chainlink/optimism/chainlink_optimism_price_feeds_hourly.sql
```diff
@@ -122,7 +122,7 @@ FROM (
                 AVG(price_feeds.underlying_token_price) AS underlying_token_price_avg
             FROM hourly_sequence 
             LEFT JOIN {{ ref('chainlink_optimism_price_feeds') }} price_feeds 
-            ON hourly_sequence.hr = date_trunc('hour', price_feeds.block_time)
+            ON hourly_sequence.hr = date_trunc('hour', timestamp price_feeds.block_time)
```
jkylling, line 125:
```suggestion
            ON hourly_sequence.hr = date_trunc('hour', price_feeds.block_time)
```

models/chainlink/polygon/chainlink_polygon_price_feeds_hourly.sql
```diff
@@ -122,7 +122,7 @@ FROM (
                 AVG(price_feeds.underlying_token_price) AS underlying_token_price_avg
             FROM hourly_sequence 
             LEFT JOIN {{ ref('chainlink_polygon_price_feeds') }} price_feeds 
-            ON hourly_sequence.hr = date_trunc('hour', price_feeds.block_time)
+            ON hourly_sequence.hr = date_trunc('hour', timestamp price_feeds.block_time)
```
jkylling, line 125:
```suggestion
            ON hourly_sequence.hr = date_trunc('hour', price_feeds.block_time)
```

jkylling:
You can see in the CI logs which model is failing https://github.com/duneanalytics/spellbook/actions/runs/5599296341/jobs/10241763951. The last failing CI run failed on:
```
13:56:07  Database Error in model chainlink_ethereum_ocr_reverted_transactions (models/chainlink/ethereum/chainlink_ethereum_ocr_reverted_transactions.sql)
13:56:07    TrinoUserError(type=USER_ERROR, name=NOT_SUPPORTED, message="unsupported type: timestamp(3) with time zone", query_id=20230[71](https://github.com/duneanalytics/spellbook/actions/runs/5599296341/jobs/10241763951#step:12:72)9_135546_13393_6tppc)
13:56:07    compiled Code at target/run/spellbook/models/chainlink/ethereum/chainlink_ethereum_ocr_reverted_transactions.sql
```
This tells us that something is up with the `chainlink_ethereum_ocr_reverted_transactions` model. Unfortunately the logs are not very verbose, but this type of error message is caused by partitioning on a `timestamp(3) with time zone` column, which is unsupported.

models/chainlink/bnb/chainlink_bnb_price_feeds_hourly.sql
```diff
@@ -24,7 +24,7 @@ WITH
    FROM
     {% if not is_incremental() %}
       unnest(
-        sequence(0, CAST(date_diff('hour', timestamp '{{project_start_date}}', current_timestamp) AS INTEGER))
```
jkylling: On the other hand, this is required.

I think we need to do a primer on SQL syntax for literals:
You can specify timestamp _literals_ in two ways:
1. `timestamp '2023-01-01'`
2. cast('2023-01-01' as timestamp)

This syntax is only valid for _literals_. For general timestamp conversions you must use casts. That is, `cast(some_column as timestamp)`.

The rule is:
Use `timestamp '2023-01-01'` for hard coded values (which project_start_date is), use `cast(column_name as timestamp)` for converting columns. You can use casts in both cases, but it's cleaner to use timestamp literals for hard coded values.

jkylling: Yes, this is correct, except for the extra parenthesis in the cast.

It's true that Trino is a bit more picky on types than Postgres or Spark. In Postgres or Spark you are probably used to being able to compare strings directly with date and timestamp types, while this is not possible in Trino. For instance,
```
select block_time > '2023-01-01'
```
is valid in Postgres or Spark, but must be changed to
```
select block_time > timestamp '2023-01-01'
```
in Trino.

Similarly, if you happen to have a column with dates as varchar (though, this should not happen), you would need to do a cast before comparing, so we would have to change
```
select  block_time > some_column_with_varchar_dates
```
to
```
select  block_time > cast(some_column_with_varchar_dates as timestamp)
```

Except for that it should work mostly the same. Between different timestamp and date types Trino does automatic conversions, similar to Postgres and Spark. It's only between varchar types and timestamp types that Trino does not do automatic conversions, while Postgres and Spark do.

models/chainlink/optimism/chainlink_optimism_price_feeds_hourly.sql
```diff
@@ -1,115 +1,141 @@
...
 
-WITH gs AS (
+WITH
+  hourly_sequence_meta as (
+   SELECT
+    * 
+   FROM
+    {% if not is_incremental() %}
+      unnest(
+        sequence(0, CAST(date_diff('hour', '{{project_start_date}}', current_timestamp) AS INTEGER))
```
jkylling, line 27:
```suggestion
        sequence(0, CAST(date_diff('hour', timestamp '{{project_start_date}}', current_timestamp) AS INTEGER))
```

models/chainlink/bnb/chainlink_bnb_price_feeds.sql
```diff
@@ -1,59 +1,59 @@
...
-    alias = alias('price_feeds'),
-    partition_by = ['block_date'],
-    materialized = 'incremental',
-    file_format = 'delta',
-    incremental_strategy = 'merge',
-    unique_key = ['blockchain', 'block_number', 'proxy_address','underlying_token_address'],
+{{
+  config(
+    tags=['dunesql'],
+    alias='price_feeds',
```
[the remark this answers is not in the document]
jeff-dude: this is for dunesql versions. the legacy files should already have the alias properly configured.

jeff-dude:
@anon-r-7 while starting the review, i'm finding i have some changes to request, but i think this will go much quicker if we break this PR out into multiple PRs. it'll help us both move at a quicker pace. i'm struggling a bit on the 120+ edited files, new spells, migrated spells, removed spells, etc.

ideas for breaking this up:
- PR for any spell removals (i thought i saw some bnb ones being removed?)
- PR per blockchain
- PR for migration only, no new spells
- PR for new spells only, dunesql only

ideally we keep the modified count smaller per PR, so i can help you get through quicker. i'll leave it up to you how to break it up with the ideas above. of course, you can keep this branch on your fork, then just push commits and/or move code around various branches to accomplish how you break it up.

models/chainlink/bnb/chainlink_bnb_ocr_fulfilled_transactions.sql
```diff
@@ -0,0 +1,53 @@
+{{
+  config(
+    tags=['dunesql'],
+    alias='ocr_fulfilled_transactions',
```
jeff-dude, line 4: apply alias macro on all alias properties:
`alias = alias('<tablename>'),`
similar to approach on legacy files, but we default the second param to false to indicate dunesql

models/chainlink/bnb/chainlink_bnb_ocr_fulfilled_transactions.sql
```diff
@@ -0,0 +1,53 @@
...
+    materialized='view',
```
jeff-dude, line 5: you materialize as a view, yet use incremental logic below. this won't do anything. it'll simply ignore all incremental logic. you should either:
- make this incremental, then add the other config properties needed (file type delta, unique keys, merge approach, etc) -- there are plenty of examples in spellbook to follow
- keep as view, but remove anything incremental in the model

models/chainlink/bnb/chainlink_bnb_ocr_fulfilled_transactions.sql
```diff
@@ -0,0 +1,53 @@
...
+  ),
+  ocr_fulfilled_transactions AS (
+    SELECT
+      MAX(tx.block_time) as block_time,
+      tx."from" as "node_address",
+      MAX((cast((gas_used) as double) / 1e18) * gas_price) as token_amount,
+      MAX(bnb_usd.usd_amount) as usd_amount
+    FROM
+      {{ source('bnb', 'transactions') }} tx
+      RIGHT JOIN {{ ref('chainlink_bnb_ocr_gas_transmission_logs') }} ocr_gas_transmission_logs ON ocr_gas_transmission_logs.tx_hash = tx.hash
```
jeff-dude, line 36: we've seen some poor performance on joins like this. you don't notice in CI tests because it's simply building a view, not running the query. you would likely see the poor performance later downstream.

you're essentially joining base transactions to base logs, both massive tables, especially on bnb. it would be good to apply more join conditions as well -- block number? date filters (i.e. incremental block filter on join conditions for both historical run and incremental)?

jeff-dude (changes requested):
i only left a few comments, but i'm seeing common themes throughout bnb spells here. most are views, with incremental logic that won't do anything based on config. the views are also reading from the massive bnb base tables. i'd advise starting at the most upstream spells in the lineage (i.e. spells which read from sources, not refs) and look to make fully incremental.

general pattern to think of:
- materialize upstream spells that read from large source tables
- if using views, save those for spells which read from other spells in step 1 (even then, materializing still makes sense if it's joining a bunch of tables)

when materializing, we should consider filters and join conditions as much as possible to speed up the read and write.
- joining base tables? in addition to hash join, look to filter on dates as much as possible, plenty of examples in spellbook (both incremental and non_incremental)

left a few other minor comments as well

jeff-dude:
i'm back to thinking breaking up this PR is the best bet for both of us.
- one PR for solely migrating existing spells to dunesql, no other changes included
- one PR for new spells, dunesql only, focus on only one blockchain first, also focus on just the most upstream spells in the new lineage so we can set a solid foundation before adding the downstream spells

this is a lot of work jammed into one PR and can easily get lost along the way

jeff-dude:
> I've followed the best practice recommendations from dbt which is overwhelmingly "start with views, only use incremental when necessary".

you're more than welcome to leave as views and remove the incremental jinja logic. it actually makes it easier on our backend orchestration, as the views are built once and then it's all on the experience of querying the view in the dune app downstream. this would expedite the dev process here too.

i'm only trying to save you time by recommending incremental, as i've seen in the past when views are built in this way on such large base tables, they can be slow to query.

if you want to start with views, then attempt to write queries and dashboards on top, we can do that. you can always come back later and refactor if it doesn't work out well.

jeff-dude:
> > I've followed the best practice recommendations from dbt which is overwhelmingly "start with views, only use incremental when necessary".
> 
> you're more than welcome to leave as views and remove the incremental jinja logic. it actually makes it easier on our backend orchestration, as the views are built once and then it's all on the experience of querying the view in the dune app downstream. this would expedite the dev process here too.
> 
> i'm only trying to save you time by recommending incremental, as i've seen in the past when views are built in this way on such large base tables, they can be slow to query.
> 
> if you want to start with views, then attempt to write queries and dashboards on top, we can do that. you can always come back later and refactor if it doesn't work out well.

whichever approach you take here, i'd still suggest separate PRs as noted above. it's going to be easier for me to push through for you.

i didn't realize you didn't write the original price feeds spells. i appreciate you taking the time to help migrate them in order to unblock yourself. the other option is to wait for us to get around to them, but i think your path is likely preferred to expedite it all.

PR #3700, opened 2023-07-07
T-31518 `prices_tokens` lineage migration

jeff-dude:
one major open question:
how to handle the union of solana into this, as its addresses are varchar everywhere, yet others are varbinary?
cc @andrewhong5297 

fyi @Hosuke on a few labels models switching ref to `_legacy` version, until we can finalize those migrations and change back to remove `_legacy`

fyi @MSilb7 on optimism ones included


jeff-dude:
@aalan3 anything you can think of that would affect downstream `prices.usd` if this moves to dunesql?

MSilb7:
For Optimism, we should be able to just to `prices_tokens` without sub files for curated, bridged, etc. But it seems like those files already existed.

Hosuke:
> one major open question: how to handle the union of solana into this, as its addresses are varchar everywhere, yet others are varbinary? cc @andrewhong5297
> 
> fyi @Hosuke on a few labels models switching ref to `_legacy` version, until we can finalize those migrations and change back to remove `_legacy`
> 
> fyi @MSilb7 on optimism ones included

The solution I used in past migration.
```
        , CASE
            WHEN blockchain = 'solana'
            THEN from_base58(contract_address)
            ELSE contract_address
        END AS contract_address
```

models/prices/optimism/prices_optimism_tokens_bridged.sql
```diff
@@ -1,17 +1,17 @@
...
+        materialized = 'table',
         file_format = 'delta',
-        tags=['static']
+        tags = ['dunesql']
         )
 }}
 SELECT
     e.token_id
-    , 'optimism' as blockchain
+    , LOWER(TRIM('optimism')) as blockchain
```
jkylling, line 11:
```suggestion
    , 'optimism' as blockchain
```

models/prices/prices_native_tokens.sql
```diff
@@ -1,57 +1,57 @@
...
+        materialized = 'table',
         file_format = 'delta',
-        tags=['static']
+        tags = ['static', 'dunesql']
         )
 }}
 SELECT 
     TRIM(token_id) as token_id
-    , LOWER(TRIM(blockchain)) as blockchain
+    , CAST(LOWER(TRIM(blockchain)) as VARCHAR) as blockchain
```
jkylling, line 11:
```suggestion
    , blockchain
```

models/prices/prices_native_tokens.sql
```diff
@@ -1,57 +1,57 @@
...
     , TRIM(symbol) as symbol
-    , LOWER(TRIM(contract_address)) as contract_address
-    , CAST(decimals as int)
+    , CAST(LOWER(TRIM(contract_address)) as VARBINARY) as contract_address
```
jkylling, line 13:
```suggestion
    , contract_address
```

jkylling (commented):
It should be possible to remove all usage of LOWER and TRIM.

For Solana addresses we should use from_base58 in the source or when unioning.

jeff-dude:
> For Optimism, we should be able to just to `prices_tokens` without sub files for curated, bridged, etc. But it seems like those files already existed.
> 
> Re-read again: my vote is to kill the curated/bridged table and just select from `tokens_optimism_erc20` where decimals is not null (which means we mapped it somehow). There are multiple potential tokens sources, so it's weird to hardcode the 2 input files.

@MSilb7 can you confirm [this commit](https://github.com/duneanalytics/spellbook/pull/3700/commits/bb9f3a14cc302a25d56de4cacf93b733769efa15)?

Hosuke:
> it looks like base / decoded solana tables still have varchar values. is the expectation that when joining to this table, they'd cast in the join condition?

I used `typeof()` function to manually check and find the varchar columns in source tables.
Maybe we could consider to add type assurance macro test in dune sql spells.

models/prices/optimism/prices_optimism_tokens_curated.sql
```diff
@@ -1,119 +0,0 @@
...
-    , 'optimism' as blockchain
-    , TRIM(symbol) as symbol
-    , LOWER(TRIM(contract_address)) as contract_address
-    , decimals
-FROM
-(
-    VALUES
-
-    --tokens not yet supported or are not active on coinpaprika are commented out
-    ("op-optimism", "OP", "0x4200000000000000000000000000000000000042", 18),
```
aalan3, line 20: Tried to wrap my head around this but not sure how optimism prices work. but it looks like this table should be included in the tokens list?

models/prices/optimism/prices_optimism_tokens.sql
```diff
@@ -3,22 +3,19 @@
...
+        tags = ['static', 'dunesql']
         )
 }}
 SELECT
-    token_id
-    , blockchain
-    , symbol
-    , contract_address
-    , decimals
-FROM {{ ref('prices_optimism_tokens_curated') }}
```
aalan3, line 15: I think we need to include curated here still

aalan3 (commented):
Had a couple of comments regarding optimism prices. We should also try to fix the script that's failing?

jeff-dude:
> We should also try to fix the script that's failing?

the prices automation script could use some love for sure, but it mostly works well on one-off additions. there are some known scenarios where it returns a failure, but we can still override and merge with manual proof. for instance, a `token_id` found in coinpaprika only returns ethereum info, but really that token is on more blockchains. we can still use the output from ethereum, as long as the addresses are correct for the specific blockchain. in those scenarios, we ask the users to link to block explorer for proof.

models/prices/optimism/prices_optimism_tokens.sql
jeff-dude: per @MSilb7 recommendation in earlier comments on this PR, we can remove these. optimism will rely on curated manual adds from the ethereum file and inner join to his other spell for bridged mappings of addresses to filter down the results optimism cares about.

@MSilb7 can you confirm once again we're good here for optimism?

aalan3: Regarding selecting only from `tokens_optimism_erc20`, we still need the coinpaprika ids. But I agree that it could be just one file. If we do it the way we have now, people can only see prices for tokens that are in ethereum. So if there's a token that exists on optimism but not ethereum it's not going to work, or if the address is different.

MSilb7: cc @aalan3 & @jeff-dude , the optimism prices file should stay as it was before. I had the erc20 tokens file and the prices file confused.

jeff-dude: thank you. i've reverted OP back @aalan3

aalan3 (approved):

Hosuke:
> > We should also try to fix the script that's failing?
> 
> the prices automation script could use some love for sure, but it mostly works well on one-off additions. there are some known scenarios where it returns a failure, but we can still override and merge with manual proof. for instance, a `token_id` found in coinpaprika only returns ethereum info, but really that token is on more blockchains. we can still use the output from ethereum, as long as the addresses are correct for the specific blockchain. in those scenarios, we ask the users to link to block explorer for proof.
> 
> i'd suggest doing any fixes in its own branch / PR as a followup, as we know the current lists work

I try to upgrade the token-checker to support dune sql `prices.tokens`:
https://github.com/duneanalytics/spellbook/pull/3987

PR #3701, opened 2023-07-08
Add Velodrome V2 to DEXs

MSilb7:
Good to go.

- Adds Velodrome V2 to dex trades and incentive contract mappings.
- Includes legacy & Dunesql versions where relevant.


MSilb7:
**NEEDS HELP: Something with Seeds configuration is causing fails. It can't find the seeds schema.**

0xRobin:
@MSilb7 when this is merged: https://github.com/duneanalytics/spellbook/pull/3702
It should be as easy as just changing the test from `check_dex_seed_legacy` to `check_dex_seed` in your model .yml schema. 

You did not find a schema because there was non (it's only really needed when you want to override the column types)

MSilb7:
> #3702

Awesome! Made the yml change now, so it should be good once that gets merged and we update the branch. @0xRobin

MSilb7:
Also having permissions errors here.

+ following up on the dex seeds issue (if this is resolved) @0xRobin

MSilb7:
I don't believe this is blocked anymore

cc @0xRobin for the comment on dex seed

models/velodrome/optimism/velodrome_optimism_trades.sql
```diff
@@ -52,14 +81,14 @@ SELECT
...
         else concat(erc20a.symbol, '-', erc20b.symbol)
     end as token_pair
-    ,dexs.token_bought_amount_raw / power(10, erc20a.decimals) AS token_bought_amount
-    ,dexs.token_sold_amount_raw / power(10, erc20b.decimals) AS token_sold_amount
-    ,dexs.token_bought_amount_raw  AS token_bought_amount_raw
-    ,dexs.token_sold_amount_raw  AS token_sold_amount_raw
+    ,cast(dexs.token_bought_amount_raw as double) / cast( power(10, erc20a.decimals) as double) AS token_bought_amount
+    ,cast(dexs.token_sold_amount_raw as double) / cast( power(10, erc20b.decimals) as double) AS token_sold_amount
+    ,CAST(dexs.token_bought_amount_raw AS double) AS token_bought_amount_raw
+    ,CAST(dexs.token_sold_amount_raw AS double) AS token_sold_amount_raw
```
jeff-dude, line 87: we shouldn't cast these, unless necessary to get source types to `uint256`, which is what we use for raw amounts in `dex.trades`

models/velodrome/optimism/velodrome_optimism_trades.sql
```diff
@@ -52,14 +81,14 @@ SELECT
         when lower(erc20a.symbol) > lower(erc20b.symbol) then concat(erc20b.symbol, '-', erc20a.symbol)
         else concat(erc20a.symbol, '-', erc20b.symbol)
     end as token_pair
-    ,dexs.token_bought_amount_raw / power(10, erc20a.decimals) AS token_bought_amount
-    ,dexs.token_sold_amount_raw / power(10, erc20b.decimals) AS token_sold_amount
-    ,dexs.token_bought_amount_raw  AS token_bought_amount_raw
-    ,dexs.token_sold_amount_raw  AS token_sold_amount_raw
+    ,cast(dexs.token_bought_amount_raw as double) / cast( power(10, erc20a.decimals) as double) AS token_bought_amount
+    ,cast(dexs.token_sold_amount_raw as double) / cast( power(10, erc20b.decimals) as double) AS token_sold_amount
```
jeff-dude, line 85: these likely don't need casted either

models/velodrome/optimism/velodrome_optimism_trades.sql
```diff
@@ -52,14 +81,14 @@ SELECT
...
+    ,CAST(dexs.token_bought_amount_raw AS double) AS token_bought_amount_raw
+    ,CAST(dexs.token_sold_amount_raw AS double) AS token_sold_amount_raw
     ,coalesce(
         dexs.amount_usd
-        ,(dexs.token_bought_amount_raw / power(10, p_bought.decimals)) * p_bought.price
-        ,(dexs.token_sold_amount_raw / power(10, p_sold.decimals)) * p_sold.price
+        ,(cast(dexs.token_bought_amount_raw as double) / cast(power(10, p_bought.decimals) as double) ) * p_bought.price
+        ,(cast(dexs.token_sold_amount_raw as double) / cast(power(10, p_sold.decimals) as double) ) * p_sold.price
```
jeff-dude, line 91: same here

jeff-dude (changes requested):

models/velodrome/optimism/velodrome_optimism_trades.sql
MSilb7: If we don't cast, I think we'll lose the decimals (basing on queries in Dune UI, I always have to cast as double)

jeff-dude: we don't do it on other dex models, i.e. here:
https://github.com/duneanalytics/spellbook/blob/main/models/uniswap/ethereum/uniswap_v3_ethereum_trades.sql#L56-L57

when you merge into `dex.trades`, it may fail on the union on data type potentially.

we have a beta `dex.trades` until migration is final (only two projects away), you should try adding it in this PR to include this project into sector spell and see if union works fine
https://github.com/duneanalytics/spellbook/blob/main/models/dex/dex_trades_migration_beta.sql

0xRobin: You'll only perform integer division (which loses the decimals) if both sides are integer types. 
`pow(x,y)` is double by default so removing the casts shouldn't be a problem. :ok_hand: 
https://dune.com/queries/2982847

jeff-dude (approved):

PR #3702, opened 2023-07-10
Dex trades legacy seed

0xRobin:
This adds a legacy version of the dex trades seed test (and seed itself)


jkylling (approved):

jeff-dude (approved):
@couralex6 was messing with this a bit too during uniswap migration. just tagging to get his approval and awareness.

jeff-dude (approved):
this assumes @couralex6 has the new `tests/generic/check_dex_seed.sql` added when migrating (and translated)? so tests pass in CI?

then since we aren't modifying schema files, we won't ever use `check_dex_seed_legacy` test?

couralex6:
I already have most of these changes in #3667 . Looks like I'm missing the some `'dex_trades_seed'` -> `'dex_trades_seed_legacy'`. Do you mind if I just import those in the uniswap PR?

0xRobin:
> then since we aren't modifying schema files, we won't ever use `check_dex_seed_legacy` test?

correct for `dunesql` models, `check_dex_seed_legacy` will still be used for non-migrated spark models.

0xRobin:
> Do you mind if I just import those in the uniswap PR?

you can definitely merge this into your PR, but I still suggest we merge this one into main asap as there are other PRs that need this change.

tests/generic/check_dex_seed_legacy.sql
jeff-dude, line 1: this is what i was referring to -- we don't have the dunesql version of this macro in this PR? is that coming in another PR?

jeff-dude: we have the seed itself, but not the test

0xRobin: oh @jeff-dude I get your comment now!

we already have both `check_dex_seed` (dunesql) and `check_dex_seed_legacy` on main.
https://github.com/duneanalytics/spellbook/blob/main/tests/generic/check_dex_seed.sql

They where just using both the same seed which was an initial oversight.

jeff-dude: ahhh okay, my fault, thank you

jeff-dude (approved):

jeff-dude:
merging now @couralex6

PR #3703, opened 2023-07-10
[Easy] [ERC20] Add missing tokens

gentrexha:
Usual missing tokens update.

cc @bh2smith


gentrexha:
I'm really confused by the error I'm receiving in the CI-CD pipeline:

```
TrinoUserError(type=USER_ERROR, name=SYNTAX_ERROR, message="line 28432:55: Binary literal must contain an even number of digits", query_id=20230711_090012_08119_buw8w)
```

All of the tokens that have been introduced are 42 digits long. This can be confirmed by this regex: `(0x.{40})`. @jeff-dude Is there anything else I'm missing here?

bh2smith (approved):
Noice!

jeff-dude:
it's still running, so not 100% sure it's resolved, but i pushed a commit to fix symbols which were missing single quotes. it looks like the trend is:
- symbol starts with `0x`, the quotes aren't used, dunesql engine yells at us since it wants to assume they are varbinary

can you alter the script to fix this?

jeff-dude:
> it's still running, so not 100% sure it's resolved, but i pushed a commit to fix symbols which were missing single quotes. it looks like the trend is:
> 
> * symbol starts with `0x`, the quotes aren't used, dunesql engine yells at us since it wants to assume they are varbinary
> 
> can you alter the script to fix this?

looks like it worked! for reference, i just compiled the query, ran on dune app, then it gave me the exact line number of the issue and the symbol stuck out right away

jeff-dude (approved):

gentrexha:
> > it's still running, so not 100% sure it's resolved, but i pushed a commit to fix symbols which were missing single quotes. it looks like the trend is:
> > 
> > * symbol starts with `0x`, the quotes aren't used, dunesql engine yells at us since it wants to assume they are varbinary
> > 
> > can you alter the script to fix this?
> 
> looks like it worked! for reference, i just compiled the query, ran on dune app, then it gave me the exact line number of the issue and the symbol stuck out right away

Thank you for fixing this. We'll adjust the script to handle these type of things in the future.

PR #3705, opened 2023-07-10
Use partition_by in DuneSQL create table statements

jkylling:
So far we have been running without partition_by when creating tables using DuneSQL. This impacts the data layout and could influence data skipping, although most table are likely overpartitioned, and we should partition on block_month instead. Note that we need to cast block_date explicitly to date when the column is used for partitioning.

### DIFF
--- a/macros/dune/adapters.sql
+++ b/macros/dune/adapters.sql
@@ -1,5 +1,8 @@
 {% macro trino__create_table_as(temporary, relation, sql) -%}
-  {%- set _properties = config.get('properties') -%}
+  {%- set _properties = {} -%}
+  {%- if config.get('partition_by', None) != None -%}
+    {%- do _properties.update({'partitioned_by': "ARRAY['" + (config.get('partition_by') | join("', '") )  + "']"}) -%}
+  {%- endif -%}
   create or replace table {{ relation }}
     {{ create_table_properties(_properties, relation) }}
   as (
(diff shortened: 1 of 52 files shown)


jeff-dude:
> So far we have been running without partition_by when creating tables using DuneSQL. This impacts the data layout and could influence data skipping, although most table are likely overpartitioned, and we should partition on block_month instead. Note that we need to cast block_date explicitly to date when the column is used for partitioning.

this is a good catch, though i'm surprised `date_trunc()` function doesn't already cast to date?

also, if the majority of spells are better off at month partitions, i'd vote that we move on that now and stay consistent. we originally had done `block_date` because i believe decoded/base did that and we matched?

jeff-dude (approved):
approved with assumption i can merge with modified off for spark, on for trino

jeff-dude:
my takeaway on this:
- moving forward, we will update the partition column universally to be at the month level of granularity (will require edit to already migrated models and tweaks to ongoing migrations)
  - column name: `block_month` or just `month`?
- CI tests should fail if the config block doesn't contain list format
  - good: `partition_by = ['column_name']`
  - bad: `partition_by = 'column_name`
- `date_trunc()` isn't sufficient to cast as date, need to explicitly cast (we could explore new function to handle both at once)

fyi @couralex6 @0xRobin @Hosuke

Hosuke:
>   * column name: `block_month` or just `month`?

If we keep consistent with the past, we may use `block_month` to replace `block_date`.

0xRobin:
+1 for `block_month` 
@jkylling I don't understand why partitioning won't work on a `timestamp` column but would work on a `date` column. 
Can you elaborate?

PR #3706, opened 2023-07-10
updated destination chainlist adding official endpoint addresses and comments on some deprecated chains

0xChester:
Hello, I've updated the layerzero_chain_list.sql file with 5 new Endpoint Addresses per LayerZero Labs docs: https://layerzero.gitbook.io/docs/technical-reference/mainnet/supported-chain-ids#tenet and added comments on 2 deprecated chains per same documentation.

PR #3708, opened 2023-07-10
Migrate layerzero

jkylling:


jkylling:
Data matches on the models we are able to run. I'm looking into the memory issues.

One awkward issue is that the types are wrong in the current layerzero.send model, i.e., they are all varchar. We are missing view overrides for this model. Our options are basically:
1. Keep the invalid types and materialize the table with varchar.
2. Keep the invalid types in a view on top a table materialized with varbinary.
3. Do a breaking schema change. This can break some queries.

jeff-dude:
can you integrate these changes (#3706), so i can close that PR? then you'd have on legacy and new here?

PR #3713, opened 2023-07-11
[CoW] migrate and add new field to cow.order_rewards

bh2smith:
This PR migrates a basic spell to DuneSQL - while simultaneously adding a new field. This is already demonstrated to work as expected in the following PoC query.

Those with non-null new field: https://dune.com/queries/2714240

Regular PoC Query: https://dune.com/queries/1752782

cc @fhenneke, @harisang from the solver team and @gentrexha

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
@@ -1,4 +1,5 @@
 {{ config(alias='order_rewards',
+        tags = ['dunesql'],
         post_hook='{{ expose_spells(\'["ethereum"]\',
                                     "project",
                                     "cow_protocol",
@@ -6,11 +7,12 @@
 )}}
 
 -- PoC Query here - https://dune.com/queries/1752782
-select
-    distinct order_uid,
-    block_number,
-    tx_hash,
-    solver,
-    data.amount as cow_reward,
-    cast(data.surplus_fee as double) as surplus_fee
-from {{ source('cowswap', 'raw_order_rewards') }}
+SELECT
+  DISTINCT from_hex(order_uid) as order_uid,
+  block_number,
+  from_hex(tx_hash) as tx_hash,
+  from_hex(solver) as solver,
+  from_hex(data.quote_solver) as quote_solver,
+  cast(data.amount as uint256)  AS cow_reward,
+  cast(data.surplus_fee as uint256) AS surplus_fee
+FROM {{ source('cowswap', 'raw_order_rewards') }}
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
@@ -1,4 +1,4 @@
-{{ config(alias='order_rewards',
+{{ config(alias=alias('order_rewards'),
         tags = ['dunesql'],
         post_hook='{{ expose_spells(\'["ethereum"]\',
                                     "project",
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
@@ -228,7 +228,9 @@ models:
         to the solver competition on the granularity level of orders.
     columns:
       - *tx_hash
-      - *solver_address
+      - &solver
+        name: solver
+        description: "Address of the solver who settled the order (i.e. the from address of the corresponding transaction)"
       - *block_number
       - &cow_reward
         name: cow_reward
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
@@ -224,10 +224,8 @@ models:
     config:
       tags: ['ethereum','cow_protocol','app_data', "metadata"]
     description: >
-        Order Rewards contains the data necessary to compute Weekly Solver Payouts.
-        In particular, how much COW token solvers are entitled for each settled order.
-        Furthermore, since the introduction of limit orders into the protocol, 
-        this table also contains the surplus fee.
+        Order Rewards (or more appropriately named Order Meta) contains off-chain meta data relevant 
+        to the solver competition on the granularity level of orders.
     columns:
       - *tx_hash
       - *solver_address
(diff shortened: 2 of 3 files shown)


models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
```diff
@@ -1,16 +1,18 @@
...
-    data.amount as cow_reward,
-    cast(data.surplus_fee as double) as surplus_fee
-from {{ source('cowswap', 'raw_order_rewards') }}
+SELECT
+  DISTINCT from_hex(order_uid) as order_uid,
+  block_number,
+  from_hex(tx_hash) as tx_hash,
+  from_hex(solver) as solver,
+  from_hex(data.quote_solver) as quote_solver,
+  cast(data.amount as uint256)  AS cow_reward,
```
harisang, line 16: Do we have the cow_reward and surplus_fee for backwards compatibility reasons here?

bh2smith: `surplus_fee` still exists as far as I understand and yea I guess you cold say `cow_reward` is for backwards compatibility... Also note that the back end is still populating this field... with zeros I think.

bh2smith:
> Can we actually rename the file, as order_rewards might be a bit misleading?

We could, but this is a much more involved change that requires coordination with the Dune team and also tracking down all the queries that use it. Would much rather prefer not even attempting to rename it, but if you insist we can make it happen.

harisang:
> > Can we actually rename the file, as order_rewards might be a bit misleading?
> 
> We could, but this is a much more involved change that requires coordination with the Dune team and also tracking down all the queries that use it. Would much rather prefer not even attempting to rename it, but if you insist we can make it happen.

Ok, no worries then. No problem keeping it as is

harisang (approved):

jeff-dude:
> > > Can we actually rename the file, as order_rewards might be a bit misleading?
> > 
> > 
> > We could, but this is a much more involved change that requires coordination with the Dune team and also tracking down all the queries that use it. Would much rather prefer not even attempting to rename it, but if you insist we can make it happen.
> 
> Ok, no worries then. No problem keeping it as is

yeah, just to clarify here, you could run:
`dbt ls --resource-type model --output path --select +cow_protocol_ethereum_order_rewards+`
to see dbt lineage, and notice there is nothing upstream or downstream of this spell.

changing the file name would be fine from a dbt perspective. however, if you also changed the schema or table name, you would then risk queries on the dune app breaking, even though dbt would be fine.

jeff-dude (changes requested):
for migrating an existing spell, such as here, please follow the steps here:
https://github.com/duneanalytics/spellbook#existing-spells

your PR should end up with two files, one spark legacy and one dunesql. the biggest things to look for are:
- adding tags for dunesql (which is done here in the PR)
- updating how we call the alias property in both legacy/dunesql spell
- ensuring dunesql spell syntax is good (which is set here from your query proof)

the main reason to keep the legacy files is to ensure we keep running spark spells until the engine is fully deprecated.

bh2smith:
Thanks @jeff-dude -- I did the legacy thing.

gentrexha (approved):
Left a schema comment. Other than that, LGTM.

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
```diff
@@ -224,10 +224,8 @@ models:
...
     description: >
-        Order Rewards contains the data necessary to compute Weekly Solver Payouts.
-        In particular, how much COW token solvers are entitled for each settled order.
-        Furthermore, since the introduction of limit orders into the protocol, 
-        this table also contains the surplus fee.
+        Order Rewards (or more appropriately named Order Meta) contains off-chain meta data relevant 
+        to the solver competition on the granularity level of orders.
     columns:
       - *tx_hash
       - *solver_address
```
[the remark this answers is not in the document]
bh2smith: Fixed.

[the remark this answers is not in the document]
gentrexha: Thanks!

models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
```diff
@@ -1,16 +1,18 @@
 {{ config(alias='order_rewards',
```
jeff-dude, line 1: can you make the alias property:
`alias=alias('order_rewards'),`

notes from readme:
```
update alias property again, this time to simplify and remove legacy flag
alias = alias('blocks'),
```

this leverages a new alias macro that helps us differentiate table naming in our metastores

jeff-dude: this will occur in all new models for dunesql, so the three total steps:
- tag for dunesql
- alias macro in config block
- translated syntax

jeff-dude (changes requested):

models/cow_protocol/ethereum/cow_protocol_ethereum_order_rewards.sql
bh2smith: So, always use alias macro. No problemo - just need to go change all my PRs.

bh2smith: Done here.

bh2smith: Update -- also now done in all open PRs.

jeff-dude: we assign a default value for the legacy flag in the macro to `False`, so that's why we don't use on these translated ones, and only override in legacy files

bh2smith: Sounds good to me... as long as everything is correct now.

jeff-dude (approved):
yep -- this looks good! nice work 🤝 

this will repeat across any migration. i'll look at other PRs now too, but the other main thing to look out for is downstream impacts, using the example from earlier comment:
`dbt ls --resource-type model --output path --select cow_protocol_ethereum_order_rewards+`

if this returns other models downstream, we'd need to edit those to change:
`ref(model_name)` to `ref(model_name_legacy)`

i can help look out for that in other PRs

bh2smith:
Thanks @jeff-dude -- I think I've got the hang of it. Once we land the open ones and Trades I think the last couple will be easy-peasy.

PR #3714, opened 2023-07-11
[CoW] Migrate and add utm to cow.app_data

bh2smith:
This spell migrates the query to dunesql and adds a couple recent columns to the table: 

Check out the PoC Query here: https://dune.com/queries/1751965

Note that this cow.referrals depends on this, so we had to migrate it first. Funny thing is that it did not require any changes to migrate so we only added the tag here. However we will have to migrate `trade_slippage` as well/separately.

cc @gentrexha

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_app_data.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_app_data.sql
@@ -1,4 +1,4 @@
-{{ config(alias='app_data',
+{{ config(alias=alias('app_data'),
         tags=['dunesql'],
         post_hook='{{ expose_spells(\'["ethereum"]\',
                                     "project",
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_app_data.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_app_data.sql
@@ -8,39 +8,37 @@
 
 -- Find the PoC Query here: https://dune.com/queries/1751965
 with
-  partially_unpacked_app_content AS (
-    SELECT
-      app_hash,
-      content.appCode AS app_code,
-      content.environment,
-      content.metadata.orderClass.orderClass AS order_class,
-      content.metadata.quote,
-      content.metadata.referrer,
-      content.metadata.utm
-    FROM {{ source('cowswap', 'raw_app_data') }}
+  partially_unpacked_app_content as (
+    select
+        distinct app_hash,
+        content.appCode AS app_code,
+        content.environment,
+        content.metadata.orderClass.orderClass as order_class,
+        content.metadata.quote,
+        content.metadata.referrer,
+        content.metadata.utm
+    from {{ source('cowswap', 'raw_app_data') }}
   ),
-  unpacked_referrer_app_data AS (
-    SELECT
-      app_hash,
-      app_code,
-      environment,
-      order_class,
-      quote,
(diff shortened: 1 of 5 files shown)


models/cow_protocol/ethereum/cow_protocol_ethereum_app_data.sql
```diff
@@ -7,41 +8,37 @@
...
+  unpacked_referrer_app_data as (
     select
         app_hash,
         app_code,
         environment,
         order_class,
         quote,
-        -- different app data versions put referrer in two possible places.
-        lower(coalesce(referrer.address, referrer.referrer)) as referrer
+        coalesce(referrer.address, referrer.referrer) as referrer,
```
gentrexha, line 29: Missing `lower` function here compared to previous version and POC query.

models/cow_protocol/ethereum/cow_protocol_ethereum_app_data_legacy.sql
gentrexha, line 1: Can you please remind me why we need to keep these legacy spells? It seems identical to the new one?

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
gentrexha, line 1: Does the name ` name: cow_protocol_ethereum_app_data` need to be changed to legacy here as well?

gentrexha (approved):
LGTM.

Left some discussion comments and questions.

models/cow_protocol/ethereum/cow_protocol_ethereum_schema.yml
bh2smith: not according to the readme - https://github.com/duneanalytics/spellbook#existing-spells

CI is passing - the rest is up to Dune.

bh2smith: Also, it makes it possible to migrate things downstream (or upstream IDFK) of dependencies before tables which depend on them (independently) by "lazily" adding a `_legacy` to them.

models/cow_protocol/ethereum/cow_protocol_ethereum_app_data_legacy.sql
jeff-dude: it helps us keep orchestration going in prod for both spark and dunesql

bh2smith:
Results of

```sh
(spellbook) ➜  spellbook git:(cow/app_data) dbt ls --resource-type model --output path --select cow_protocol_ethereum_app_data+     
models/cow_protocol/ethereum/cow_protocol_ethereum_app_data.sql
```

models/cow_protocol/ethereum/cow_protocol_ethereum_app_data.sql
jkylling: This has type varchar: https://dune.com/queries/2720831 Could it be converted to varbinary using `from_hex` on line 18? This might also apply to other fields, like app_hash?

jkylling (commented):
@jeff-dude could you check that the data in the migrated spell matches the data in the legacy spell?

models/cow_protocol/ethereum/cow_protocol_ethereum_app_data.sql
bh2smith: Fix this in both cases.

jeff-dude (approved):
thank you 🙌

PR #3694, opened 2023-07-07
Migrate NFT beta base_trades

0xRobin:
Following this PR: https://github.com/duneanalytics/spellbook/pull/3664

This migrates all other `base_trades` NFT platform models. 

known issues to resolve:

- [x] ~Blur model: no royalty fee~
- [x] ~Sudoswap model: no results~

PR #3733, opened 2023-07-13
Blur v2 Base Trades

hildobby:


0xRobin:
@hildobby you can do that and then use `>= TIMESTAMP '{{blur_v2_start_date}}'`

0xRobin:
To follow the new NFT (beta) structure please:
1. move the model into `models/_sector/nft/trades/ethereum/platforms`
2. add the schema in `models/_sector/nft/trades/ethereum/platforms/_schema.yml`
3. add the sources to `models/_sector/nft/trades/ethereum/platforms/_sources.yml`
4. add the model to the list in `models/_sector/nft/trades/nft_ethereum_trades_beta.sql` to have it included in the downstream nft sector spells. (now possible since https://github.com/duneanalytics/spellbook/pull/3694 is merged)
5. add a new seed file (example `seeds/_sector/nft/blur_ethereum_base_trades_seed.csv`) following the same standard and the `check_seed` test in the model schema.

jensenity (approved):
Nice!! Been waiting for this for my dashboard ❤️

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+    WHERE evt_block_time >= {{blur_v2_start_date}}
+    {% endif %}
+    )
+
+SELECT CAST(date_trunc('day', bt.block_time) AS date) AS block_date
+, bt.block_time
+, bt.block_number
+, bt.tx_hash
+, bt.evt_index AS sub_tx_trade_id
+, CASE WHEN txs."from" = bt.trader THEN 'Sell' ELSE 'Buy' END AS trade_category
```
0xRobin, line 89: could we use the "Side" from `collectionPriceSide` here to determine the correct trade_category?

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+, 'secondary' AS trade_type
+, CASE WHEN txs."from" = bt.trader THEN bt.trader ELSE txs."from" END AS buyer
```
0xRobin, line 91: FYI, this always evals to `txs."from"`
See comments on the seed examples.

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+, CASE WHEN txs."from" = bt.trader THEN txs."from" ELSE bt.trader END AS seller
```
0xRobin, line 92: FYI, this always evals to `bt.trades`
See comments on the seed examples.

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+, CAST(1 AS UINT256) AS nft_amount
+, CAST(bt.price*POWER(10, 18) AS UINT256) AS price_raw
+, CASE WHEN bt.order_type = 0 THEN 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 ELSE 0x0000000000a39bb272e79075ade125fd351887ac END AS currency_contract
+, bt.project_contract_address
+, CAST(0 AS uint256) AS platform_fee_amount_raw
+, CAST(NULL AS varbinary) AS platform_fee_address
+, CAST(ROUND(bt.price * POWER(10, 18) * bt.fee) AS UINT256) AS royalty_fee_amount_raw
+, bt.royalty_fee_address
+FROM blur_v2_trades bt
+INNER JOIN {{ source('ethereum', 'transactions') }} txs ON txs.block_number=bt.block_number
```
0xRobin, line 104: If possible, we should try and remove the join with transactions to make the query more optimized. 
Currently this is only used to determine:
- trade_category
- buyer
- seller

Is it possible to source these columns from the event data?

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+    incremental_strategy = 'merge',
+    unique_key = ['block_number','tx_hash','sub_tx_trade_id'],
+    )
+}}
+
+{% set blur_v2_start_date = "cast('2023-07-05' as timestamp)" %}
+
+WITH blur_v2_trades AS (
+    SELECT evt_tx_hash AS tx_hash
+    , CAST(ROUND((bitwise_right_shift(collectionPriceSide, 160) - (bitwise_right_shift(collectionPriceSide, 248) * CAST(power(2, 88) AS UINT256)))/POWER(10, 18), 8) AS double) AS price
```
0xRobin, line 17:
```suggestion
    , bytearray_to_uint256(bytearray_substring(cast(collectionPriceSide as varbinary),2,11)) AS price
```
(note this change will be full decimals, so you'd also need to remove the `*POWER(10, 18)` in the final select)

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+    , evt_block_time AS block_time
+    , evt_block_number AS block_number
+    , NULL AS fee_side
+    , evt_index
+    , contract_address AS project_contract_address
+    , CAST(bitwise_right_shift(collectionPriceSide, 248) AS BIGINT) AS order_type
```
0xRobin, line 23:
```suggestion
    , bytearray_to_bigint(bytearray_substring(cast(collectionPriceSide as varbinary),1,1))
```

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+    , from_hex('0x' || LOWER("RIGHT"(CAST(to_hex(CAST(collectionPriceSide AS varbinary)) AS varchar), 40))) AS nft_contract_address
```
0xRobin, line 24:
```suggestion
    , bytearray_substring(cast(collectionPriceSide as varbinary),13,20) AS nft_contract_address
```

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+    , orderHash AS order_hash
+    , bitwise_right_shift(tokenIdListingIndexTrader, 168) AS nft_token_id
```
0xRobin, line 26:
```suggestion
    , bytearray_to_uint256(bytearray_substring(cast(tokenIdListingIndexTrader as varbinary),1,11))
```

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+    , from_hex('0x' || LOWER("RIGHT"(CAST(to_hex(CAST(tokenIdListingIndexTrader AS varbinary)) AS varchar), 40))) AS trader
```
0xRobin, line 27:
```suggestion
    , bytearray_substring(cast(tokenIdListingIndexTrader as varbinary),13,20) AS trader
```

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+    alias = alias('base_trades'),
+    partition_by = ['block_date'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['block_number','tx_hash','sub_tx_trade_id'],
+    )
+}}
+
+{% set blur_v2_start_date = "cast('2023-07-05' as timestamp)" %}
```
0xRobin, line 13: as a general standard we prefer:
`{% set blur_v2_start_date = '2023-07-05' %}`
with
`WHERE evt_block_time >= TIMESTAMP {{blur_v2_start_date}}`

Nothing wrong with your approach, just trying to use some standards across the repo.

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
0xRobin, line 1: Hey @hildobby,
I've proposed some syntax changes that make better use of the existing varbinary functions to retrieve the different elements out of the event fields. 
Here's a dune query that has both your original syntax and my proposed "fixes":
https://dune.com/queries/2739951

(the comments should also apply to the other selects in the union)

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
0xRobin: I still have to go through a lot of the marketplaces to change this (as they were written before we decided on a standard), so we can leave this as is for now.

jensenity:
So I checked the reservoir's indexer code on their way to interpret BlurV2 trades, it's exactly how @hildobby built it...I think with what @0xRobin suggested....if a transaction table slows down the query...I suggest a dimension table in the middle for just tx_hash with from address(seller) and to address(buyer)...I don't know whether it's a good idea, but just trying to suggest something for this to move forward.


### Reservoir interpretation
https://github.com/reservoirprotocol/indexer/blob/4624a6da2b0a97d4ab0b3c9a6989fbc6ec3aad87/packages/indexer/src/sync/events/data/blur-v2.ts#L8
https://github.com/reservoirprotocol/indexer/blob/4624a6da2b0a97d4ab0b3c9a6989fbc6ec3aad87/packages/indexer/src/sync/events/handlers/blur-v2.ts

jensenity:
> I believe Hildobby will continue work on this this week.

Thanks @0xRobin I will try to work on new stuffs on dune then^^ Thanks for the quick reply

hildobby:
> ts, I don't think an intermediate table is needed.
> If we can do without the join on transactions, perfect. If we need it to source the correct fields, not a big problem.
> I believe Hildobby will continue work on this this week.



> So I checked the reservoir's indexer code on their way to interpret BlurV2 trades, it's exactly how @hildobby built it...I think with what @0xRobin suggested.....it could smoother out the queries....if a transaction table slows down the query...I suggest a dimension table in the middle for just tx_hash with from address(seller) and to address(buyer)...I don't know whether it's a good idea, but just trying to suggest something for this to move forward. I kinda want to work on this branch...to help out
> 
> ### Reservoir interpretation
> https://github.com/reservoirprotocol/indexer/blob/4624a6da2b0a97d4ab0b3c9a6989fbc6ec3aad87/packages/indexer/src/sync/events/data/blur-v2.ts#L8 https://github.com/reservoirprotocol/indexer/blob/4624a6da2b0a97d4ab0b3c9a6989fbc6ec3aad87/packages/indexer/src/sync/events/handlers/blur-v2.ts

Hey @jensenity! Fyi, Reservoir uses the API provided by marketplaces to fetch trade data while I use onchain data. Onchain data is a looot harder to fake than any API data so I trust the former infinitely more than the latter. The complexity here was around how the data was packed in emitted events to save gas, which I did not see anyone tackle before I did. 

Just wanted to inform about the differences between the two datasets, reservoir is good to compare against and 99.99% likely to be accurate but shouldn't be treated as the source of truth. However that also means that reservoir has trades for new marketplace versions straight away while those using onchain data (ie nft.trades) need to manually account for those.

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
hildobby: I used that method because that was the only one that seemingly passed tests, I switched back to your method and now tests are failing again, any clues why?

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
hildobby: fixed it

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
hildobby: fixed it

models/_sector/nft/trades/ethereum/platforms/blur_v2_ethereum_base_trades.sql
hildobby: done

0xRobin (approved):
:heavy_check_mark: good stuff!

0xRobin:
@jeff-dude I believe this is an example that would need the dummy spark model for the new schema?

jeff-dude:
> @jeff-dude I believe this is an example that would need the dummy spark model for the new schema?

i think you're right -- i merged in main, which has a new test in CI to check this. i think we should expect it to fail.

PR #3735, opened 2023-07-13
All non-translated files moved to _legacy.sql

couralex6:


jeff-dude:
need to ensure we update the `readme` either with this or after, to ensure we aren't asking users to create `_legacy` files anymore as it's already done

jeff-dude (approved):

PR #3719, opened 2023-07-11
[CoW] Migrate Trades Table

bh2smith:
Migration of CoW Protocol Trades table. This was a bit of a big one, but somehow trades is a relatively low level table with very few dependencies. This should pave the way to migrate nearly all the rest of the cow spells.

cc @gentrexha

PR #3375, opened 2023-05-19
[Bug] Fix Trade Surplus for Partial Fills

olgafetisova:
Brief comments on the purpose of your changes:

Fixing a bug for surplus calculation.
The fix adjusts the surplus that has been previously computed wrongly.

The current calculation already includes the filled proportion amount and doesn't need to be multiplied again by the proportion fill.

Brief comments on the purpose of your changes:
Example:
This [transaction](https://explorer.cow.fi/orders/0x1bcbe704226db32ab68b39d13b8650a1cb3ec4518d9516b92b8e4d1649b037477f2a72b087939917caa1011de5a18302b3f21d0a646a7739/?tab=fills) has approximately 109 ICSA in surplus, which is ~$54, our previous calculation was giving only a percentage of that value due to the fact that the order has been filled in 2 transactions.

**For Dune Engine V2**

I've checked that:

### General checks:
* [X] I tested the query on dune.com after compiling the model with dbt compile (compiled queries are written to the target directory)
* [ ] I used "refs" to reference other models in this repo and "sources" to reference raw or decoded tables 
* [ ] if adding a new model, I added a test
* [ ] the filename is unique and ends with .sql
* [ ] each sql file is a select statement and has only one view, table or function defined  
* [ ] column names are `lowercase_snake_cased`
* [ ] if adding a new model, I edited the dbt project YAML file with new directory path for both models and seeds (if applicable)
* [ ] if wanting to expose a model in the UI (Dune data explorer), I added a post-hook in the JINJA config to add metadata (blockchains, sector/project, name and contributor Dune usernames)

### Pricing checks:
* [ ] `coin_id` represents the ID of the coin on coinpaprika.com
* [ ] all the coins are active on coinpaprika.com (please remove inactive ones)

### Join logic:
* [ ] if joining to base table (i.e. ethereum transactions or traces), I looked to make it an inner join if possible

### Incremental logic:
* [ ] I used is_incremental & not is_incremental jinja block filters on both base tables and decoded tables
  * [ ] where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to base table (i.e. ethereum transactions or traces), I applied join condition where block_time >= date_trunc("day", now() - interval '1 week')
* [ ] if joining to prices view, I applied join condition where minute >= date_trunc("day", now() - interval '1 week')

PR #3745, opened 2023-07-14
[CoW] Hotfix - Reintroduce block_number into trades

bh2smith:
Not sure what happened here, but I somehow left out the `block_number` field from the trades table (here #3719). Note easy to tell from the diff, but now its all fixed. Also managed to revert a fix previously made to the `surplus_usd` field https://github.com/duneanalytics/spellbook/pull/3375.

cc @gentrexha

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
@@ -21,6 +21,7 @@ WITH
 trades_with_prices AS (
     SELECT cast(date_trunc('day', evt_block_time) as date) as block_date,
            evt_block_time            as block_time,
+           evt_block_number          as block_number,
            evt_tx_hash               as tx_hash,
            evt_index,
            trade.contract_address    as project_contract_address,
@@ -61,6 +62,7 @@ trades_with_prices AS (
 trades_with_token_units as (
     SELECT block_date,
            block_time,
+           block_number,
            tx_hash,
            evt_index,
            project_contract_address,
@@ -173,6 +175,7 @@ eth_flow_senders as (
 valued_trades as (
     SELECT block_date,
            block_time,
+           block_number,
            tx_hash,
            evt_index,
            CAST(NULL as array<bigint>) as trace_address,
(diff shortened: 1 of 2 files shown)


models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
```diff
@@ -245,5 +248,5 @@ valued_trades as (
 select *,
   -- Relative surplus (in %) is the difference between limit price and executed price as a ratio of the limit price.
   -- Absolute surplus (in USD) is relative surplus multiplied with the value of the trade
-  fill_proportion * usd_value * (atoms_bought * limit_sell_amount - atoms_sold * limit_buy_amount) / (atoms_bought * limit_sell_amount) as surplus_usd
+  usd_value * (atoms_bought * limit_sell_amount - atoms_sold * limit_buy_amount) / (atoms_bought * limit_sell_amount) as surplus_usd
```
bh2smith, line 251: cc @harisang -- this here

gentrexha: Why are we reverting this?

gentrexha (approved):
Left a discussion comment. LGTM.

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
bh2smith: Did you read the PR description?

gentrexha: Cool. Thanks!

jeff-dude (approved):
thank you, we actually got a few failures on our internal tests around this (missing block number)

models/cow_protocol/ethereum/cow_protocol_ethereum_trades.sql
harisang: Out of curiosity, `usd_value` is equal to the value of the trade? So, for partial fills, this still works because it again describes the value of what was traded?

If so, how is this measured? Is the price of the sell token used, the price of the buy token, an average of the two maybe?

bh2smith: We use the max of the (sell_value, buy_value)

PR #3752, opened 2023-07-14
Update ci to reflect legacy tag changes

couralex6:
With the addition of the `legacy` tag for all _legacy.sql models, we need to change the CI to select `tag:legacy` instead of excluding `tag:dunesql`.


jeff-dude (approved):

jeff-dude:
@0xRobin @Hosuke fyi

0xRobin:
Awesome!! 
we can add a small test to check there are no models without a tag, and no models with both tags.

PR #3762, opened 2023-07-15
Create OP Worldcoin Safes

MSilb7:
Spell to identify Worldcoin account Safes & add them to labels.

This can be used for attribution later on (i.e. these could count as both Worldcoin and Safe, whereas contract creator logic would only count these as Safe).


MSilb7:
Marking as draft to update with better logic from Worldcoin team.

models/labels/addresses/__single_category_labels__/worldcoin/labels_worldcoin_accounts.sql
```diff
@@ -0,0 +1,32 @@
...
+{% for model in models %}
+
+SELECT
+    '{{ model[0] }}' as blockchain,
+    account_address as address,
+    'Worldcoin Account' AS name,
+    'worldcoin' AS category,
+    'msilb7' AS contributor,
+    'query' AS source,
+    from_iso8601_timestamp('2023-07-15') as created_at,
```
Hosuke, line 21: `timestamp` function is supported now.
```suggestion
    timestamp '2023-07-15' as created_at,
```

jeff-dude (approved):

PR #3763, opened 2023-07-16
uniswap_pools add optimism and bnb

mtitus6:


models/uniswap/optimism/uniswap_optimism_ovm1_pool_mapping.sql
mtitus6, line 1: These pools were created at GENESIS.  We need the creation date, creation block, and contract address to union with other models.  Please review what I have used and suggest any changes.

models/uniswap/optimism/uniswap_optimism_pools.sql
mtitus6, line 1: I was not sure the best method to implement an incremental load here due to the ovm1 static model.  Should I just add to each model separately where I union them?

mtitus6:
@Hosuke I think this is Ready for Review.  I added comments on a couple files where I would appreciate some feedback.   Thanks for your help 🙏

models/uniswap/bnb/uniswap_bnb_pools.sql
```diff
@@ -0,0 +1,29 @@
...
+    schema = 'uniswap_bnb',
+    alias = 'pools',
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['pool'],
+    post_hook='{{ expose_spells(\'["bnb"]\',
+                                "project",
+                                "uniswap",
+                                \'["hildobby"]\') }}'
```
Hosuke, line 12:
```suggestion
                                \'["mtitus6"]\') }}'
```

models/uniswap/optimism/uniswap_optimism_pools.sql
Hosuke: For static views we do not need to make incremental filters for them.
However we may add a tag to indicate them.
i.e.
```
tags = ['static', 'dunesql'],
```

mtitus6: OK. Will it cause a problem if apply an incremental load to Factory_evt_PoolCreated and not to uniswap_optimism_ovm1_pool_mapping when I am combining them together?  
I assumed that would cause duplicate data from static table to be added.

Hosuke (approved):
LGTM.✅ for the dune sql updates.
Thank you @mtitus6

mtitus6:

mtitus6:
recheck

models/uniswap/bnb/uniswap_bnb_pools.sql
```diff
@@ -0,0 +1,29 @@
+{{ config(
+    tags=['dunesql'],
+    schema = 'uniswap_bnb',
+    alias = 'pools',
```
jeff-dude, line 4: we need to ensure to use new alias macro on all config blocks. please use:
`alias('pools')`, similar to arbitrum version

models/uniswap/bnb/uniswap_bnb_pools.sql
```diff
@@ -0,0 +1,29 @@
+{{ config(
+    tags=['dunesql'],
+    schema = 'uniswap_bnb',
```
jeff-dude, line 3: fyi @Hosuke -- in case i haven't explicitly called this out. when creating a new spell on dunesql, we need to ensure the schema used is already created on spark too -- even if it's a different spell, just need the schema to exist.

i'm viewing on databricks directly, so you'd have to double check on the dune app instead likely, but here is proof for this schema:
image

this CI step `check for missing legacy models` also tries to help us catch it: 
https://github.com/duneanalytics/spellbook/actions/runs/5800389627/job/15722400771?pr=3763

jeff-dude: long way of me saying: we should be good here :)

jeff-dude (changes requested):
one minor change, otherwise just a note for reviews as well

models/uniswap/bnb/uniswap_bnb_pools.sql
Hosuke: So the schema need to be existed in spark legacy sql first, then we can use it in the dune sql?

jeff-dude: not necessarily first, just in addition to dunesql. it has to do with how `dbt run` works behind-the-scenes, where it'll first run queries on the information schema to search for all schemas in a dbt project. since we share engines in the project, it runs on both for all models, legacy or not. we haven't found a quick, simple way to pass in the `tag:dunesql` type filter to that process like we do elsewhere. this is just our quick workaround fix.

jeff-dude (approved):

PR #3768, opened 2023-07-17
add small CI checks

0xRobin:
this PR adds checks in the CI to catch:
1. models with no custom schema
2. models with no `dunesql` or `legacy` tag
3. models with both `dunesql` or `legacy` tag


jeff-dude (approved):
i like the intention here 👍 

i'll let sit in approved and give @couralex6 a chance to look

couralex6 (approved):

PR #3783, opened 2023-07-18
Create Astaria lending spell - ready for review

henrystats:


henrystats:
apologies for the tag again here ser @jeff-dude, this pr was working fine a few days ago when I put it up but now it's failing as I'm getting an access deny error to the tokens_nft table (originally used the tokens_ethereum_nft) then I switched to tokens_nft incase that would magically work but still the same error.

jeff-dude:
> apologies for the tag again here ser @jeff-dude, this pr was working fine a few days ago when I put it up but now it's failing as I'm getting an access deny error to the tokens_nft table (originally used the tokens_ethereum_nft) then I switched to tokens_nft incase that would magically work but still the same error.

interesting. i'm going to tag @0xRobin since i know he is calling `tokens_nft` as a ref in nft migration PRs, and not hitting this access issue. typically, if the spell isn't modified in the PR, then we read it from prod. i'm not sure why one PR could read it, while another can't? or maybe i'm missing something at first glance.

jeff-dude:
> hola ser @jeff-dude 👋 - still facing issues here with the incremental logic I put in this code :(... the initial models run correctly and the results & tests pass..
> 
> but not sure why the incremental run is failing..

its a bit tough to say just eyeballing it. can you try to see if you can run `dbt compile` locally and pass an extra flag to force it to output incremental logic? that way you can copy/paste into dune app and find error?

maybe `{{ this }}` is case sensitive and needs to be lower?

otherwise, error looks like an extra `)` somewhere

henrystats:
thank you ser, looks like it was case sensitive... works now although the the other incremental model is failing because of a missing column which is weird since the initial model is running fine... will continue to check.

henrystats:
really have no idea why this is failing... the underlying table returns has the columns that say it can't find.

for some reason it's combining evt_index and lien_id as one column... not sure how and why that is happening.. 

the initial table runs correctly with both evt_index & lien_id...

henrystats:
@jeff-dude, there was a missing "," in the unique keys hence why the incremental run was failing.. 

I've fixed that now and this is now ready for review!

henrystats:
@jeff-dude saw that this was now in review... since we're now partitioning by block_month, should I submit a quick pr to partition this by block_month?

henrystats:
fixed the partitioning!

jeff-dude (approved):
looks good!

and yes, thank you for being proactive on the new partition 🤝

PR #3791, opened 2023-07-18
oneinch: init fantom calls transfers & migrate upstream spells

grkhr:
Hi folks! 

We are starting a process of spellbook'ing our 1inch dashboards to unify queries and optimize workloads (which will benefit both sides).

This view is a proof-of-concept step.


models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
```diff
@@ -0,0 +1,126 @@
+{{ config( 
+    alias = alias('fantom_calls_transfers'),
```
jeff-dude, line 2: please supply a `schema` property in addition to alias
`schema = '1inch_fantom',`
then alias can simplify:
`alias = alias('calls_transfers'),`

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
```diff
@@ -0,0 +1,126 @@
...
+    partition_by = ['block_date'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['tx_hash', 'start', 'transfer_trace_address']
+    )
+}}
+
+{% set project_start_date = "timestamp '2020-11-13'" %} 
+{% set lookback_days = -2 %}
```
jeff-dude, line 13: our spellbook-wide standard is to look back 7 days. it's mostly as a safeguard for late arriving blocks and fixing bugs. you can open it up a few more days to the standard

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
```diff
@@ -0,0 +1,126 @@
...
+            , min_by(success, trace_address) as call_success
+            , min_by({{selector}}, trace_address) as call_selector
+        from {{ source('fantom', 'traces') }}
+        where 
+            {% if is_incremental() %}
+                block_time >= date_add('day', {{lookback_days}}, now())
+            {% else %}
+                block_time >= {{ project_start_date }}
+            {% endif %}
+            and "to" in (
```
jeff-dude, line 56: you may get better performance switching this to an inner join instead, just a thought

models/oneinch/fantom/oneinch_fantom_schema.yml
```diff
@@ -0,0 +1,47 @@
+version: 2
+
+models:
+  - name: oneinch_fantom_calls_transfers
```
jeff-dude, line 4: you can add tests in this schema file on the unique keys, to ensure no duplicates. there are lots of examples in other schema files in the repo. it's the "unique combo of columns" test, where you supply the unique keys specified in the model

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
```diff
@@ -0,0 +1,126 @@
...
+                select contract_address from {{ ref('oneinch_contract_addresses') }}
```
jeff-dude, line 57: i believe this is where you're failing the CI test. since this spell is dependent on upstream spell(s), you need to also migrate those. we should focus on most upstream first, then downstream. if this ref is a simple spell, feel free to add the migrated version to this PR.

please be sure to modify the non-legacy file -- we generated legacy versions to maintain spark logic. there should be this file without legacy suffix.

jeff-dude (changes requested):
thanks for the clean code here 🤝 

left a few comments below on how to proceed

this is looking awfully close to how some of the existing 1inch spells are built, i'm excited to see how we can progress this lineage and fix the existing 1inch spells!

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
grkhr: could it be just in current `oneinch` schema? (actually i've tried to pass schema and removed, thought that was causing the problem lol)

it'd be like
`oneinch.fantom_calls_transfers`
`oneinch.ethereum_calls_transfer`
`...`

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
grkhr: fair enough, will change to 7

models/oneinch/fantom/oneinch_fantom_schema.yml
grkhr: i was thinking to add uniq columns test, but seems that it might be a bit heavy for compute. i'll add after your approve

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
jeff-dude: we have a pattern spellbook-wide to write schemas in this format:
`<project/sector_name>_<version>_<blockchain>.<table_alias>`
where:
- version is only if applicable
- blockchain is included when we start to build cross-chain spells, so we can keep a clean lineage per blockchain and then into cross-chain

after you build all the blockchain-specific schemas and spells, the cross-chain will just be `1inch`

it also follows our directory path naming standard

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
grkhr: if by "migration" of that upstream spell you mean adding tag and alias — it's done in the earlier PR waiting for merge https://github.com/duneanalytics/spellbook/pull/3784/files

(also, should we remove that workaround with `'0x...'` quotes and pass varbinary normally, without quotes? if yes i'd just do all of these things in this PR and ask to cancel 3784)

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
grkhr: got you, thanks for the clear explanation!

models/oneinch/fantom/oneinch_fantom_schema.yml
jeff-dude: it's good to see that test pass in the CI phase, since we don't run the test in prod. it'll help ensure no duplicates in historical run, otherwise incremental will only read 7 days and anyhting prior to that may have already written dupes. you can add at the end of your dev process here, if you want the CI to remain quick(er)

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
jeff-dude: yes exactly, that spell.

i would recommended pulling those changes into this PR, since it's small. then we can handle the few dependencies together quickly.

you can simply remove the single quotes on all the addresses, and dunesql engine will write as varbinary. also ensure you remove any functions on the address column like lower/cast/whatever else. just leave the hardcoded address as-is without quotes

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
grkhr: As far as I know from the experience with other columnar databases, inner-join is more effective than semi-join in cases when there are lot of rows in the second table. In this particular case there are just 30 addresses, so it shouldn't make a difference

btw tested both cases, ran at the same time

large cluster
image

medium cluster
image

I think it won't make a difference, wdyt?

models/oneinch/fantom/oneinch_fantom_schema.yml
grkhr: got you, thanks again for the explanation! adding the test 🤝

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
jeff-dude: thanks for the evidence -- i'm good with as-is

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
grkhr: thanks! done

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
```diff
@@ -0,0 +1,127 @@
...
+            and call_type = 'call'
+        group by tx_hash
+    ) gr_traces using(tx_hash)
+)
+
+, merged as (
+    select 
+        calls.tx_hash
+        , transfers.block_time
+        , date_trunc('day', transfers.block_time) as block_date
```
jeff-dude, line 70: ah, this is another oddity we found in dunesql. you need to explicitly cast this field to date in order to use in the `parition_by` column above in config. without the cast, it just takes the `block_time` timestamp data type and we can't partition on timestamp

jeff-dude: also, we're looking to make our partitions larger universally post-migration. maybe you can get a head start and truncate this to month, then rename to `block_month`, and partition by that? unless you use the field for something else too?

grkhr: we don't need that field tbh, so i'll just change to month and partitioning as well! also was curious why are you using daily partitions instead of monthly hah

grkhr:
hey @jeff-dude 

seems good to go for me, but i'm again geting a CI error
```
Failure in test dbt_utils_unique_combination_of_columns_oneinch_fantom_calls_transfers_tx_hash__start__transfer_trace_address (models/oneinch/fantom/oneinch_fantom_schema.yml)
22:24:37    Got 2133 results, configured to fail if != 0
```

on production everything is great
image

models/oneinch/fantom/oneinch_fantom_calls_transfers.sql
```diff
@@ -0,0 +1,127 @@
+{{ config( 
+    schema = 'oneinch_fantom',
+    alias = alias('calls_transfers'),
+    tags = ['dunesql'],
+    partition_by = ['_partition'],
```
jeff-dude, line 5: we're looking to keep a universal name for this column, so if you don't mind going back to `block_month`

jeff-dude: while you may not use the column, some spells might and it'll be better to query on the app

jeff-dude:
> hey @jeff-dude
> 
> seems good to go for me, but i'm again geting a CI error
> 
> ```
> Failure in test dbt_utils_unique_combination_of_columns_oneinch_fantom_calls_transfers_tx_hash__start__transfer_trace_address (models/oneinch/fantom/oneinch_fantom_schema.yml)
> 22:24:37    Got 2133 results, configured to fail if != 0
> ```
> 
> on production everything is great image
> 
> i'll try to add more keys to pass this, but...

we've seen similar issues on other spells, where if any of these fields have a `null` value, the join condition in the generated merge statement fails and it inserts on the incremental runs, causing duplicates. we're exploring overriding the default dbt-core macro for merge statements to make `null = null` work in join condition, but for now, it'll fail and write dupes. is it possible you have nulls in one or more of the keys?

grkhr:
@jeff-dude Yes, it's possible. Added a workaround for that, let's see

> if yes, is that key truly needed?
Yes. But, as I said in the first comment, it's proof-of-concept view. We'll deploy that, run some tests, discuss with the team. Then we can make it more production mature when we'll be scaling it on the other networks

grkhr:
all check have passed finally, ready for merge i hope! @jeff-dude

grkhr:
@jeff-dude all set now, ready for merge.

jeff-dude (approved):
great work here @grkhr -- look forward to continuing this path!

PR #3792, opened 2023-07-18
zeroex - activate poly deduped

RantumBits:


jeff-dude:
> @jeff-dude can you review when you have a chance? Ii've updated legacy only here so far. Is this sufficient?

bug fixes are fine on legacy only. keeps the review simple.

with that in mind, what is the fix here? it's only removing the exclusion, what was the issue causing dupes?

RantumBits:
> > @jeff-dude can you review when you have a chance? Ii've updated legacy only here so far. Is this sufficient?
> 
> bug fixes are fine on legacy only. keeps the review simple.
> 
> with that in mind, what is the fix here? it's only removing the exclusion, what was the issue causing dupes?

The underlying model was updated in PR #3356

jeff-dude (approved):

jeff-dude:
> > > @jeff-dude can you review when you have a chance? Ii've updated legacy only here so far. Is this sufficient?
> > 
> > 
> > bug fixes are fine on legacy only. keeps the review simple.
> > with that in mind, what is the fix here? it's only removing the exclusion, what was the issue causing dupes?
> 
> The underlying model was updated in PR #3356

thanks for clarification

PR #3793, opened 2023-07-19
Migrate aragon & dao creator tool spells 

henrystats:


henrystats:
> @henrystats at first glance, it looks like your set up for the seed tests are all good. we can also see that the seeds are being built and the test query runs, it just fails the actual test within the query.
> 
> first thing to try:
> 
> * run `dbt compile`
> * grab the test query (should be under models/aragon//schema/ -- that may not be exact, but general path)
> * you should see various check seed named sql files, they should contain the compiled sql that builds the test
> * copy/paste to dune app
> * confirm it's returning rows (it shouldn't)
> * investigate the query as to why -- could it be null values in the join conditions? could join conditions be wrong? were the check columns written correctly to seed and model?

Thank you very much ser... I'll follow these steps and check to see what's wrong... I also noticed that there were some missing results which was weird because it's the exact same query...

Just a tiny bit of missing results too.

jeff-dude:
@henrystats i'd highly recommended breaking out this PR across multiple branches / PRs. this is a bit difficult to parse through. can you do a PR for each project? i think best case scenario:
- PR for simply migrating to dunesql, no changes to code
- PR for new project (one PR per project), dunesql only

henrystats:
Okie ser.. will do that, tyvm ser!

henrystats:
Thank you very much ser @jeff-dude , all pr's now opened... Wanted to close this and saw that I didn't reply the last text...

PR #3797, opened 2023-07-19
migrate NFT arbitrum models

0xRobin:
Migrates Arbitrum NFT models


jeff-dude (commented):
code looks 🔥 per usual, but curious why you made tofu_v1 a macro, if it's not exactly set up to be used for multiple models? the macro itself hardcodes version value and such. is this something you plan to enhance in the future, as more versions are released?

models/_sector/nft/trades/old/platforms/stealcam_arbitrum_events.sql
```diff
@@ -1,7 +1,7 @@
 {{ config(
     schema = 'stealcam_arbitrum',
     alias = alias('events'),
-    partition_by = ['block_date'],
```
jeff-dude, line 4: also, how come we're removing partitions on the arbritrum spells?

0xRobin: following this guideline on delta lake partitioning:
https://docs.databricks.com/tables/partitions.html#do-small-tables-need-to-be-partitioned
I'm going to remove partitioning from the smaller marketplace models. (this specific one is only 50k records)

0xRobin:
> if it's not exactly set up to be used for multiple models?

The macro will be re-used for tofu v1 models on bnb, polygon and optimism :)

models/_sector/nft/trades/old/platforms/stealcam_arbitrum_events.sql
jeff-dude: good point, we should eventually do this check universally. i'll add to list for post-migration.

PR #3818, opened 2023-07-20
Revert "Removes trace address from legacy files aswell" on uniswap lineage

jeff-dude:
fyi @couralex6 -- i need to add this back on legacy files, as all other dexes still use it and causes downstream failures. we will need to just let the column live on spark for now, shouldn't hurt anything.


couralex6:
Sounds good!

PR #3820, opened 2023-07-20
Quick fix dao balances - ready for review 

henrystats:

### DIFF
--- a/models/dao/balances/dao_balances.sql
+++ b/models/dao/balances/dao_balances.sql
@@ -92,6 +92,7 @@ LEFT JOIN
 LEFT JOIN 
     {{ source('prices', 'usd') }} e 
     ON db.contract_address = lower('0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee')
+    AND d.day = e.minute
     AND db.blockchain = 'ethereum'
     AND e.symbol = 'WETH'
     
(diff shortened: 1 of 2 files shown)


henrystats:
hola ser @jeff-dude, I just tried to do a very tiny change cause it seems ETH prices on the prices.usd table (for ethereum) is down... so wanted to add a coalesce here but then that would be fixed soon soo probably shouldn't bother adding this...

But it's saying I need to edit the tags and the tag for the legacy file is already there and I can't edit the dune-sql one since I'd need to migrate the spell (already migrating in another spell)...

But curious why I can't make the small edit.

jeff-dude:
> hola ser @jeff-dude, I just tried to do a very tiny change cause it seems ETH prices on the prices.usd table (for ethereum) is down... so wanted to add a coalesce here
> 
> But it's saying I need to edit the tags and the tag for the legacy file is already there and I can't edit the dune-sql one since I'd need to migrate the spell (already migrating in another spell)...
> 
> But curious why I can't make the small edit.

there is no need to edit the non-legacy file until you are ready to migrate it. if you are looking for a hotfix on spark only, then only edit the legacy file. that check in CI is making sure anything going to prod has one of the tags, but the intention is for that file to sit idle until migrated, then the dunesql tag will be applied

henrystats:
okee ser... thank you very much..

jeff-dude (approved):

PR #3829, opened 2023-07-21
ethereum pools lido liquidity migration

ppclunghe:


jeff-dude:
@ppclunghe looks like this one may need broken out further? seems some dependencies are missing, since they aren't all merged yet

ppclunghe:
@jeff-dude yes, seems that upstream spell doesn't pass the tests because not all downstream spells are merged yet, can we merge it without tests passing? or what else we can do with it?

models/lido/lido_liquidity.sql
jeff-dude, line 1: can you remove this one from this PR?
then we will follow-up with a PR for the cross-chain/cross-project spell once all others are merged and stable
CI should pass without this one here

ppclunghe: I removed upstream spell, CI passed

jeff-dude, line 1: this file should remain in the repo, but just don't make any modifications on it so that it doesn't appear in the PR here

models/lido/lido_schema.yml
jeff-dude, line 1: similar, you can keep this in the schema file. it'll be there and ready when you modify the cross-chain spell in follow-up PR soon

jeff-dude (changes requested):
apologies for any confusion, but see comments below. the idea is that the non-legacy file will sit in the repo with no modifications, until a follow-up PR later when ready, where you'll apply the dunesql tag and such. without tags, it'll just sit there, which is fine.

models/lido/lido_liquidity.sql
ppclunghe: I have tried to keep this file non-legacy lido_liquidity file without tag, but test doesn't pass in this way
https://github.com/duneanalytics/spellbook/actions/runs/5677885801/job/15387001473#step:10:14
so I decided to delete this file to get tests passed
am I right that once downstreem PRs will be merged I can re-create lido_liquidity ?

jeff-dude: but if the file is left in the repo, and not modified in your branch, it shouldn't show up in the PR, therefore it won't be checked in that `check tags` step. we want it to have no tag for now, until it's ready (after this PR and all other liquidity ones)

ppclunghe: ok, I am going to close this PR and create the new one where lido_liquidity file will not be modified

PR #3830, opened 2023-07-21
Create spell for the latest usd price of every token in dex.trades

bernatfp:
This spell aims to offer a solution to obtaining USD prices from tokens which don't get traded often enough to reliably show up in the Coinpaprika price feed.

Examples of tokens which sometimes don't have a price in Coinpaprika:
- vBNT: https://coinpaprika.com/coin/vbnt-bancor-governance-token/
- CRETH2: https://coinpaprika.com/coin/creth2-cream-eth-token/

For vBNT (`0x48fb253446873234f2febbf9bdeaa72d9d387f94`), we can find it through the query in this PR on Dune:
image

This spell includes the USD value calculation of raw token units for cases where we don't know the number of decimals of the token.

### DIFF
--- a/models/dex/dex_prices_latest.sql
+++ b/models/dex/dex_prices_latest.sql
@@ -35,7 +35,7 @@ FROM (
             token_bought_amount AS token_amount,
             token_bought_amount_raw AS token_amount_raw,
             amount_usd
-        FROM dex.trades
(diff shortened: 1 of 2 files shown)


jeff-dude:
closing due to #3846 -- can revisit migration when dependencies are ready

PR #3832, opened 2023-07-21
Migrate Chainlink PriceFeeds to DuneSQL

anon-r-7:
- Migrate Chainlink Price Feeds to DuneSQL

@jeff-dude This is the first of potentially a few PRs for Chainlink as we discussed. This focuses just on migrating the Chainlink model to DuneSQL. I've backtested the data from this PR against the legacy and everything lines up. A few major changes were needed due to some functions no longer being supported.

You'll notice I've renamed the oracle_addresses files to be prefixed with `price_feed`. This is to maintain clear structure given we are going to be adding many more types of metrics for Chainlink in the future, the first of which being OCR, followed by FM, Keepers, Automation, CCIP and LLO. Each set of queries that support a given metric should be prefixed accordingly. 

In a followup PR, we will also be adding extending the price feed models by (a) updating the oracle addresses to include all known feeds, and (b) adding support for all evm chains. Here's a roadmap for PRs you should expect to see from us between today and the next few weeks

- PR#1: PriceFeed DuneSQL Migration (code complete, PR submitted)
- PR#2: OCR for all EVM chains (code complete, waiting to submit PR until this one is finished)
- PR#3: Extend PriceFeeds (WIP)
- PR#4: FM for all EVM chains
- PR#5: Keeper for all EVM chains


models/chainlink/bnb/chainlink_bnb_price_feeds.sql
```diff
@@ -1,59 +1,59 @@
...
-    alias = alias('price_feeds'),
-    partition_by = ['block_date'],
-    materialized = 'incremental',
-    file_format = 'delta',
-    incremental_strategy = 'merge',
-    unique_key = ['blockchain', 'block_number', 'proxy_address','underlying_token_address'],
+{{
+  config(
+    tags=['dunesql'],
+    alias='price_feeds',
```
jeff-dude, line 4: please keep the alias macro here:
`alias = alias('price_feeds'),`

notice the non-legacy files don't pass the second param in, as it defaults to false. while the legacy files pass in the second flag and set to true.

this is needed for our orchestration to configure behind-the-scenes

jeff-dude: this will apply universally, only commenting here

models/chainlink/bnb/chainlink_bnb_price_feeds.sql
```diff
@@ -1,59 +1,59 @@
...
-                  /POWER(10, cfa.decimals)
-              )                                         AS oracle_price,
-	       cfa.proxy_address,
-           cfa.aggregator_address
-	FROM {{ source('bnb', 'logs') }} l
-	INNER JOIN {{ ref('chainlink_bnb_oracle_addresses') }} cfa ON l.contract_address = cfa.aggregator_address
-	WHERE l.topic1 = '{{answer_updated}}'
+    SELECT
+        l.block_time,
+        cast(date_trunc('day', l.block_time) as date) as block_date,
```
jeff-dude, line 34: if `block_date` is used solely for the partition, we are looking to universally change this to a higher level of granularity, `block_month`.

can you update to be `'month'` instead, then rename column and partition in config to `block_month`?

if you use `block_date` for more than the partition, then you can simply add new column instead. in fact, that may be safer if downstream queries on the app depend on this column for whatever reason.

jeff-dude: this would also apply universally, only commenting here

models/chainlink/bnb/chainlink_bnb_schema.yml
```diff
@@ -5,44 +5,39 @@ models:
...
-          match_columns:
-            - blockchain
-            - block_number
-            - proxy_address
-            - underlying_token_address
-          check_columns:
-            - aggregator_address
-            - feed_name
-            - oracle_price
+    # Tempoarily Disable, test_data not found.
```
jeff-dude, line 13: we can re-enable this by forcing an update to the seed file (`chainlink_get_price_seed.sql`), so it pushes to this PR.

then for the columns listed, both match and check, you can create a schema file in the seeds directory path which calls out their datatype and applies a `tag:dunesql`. here is an example: 
https://github.com/duneanalytics/spellbook/blob/main/seeds/_sector/nft/_schema.yml#L1-L15

jeff-dude: you only need to add columns which leverage unique dunesql data types typically (varbinary, uint256, etc)

jeff-dude (changes requested):
submitting initial feedback on bnb and cross-chain spells.

assuming other chains are similar.

seeds/chainlink/_schema.yml
```diff
@@ -0,0 +1,69 @@
+version: 2
+
+seeds:
+  - name: chainlink_bnb_price_feeds
```
jeff-dude, line 4: this must be the name of the seed file, not model name -- is that set correctly?

models/chainlink/bnb/chainlink_bnb_price_feeds.sql
```diff
@@ -1,59 +1,61 @@
...
-    partition_by = ['block_date'],
-    materialized = 'incremental',
-    file_format = 'delta',
-    incremental_strategy = 'merge',
-    unique_key = ['blockchain', 'block_number', 'proxy_address','underlying_token_address'],
+{{
+  config(
+    tags=['dunesql'],
+    alias=alias('price_feeds'),
+    partition_by=['block_date'],
```
jeff-dude, line 5: looks like you built `block_month`, but forgot the partition on some?

jeff-dude:
one final comment above, but otherwise it's looking good!
thanks again for the patience and help pushing these along.

jeff-dude (approved):
let's give this a go!

one last thank you for the patience and working with us here to keep the migration clean. look forward to seeing your further contribution.

MSilb7:
Update here: This broke some downstream queries because `underlying_token_address` column was dropped. Is it possible to edit the spell to re-add this column in? @anon-r-7

jeff-dude:
also tagging @AnonJon as another chainlink spell contributor

MSilb7:
> > Update here: This broke some downstream queries because `underlying_token_address` column was dropped. Is it possible to edit the spell to re-add this column in? @anon-r-7
> 
> Hey mslib. That particular field was removed because it was not relevant in 99.9999% of use cases. We've refactored this such that the equivalent prices are calculated without requiring the manual work to find this address. Reason for this decision is that it wasn't practical to find, and keep up to date, this address manually for every feed on every network. And in fact, for some of the feeds, the underlying address was incorrect. This is more of a data availability and verification problem then it is a code problem. The trade off decision here was (1) stick to keeping it and have support for just a few feeds on 3 chains or (2) provide support for all feeds on all chains. Given that in many cases it's not relevant we went with option two. The benefit is it no longer requires manual work to add or keep up to date these feeds, a script runs that can do all this programmatically say once a month and then create a PR. I recommend updating downstream tables accordingly.
> 
> We did think through the use cases for this and therefore added columns for base and quote so the user could join based on some pair of they wanted to without having the underlying address in the table (or other such use cases).

got it, yeah maybe i'll just remove this from queries. the old logic for token prices was:

1. Check in prices.usd
2. If doesn't exist, check in Chainlink feeds.
3. If doesn't exist, check in dex trades prices (high error rate).

So I can just cut out the Chainlink step.

PR #3847, opened 2023-07-24
Revert "oneinch: init fantom calls transfers & migrate upstream spell…

jeff-dude:
…s (#3791)"

This reverts commit 86e19c1aa92e56474cd18660ef5cebde907e50d5.

sorry @grkhr -- we hit a bug where schemas which are new to dunesql only cause spark prod to fail, as the "pre-run" steps in `dbt run` are querying for all schemas in the dbt project, yet can't find on spark. we are looking into this bug and will look to fix and add this back asap. fwiw, it built on dunesql, it just won't refresh for now.


grkhr:
@jeff-dude no worries! I can imagine. looking forward to the fix

PR #3854, opened 2023-07-25
[waiting for the fix] oneinch: bnb_calls_transfers init

grkhr:
This PR is exactly the same as https://github.com/duneanalytics/spellbook/pull/3791 but for BNB.

This is step 2 in our proof-of-concept spellbook'ing everything, further these views will be deleted.


grkhr:
cc: @jeff-dude strange error here

PR #3864, opened 2023-07-26
[easy] OP Static Updates

MSilb7:
This is ready to go whenever, but as usual, I'll keep adding more mapping updates to this PR as things come in.

- New contract creator addresses: Our current workflow is 1) Identify contracts that meet some usage threshold, 2) Try to map the deployers we don't already have mapped.

I'm also still updating across legacy versions as well, but it would be nice to kill the legacy files at some point.


jeff-dude:
> I'm also still updating across legacy versions as well, but it would be nice to kill the legacy files at some point.

you can kill off the legacy edits when all downstream spells that rely upon it are migrated

jeff-dude (approved):

PR #3866, opened 2023-07-26
Update to Tigris Trades - Ready for review

henrystats:


henrystats:
hola ser @jeff-dude, apologies for yet another tag :(

I'm trying to add tests to this spell to check for when there are missing transactions so they would fail and can be fixed if it ever happens...

I added a test file and It isn't showing in test_initial_model... I manually entered a wrong result so it would fail but it's still passing which means the test isn't working... 

I checked other test files written and I couldn't find where the mistake I was making was.

jeff-dude (approved):
whew, quite the overhaul here but i think it all looks good!

PR #3869, opened 2023-07-26
Migrate oneinch_v1_ethereum_trades to DuneSQL

antonio-mendes:
Attempting to migrate the `oneinch_v1_ethereum_trades` spell to use DuneSQL.


models/oneinch/ethereum/oneinch_v1_ethereum_trades.sql
```diff
@@ -30,218 +31,218 @@
...
-        toToken as to_token,
-        tokensAmount as from_amount,
-        minTokensAmount as to_amount,
-        call_tx_hash as tx_hash,
-        call_block_time as block_time,
-        call_trace_address as trace_address,
-        CAST(-1 as integer) AS evt_index,
-        contract_address
+      call_block_number AS block_number,
+      TRY_CAST(NULL AS VARCHAR) AS taker,
```
jeff-dude, line 38: was `try_cast` preferred in harmonizer?

jeff-dude: also, `taker` and `maker` fields should always be `varbinary`

models/oneinch/ethereum/oneinch_v1_ethereum_trades.sql
```diff
@@ -30,218 +31,218 @@
...
+      call_block_time AS block_time,
+      call_trace_address AS trace_address,
+      TRY_CAST(-1 AS INTEGER) AS evt_index,
+      contract_address
     FROM
         {{ source('oneinch_ethereum', 'exchange_v1_call_aggregate') }}
     WHERE
         call_success
         {% if is_incremental() %}
         AND call_block_time >= date_trunc("day", now() - interval '1 week')
```
jeff-dude, line 53: i believe this should fail, we had to reformat the interval syntax

jeff-dude: when you ran `dbt compile`, it pulled the `non-incremental` jinja logic since you haven't built the spell locally yet (or at the time), but if you ran the spell, then compiled, it would spit out the incrmental logic. this is likely why you've missed it here.

jeff-dude (changes requested):
few comments at quick first glance

antonio-mendes:
No need to review still, I just spotted a few errors

models/oneinch/ethereum/oneinch_v1_ethereum_trades.sql
antonio-mendes: Yeah the harmonizer sometimes uses `try_cast` rather than cast

antonio-mendes: I will take some time maybe to look into the harmonizer code and change it up so it always tries t make `taker` and `maker` varbinary rather than `varchar`

models/oneinch/ethereum/oneinch_v1_ethereum_trades.sql
antonio-mendes: You're right! Good catch! Also just realized that trino apparently doesn't allow `- interval '1' week` so will re-define this in days

PR #3876, opened 2023-07-27
Check unique dunesql spells

couralex6:
Since, we can't support new DuneSQL models that don't have a `_legacy.sql` counterpart, we need to check for those missing legacy models in the CI. 

The script runs dbt ls on both dunesql and spark and checks for models that exist in dunesql but not spark.


couralex6:
Ready to merge

jeff-dude (approved):

PR #3707, opened 2023-07-10
Solana Orca Whirlpool

andrewhong5297:
Adding Solana Orca Whirlpool using basically the `dex.trades` schema. I'm not including it into `dex.trades` now because I think Solana trades (for both NFTs and DEXs) will require a different table schema than how we handle EVM ones.

PR #3895, opened 2023-07-30
chore: update to sharky v6

ennnas:
I'm adding support for the latest version of the Sharky program (v6.0.0) and migrating the model to Dune


ennnas:

Hosuke (approved):
LGTM.✅ for the dune sql updates
Thank you @ennnas 

[Sample sharky solana events](https://dune.com/queries/2798485)

jeff-dude:
hi @ennnas -- thank you for initiating the migration here! since it's an existing spell, and will run on both engines simultaneously until spark is decommissioned, we need to take action on the `_legacy` version of the spell too. there are two choices imo:
- apply the same updates to the spark spell, to ensure data in sync
- OR override all of the legacy script with a simple `select 1`, so that it runs and keeps our `dbt compile` commands working cross-engine, but also saves resources on the backend orchestration since all interactions on the dune app will read from dunesql version now anyway
  - example here: https://github.com/duneanalytics/spellbook/pull/3707/files#diff-f23e35a88fc95756720c5c81ef980824f64139373773348277416683243de81a

my vote is option 2 above, for simplicity and saving the compute time.

jeff-dude:
> hi @ennnas -- thank you for initiating the migration here! since it's an existing spell, and will run on both engines simultaneously until spark is decommissioned, we need to take action on the `_legacy` version of the spell too. there are two choices imo:
> 
> * apply the same updates to the spark spell, to ensure data in sync
> * OR override all of the legacy script with a simple `select 1`, so that it runs and keeps our `dbt compile` commands working cross-engine, but also saves resources on the backend orchestration since all interactions on the dune app will read from dunesql version now anyway
>   
>   * example here: https://github.com/duneanalytics/spellbook/pull/3707/files#diff-f23e35a88fc95756720c5c81ef980824f64139373773348277416683243de81a
> 
> my vote is option 2 above, for simplicity and saving the compute time.

option 2 seems feasible as the spell looks to be a standalone without downstream impact, based on some `dbt ls` commands i ran to test that out.

models/sharky/solana/sharky_solana_events.sql
```diff
@@ -1,4 +1,5 @@
 {{ config(
+    tags=['dunesql'],
     alias = alias('events'),
     partition_by = ['block_date'],
```
jeff-dude, line 4: could you also increase partition size to monthly and apply a prehook, like here?
https://github.com/duneanalytics/spellbook/pull/3707/files#diff-0576be6634b68cdb81185a4376f442e9be34f967a6f182656a63b89c87a1f3d0R6-R10

we're using that prehook on larger data volume spells on trino (soon will be universally built at project level so we don't have to config each model moving forward). as for the partition, we're looking to increase universally. if `block_date` is used at all on the dune app for queries, you can add a new column to be safe, otherwise feel free to override existing

ennnas: Yeah `block_date` is used so I'll add a new one

jeff-dude:
> @jeff-dude I added the `block_month` and dropped the legacy query. I tested everything on 1 month worth of data and the results are correct. I'm now reverting to the original start time

thank you for the updates! from a spell building perspective, this looks good.

are you familiar with solana decoding live on dune now? this was built prior to that and reads from the base tables. i wonder if its possible to update this during migration to read from decoded tables? quick glance at the data explorer on the app shows that sharky appears to be decoded and there are various tables.

i get a little nervous when i see the runtime so high on the CI tests. i know solana is large data volume regardless, but i imagine the decoded table reads should help and save a lot of the heavy logic used here.

as of now, our dunesql cluster used on CI tests via PRs is shared with prod orchestration, so failures we see here will more than likely occur in prod.

ennnas:
Yeah I looked a bit at the decoded data for Solana and for Sharky but I remember that the data still had many decoding issues probably because the slots were not aligning perfectly. If we want to go that way I would feel more comfortable in creating a separate spell for now and use this one as a ground truth to run some comparisons in the future and eventually use only the one with decoded tables. I could take care of creating a new spell after we merge this one if you want

jeff-dude:
> Yeah I looked a bit at the decoded data for Solana and for Sharky but I remember that the data still had many decoding issues probably because the slots were not aligning perfectly. If we want to go that way I would feel more comfortable in creating a separate spell for now and use this one as a ground truth to run some comparisons in the future and eventually use only the one with decoded tables. I could take care of creating a new spell after we merge this one if you want

that's fair. let's go with that plan since this already exists.

ennnas:
@jeff-dude ok sure, do you need anything else from my side?

ennnas:
> 

@jeff-dude Yes I will start working on it as soon as the decoded tables are updated since the latest smart contract is not yet decoded on Dune

models/sharky/solana/sharky_solana_events.sql
```diff
@@ -77,65 +83,118 @@ WITH sharky_txs AS (
         INNER JOIN filtered_txs
```
jkylling, line 83: Could we try switching the join order to
```
FROM filtered_txs
INNER JOIN sharky_txs
```
?

belen-pruvost: This does the trick - see [sample PR](https://github.com/duneanalytics/spellbook/pull/3983)

ennnas: Excellent 🚀 Do you want to commit the changes yourself or should I do it?

belen-pruvost: Feel free to commit the change yourself 🙌 thank you!

Restuta:
🥁 can't wait for this to go live

jeff-dude:
closing for #4055

PR #3896, opened 2023-07-30
Creator Mapping -> All EVMS

MSilb7:
Extend optimism creator mappings to all EVMs.

This includes:
- Modified macro to include all EVM chains
- Iterating through chains for mapping spells (testing this)

Still Todo:
- We need to ensure that we have a spell to aggregate all ERC20 and NFT tokens across chains. The existing ERC20 and NFT spells are not sufficient:
1. There is no table for all ERC20s (Todo and add here)
2. The NFT tokens table depends on a spell where chains are manually entered


MSilb7:
Got an error for the query exceeding planning time. Maybe due to the erc20 token logic?

We could revert / readd other spells to generate erc20 tokens tables if that's faster. cc @jeff-dude

MSilb7:
attempting a simplifaction until we get an "all erc20s" table

MSilb7:
Getting planning error. Likely related to window functions, we've observed Trino planning errors in queries for window functions. Open to any recommendations, I ~think this is optimized about as much as possible. @jeff-dude

MSilb7:
seems like the intial build will still be super heavy with the code ranks. I wonder if there's a way to split it in to chunks? Now that there's logic where we only re-do the ranks for "new" contracts?

MSilb7:
seems like the intial build will still be super heavy with the code ranks. I wonder if there's a way to split it in to chunks? Now that there's logic where we only re-do the ranks for "new" contracts?

models/contracts/contracts_contract_mapping.sql
jeff-dude, line 1: unfortunately, this may be a bit too ambitious. you will need to build one spell per blockchain, then do the normal cross-chain spell on top of each.

also, we've learned recently that dunesql engine would prefer the `FROM` to be the larger table in the joins, so if `transactions` is larger than `creation_traces` for example, i would switch the order

jeff-dude (changes requested):

models/contracts/contracts_contract_mapping.sql
MSilb7: do you think using the `evms` schema tables would work well, or still have issues?

One spell per chain feels tough to maintain as more chains get added (unless there's a macro-like workaround that creates a "add a chain to a list -> get it included in a table" experience)

MSilb7: there are also bnb & fantom tables with the same name, but completely different (made by someone else), so maybe we'll have to use different table names for those (unless we can override the existing ones).

MSilb7: interesting note on the from vs join table selection.

jeff-dude: same design approach in `erc_4337` spells

MSilb7: yeah I think I'll try the nft_transfers approach.

~eventually, I'm a fan of unified chains tables (maybe pick OP Chains [Base, OP, etc] as a start since these have the exact same schema?)

jeff-dude: what do you mean exactly by unified chains? as in the current design, where you run multiple blockchains in one model?

i'm curious how that differs from taking an extra step to build per blockchain, then cross-chain on top? only query that cross-chain one?

MSilb7: one sql file, where all you have to do is add a chain to the chain macro - versus 10+ sql files where you have to remember to add the chain to every spell as it gets added.

Like for the next EVM chain that gets added - how many spells need to be manually updated?

MSilb7: Maybe said another way: How can you make overhead of a new chain as minimal as possible, but have everything "work"

jeff-dude: gotcha. imo, there isn't a huge difference between:
- adding blockchain to macro -- then existing spell would work with new blockchain
- adding new spell with params to pass into macro, but all logic is done just once

the difference in processing/compute time and resources is enough to justify option two.

if you want to explore one spell approach, we'd need to look into jinja for loops that loop through each chain and load it one at a time. that would be a fun experiment, but likely not high on our internal priorities until migration is complete

MSilb7:
Getting errors for "too many requests" when this runs the initial build for Polygon, but other chains are passing.

MSilb7:
will segment out 1 chain at a time

jeff-dude:
closing this, as we did one separate for ethereum and should follow suit on others

PR #3897, opened 2023-07-30
Init Base NFT Folder

MSilb7:
Copy Optimism NFT -> Base.

Find and replace Optimism for Base. This is a prereq for https://github.com/duneanalytics/spellbook/pull/3896

This depends on: https://github.com/duneanalytics/spellbook/pull/3898 (Some necessary files copied over)

PR #3898, opened 2023-07-30
Add Base to Transfers + Chains Macro

MSilb7:
~This adds a table for "all" ERC20 tokens (not just ones we've identified) and a total chains ERC20 transfers table with other relevant information.~

~This is a prereq for https://github.com/duneanalytics/spellbook/pull/3896 & #3897  since we need to be able to identify what is and isn't a token contract.~

~This also fixes the chains macro (duped in #3896 as well)~

Simplified to just be base transfers + Chain macro update


MSilb7:
The Macro created here can be used in the https://github.com/duneanalytics/spellbook/pull/3653 upgrade/simplification as well.

MSilb7:
Tags issue carrying over here as well. the base spells are all net new, so dependency errors are odd.

MSilb7:
> Tags issue carrying over here as well. the base spells are all net new, so dependency errors are odd.

Yeah, this `transfers_base_erc20` model that's throwing tags errors only depends on itself, and there's a legacy model.

May be a test issue?

MSilb7:
spell initial build times out, but it should run fine.

macros/public/all_evm_testnet_chains.sql
jeff-dude, line 1: this macro can be stored in the same sql file as the mainnet macro. as long as macro names remain the same and unique, we can group them in same file and then rename file to something more generic. there should be other examples of this in macros you can see.

macros/public/all_evm_mainnets_testnets_chains.sql
jeff-dude, line 1: ditto to this macro, all can be one file

profiles.yml
jeff-dude, line 1: this seems accidental -- can you revert?

models/transfers/transfers_erc20.md
jeff-dude, line 1: if we already have all the blockchain-specific erc20 transfers spells, why are we reading from the sources again here in a loop? i don't think we'll ever be able to build this as-is. we'll need to do the normal blockchain version spell for each, then a union of them after they are materialized

jeff-dude: can we remove all markdown files and just push them to another branch on your fork, save for later? this will take some more consideration

MSilb7: a benefit here would be to not have to copy and paste the same spell for every chain (i.e. some are missing from the chain-specific folders). But yeah I'll port out to a new branch.

As a general thought - I'd be a fan of having chain names in a macro, and we run that through spells, versus having folders for each chain (a lot of overhead to manage).

MSilb7: maybe for now we can add the missing chains from erc20 transfers (Base, Gnosis, Goerli)

MSilb7: The purpose for the overall transfers table and token was replaced with hitting the transfers event table every time in https://github.com/duneanalytics/spellbook/pull/3896 - so extended really isn't necessary yet (although it's easy to copy and paste

(So we don't have to create tables that collide with future token automation)

models/transfers/base/transfers_base_eth_legacy.sql
jeff-dude, line 1: why do we need to build this out fully on spark, rather than a dummy approach similar to other legacy in this PR?

jeff-dude (changes requested):

models/transfers/base/transfers_base_eth_legacy.sql
MSilb7: changing. just a leftover from the source folder

models/transfers/base/transfers_base_schema.yml
```diff
@@ -0,0 +1,82 @@
...
+    meta:
+      blockchain: base
+      sector: transfers
+      project: erc20
+      contributors: soispoke, dot2dotseurat, tschubotz
+    config:
+      tags: ['transfers', 'base', 'erc20']
+    description: "ERC20 Token Transfers on base. This table is updated every 30 minutes."
+    columns:
+      - name: unique_transfer_id
```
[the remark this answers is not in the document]
MSilb7: adding

jeff-dude (approved):

PR #3912, opened 2023-07-31
Migrate NFT Events

MSilb7:
Migrate NFT Events table to remove blockers on transfers and downstream tables.


MSilb7:
seems like there's still a lot of legacy debt here beyond my knowledge. closing for now, but this seems like a requirement for any nft-related downstream spells.

PR #3923, opened 2023-08-02
[EASY] Seaport Spell typo fix

pandajackson42:


jeff-dude:
@0xRobin can you validate this against your recent work on seaport?

0xRobin (approved):
this change is fine. Though the `base_pairs` models aren't used anymore in the NFT trades models for opensea, @pandajackson42 are you using these models on Dune? If not I'd suggest removing the `expose_spell` macro from them.

pandajackson42:
> are you using these models on Dune? 
Yes I use

> the base_pairs models aren't used anymore in the NFT trades models for opensea
Saw your new code, thx for noticing!

jeff-dude (approved):

jeff-dude (changes requested):
actually, this PR will need some work if we want to finalize still.

the non-legacy files will need migrated to dunesql syntax. is the plan to migrate these to dunesql and then continue to see them updated in prod? based on the answer in PR, i assume yes, which means we need to continue with full migration, as noted above

pandajackson42:
> actually, this PR will need some work if we want to finalize still.
> 
> the non-legacy files will need migrated to dunesql syntax. is the plan to migrate these to dunesql and then continue to see them updated in prod? based on the answer in PR, i assume yes, which means we need to continue with full migration, as noted above


gotcha. Do I need to do the migration by myself?

jeff-dude:
> > actually, this PR will need some work if we want to finalize still.
> > the non-legacy files will need migrated to dunesql syntax. is the plan to migrate these to dunesql and then continue to see them updated in prod? based on the answer in PR, i assume yes, which means we need to continue with full migration, as noted above
> 
> gotcha. Do I need to do the migration by myself?

i think so. while rob is handling the full `nft.trades` lineage, these are no longer part of that. it would be helpful if you could try the migration here yourself 👍

jeff-dude:
closing due to inactivity & no longer part of nft sector lineage

PR #3928, opened 2023-08-02
T-32005 Set join_distribution_type on materialized and partitioned models

couralex6:
Set `join_distribution_type` session property on every model that is both materialized and partitioned on entire project.

Tested by running 

`dbt -d run  --select +nft_polygon_transfers,tag:dunesql -f`

The log `On model.spellbook.nft_polygon_transfers: SET SESSION join_distribution_type='PARTITIONED'` only appears before executing `nft_polygon_transfers`, and not on every introspection query like it did before.


couralex6:
Cancelling CI run since dbt assumes all files are modified

aalan3 (approved):

PR #3929, opened 2023-08-02
oneinch: calls_transfers prod, all chains

grkhr:
- `{blockchain}/oneinch_{blockchain}_ar_calls_transfers.sql` are identical except some constants, they are all calling main macros `macros/models/_project/oneinch/oneinch_ar_calls_transfers_macro.sql`
- then they are unioning to non-materialized views `oneinch_ar_calls.sql` and `oneinch_ar_calls_transfers_amounts.sql`
- new `oneinch.protocols` (simple dict, ~500 records) materialized as table
- other changes are minor / schemas / etc

and keep in mind to look closer to `abi.signatures` as I don't want to break something


grkhr:

macros/models/_project/oneinch/oneinch_ar_calls_transfers_macro.sql
```diff
@@ -47,83 +50,102 @@ calls as (
...
-        and slice(transfers.trace_address, 1, cardinality(calls.start)) = calls.start
-        
+        and slice(transfers.trace_address, 1, cardinality(calls.start)) = calls.start 
 )
 
-select * from merged
+
+select 
+    *
+    , cast(tx_hash as varchar)||'--'||
```
jeff-dude, line 141: do you have a strong opinion on why you'd generate a surrogate key like this to use as unique key, rather than just using multiple columns and not introducing a new one?

macros/models/_project/oneinch/oneinch_ar_calls_transfers_macro.sql
```diff
@@ -47,83 +50,102 @@ calls as (
...
-                block_time >= cast(date_add('day', {{lookback_days}}, now()) as timestamp)
+                block_time >= cast(date_add('day', {{ lookback_days }}, now()) as timestamp)
             {% else %}
                 block_time >= {{ project_start_date }}
             {% endif %}
             and "to" in (
-                select distinct contract_address from {{ ref('oneinch_contract_addresses') }}
-                where project = '1inch'
+                select distinct contract_address from {{ ref('oneinch_protocols') }}
+                where protocol = 'AR' and blockchain = '{{ blockchain }}' and main
```
jeff-dude, line 65: without digging deep, what exactly is `and main` at the end here?

jeff-dude (commented):
awesome jinja magic in here!

left minor comments below

jeff-dude:
> and keep in mind to look closer to `abi.signatures` as I don't want to break something

i'll check this now

macros/models/_project/oneinch/oneinch_ar_calls_transfers_macro.sql
grkhr: To keep `{blockchain}/oneinch_{blockchain}_ar_calls_transfers.sql` code and tests simpler. Looked at **nft_transfers** and did the same :)

macros/models/_project/oneinch/oneinch_ar_calls_transfers_macro.sql
grkhr: it's equal to `and main = true`, main is bolean

grkhr: image

macros/models/_project/oneinch/oneinch_ar_calls_transfers_macro.sql
grkhr: also, otherwise I'd need to create a "system" column like [here](https://github.com/grkhr/spellbook/blob/3a3fb96af389a402699125081d29caaa1f7c8750/models/oneinch/fantom/oneinch_fantom_calls_transfers.sql#L85) because it's nullable but be in primary key?

jeff-dude: makes sense -- thank you

jeff-dude (approved):
thanks again!

PR #3945, opened 2023-08-04
Incremental predicates

aalan3:
Testing incremental predicates on both dunesql and legacy


aalan3:
I tested this on both spark and dune sql. We should go for it. Dune sql is not really faster but at least uses less resources, I want to see what happens in a real run.

couralex6 (approved):

PR #3950, opened 2023-08-04
Add incremental strategy for dao.transactions

aalan3:
Follow up on https://github.com/duneanalytics/spellbook/pull/3948/files

### DIFF
--- a/models/dao/transactions/dao_transactions.sql
+++ b/models/dao/transactions/dao_transactions.sql
@@ -4,7 +4,7 @@
     materialized = 'incremental',
     file_format = 'delta',
     incremental_strategy = 'merge',
-    unique_key = ['tx_hash', 'tx_index'],
+    unique_key = ['tx_hash', 'tx_index', 'trace_address'],
     post_hook='{{ expose_spells(\'["ethereum", "gnosis", "polygon"]\',
                                 "sector",
                                 "dao",
--- a/models/dao/transactions/dao_transactions.sql
+++ b/models/dao/transactions/dao_transactions.sql
@@ -3,6 +3,7 @@
     alias = alias('transactions'),
     materialized = 'incremental',
     file_format = 'delta',
+    incremental_strategy = 'merge',
     unique_key = ['tx_hash', 'tx_index'],
     post_hook='{{ expose_spells(\'["ethereum", "gnosis", "polygon"]\',
                                 "sector",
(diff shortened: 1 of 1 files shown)


jeff-dude (approved):

models/dao/transactions/dao_transactions.sql
```diff
@@ -4,7 +4,7 @@
     materialized = 'incremental',
     file_format = 'delta',
     incremental_strategy = 'merge',
-    unique_key = ['tx_hash', 'tx_index'],
+    unique_key = ['tx_hash', 'tx_index', 'trace_address'],
```
jeff-dude, line 7: was this failing?

the ideal scenario:
- this model + all models in the dict below all have the same unique keys, since they are being unioned together

jeff-dude: @henrystats may be able to help us standardize here

henrystats: hmmm... they have the same unique keys but one... for the eth transactions table we need the **trace_address** for the unique keys...

for erc20 tables we don't but it was originally there and it caused duplicates because that field is null... soo I guess best option is to actually put a value in the trace address for erc20 transactions (the column isn't used anyways)....

but we need all the other columns in the downstream unique keys else there are going to be duplicates.

jeff-dude: why is unique keys [here](https://github.com/duneanalytics/spellbook/blob/main/models/dao/transactions/dao_transactions.sql#L6) different than unique keys [here](https://github.com/duneanalytics/spellbook/blob/main/models/dao/transactions/ethereum/dao_transactions_ethereum_erc20.sql#L8)?

ultimately, all spells in a roll-up like this should share the same unique keys. otherwise, they shouldn't fit into the same table

henrystats: it didn't have any unique keys before because it was materialized as table.. 

so the pr that made it incremental added unique keys to it... the one that was recently merged... 

the transactions downstream spells were already incremental

jeff-dude (approved):

models/dao/transactions/dao_transactions.sql
aalan3: Logs (erc20 transfers) and traces are being mixed here which is part of the reason why, `evt_index` is set to `tx_index` but these are different values. `evt_index` is the event index of a log within a transaction, `tx_index` is just the transaction index within a block.

PR #3961, opened 2023-08-07
Element polygon trades missing from nft.trades

henrystats:


henrystats:
hola ser @jeff-dude @0xRobin, I'm not sure why element_polygon_trades are being dropped from nft_trades...

I checked the spells and the model (element_polygon_trades) is included, also checked on dune itself that the model exists and it works... 

Select * from element_polygon_trades works 

but it isn't showing on the explorer (element tables) and also in nft.trades.

jeff-dude:
> hola ser @jeff-dude @0xRobin, I'm not sure why element_polygon_trades are being dropped from nft_trades...
> 
> I checked the spells and the model (element_polygon_trades) is included, also checked on dune itself that the model exists and it works...
> 
> Select * from element_polygon_trades works
> 
> but it isn't showing on the explorer (element tables) and also in nft.trades.

gm!

thank you for bringing this up. here is what i found:
[this where clause](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/nft/trades/nft_events_legacy.sql#L100) looks to filter out any data which was included in first part of the union. however, it leaves blockchain out of consideration. element also exists on ethereum, so when it does this filter, it finds a match to the ethereum version based on just project/version values.

can you look to repurpose this PR to alter the linked spell to fix that filter? add blockchain as a third condition?

henrystats:
ouuu okaie ser.. thank you very much will do!

henrystats:
done ser... do I need to update the dune-sql version of the spell? due to the migration ongoing I was thinking I'd not be able to do that as the tag is currently not defined so the dune-sql test would fail If I Make any edits..

jeff-dude:
> done ser... do I need to update the dune-sql version of the spell? due to the migration ongoing I was thinking I'd not be able to do that as the tag is currently not defined so the dune-sql test would fail If I Make any edits..
> 
> should I make the edit regardless? so when spark gets deprecated and dune-sql is the only version left, the changes wouldn't have to be redone.

yeah, let's make the changes on both. just don't add a tag for dunesql yet, then nothing will run (as we want).

henrystats:
okaie ser will update now, thank you!

henrystats:
tests are failing 👀

henrystats:
done ser @jeff-dude

jeff-dude (approved):

0xRobin:
Thanks for looking into this! I'll make sure to adapt this on the dunesql side during the migration.

PR #3963, opened 2023-08-07
create uniswap bnb pools spell and add it to uniswap pools spell

lequangphu:


models/uniswap/bnb/uniswap_bnb_pools.sql
```diff
@@ -0,0 +1,29 @@
...
+, pool
+, fee
+, token0
+, token1
+, evt_block_time AS creation_block_time
+, evt_block_number AS creation_block_number
+, contract_address
+FROM {{ source('uniswap_v3_bnb', 'Factory_evt_PoolCreated') }}
+{% if is_incremental() %}
+WHERE evt_block_time >= date_trunc("day", now() - interval '1 week')
```
jeff-dude, line 28: using double quotes on dunesql doesn't work as intended here. with double quotes, the engine thinks you're pulling a column named `day` in the `date_trunc` function.

please see here for example of new incremental filter logic:
https://github.com/duneanalytics/spellbook/blob/main/models/uniswap/ethereum/uniswap_v1_ethereum_trades.sql#L39-L41

jeff-dude: i would recommend you run `dbt compile` locally, grab your output code and run on the dune app. then i'd also run `dbt compile` but pass in a param flag to force incremental output, then run that query too. it's a good practice to ensure your syntax works.

jeff-dude:
@lequangphu just noticed #3763 is doing the same as here and further along in the review. we will merge that shortly. closing this with that in mind

PR #3996, opened 2023-08-09
DuneSQL Migration: Balancer BPT Transfers Spells

thetroyharris:
# Migrate all `balancer_v2_blockchain.transfers_bpt` to DuneSQL. Additionally, add this spell to Base, Gnosis, and Avalanche networks.

### DIFF
--- a/models/balancer/arbitrum/balancer_arbitrum_schema.yml
+++ b/models/balancer/arbitrum/balancer_arbitrum_schema.yml
@@ -57,9 +57,9 @@ models:
     meta:
       blockchain: arbitrum
       project: balancer_v2
-      contributors: victorstefenon
+      contributors: victorstefenon, thetroyharris
     config:
-      tags: ['arbitrum', 'bpt', 'prices']
+      tags: ['arbitrum', 'bpt', 'transfers']
     description: >
       Balancer Pool Token (BPT) transfer logs on Balancer, an automated portfolio manager and trading platform, on Arbitrum.
     tests:
@@ -68,20 +68,10 @@ models:
             - evt_tx_hash
             - evt_index
             - block_date
-      - check_seed_legacy:
-          seed_file: ref('balancer_transfers_bpt_seed')
-          filter:
-            blockchain: arbitrum
-          match_columns:
-            - evt_tx_hash
-            - evt_index
-            - block_date
-          check_columns:
-            - contract_address
-            - evt_block_number
-            - from
-            - to
     columns:
+      - &blockchain
+        name: blockchain
+        description: "Blockchain"
       - *contract_address
       - &block_date
         name: block_date
@@ -89,33 +79,21 @@ models:
       - &evt_tx_hash
         name: evt_tx_hash
         description: 'Transaction hash of transfer event'
-        tests:
-          - not_null
       - &evt_index
         name: evt_index
         description: 'Event index'
-        tests:
-          - not_null
       - &evt_block_time
         name: evt_block_time
         description: 'Block time of transfer event'
-        tests:
-          - not_null
       - &evt_block_number
         name: evt_block_number
         description: 'Block number of transfer event'
-        tests:
-          - not_null
       - &from
         name: from
         description: 'Address of BPT provider of transfer event'
-        tests:
-          - not_null
       - &to
         name: to
         description: 'Address of BPT receiver of transfer event'
-        tests:
-          - not_null
       - &value
         name: value
         description: 'Amount of BPT transferred in transfer event'
@@ -185,9 +163,7 @@ models:
           project: balancer
           version: 2
     columns:
-      - &blockchain
-        name: blockchain
-        description: "Blockchain"
+      - *blockchain
       - &project
         name: project
         description: "Project name (balancer)"
--- a/models/balancer/arbitrum/balancer_v2_arbitrum_transfers_bpt.sql
+++ b/models/balancer/arbitrum/balancer_v2_arbitrum_transfers_bpt.sql
@@ -1,6 +1,7 @@
 {{
     config(
-        schema='balancer_v2_arbitrum',
+        schema = 'balancer_v2_arbitrum',
+        tags = ['dunesql'],
         alias = alias('transfers_bpt'),
         partition_by = ['block_date'],
         materialized = 'incremental',
@@ -10,7 +11,7 @@
         post_hook='{{ expose_spells(\'["arbitrum"]\',
                                     "project",
                                     "balancer_v2",
-                                    \'["stefenon"]\') }}'
+                                    \'["stefenon", "thetroyharris"]\') }}'
     )
 }}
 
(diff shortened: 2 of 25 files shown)


thetroyharris:

thetroyharris:
recheck

models/balancer/arbitrum/balancer_v2_arbitrum_pools_fees.sql
```diff
@@ -9,36 +10,38 @@
...
 SELECT
+    'arbitrum' AS blockchain,
     logs.contract_address,
     logs.tx_hash,
     logs.tx_index,
     logs.index,
     logs.block_time,
     logs.block_number,
-    bytea2numeric_v3 (SUBSTRING(logs.data FROM 32 FOR 64)) * 1 AS swap_fee_percentage
+    bytea2numeric(bytearray_ltrim(logs.data)) AS swap_fee_percentage
```
jeff-dude, line 37: `byte2numeric` should no longer be needed on dunesql. you can find a good replacement here:
https://dune.com/docs/query/DuneSQL-reference/Functions-and-operators/varbinary/

models/balancer/arbitrum/balancer_v2_arbitrum_pools_fees.sql
```diff
@@ -9,36 +10,38 @@
...
 FROM
     {{ source ('arbitrum', 'logs') }}
     INNER JOIN registered_pools ON registered_pools.pool_address = logs.contract_address
-WHERE logs.topic1 = '{{ event_signature }}'
+WHERE CAST(logs.topic0 AS VARCHAR) = '{{ event_signature }}'
```
jeff-dude, line 41: rather than cast to varchar, you should be able to leave field as-is and use `{{ event_signature }}` without single quotes

models/balancer/arbitrum/balancer_v2_arbitrum_pools_fees.sql
```diff
@@ -9,36 +10,38 @@
...
     {% if not is_incremental() %}
-    AND logs.block_time >= '{{ project_start_date }}'
+    AND logs.block_time >= CAST('{{ project_start_date }}' AS TIMESTAMP)
```
jeff-dude, line 43: the preferred approach here:
`TIMESTAMP '{{ project_start_date }}` rather than casting

rule of thumb:
- if using a string literal value hardcoded, use datatype in front (i.e. `UINT256 '0'` or `TIMESTAMP '<date>'`
- if a column, `CAST()` is preferred

models/balancer/arbitrum/balancer_v2_arbitrum_transfers_bpt.sql
```diff
@@ -23,27 +24,28 @@ WITH registered_pools AS (
...
-        CONCAT('0x', SUBSTRING(logs.topic2, 27, 40)) AS from,
-        CONCAT('0x', SUBSTRING(logs.topic3, 27, 40)) AS to,
-        bytea2numeric(SUBSTRING(logs.data, 32, 64)) AS value
+        bytearray_substring(topic1, 13) AS "from",
+        bytearray_substring(topic2, 13) AS to,
+        bytearray_to_uint256(logs.data) AS value
     FROM {{ source('arbitrum', 'logs') }} logs
     INNER JOIN registered_pools p ON p.pool_address = logs.contract_address
-    WHERE logs.topic1 = '{{ event_signature }}'
+    WHERE CAST(logs.topic0 AS VARCHAR) = '{{ event_signature }}'
```
jeff-dude, line 45: similar idea here, no need to cast

models/balancer/arbitrum/balancer_v2_arbitrum_transfers_bpt.sql
```diff
@@ -23,27 +24,28 @@ WITH registered_pools AS (
...
         {% if not is_incremental() %}
-        AND logs.block_time >= '{{ project_start_date }}'
+        AND logs.block_time >= CAST('{{ project_start_date }}' AS TIMESTAMP)
```
jeff-dude, line 47: similar idea here, on preferred cast method

jeff-dude (changes requested):

models/balancer/arbitrum/balancer_v2_arbitrum_pools_fees.sql
thetroyharris: Perfect, I have changed it. However, I end up casting it to double to keep the original data type.

models/balancer/arbitrum/balancer_v2_arbitrum_pools_fees.sql
thetroyharris: Appreciate you letting me in on the rule of thumb, I will remember that.

models/balancer/arbitrum/balancer_v2_arbitrum_pools_fees.sql
thetroyharris: This is good to know moving forward. Thank you.

models/balancer/arbitrum/balancer_arbitrum_schema.yml
```diff
@@ -68,54 +69,32 @@ models:
             - evt_tx_hash
             - evt_index
             - block_date
-      - check_seed_legacy:
```
jeff-dude, line 71: do you have any particular reasoning to remove tests? both here and the unique ones?

i'm okay with the removal ultimately, but curious the reasoning behind it. if it's difficulty migrating the test, we have a process down to help migrate seeds and tests.

if you ultimately want to remove this still, i'd say go ahead and remove the seed file itself as well

models/balancer/arbitrum/balancer_v2_arbitrum_transfers_bpt.sql
```diff
@@ -1,6 +1,7 @@
 {{
     config(
-        schema='balancer_v2_arbitrum',
+        schema = 'balancer_v2_arbitrum',
+        tags = ['dunesql'],
         alias = alias('transfers_bpt'),
         partition_by = ['block_date'],
```
jeff-dude, line 6: we're looking to increase our partition size -- can you add a `block_month` column to the model, just like `block_date`, then update value here?

note, this would apply anywhere you partition. you could also explore removing partition altogether if the dataset isn't v large

models/balancer/arbitrum/balancer_arbitrum_schema.yml
thetroyharris: The removal of the test was due to the difficulty migrating the tests and addressing the errors returned durning the migrate process. What is the process you all have?

jeff-dude: since all blockchain spells are in this PR, you'll be safe just to add here in same PR. in other scenarios, if migrating just one blockchain for instance, it'd be easier to do in separate PR to avoid seed tests running on chains not included. anyway, with that in mind, you can add a schema file for the seed in the same directory, like [here](https://github.com/duneanalytics/spellbook/pull/3886/files#diff-ed19154a4e0a755b19a400b89b091cd7b2c4b3a18d39969a9dbe3d913f1bcda1). in this file, you can force columns to a certain datatype to match your spell datatypes, most importantly the join condition ones since the join would fail if they differ

then you can force a change to your seed file itself, so it appears in the PR and builds in CI process.

the check seed macro * should * be fine as-is, you can leave that out for now

PR #4008, opened 2023-08-10
T-32073 - Conditional property for join_distribution_type = PARTITIONED

belen-pruvost:
Follow up from [this discussion](https://dune.height.app/T-32073#e2b19a18-daac-4412-bd54-b4537097a994)

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -49,9 +49,6 @@ seeds:
 # Full documentation: https://docs.getdbt.com/docs/configuring-models
 models:
   spellbook:
-    +pre-hook:
-      - sql: "{{ set_trino_session_property(is_materialized(model) and is_partitioned(model), 'join_distribution_type', 'PARTITIONED') }}"
-        transaction: true
     +post-hook:
       - sql: "{{ set_trino_session_property(is_materialized(model), 'writer_min_size', model.config.get('writer_min_size', '500MB')) }}"
         transaction: true
--- a/macros/set_trino_session_property.sql
+++ b/macros/set_trino_session_property.sql
@@ -8,7 +8,3 @@
 {%- macro is_materialized(model) -%}
   {% do return(model.config.materialized in ('table', 'incremental')) %}
 {%- endmacro -%}
-
-{%- macro is_partitioned(model) -%}
-  {% do return(model.config.materialized in ('table', 'incremental') and model.config.partition_by and model.config.partition_by|length > 0) %}
-{%- endmacro -%}
--- a/macros/enforce_join_distribution.sql
+++ b/macros/enforce_join_distribution.sql
@@ -1,0 +1,3 @@
+{% macro enforce_join_distribution(value) %}
+  {{ set_trino_session_property(true, 'join_distribution_type', value) }}
+{% endmacro %}
--- a/models/nft/chains/nft_polygon_transfers.sql
+++ b/models/nft/chains/nft_polygon_transfers.sql
@@ -5,7 +5,8 @@
         partition_by=['block_month'],
         materialized='incremental',
         file_format = 'delta',
-        unique_key = ['unique_transfer_id']
+        unique_key = ['unique_transfer_id'],
+        pre_hook='{{ enforce_join_distribution("PARTITIONED") }}',
 )
 }}
 
--- a/models/orca_whirlpool/orca_whirlpool_trades.sql
+++ b/models/orca_whirlpool/orca_whirlpool_trades.sql
@@ -8,6 +8,7 @@
         file_format = 'delta',
         incremental_strategy = 'merge',
         unique_key = ['tx_id', 'outer_instruction_index', 'inner_instruction_index', 'tx_index','block_month'],
+        pre_hook='{{ enforce_join_distribution("PARTITIONED") }}',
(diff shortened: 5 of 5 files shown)


belen-pruvost:
I believe this model will likely cause all models to be affected

dbt_project.yml
```diff
@@ -50,7 +50,7 @@ seeds:
 models:
   spellbook:
     +pre-hook:
-      - sql: "{{ set_trino_session_property(is_materialized(model) and is_partitioned(model), 'join_distribution_type', 'PARTITIONED') }}"
+      - sql: "{{ set_trino_session_property(is_materialized(model) and is_partitioned_enforced(model), 'join_distribution_type', 'PARTITIONED') }}"
```
jkylling, line 53: Nit: Could we make this more flexible and allow for configuring any join distribution type on the model. E.g.,
```
{{ set_trino_session_property(model.config.get('join_distribution_type', false), 'join_distribution_type', model.config.get('join_distribution_type'))) }}
```

jkylling: Or add a new macro to support a simpler syntax.

belen-pruvost: Yeah - this makes more sense 👍 I've opened this property to all of the property's possible values.

jeff-dude:
manually canceling CI, as it kicks off large subset (should do that each commit on these types of PRs)

macros/set_trino_session_property.sql
```diff
@@ -9,6 +9,6 @@
   {% do return(model.config.materialized in ('table', 'incremental')) %}
 {%- endmacro -%}
 
-{%- macro is_partitioned(model) -%}
-  {% do return(model.config.materialized in ('table', 'incremental') and model.config.partition_by and model.config.partition_by|length > 0) %}
+{%- macro is_join_distribution_enforced(model) -%}
+  {% do return(model.config.join_distribution_type in ('AUTOMATIC', 'PARTITIONED', 'BROADCAST')) %}
```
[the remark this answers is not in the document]
belen-pruvost: We leave the option open for the future to use different partitioning types.
However, for now, you would just need to set `join_distribution_type='PARTITIONED'` to the models that result into `EXCEEDED_LOCAL_MEMORY_LIMIT` (like `nft_polygon.transfers`)

belen-pruvost:
> manually canceling CI, as it kicks off large subset (should do that each commit on these types of PRs)

I still need to test if this is setting the session property as expected, I will limit the model run step to just this model then. Are there other options otherwise?

And besides that, for new properties on config (in this case `join_distribution_type` set for nft_polygon.transfers), do I need to define them anywhere else?

dbt_project.yml
```diff
@@ -50,7 +50,7 @@ seeds:
 models:
   spellbook:
     +pre-hook:
-      - sql: "{{ set_trino_session_property(is_materialized(model) and is_partitioned(model), 'join_distribution_type', 'PARTITIONED') }}"
+      - sql: "{{ set_trino_session_property(is_join_distribution_enforced(model), 'join_distribution_type', model.config.get('join_distribution_type', 'AUTOMATIC')) }}"
```
jkylling, line 53: We don't need a fallback as we never use it.
```suggestion
      - sql: "{{ set_trino_session_property(is_join_distribution_enforced(model), 'join_distribution_type', model.config.get('join_distribution_type')) }}"
```

macros/set_trino_session_property.sql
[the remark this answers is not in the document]
belen-pruvost: We can, yes 👍 I'll change this

belen-pruvost:
Besides the fallback feedback, which I'll address, no property is being set on the model currently so I believe the new model property is perhaps not correctly configured. Any help is appreciated 😅 
Screenshot 2023-08-10 at 16 27 48

belen-pruvost:
I've tested this locally and the session property is correctly set for nft_polygon_transfers (I added some logs to prove it)
Screenshot 2023-08-11 at 15 20 13

The CI also works fine  (by commenting out the check tags step). I will remove that `do not merge` commit

However, since this removes a prehook it implies an update on all dbt models. Not sure if there is something that needs to be done before merging the PR.

@jeff-dude + @couralex6 let me know what you think 🙏

aalan3 (approved):

belen-pruvost:
Nice - thanks @aalan3 🙌

jeff-dude:
this approach works for me. do we have others we want to apply to now too? there is a solana spell under orca_whirpool that we focused on last time too

jeff-dude:
let's pay attention to #4055 here too (merge timing and when to change pre-hook there)

belen-pruvost:
> this approach works for me. do we have others we want to apply to now too? there is a solana spell under orca_whirpool that we focused on last time too

Good call out - I'll add the prehook to this model too. Regarding https://github.com/duneanalytics/spellbook/pull/4055 I think we could go in sequence:
1) Merge this PR
2) Rebase https://github.com/duneanalytics/spellbook/pull/4055
3) If we get an exceeded local memory error on the sharky models, add the prehook

jeff-dude (approved):
sounds good, i'll look to prep prod for this merge and get it going

PR #4036, opened 2023-08-10
Base - add token prices

RantumBits:
add token prices on Base chain


models/prices/base/prices_base_tokens.sql
```diff
@@ -0,0 +1,24 @@
...
+    , symbol
+    , contract_address
+    , decimals
+FROM
+(
+    VALUES
+
+--    ('weth-weth', 'base', 'WETH', 0x4200000000000000000000000000000000000006, 18), --requested add to coinPaprika 2023-08-10
+--    ('axl-axelar', 'base', 'AXL', 0x467719aD09025FcC6cF6F8311755809d45a5E5f3, 6), --requested add to coinPaprika 2023-08-10
+    ('bald-bald', 'base', 'BALD', 0x27D2DECb4bFC9C76F0309b8E88dec3a601Fe25a8, 18)
```
Hosuke, line 21: SCAN: https://base.blockscout.com/address/0x27D2DECb4bFC9C76F0309b8E88dec3a601Fe25a8
API: https://api.coinpaprika.com/v1/coins/bald-bald

Hosuke (approved):
LGTM.✅
Thank you @RantumBits 

Not sure if we need to make legacy spell as select 1.

models/prices/base/prices_base_tokens.sql
Hosuke: However currently base chain is not supported in coinpaprika yet:
https://api.coinpaprika.com/v1/contracts/

We may need to wait for a while.

jeff-dude: i think it's fine as long as the token_id returns a response and we manually assign address here

jeff-dude: note that the token is invalid on the API though, so we won't get ongoing pricing data

jeff-dude:
i added to `prices_tokens` cross-chain spell & since this full lineage is running on dunesql, i overrode the legacy to be a dummy file

jeff-dude (approved):

PR #4050, opened 2023-08-11
experimental: Adapt nft.transfers to support denormalization

aalan3:
Since we now have `tx_from` on base decoded tables I wanted to try this and see the speedup we can get.


jeff-dude (approved):
nice. we can then continue to add blockchains to the `in` statement as more denormalize, then eventually remove altogether and finalize across all blockchains

fyi @0xRobin

PR #4055, opened 2023-08-12
chore: use sharky decoded tables

ennnas:
PR to move the sharky spell to decoded tables
reference: https://github.com/duneanalytics/spellbook/pull/3895#issuecomment-1665810110


jeff-dude:
thanks again for writing on decoded tables -- this looks clean!

i pushed a commit to clean up just a little bit:
- move schema definition to models themselves (this is a longer term goal of ours to cleanup project file, just doing it while i'm here)
- add column aliases on the join between base table and decoded CTEs
- add incremental filter logic on the base table

models/sharky/solana/sharky_solana_events.sql
```diff
@@ -35,107 +30,333 @@ WITH sharky_txs AS (
...
+            call_block_time,
+            call_tx_signer,
+            call_tx_id,
+            account_loan,
+            'Offer' AS evt_type,
+            account_escrow,
+            account_orderBook,
+            NULL AS account_collateralMint,
+            ARRAY[CAST(ROW(call_data, call_account_arguments) AS ROW(data VARBINARY, account_arguments ARRAY<VARCHAR>))] AS sharky_instructions,
+            CAST(principalLamports AS BIGINT) AS amount_raw
```
jeff-dude, line 51: suggestion: use dunesql datatype `uint256` instead? or is that just an evm concept and solana would prefer bigint?

this would apply to raw amounts throughout, just an open question

models/sharky/solana/sharky_solana_events.sql
```diff
@@ -35,107 +30,333 @@ WITH sharky_txs AS (
...
+    'solana' AS blockchain,
+    'sharky' AS project,
+    call_tx_id AS tx_hash,
+    call_block_date AS block_date,
+    CAST(date_trunc('month', call_block_date) AS DATE) AS block_month,
+    call_block_time AS block_time,
+    call_block_slot AS block_number,
+    (amount_raw / 1e9) * p.price AS amount_usd,
+    amount_raw / 1e9 AS amount_original,
+    CAST(amount_raw AS DECIMAL(38, 0)) AS amount_raw,
```
jeff-dude, line 353: similar to question above, but this time it's decimal here?

jeff-dude:
one other note:
we're actively changing the pre-hook approach here #4008 

we will likely merge that monday, then we would want to apply the approach from the model config in that PR here too since solana has large data volumes

models/sharky/solana/sharky_solana_events.sql
ennnas: if it's a smaller data type we can consider using `uint256` here

models/sharky/solana/sharky_solana_events.sql
ennnas: here we can't use `uint256` because `amount_raw` might be negative for example when extending loans with a `new loan amount < old loan amount`

models/sharky/solana/sharky_solana_events.sql
jeff-dude: there is also `int256`
https://dune.com/docs/query/DuneSQL-reference/datatypes/#int256-dune-sql

again -- just a suggestion, as this is a standalone spell. if in the future it rolls into a sector-level spell, we would look to standardize

ennnas: I would keep it like this for now and if a best practice arises I will take care of updating it to follow them 👍🏾

ennnas: I would keep it like this for now and if a best practice arises I will take care of updating it to follow them 👍🏾

models/sharky/solana/sharky_solana_events.sql
```diff
@@ -35,107 +30,333 @@ WITH sharky_txs AS (
...
+       events.*,
+       CASE
+            -- For take and rescind instructions we can take the exact amount available in the escrow account
+            WHEN (
+                    events.evt_type = 'Take'
+                    OR events.evt_type = 'Rescind'
+                ) THEN abs(st.post_balances[array_position(st.account_keys, events.account_escrow)] - st.pre_balances[array_position(st.account_keys, events.account_escrow)] )
+            ELSE (abs(st.post_balances[1] - st.pre_balances[1]))
+       END AS amount_raw
+    FROM  {{ source('solana','transactions') }} st
```
jkylling, line 319: Is it possible to avoid the join with solana.transactions here? Why is it needed?

ennnas: We need to compute the balance change of either the tx signer or the escrow account

models/sharky/solana/sharky_solana_events.sql
jeff-dude: sounds good!

jeff-dude:
looking good after the latest config applied!

jeff-dude (approved):

PR #4073, opened 2023-08-14
Create a optimized version of prices_usd_latest

aalan3:
This version should be faster than what currently exists.


models/prices/prices_usd_latest_old.sql
```diff
@@ -0,0 +1,18 @@
+{{ config(
```
jkylling, line 1: Can this be made incremental?

aalan3: Yeah I guess, it would make the runs a bit faster.

models/prices/prices_usd_latest_old.sql
```diff
@@ -0,0 +1,22 @@
...
+SELECT
+  pu.blockchain
+, pu.contract_address
+, pu.decimals
+, pu.symbol
+, max(pu.minute) as minute
+, max_by(pu.price, pu.minute) as price
+FROM {{ source('prices', 'usd') }} pu
+{% if is_incremental() %}
+    WHERE minute >= date_trunc('day', now() - interval '2' day)
```
jkylling, line 20: How does this work? If it's an incremental run, then only compute this subset, and try to merge it into the table? How does it know to use blockchain, contract_address, decimals and symbol as a key, and minute, price as a value?

aalan3: was a bit too quick here, I forgot to add `unique_key = [...]` in the config

models/prices/prices_usd_latest_old.sql
```diff
@@ -0,0 +1,23 @@
+{{ config(
+        schema='prices',
+        alias = alias('usd_latest_old'),
+        tags= ['static', 'dunesql'],
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'merge',
+        unique_key = ['blockchain','contract_address','decimals','symbol']
```
aalan3, line 8: @jkylling by omitting `price` and `minute` this should update with the latest minute every time and the table should have one row per combination.

jkylling (approved):
Have you tested the runtime of this vs. the current version?

models/prices/prices_usd_latest_old.sql
```diff
@@ -0,0 +1,23 @@
+{{ config(
+        schema='prices',
+        alias = alias('usd_latest_old'),
+        tags= ['static', 'dunesql'],
```
jeff-dude, line 4: if we're doing incremental, i would remove `static` here

models/prices/prices_usd_latest_old.sql
jeff-dude: just `blockchain` and `contract_address` should be enough here, as we enforce unique on `contract_address` for each blockchain-specific spell, then when we go cross-chain, `blockchain` will help differentiate

models/prices/prices_usd_latest_old.sql
jeff-dude: yep, this is right

models/prices/prices_usd_latest_old.sql
```diff
@@ -0,0 +1,23 @@
+{{ config(
+        schema='prices',
+        alias = alias('usd_latest_old'),
```
jeff-dude, line 3: nit: can we change `old` to `historical` on both file name / spell table name? the `_old` on a file / table make me think "this isn't a relevant table anymore"

models/prices/prices_usd_latest_old.sql
aalan3: Updated.

models/prices/prices_usd_latest_old.sql
aalan3: for native tokens these can be null which would cause issues in that case?

models/prices/prices_usd_latest_historical.sql
```diff
@@ -0,0 +1,23 @@
+{{ config(
+        schema='prices',
+        alias = alias('usd_latest_old'),
```
jeff-dude, line 3: this alias needs changed too

models/prices/prices_usd_latest_old.sql
jeff-dude: fair, it doesn't hurt to keep

jeff-dude (approved):
looks good now 🚀

PR #4076, opened 2023-08-14
Dex Aggregator Migration - #4 - Bebop - ready for review

henrystats:


henrystats:
@jeff-dude I couldn't find the original creator of this spell, I'm trying to tag here but I'm not sure it's the same person...

I was migrating this spell to dune-sql and I made a small change and introduced a new column (trade type)...

Previously.. when there's a multi-trade transaction, the side of the trade that only occurs once is divided equally into the other side... 

This is then used for the amount_usd column, but I suspect that it's breaking somewhere and sometimes it's using the full amount (for amount_usd) in the deduped trades which inflates the volume... 

I adjusted the query so it only uses the amount_usd of the deduped trades and it keeps the original value of the single token traded on the other side of the trade (so that isn't divided)... 

And then for simple straight swaps, we can default back to our use of coalesce(token_bought, token_sold)... 


I wanted to discuss this change as I suspect that there would be other models that allow multi transactions and since most people using the dex_aggregator.trades table will be using the amount_usd column it'd be best to not have that inflated...


The trade_type column would also help to differentiate...

jeff-dude:
> @jeff-dude I couldn't find the original creator of this spell, I'm trying to tag here but I'm not sure it's the same person...
> 
> I was migrating this spell to dune-sql and I made a small change and introduced a new column (trade type)...
> 
> Previously.. when there's a multi-trade transaction, the side of the trade that only occurs once is divided equally into the other side...
> 
> This is then used for the amount_usd column, but I suspect that it's breaking somewhere and sometimes it's using the full amount (for amount_usd) in the deduped trades which inflates the volume...
> 
> I adjusted the query so it only uses the amount_usd of the deduped trades and it keeps the original value of the single token traded on the other side of the trade (so that isn't divided)...
> 
> And then for simple straight swaps, we can default back to our use of coalesce(token_bought, token_sold)...
> 
> I wanted to discuss this change as I suspect that there would be other models that allow multi transactions and since most people using the dex_aggregator.trades table will be using the amount_usd column it'd be best to not have that inflated...
> 
> The trade_type column would also help to differentiate...

thanks for the info. the original creator of bebop spells is @B1boid 

@B1boid do you have any input on the above comment, with changes made during migration to dunesql?

jeff-dude:
@henrystats if you want to fix the seed file here, i think we can safely move forward and merge

henrystats:
yes sure ser, fixed now @jeff-dude thank you!

jeff-dude:
> yes sure ser, fixed now @jeff-dude thank you!

looks like maybe a bug introduced?

henrystats:
yep... weird, looks like it's from the oneinch spells? that's where the split functions were used... but the oneinchspells were running just fine earlier so I'm confused why that would be causing issues now.

henrystats:
@jeff-dude, SELECT * FROM oneinch_ethereum.trades on dune.com says that there is some issue somewhere in the UNION... 

Could it be that all the models haven't ran yet in dunesql?

I'm working through it now and will submit a fix pr, as there might be some other issues too.

henrystats:
soo dbt compile works, for the first table causing issues on the join... the onesplit table. I think it also has 24 columns instead of 25 (which I'm assuming it's because we added block_month)...

I tested the compiled version of the query on dune.com and it returns accurate results so I'm guessing that the issue here is that everything hasn't run in prod yet? do I have to do anything?

jeff-dude:
> soo dbt compile works, for the first table causing issues on the join... the onesplit table. I think it also has 24 columns instead of 25 (which I'm assuming it's because we added block_month)...
> 
> I tested the compiled version of the query on dune.com and it returns accurate results so I'm guessing that the issue here is that everything hasn't run in prod yet? do I have to do anything?

1inch was delayed in prod, but it's done now. i just ran:
`SELECT * FROM oneinch_ethereum.trades limit 10`
and it returned successfully

henrystats:
thank you ser @jeff-dude... I'm still unable to figure out what the issue here is.

`            AND COALESCE(CAST(split(seed.{{column_name}}, ',') as array<bigint>), ARRAY[]) = model_sample.{{column_name}}` this is the line in check_macro seed that uses the split command. 

The function is used on seed.trace_address.. I believe a pr was merged recently to update the check_macro test helper so I wonder if that is what is causing it to break now?

jeff-dude (approved):

PR #4083, opened 2023-08-15
Migrate NFT opensea models

0xRobin:
Migrate NFT opensea models


0xRobin:

0xRobin:
recheck

0xRobin:
all non-seaport models passing :heavy_check_mark:

models/opensea/ethereum/opensea_v4_ethereum_events.sql
```diff
@@ -1,558 +1,45 @@
 {{ config(
     schema = 'opensea_v4_ethereum',
     alias = alias('events'),
-    partition_by = ['block_date'],
+    tags = ['dunesql'],
     materialized = 'incremental',
     file_format = 'delta',
     incremental_strategy = 'merge',
     unique_key = ['block_date', 'unique_trade_id']
```
jeff-dude, line 8: looks like unique keys differ here compared to arbitrum version. is that temporary until the field is fully deprecated later?

jeff-dude, line 8: looks like unique keys differ here compared to arbitrum version. is that temporary until the field is fully deprecated later?

macros/models/_sector/nft/platforms/seaport_v3_trades.sql
```diff
@@ -0,0 +1,724 @@
...
+            ,a.consideration_cnt
+            ,a.order_hash
+            ,a.is_private
+            ,a.is_self_trans
+            ,a.is_platform_fee
+            ,a.eth_erc_idx
+            ,a.nft_cnt
+            ,a.erc721_cnt
+            ,a.erc1155_cnt
+            ,date_trunc('day', a.block_time) as block_date
```
jeff-dude, line 328: minor: cast this to date? otherwise, it'll pull through timestamp type from `block_time`

jeff-dude: (i'll only comment on this instance)

jeff-dude (commented):
minor comments below, otherwise looking great 🚀

models/opensea/ethereum/opensea_v4_ethereum_events.sql
0xRobin: That's not intentional, it was copied over from the legacy files.
But good catch, we should change it while we're in here. :+1:

macros/models/_sector/nft/platforms/seaport_v3_trades.sql
0xRobin: There's a lot of these still lingering in the NFT dependency tree.
They should indeed all be cast to date. :+1:

jeff-dude (approved):

PR #4091, opened 2023-08-16
Replace UNION with UNION ALL in transfers_bnb.bep20

jkylling:
The rows are already all distinct.
We should also consider dropping the unique_transfer_id.

### DIFF
--- a/models/transfers/bnb/bep20/transfers_bnb_bep20_legacy.sql
+++ b/models/transfers/bnb/bep20/transfers_bnb_bep20_legacy.sql
@@ -53,15 +53,15 @@ with
         from
             {{ source('bnb_bnb', 'WBNB_evt_Withdrawal') }}
     )
-    
+
 select unique_transfer_id, 'bnb' as blockchain, wallet_address, token_address, evt_block_time, CAST(amount_raw AS VARCHAR(100)) as amount_raw
 from sent_transfers
-union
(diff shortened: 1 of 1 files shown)


aalan3 (approved):

models/transfers/bnb/bep20/transfers_bnb_bep20.sql
```diff
@@ -1,4 +1,5 @@
 {{ config(materialized='view', alias = alias('bep20'),
+        tags = ['dunesql'],
```
jeff-dude, line 2: if we go down this path, we should for sure enhance the whole model

jeff-dude:
if we want immediate merge, changes need made on legacy file. if you complete the migration to dunesql, we should consider enhancing this one & all other blockchain versions (same logic)

models/transfers/bnb/bep20/transfers_bnb_bep20.sql
aalan3: what does enhance mean here?

aalan3:
to keep it simple I'll just do the legacy updates here. We can migrate the rest later. I also want to get rid of the unique transfer id here, not sure why it's there in the first place since this isn't being materialized.

jeff-dude (approved):

PR #4092, opened 2023-08-16
Change UNION to UNION ALL in transfers models

jkylling:
From what I can tell all rows are already distinct, so the UNION can safely be replaced by UNION ALL.


aalan3 (approved):

aalan3:
Updated this to be for legacy models transfers. Also updated base (non-legacy) to use `union all` as well

jeff-dude (approved):

Issue #4089, opened 2023-08-16
[BUG] `prices.usd_latest` showing dupe prices for native tokens


nazihkalo:
### Description
`prices.usd_latest` showing dupe prices for native tokens

### Current behavior
[Query Link](https://dune.com/queries/2860854)
<img width="930" alt="image" src="https://github.com/duneanalytics/spellbook/assets/49926988/a523765e-5cd2-4a34-bc11-8da2d3e8600b">

### Expected behavior
Most queries assume that there should be at most 1 row per symbol `where blockchain is null`. Seems like a dbt model bug

### Impacted model(s)
```
prices.usd_latest
```

### Possible solution
Revert this PR https://github.com/duneanalytics/spellbook/pull/4073 

@jeff-dude @aalan3

nazihkalo:
OK I think I've identified the root cause. 

This addition of  incremental model here https://github.com/duneanalytics/spellbook/pull/4073/commits/4306d9d9217c5253ae97adbfb7367cfce869233b
<img width="909" alt="image" src="https://github.com/duneanalytics/spellbook/assets/49926988/cded0db0-19ec-4434-b29d-f9acbbb7425b">

uses `unique_key = ['blockchain','contract_address','decimals','symbol']` as the incremental key, but according to this dbt issue ->
https://github.com/dbt-labs/dbt-core/issues/5691#issuecomment-1222358247

> In cases where multiple columns are needed to uniquely identify a record (= unique_key is a list), none of those columns can be null. The equality check for upserts / merges doesn't have handling for null values built in—nor do I think it should, since this obfuscates more logic in an already difficult-to-debug place (materialization logic, rather than transformation logic).
> If there's any risk of one or more columns being null, you're better off creating a single surrogate (hash) key. The [dbt_utils.surrogate_key](https://github.com/dbt-labs/dbt-utils#surrogate_key-source) macro handles null values, and will return the same hashed result for the same null input(s).

aalan3:
Hello, thank you for reporting this! I will look into it and fix it.

PR #4094, opened 2023-08-16
Fix duplicates by adding unique key using surrogate_key

aalan3:
Fixes https://github.com/duneanalytics/spellbook/issues/4089


jkylling (approved):
This will do a merge into on the unique_key? Is there a way for dbt to do the merge into statement on multiple columns?

aalan3:
Yes it will use the unique key to merge into. We can add multiple columns next to `unique_key` to merge into on that as well (like we previously did). But that doesn't seem to handle nulls properly.

PR #4096, opened 2023-08-16
DuneSQL Migration: Balancer `transfers_bpt` &  `pools_fees`

thetroyharris:


thetroyharris:
Moving PR #3996 to this PR (#4096) @jeff-dude, apologies for the inconvenience.

thetroyharris:
@jeff-dude

Any idea on this error: https://github.com/duneanalytics/spellbook/actions/runs/5881213344/job/15949223006?pr=4096#step:14:192

Seems to be an issue with naming the column "from" however I am not sure where specifically to address this issue or how to resolve it.

models/balancer/arbitrum/balancer_arbitrum_schema.yml
```diff
@@ -82,6 +83,9 @@ models:
             - from
```
jeff-dude, line 83: it is likely this `from` being passed in to the test for comparison. on dunesql, it needs to be `"from"` with double quotes, or it looks for the normal `from` sql syntax

first thing to try, add double quotes directly here?
if that doesn't work, we can look to edit the test to handle adding double quotes if column = from

thetroyharris: I added double quotes to the from on in the schema file and the seed file column name. It returned the same error. See  commit here: https://github.com/duneanalytics/spellbook/pull/4096/commits/e66754b87f47d3082288d2a7c40a0539f0600389

jeff-dude: okay, let me pull your branch and give something a shot

jeff-dude: we're on the right path here. looks like the seed file itself needs modified to use actual raw values, rather than the current format, so we can take advantage of the large int data types (`uint256` in the schema here right now)

can you try to fix that in the seed?

thetroyharris:
This PR has moved to #4113

PR #4105, opened 2023-08-17
Use writer_scaling_min_data_processed instead of writer_min_size after 423 upgrade

jkylling:
writer_min_size is removed and replaced by writer_scaling_min_data_processed. Since we are doing multiple version bumps it does not look like we can do a graceful upgrade.

### DIFF
--- a/dbt_project.yml
+++ b/dbt_project.yml
@@ -50,7 +50,7 @@ seeds:
 models:
   spellbook:
     +post-hook:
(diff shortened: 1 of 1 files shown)


aalan3 (approved):
are there any practical differences we should consider?

jkylling:
> are there any practical differences we should consider?

We might need to tweak the configuration a bit, as there are improvements to the optimize algorithm in 423. But the current configuration should be safe to use.

PR #4107, opened 2023-08-17
Dex trades beta dunesql

aalan3:
This is to test dex trades under the alias `trades_migration_beta` to make sure that everything builds correctly and to check the data.


models/dex/dex_trades_migration_beta.sql
```diff
@@ -0,0 +1,133 @@
+{{ config(
+        tags=['dunesql'],
+        alias = alias('trades_migration_beta'),
+        partition_by = ['block_date'],
```
jeff-dude, line 4: `block_month` here to match pattern of upstream spells?

jeff-dude (approved):
good to go outside of partition question

aalan3:
Good catch changed to `block_month` now.

antonio-mendes (approved):
Beautiful, looks good to me too

aalan3:
This can't be merged until `block_time` exists on all models that are being referenced. I will create some PRs to fix this.

couralex6 (approved):

aalan3:
I figured out the issue! The incremental part of statement was incorrectly formated due to the loop, so the last model didn't have the time filter (fixed in https://github.com/duneanalytics/spellbook/pull/4107/commits/b29627f1ddc32560733d47a527d22f1aea19c189) . This was not a problem before but since I added the `incremental_predicate`, the last model was inserting the entire model which caused duplicates because the destination has a time filter now.

jeff-dude (approved):

PR #4119, opened 2023-08-18
Uniswap block month partitioning

aalan3:
This PR is a requirement for https://github.com/duneanalytics/spellbook/pull/4107.

I've done the following for the models
* Partition by `block_month`
* Added `CAST(date_trunc('month', dexs.block_time) AS date) AS block_month` to each model
* Added `block_month` to the yaml files
* Changed so that the models variables are defined using `ref` (this makes it possible to cmd + click in VSCode)
* There are a bunch of whitespace changes which is my editor removing it automatically

### DIFF
--- a/models/uniswap/arbitrum/uniswap_arbitrum_schema.yml
+++ b/models/uniswap/arbitrum/uniswap_arbitrum_schema.yml
@@ -28,12 +28,15 @@ models:
       - &blockchain
         name: blockchain
         description: "Blockchain which the DEX is deployed"
-      - &project 
+      - &project
         name: project
-        description: "Project name of the DEX"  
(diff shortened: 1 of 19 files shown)


jeff-dude (approved):
works for me!

side note: we left `try_cast` alone for now, but post-migration planned on universal change to `cast`, but obviously fine to do here now

aalan3:
I think we can drop `cast` all together but yeah for another time

jeff-dude:
> I think we can drop `cast` all together but yeah for another time

i don't think so, based on our testing. `date_trunc` on trino doesn't auto-cast, it takes from the field you pass in (which is timestamp almost every time), so it'd output timestamp

PR #4128, opened 2023-08-19
OpenSea v3 (& v4) DuneSQL `nft.trades`

hildobby:
From query here: https://dune.com/queries/2926594
Likely not perfect but seems fairly efficient.

Seaport model could also be split apart from this spell (and have it be dependant on it), but it would be cool to get the migration done asap and worry about this later!

This is Ethereum only but can be turned into a macros and have all chains since logic is the same across the board


hildobby:
Made the `seaport_{chains}.traces` tables with macros here: https://github.com/duneanalytics/spellbook/pull/4129.

Once that one is merged I can merge main and do the same with macros for OpenSea on Seaport

jeff-dude:
closing due to inactivity, feel free to reopen as needed 🤝

PR #4129, opened 2023-08-20
Create `seaport_{chain}.traces` with macros

hildobby:

### DIFF
--- a/macros/models/_project/seaport/seaport_traces.sql
+++ b/macros/models/_project/seaport/seaport_traces.sql
@@ -1,0 +1,96 @@
+{% macro nft_transfers(blockchain, seaport_events) %}
+{%- set token_standard_start = 'bep' if blockchain == 'bnb' else 'erc' -%}
+{%- set spark_mode = True -%} {# TODO: Potential bug. Consider disabling #}
+
+WITH base_data AS (
+     SELECT evt_block_time AS block_time
+    , evt_block_number AS block_number
+    , orderHash AS order_hash
+    , offerer
+    , recipient
+    , evt_tx_hash AS tx_hash
+    , contract_address AS seaport_contract_address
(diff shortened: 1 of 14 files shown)


0xRobin (commented):
Not sure there's the need to materialize and expose the traces as they basically only unnest the offers and considerations columns in the order events.

Do you see a use for this outside of the nft trades models? If not we can keep the logic in the macro and just use the macro in the nft trades models.

hildobby:
> Not sure there's the need to materialize and expose the traces as they basically only unnest the offers and considerations columns in the order events.
> 
> Do you see a use for this outside of the nft trades models? If not we can keep the logic in the macro and just use the macro in the nft trades models.

Yeah and model is pretty fast to run anyways so I don't see much of an issue. I think other marketplaces using seaport might benefit from this spell but there's barely any of those afaik (and I can port blur and looksrare's seaport trades later for an easier read on the code there)

hildobby:
I'm also looking into OTC deals, a lot of which use seaport. I would love to get this merged, even as a view would be fine since the model runs very quickly anyways

0xRobin (approved):

jeff-dude (approved):
views work for me.

is there a reason `base` one is excluded with `prod_exclude` tag?

hildobby:
> views work for me.
> 
> is there a reason `base` one is excluded with `prod_exclude` tag?

Seaport on Base was decoded under the `opensea` name (instead of `seaport`), hence why it was failing and I excluded it. I just asked @antonio-mendes to change the namespace and removed the `prod_exclude` tag here, A simple `Update branch` after the change should get CI to pass tests hopefully and we should be good!

antonio-mendes:
The rename is done @hildobby 👍

PR #4130, opened 2023-08-20
Migrate Optimism Bridge Flows - Ready for review 

henrystats:


henrystats:
@chuxinh wanted to tag you to this as well incase you wanted to review... migrated the bridge flows spell and the spells that it relied on...

tested it on dune.com that it works correctly as well.

chuxinh (commented):
Some nits. Otherwise LGTM!

henrystats:
@jeff-dude the Wallet summary spell relies on this getting merged first so I could add bridged funds to it...

This bridge flows spell is ready for review / merging.

models/chain_info/chain_info_chain_ids.sql
```diff
@@ -9,13 +10,13 @@
...
-,`chain` AS chain_shortname
-,`nativeCurrency` AS native_currency_symbol
-,`chainId` AS chain_id
-,`networkId` AS network_id
-,`infoURL` AS info_url
-,`explorer` AS explorer
+"name" AS chain_name
+,"chain" AS chain_shortname
+,"nativeCurrency" AS native_currency_symbol
+,"chainId" AS chain_id
```
jeff-dude, line 16: would it make sense to cast this value to match the data type in `optimism.logs`, as it's joined to that base table in the other spell `optimism_standard_bridge_flows`? that way, you don't need to force cast in the join condition there?

jeff-dude (commented):
one minor comment below

models/chain_info/chain_info_chain_ids.sql
henrystats: it's manually imported as integer in optimism_standard_bridge_flows.... I guess I could just changed that

from for example 10 to '10'? 

in the only other model in bridge flows (hop optimism), it's also just selecting from the chain_id table so there's no existing model that has a specific data type from optimism.logs...

henrystats: @jeff-dude

jeff-dude: i'm thinking if it's int [here](https://github.com/duneanalytics/spellbook/blob/563ff62511d692b1f31ad1858bcca249736f3c7e/models/bridge/optimism/optimism_standard_bridge_flows.sql#L49-L50), then it should be int [here](https://github.com/duneanalytics/spellbook/blob/563ff62511d692b1f31ad1858bcca249736f3c7e/models/chain_info/chain_info_chain_ids.sql#L16) too, i.e. remove the quotes and it should do automatically (or can cast in the select to be safe)

jeff-dude: seems misleading to have different types, then cast in the join condition between them

henrystats: oke ser I'll update now

henrystats: Uhmm soo it's failing with this error @jeff-dude 

`Error: Cannot cast '3125659152' to INT [Execution ID: 01H8PR4VS9VXKB5E8C8HWFZMRH].`

I casted in the select statement in chain-ids and one of the chain ids has that id... The issue is that there are 600+ id's so I can't manually remove the quotes so it's written in integer.. 

should I revert this change, and then in optimism_bridge_flows, in the select statement where we put the chain id manually, I write it as VARCHAR instead of integer so there's no need to cast in the join condition?

jeff-dude: ah, didn't notice the `chain_ids` spell has values much larger than `1` or `10` -- only saw those in the bridge script here. you can up the data type to `bigint` or even `uint256` if needed (seems doubtful, but backup plan if bigint doesn't work)

you would then cast the `1` and `10` in bridge script to match type from `chain_ids` spell

henrystats: done ser

jeff-dude (approved):
thank you!

PR #4132, opened 2023-08-21
[labels] Add more arbitrum cex address

sankinyue:
Use sql query to obtain those EOA addresses that have initiated transactions "to" many different addresses.

A hot wallet as an exchange usually has a very significant behavior to deal with a large number of different customers withdraw ETH and token.

Finally, the physical ownership of the address is analyzed based on the block browser label and the record of receiving ETH.

```
select 
      "from"
    , count(distinct to) as address_count
from arbitrum.transactions
group by 1
having count(distinct to) > 10000
order by 2 desc
```


Hosuke (approved):
LGTM.✅
Thank you @sankinyue 

[Sample cex_arbitrum.addresses]()

hildobby (approved):
Nice, great work @sankinyue !

jeff-dude (approved):
nice work -- thank you!

PR #4133, opened 2023-08-21
T-32646: Remove deprecated prices

hawkaa:


aalan3 (approved):

hawkaa:

hawkaa:
That test breaks down when there is removal of prices. The output command returns non-0 when the result is empty. Can I merge with a 🔴 test? Should not be a problem on main.

PR #4142, opened 2023-08-22
Add new chains

kinghispanvsgh:
Add Linea and Metis blockchains to layerzero.send

### DIFF
--- a/models/layerzero/layerzero_chain_list_schema.yml
+++ b/models/layerzero/layerzero_chain_list_schema.yml
@@ -3,7 +3,7 @@ version: 2
 models:
  - name: layerzero_chain_list
    meta:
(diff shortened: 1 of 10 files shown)


springzh:
@kinghispanvsgh I'm curious about how do you test the spells. Currently the two blockchains are not supported yet. So there no relevant source table as linea.traces, metis.traces.

jeff-dude:
> @kinghispanvsgh I'm curious about how do you test the spells. Currently the two blockchains are not supported yet. So there no relevant source table as linea.traces, metis.traces.

yes, this is correct, we don't have those blockchains on dune yet. we can't build spells on them, as no data available. closing PR for now.

PR #4143, opened 2023-08-22
Datasets

aalan3:
This adds the reservoir views to spellbook. We might change the folder structure later.


models/datasets/reservoir/ethereum/tokens_beta.sql
```diff
@@ -0,0 +1,41 @@
...
+    t.token_id,
+    t.updated_at
+from reservoir.tokens_0010 t
+    inner join (
+        select
+            id,
+            max(updated_at) as recent_updated_at,
+            max(filename) as last_filename
+        from
+            reservoir.tokens_0010
```
jeff-dude, line 36: we should avoid hardcoded source objects -- is the plan to establish these as sources? then introduce a source yaml file to maintain, use source jinja syntax here?

models/datasets/reservoir/ethereum/tokens_beta.sql
```diff
@@ -0,0 +1,41 @@
...
+                \'[""]\') }}'
+    )
+}}
+select
+    t.created_at,
+    t.collection_id,
+    _unsafe_0x_or_base58_to_varbinary(t.contract) as contract,
+    _unsafe_0x_or_base58_to_varbinary(t.floor_ask_maker) as floor_ask_maker,
+    _unsafe_0x_or_base58_to_varbinary(t.floor_ask_id) as floor_ask_id,
+    _unsafe_0x_or_base58_to_varbinary(t.owner) as owner,
```
jeff-dude, line 15: we haven't been using these on dunesql spells. this means the third party data is uploaded on spark, then needs converted to dunesql datatypes?

i think the approach has been `fron_base58()` function (or something like that, off top of head) to convert to varbinary

macros/expose_dataset.sql
jeff-dude, line 1: i think we should differentiate this macro naming more from the usual expose spell macro.

are we officially calling it `third party data` or `community data`?

whichever we use, i think the macro should be:
`expose_third_party_data`
or `expose_community_data`

models/datasets/reservoir/ethereum/tokens_beta.sql
jeff-dude, line 1: similar to macro, maybe we rename:
`models/datasets/`
to
`models/third_party_data/`
or `models/community_data`

models/datasets/reservoir/ethereum/tokens_beta.sql
```diff
@@ -0,0 +1,41 @@
+{{ config(
+    tags = ['dunesql'],
+    schema = 'reservoir',
+    alias = alias('tokens_beta'),
+    post_hook = '{{ expose_dataset(\'["ethereum"]\',
```
jeff-dude, line 5: this wuold change if the macro name changes (playing it safe with comment)

jeff-dude (changes requested):

models/datasets/reservoir/ethereum/tokens_beta.sql
aalan3: @dsalv is it safe to use some other function here?

macros/expose_dataset.sql
aalan3: Naming here is TBA it can change later. But I was mostly thinking that whatever is in this repo is already `community`.

models/datasets/reservoir/ethereum/tokens_beta.sql
dsalv: It should be fine, but checking using the raw data is better. I can do it tomorrow or you can also check in the explorer

aalan3: >but checking using the raw data is better.

Could you elaborate on this I didn't understand it.

dsalv: ` from_hex(t.owner)` should work. But it's better to run queries on raw tables and validate that the resulting data is same

models/datasets/reservoir/ethereum/tokens_beta.sql
jeff-dude: we lose the capability for dbt lineage without `source()` or `ref()` usage

this is experimental, so if you want to proceed and potentially revisit later, feel free. it's just not something that is done currently across spellbook

jeff-dude (approved):
i still think we should consider something more descriptive than `dataset` for the macro / directory paths, but i'm good with moving forward for sake of the POC efforts

models/datasets/reservoir/ethereum/tokens_beta_legacy.sql
```diff
@@ -0,0 +1,11 @@
+{{ config(
+  schema = 'reservoir',
+  alias = alias('tokens_beta', legacy_model=True),
```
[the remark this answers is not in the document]
aalan3: we don't this is just to make DBT happy during the migration phase

dsalv (approved):

jeff-dude:
feel free to merge on your preferred timeframe, i was not planning to merge on my side

aalan3:
I’m waiting for the data explorer issue to be fixed before merging this, otherwise I can’t test it properly

jeff-dude:
@aalan3 what's the plan here?

aalan3:
@jeff-dude let's merge this

PR #4146, opened 2023-08-23
Fix check_seeds macro

aalan3:
We need to go through some tests and see if they are still correct. This type of error is very subtle since `NULL = false` in SQL becomes just `NULL`.

### DIFF
--- a/macros/test-helpers/check_seed_macro.sql
+++ b/macros/test-helpers/check_seed_macro.sql
@@ -96,5 +96,5 @@
         from equality_tests
     ) all
     -- equality check can be null so we have to check explicitly for nulls
-    where equality_check = false or equality_check is null
+    where equality_check is distinct from true
 {% endmacro %}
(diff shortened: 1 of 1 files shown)


aalan3:
@0xRobin do you have any insight on this? (since you worked on this previously in https://github.com/duneanalytics/spellbook/pull/3676). I think the legacy version of this check `result_model = expected_seed` should have resulted in pretty much the same issue.

0xRobin:
ah yes you're right, the check is not null-safe (just as the equality operator is not). 
I propose we follow trino docs here and start using the null-safe `IS DISTINCT FROM` or `IS NOT DISTINCT FROM` 
https://trino.io/docs/current/functions/comparison.html#is-distinct-from-and-is-not-distinct-from
(as well as changing the legacy version to use `<=>`, the spark equivalent) 
https://spark.apache.org/docs/3.0.0-preview/sql-ref-null-semantics.html#comparision-operators-

We might have missed some issues without knowing due to this.
Might be best to do a full test run on spark and trino to see if we have failing tests?

couralex6 (approved):

aalan3:
Yes we should do a full test run to try to find issues. We currently only test when something is changed but we can do a full test in CI. 

I'll do a separate PR for legacy since CI is breaking here.

PR #4153, opened 2023-08-24
Dex Aggregator Migration - #7 - Migrate Oneinch

henrystats:


jeff-dude:
whew, taking on 1inch -- thank you! i did the initial migration from postgres to spark, so i'm familiar with all of the scripts and should be able to help if issues come up. note some sql comments in one of the versioned spells, where we have known issues on some tx's that break the logic. we're not 100% confident in the logic itself due to that.

henrystats:
oke ser thank you @jeff-dude! it took about 20mins for the ci to run for just the clipper and limit order models to work... 

For it to be easier to review/ it also helps as I'm migrating, would it make sense for me to submit individual pr's for each model?

I could still do that in one pr but there are a lot of models so I'm wondering If it's better to separate them out.

jeff-dude:
> oke ser thank you @jeff-dude! it took about 20mins for the ci to run for just the clipper and limit order models to work...
> 
> For it to be easier to review/ it also helps as I'm migrating, would it make sense for me to submit individual pr's for each model?
> 
> I could still do that in one pr but there are a lot of models so I'm wondering If it's better to separate them out.

i would appreciate that, yes. separate PR per version spell would expedite the process for us both

henrystats:
oke ser... will separate them out thank you ser!

PR #4154, opened 2023-08-24
Fix typos in Chainlink price feeds

aalan3:
These typos cause issues downstream.


jeff-dude:
good catch. we won't see that in CI, as we don't run those statements unless it's prod

jeff-dude:
we'd have to write some custom tests to be certain

PR #4164, opened 2023-08-25
migrate top NFT models

0xRobin:
This PR will migrate the final models in the `nft.trades` lineage to DuneSQL


0xRobin:
@jeff-dude I think we're ready!! :smiley:
The difference in the last two months is due to some (newer) models only being present in the dunesql datase. (blur v2, looksrare seaport and opensea on arbitrum)
image
image
image

models/_sector/nft/trades/nft_events.sql
```diff
@@ -1,7 +1,8 @@
 {{ config(
     schema = 'nft',
     alias = alias('events'),
-    partition_by = ['block_date'],
+    tags = ['dunesql'],
+    partition_by = ['blockchain','project','block_month'],
```
jeff-dude, line 5: could this granular of a partition lead to v small partitions?
can validate with a simple query to group by these fields and get a count, order by asc?

jeff-dude: we still have a backlog item to build a universal partition strategy post-migration, but i'd rather lean towards too big of partitions vs. too small in the short term

jeff-dude: to be fair, a universal `block_month` on all dex models likely led to many small partitions on smaller projects too

models/_sector/nft/trades/nft_events.sql
jeff-dude, line 1: i wonder if we simplify the directory to remove `trades`
`models/_sector/nft/<models>`

jeff-dude: looks like the macro directory structure doesn't use `trades` either

models/_sector/nft/trades/nft_events.sql
jeff-dude: (these comments apply throughout the PR)

jeff-dude (commented):
a few comments below, but great work as always here 🤝

models/_sector/nft/trades/nft_events.sql
0xRobin: I want to keep `trades` around as that is the top level sector spell for this dependency tree.
In the future I'd see 
`_sector/nft/lending`
`_sector/nft/mints`
`_sector/nft/approvals`
`_sector/nft/transfers`
etc. on the same level

models/_sector/nft/trades/nft_events.sql
0xRobin: This is current trade-off I saw as best.
It's speed for cross-chain queries vs speed for blockchain/marketplace specific queries.
You don't really want a query on only blur on eth mainnet to have to scan all the magiceden solana trades. And since most of the uses are still project/blockchain specific I went with this partitioning.

jeff-dude: i'm good with that 👍 
we can revisit when we get to the future backlog item, just to ensure consistency across sector spells and such

models/_sector/nft/trades/nft_events.sql
jeff-dude: fair, i can be on board with that

PR #4177, opened 2023-08-27
Oneinch trades migration - #8 - 1inch

henrystats:


henrystats:
@jeff-dude I've submitted separate pr's to migrate each oneinch model... This particular pr migrates the oneinch versions (1,2,3,4,5) soo it's still somewhat a different model...

I wanted to suggest that we should merge this first after the pr's are reviewed because 

1. 1split & 1Proto rely on it (I commented the anti left join out to test that both models work)

2. They all use one schema file and the oneinch_v1_trades is at the top of the schema file sooo if this gets merged first, I'd only need to make changes to the schema once because of the block_month column... I added it in all the models using "&" just so it's neat hehe soo once this gets reviewed/merged I'd only need to update the schema of the other pr's once since this is at the top and *block_month would work fine for all...

jeff-dude (approved):

PR #4186, opened 2023-08-28
Celo erc721/1155 transfers

tomfutago:


models/transfers/celo/erc1155/transfers_celo_erc1155_rolling_day.sql
```diff
@@ -6,7 +6,7 @@
         materialized = 'incremental',
         file_format = 'delta',
         incremental_strategy = 'merge',
-        unique_key = ['wallet_address', 'token_address', 'block_day', 'token_id'],
+        unique_key = ['wallet_address', 'token_address', 'block_day', 'token_id', 'amount'],
```
Hosuke, line 9: Since there is no incremental filters in this spell, `transfers_celo_erc1155_rolling_day` may not be an incremental view, but simply a view.

Hosuke: The unique_key is required when the model is incremental view.
Here is a more detailed doc:
https://docs.getdbt.com/docs/build/incremental-models

By the way do you think is there any better option for unique_key other than `amount`?

tomfutago: Rolling scripts updated to views now. With agg tables I realised I actually don't need `amount` as part of unique_key - it's good to go without it.

Hosuke (approved):
LGTM.✅
Thank you @tomfutago 

[Sample transfers_celo_erc1155]()
[Sample transfers_celo_erc721]()

jeff-dude (approved):

PR #4188, opened 2023-08-28
Add tag to remove spell

aalan3:
This PR adds a tag (`remove`) that can be used to mark models as removed, which means that they should no longer run in CI or DBT anymore. 

This tag should ONLY be applied to legacy models that have had their full lineage migrated to Dune sql.

This PR also updates CI.


jeff-dude (approved):
nice!

are you going to add the full lineage in this PR, or merge this for the tag and then follow up for the full lineage after?

either way, feel free to merge when you're set 👍

aalan3:
Yeah gonna merge the rest in a separate PR. Just wanted to make sure CI works properly. I also update the legacy jobs in DBT.

0xRobin:
> This tag should ONLY be applied to legacy models that have had their full lineage migrated to Dune sql.

let me add a quick check for this.

PR #4189, opened 2023-08-28
CI check for tag:remove

0xRobin:
This checks there are no child models from `tag:remove` models that are still enabled (without the remove tag)


aalan3 (approved):
This is clever thanks! We can probably use something similar to see which lineages are remaining to migrate or incomplete as well.

jeff-dude (approved):
nice! yeah, to alan's point, we were already considering using something like this later when ready but you're on top of it. much appreciated

aalan3:
Merged to test for https://github.com/duneanalytics/spellbook/pull/4190

PR #4192, opened 2023-08-29
Restructure ETH staking sector files

hildobby:
With the ever increasing number of tagged ETH staking addresses and the different methods needed, I think it will be easier with them split across different files


Hosuke (approved):
LGTM.✅
Thank you @hildobby 

[Sample staking_ethereum_entities_contracts](https://dune.com/queries/2980280)
[Sample staking_ethereum_entities_coinbase](https://dune.com/queries/2980286)
[Sample staking_ethereum_entities_binance](https://dune.com/queries/2980290)
[Sample staking_ethereum_entities_addresses](https://dune.com/queries/2980296)

models/staking/ethereum/staking_ethereum_entities_addresses_legacy.sql
jeff-dude, line 1: just a note, legacy files with dummy select 1 are only needed if:
- dunesql spell only
- _and_ the schema doesn't exist on spark from older spells

it certainly doesn't hurt to add, but to your question about stopping the legacy additions, you can in certain scenarios 🤞 

you can leave these here, since already built, but something you can consider in future builds

jeff-dude (approved):
sounds good!

models/staking/ethereum/staking_ethereum_entities_addresses_legacy.sql
hildobby: Ahh noted, thanks for the heads up!

PR #4201, opened 2023-08-29
Optimism ERC20 & ETH Balances - WIP

henrystats:


henrystats:
@MSilb7 @jeff-dude, I'm working on balances for all the evm chains for both erc20 tokens & mainnet tokens (I'll build for NFTs shortly after I'm done with this...

I have a couple of questions I'd love if you could help with regarding the spells before I start duplicating for other evm chains... 


**Transfers Table**

1. I changed to using dbtutilscombination for the unique test, I added unique keys as well and remove unique-transfer-id... I believe we used that approach when the balances spell was originally built which was a while ago and it was a similar method used for dexes & NFTs as well but that was switched because using unique-keys were a faster and less resource consuming way of testing... I wanted to be sure that I'm not breaking anything by switching to unique keys, it isn't affecting any downstream models...

I also switched to a UNION ALL, UNION ALL is faster I believe since each table we're joining is either positive or negative on the amount column so UNION doesn't do anything. 

2. Some other transfers models are ran incrementally, should we do this across all the transfer spells?

3. Would it be okay to replicate the current erc20 model for eth transfers so it'd be easier to build a balances spell..

**Balances**

1. Would a general evm.balances table be useful? it'd be a UNION ALL of all the evm models.

models/balances/optimism/erc20/balances_optimism_erc20_noncompliant.sql
```diff
@@ -0,0 +1,13 @@
...
+        materialized ='table',
+        file_format = 'delta'
+        )
+}}
+
+SELECT  
+    DISTINCT token_address
+FROM 
+{{ ref('transfers_optimism_erc20_rolling_day') }}
+WHERE round(amount/power(10, 18), 6) < -0.001
```
antonio-mendes, line 13: Good idea to check for compliance this way. We should monitor this table and see what ends up here so we can proactively identify if there's any faulty logic.

Also, give were using DuneSQL and have the INT256 and UINT256 types available, if everything runs as expected we shouldn't expect to see integer overflow or any weirdness happening. But still, better safe than sorry so let's monitor.

models/transfers/optimism/erc20/transfers_optimism_erc20_agg_day.sql
```diff
@@ -0,0 +1,27 @@
...
+    t.symbol,
+    sum(tr.amount_raw) as amount_raw,
+    sum(tr.amount_raw / power(10, t.decimals)) as amount
+FROM 
+{{ ref('transfers_optimism_erc20') }} tr
+LEFT JOIN 
+{{ ref('tokens_optimism_erc20') }} t on t.contract_address = tr.token_address
+{% if is_incremental() %}
+-- this filter will only be applied on an incremental run
+WHERE tr.evt_block_time >= date_trunc('day', now() - interval '7' Day)
```
antonio-mendes, line 25: Wonder if we could try and reduce this to maybe ~3 days or something for efficiency gains. I'm assuming these balance spells are gonna be pretty expensive to run, and 3 days would still give us quite some time to fix any issues with incremental runs failing before we would have a real problem.

What do you think @jeff-dude?

antonio-mendes: I'm writing this here but I'm ofc talking about applying this to all the other files

models/transfers/optimism/erc20/transfers_optimism_erc20_rolling_hour.sql
```diff
@@ -0,0 +1,17 @@
+{{ config(
+        tags = ['dunesql'],
+        alias = alias('rolling_hour'))
+}}
+
+        SELECT
```
antonio-mendes, line 6: I might be reading this wrong, but wouldn't we need this to be on an incremental run as well? Otherwise it will only run when we deploy it?

models/balances/optimism/erc20/balances_optimism_erc20_latest.sql
```diff
@@ -0,0 +1,30 @@
+{{ config(
+        tags = ['dunesql'],
+        alias = alias('latest'),
+        post_hook='{{ expose_spells(\'["optimism"]\',
```
antonio-mendes, line 4: It might make sense to materialize the `latest` tables as they might end up being used in a lot of queries/dashboards/models downstream? Thoughts @jeff-dude?

antonio-mendes (changes requested):
Great PR @henrystats thanks for spending the time on this, it's super valuable to everyone! I left some questions on the PR, I also pinged Jeff here and there as I'm curious to get his opinion.

Regarding the questions you asked above, same as @jeff-dude my initital thoughts is yes to everything! I think we can also if necessary have these tables in sort of a beta state like dex.trades is right now for DuneSQL and spend some time validating this against other sources.

As a final note, knowing there's nothing you can do about that hehe, I'm gonna ask internally what's the status on being able to get ETH balances as that would obviously be very helpful to get a full view of a wallet balances.

antonio-mendes (commented):
Great PR @henrystats thanks for spending the time on this, it's super valuable to everyone! I left some questions on the PR, I also pinged Jeff here and there as I'm curious to get his opinion.

Regarding the questions you asked above, same as @jeff-dude my initital thoughts is yes to everything! I think we can also if necessary have these tables in sort of a beta state like dex.trades is right now for DuneSQL and spend some time validating this against other sources.

As a final note, knowing there's nothing you can do about that hehe, I'm gonna ask internally what's the status on being able to get ETH balances as that would obviously be very helpful to get a full view of a wallet balances.

models/transfers/optimism/erc20/transfers_optimism_erc20_rolling_hour.sql
jeff-dude: it defaults to a view if we don't specify in config block, so the data will be live and just refreshed at runtime on the app

models/balances/optimism/erc20/balances_optimism_erc20_latest.sql
jeff-dude: a view is nice for most up-to-date data, but if the performance is poor, looking to make incremental is worthwhile to try

henrystats:
Thank you very much ser @antonio-mendes, I'll make the changes suggested and also make the changes related to the questions I originally asked. 

Also for ETH balances I didn't understand what you mean. I thought we could already do that through the traces table 👀

MSilb7:
Overall note - you'll also need to go through OVM1 to get accurate ERC20 balances for addresses. There are also older chain versions for SNX staking that we have a state dump for, but need to parse through. Otherwise, you'll get negative balances.

But for now, make sure to include optimism_legacy. Note that ETH was an ERC20 then (I believe 0x42000....006), and I believe there was no concept of WETH (now 0x4200....0006).

henrystats:
Thank you ser @MSilb7 I'll make the changes accordingly. 

I wrote a [query](https://dune.com/queries/3000532) on dune to find the token addresses not in op transactions but in ovm legacy and I found 6.

Was trying to quickly google a chain explorer for ovm transactions but I couldn't find any, do you know where/how I could map the contract address so I know what their respective contract addresses on op mainnet is... 

I plan to use a case when to switch the contract addresses in the ovm erc20 transfers to their contract addresses (since ovm is not going to run incrementally).

models/transfers/optimism/erc20/transfers_optimism_erc20_rolling_hour.sql
antonio-mendes: Ah ok that was a derp from me, I was thinking for a sec it would run once create the table and leave it in a stale state. Makes sense!

antonio-mendes:
Good callout on Optimism Legacy @MSilb7 , thank you!

@henrystats what I meant with ETH balances is that we'd like to have available statediffs for ETH for all chains (or BNB, MATIC and FTM for the respective chains) as that would make that balance a lot easier to keep track of. Unfortunately we're not there yet, so the proposed solution will work, it's just more resource intensive.

henrystats:
Ah I see... okeee... thank you ser @antonio-mendes

henrystats:
Hola ser @antonio-mendes, thank you ser. I understand and looking forward to when we have that, would certainly help a lot!

bump ser @MSilb7 regarding the tokens with changed contract addresses, I'll be making the final update/changes to the spell tonight so I wanted to add those tokens as well.

henrystats:
Hola ser @MSilb7, I wanted to ask for your help here ser, so I've added OVM data and there still seem to be some negative balances / slightly off balances for ETH. 

I noticed that a handful of the addresses are Self destruct contracts and others are the ones that had interactions on OVM legacy. 

Everything looks correct to me soo far so not sure what I'm doing wrong, if you could pls help take a look so see where the issue could possible be, or if you had a balances query on dune for OVM or a query with transactions & gas fees so I could try and compare to see where the issue is.

jeff-dude:

#4521 
#4523 
#4633 
#4698

PR #3774, opened 2023-07-17
Revert "Looksrare seaport-routed base trades (#3734)"

jeff-dude:
This reverts commit fe6e6c5d86d981bbfcf42481b86b4a2f8fcd01c5.

fyi @0xRobin @hildobby 
i need to revert this for the time being, as it created unintended side effects on the legacy spark orchestration in prod. running dbt compile (which dbt cloud does by default on our run command) is failing.
```
databricks.sql.exc.ServerOperationError: [SCHEMA_NOT_FOUND] The schema `looksrare_seaport` cannot be found. Verify the spelling and correctness of the schema and catalog.
If you did not qualify the name with a catalog, verify the current_schema() output, or qualify the name with the correct catalog.
To tolerate the error on drop use DROP SCHEMA IF EXISTS.
```

it may have to do with some of our dune override macros that are part of the overall compile process?

seems like it doesn't know to filter out the metastore used for trino side, and tries to list the schema on spark?

i'm thinking our gh actions pass compile successfully because we pass in flags for `tag:legacy`, while we can't do that in dbt cloud?

PR #4204, opened 2023-08-29
Looksrare seaport (rerevert)

hildobby:
This is a followup to the discussion here: https://github.com/duneanalytics/spellbook/pull/3774

As a reminder, the model was merged ~45 days ago but changes were reverted due to the then just-discovered spark bug (which I mitigated with an added dummy spark file here)


models/_sector/nft/trades/ethereum/platforms/_schema.yml
```diff
@@ -312,6 +312,34 @@ models:
...
+ - name: looksrare_seaport_ethereum_base_trades
+   meta:
+     blockchain: ethereum
+     sector: nft
+     project: looksrare
+     contributors: [ 'hildobby' ]
+   config:
+     tags: [ 'ethereum', 'nft', 'trades', 'looksrare' ]
+   description: "looksrare's seaport base trades"
+   tests:
```
jeff-dude, line 325: rob introduced a new test, as seen here:
https://github.com/duneanalytics/spellbook/blob/main/models/_sector/nft/trades/ethereum/platforms/_schema.yml#L13-L14

which you can apply here by simply adding that one row. you shouldn't need to do anything else. it will check that the columns written are the correct data type for `nft.trades` lineage

models/_sector/nft/trades/ethereum/platforms/looksrare_seaport_ethereum_base_trades.sql
```diff
@@ -0,0 +1,42 @@
+{{ config(
+    schema = 'looksrare_seaport_ethereum',
+    tags = ['dunesql'],
+    alias = alias('base_trades'),
+    partition_by = ['block_date'],
```
jeff-dude, line 5: you can remove this partition for now. rob will be doing a universal approach to partitioning across base models in nft lineage post-migration

jeff-dude (changes requested):
minor comments below

hildobby:
Just addressed both of your comments @jeff-dude

jeff-dude (approved):
sounds good 🤝

PR #4207, opened 2023-08-30
[WIP] Celo more token symbols and nft standards

tomfutago:


models/tokens/tokens_native.sql
jeff-dude, line 1: if migrating this to dunesql, we should update the syntax to use native datatypes. please remove quotes around addresses and the `lower` function on address for instance

tomfutago: makes sense, thanks for checking

tomfutago: @jeff-dude there is solana address mixed in on this list tho - is it possible to represent it as byte array? I'm getting error if I simply remove quotes for this address (so11111111111111111111111111111111111111112). Alternative would be to comment this row out for now, but not sure if that's the way to go?

jeff-dude: i believe we use `from_hex()` to convert solana addresses? you can verify in our `prices.tokens` lineage of spells, where we combine solana into other evm blockchain addresses

tomfutago: cool, thanks. prices_solana_tokens uses from_base58, so I'll go with that too.

PR #4213, opened 2023-08-31
Various fixes to `seaport.traces` & `seaport.tagging`

hildobby:
- Change from `sector: nft` to `project: seaport` when mislabelled
- Hide single chain spells, leave crosschain ones visible
- Add OpenSea hash salt marker to all other EVMS (was just on Ethereum before)


jeff-dude (approved):
once the post-hook already ran to make it visible, removing here may not remove the table properties. we may need to work with engine team to understand how to remove them

hildobby:
> t visible, removing here may not remove the table properties. we may need to work with engine team to understand how to remove t

Ah damn, it's not a huge deal tbh but I realised having all chains makes it kind of a lot in the UI vs only having the crosschain one (most only need the crosschain one and those who want the chain specific one likely know the name structure imo)

jeff-dude:
agreed, i think moving forward, you can plan to only expose cross-version or cross-chain, and then i will try to catch that in PRs too

PR #4215, opened 2023-08-31
Testing oneinch Migration

henrystats:


henrystats:
soo it looks like we just need to wait a bit for all the models to work.. a few hours ago.. select * from oneinch_onesplit_ethereum.trades didn't work on dune.com and was returning a weird error.

It works now though, and the UNION ALL works as well https://dune.com/queries/2985132

But the main table doesn't work... So I'll continue to check this to see if it all eventually works.

jeff-dude:
i think we're good now? let me know if it needs re-opened

PR #4223, opened 2023-09-01
Add Base to Contracts Mapping

MSilb7:
~Comes after https://github.com/duneanalytics/spellbook/pull/4222~ **DONE**
~& after https://github.com/duneanalytics/spellbook/pull/4240~ **DONE**

Adds Base to multi-chain creator mapping & adds handling for OP Stack Chain Predeploys


MSilb7:
Seems like there's timeout issues, but the initial build runs in < 5 mins on dune.com: https://dune.com/queries/2994348

MSilb7:
Still not sure why this is causing so many runtime issues. It completes in 10 mins here: https://dune.com/queries/2994348

jeff-dude:
> Wondering if there's any insight/ideas as to why this spell initial build times out (or things to try to modify)?
> 
> Running both self-destruct and contract mapping on dune.com (free tier) takes ~ 7 mins: https://dune.com/queries/2994348
> 
> cc @jeff-dude

replicating the query by using a CTE for the spell dependency on the app may not help us understand the query planning and performance it seems. can you push a commit to up the timeout on CI to a higher number of minutes? you can find in the gh workflows directory. let's see what happens when it runs on a higher timeout period, even though it's odd when ethereum ran in the set time just fine

MSilb7:
doubled the timeout

PR #4226, opened 2023-09-01
[easy] Velodrome - Fix Version Mapping

MSilb7:
Make version pull from each model, versus being hardcoded to 1


jeff-dude (approved):
also just noticed this today when analyzing data in dex trades between the engines -- thank you!

PR #4229, opened 2023-09-02
Celo erc20 daily balances

tomfutago:


tomfutago:
Hey @antonio-mendes, thanks for looking into this and your comments!

I'm still concerned about going with views tho.. I just tried running this on Large query engine:
`select count(*) from balances_polygon.erc20_hour`
and it timed out.. (I tried twice - same result)

It's maybe not a representative of what most users would want to use this spell for, but at the same time I'd assume it's simple enough for any spell to be able to complete without any issues, would you agree?

antonio-mendes:
I wouldn't worry too much about that failing. That query, as you said, won't really be used unless someone is making a full dump of the table, and even in that case they could batch it in shorter periods of time and make it work. It also would never work because we can only render so much that data in the UI.

The important questions are whether we can build it in the first place, if the incremental runs are fast enough where needed, and more importantly if the data results are accurate.

If you attempt to query `balances_polygon.erc20_hour` in a more "natural" manner, like checking your wallet or doing some aggregations on tokens/wallets you will find it most likely works.

antonio-mendes:
@tomfutago yeah feel free to use materialising and incremental updates for transfer tables, just maybe we won't need it for balances

tomfutago:
@antonio-mendes updated balances to build as views. I had to remove uniqueness check on hourly spell as it was timing out, which I think was a similar conclusion on other erc20 balances.

Also, a note for @henrystats - I saw you added credits to my hourly sequence query in your `balances_polygon_erc20_hour` spell - thank you! And apologies at the same time, as original version of that query contained a bug, which causes dupes for midnight on 1st Jan on every year in the middle of the sequence (so in Polygon case it's 2021-01-01 00:00, 2022-01-01 00:00 and 2023-01-01 00:00). 
Fix is to either update: `sequence(0, 9000)` to `sequence(1, 9000)`
or: `s.n <= date_diff('hour', y.year, y.year + interval '1' year)` to `s.n < date_diff('hour', y.year, y.year + interval '1' year)`

antonio-mendes:
Oh damn, good catch with the time bug @tomfutago ! @henrystats or @tomfutago would you ming opening PRs to fix that for BNB and Polygon since those are merged now?

henrystats:
yes I can set up a fix pr @antonio-mendes, will do it right away!

Thank you @tomfutago!!

jeff-dude:

#4521 
#4523 
#4633 
#4698

PR #4237, opened 2023-09-03
Create format figure macro - Need Help

henrystats:

### DIFF
--- a/macros/public/format_figure.sql
+++ b/macros/public/format_figure.sql
@@ -1,0 +1,13 @@
+{% macro format_figure() %}
+   CREATE OR REPLACE FUNCTION format_figure(figure, round_to, currency STRING)
+   RETURNS STRING
+   RETURN
+   SELECT
+      CASE
+         WHEN CAST(figure as double) >= 1000000000000 THEN currency || format(concat('%1.', CAST(round_to as VARCHAR), 'f'), CAST(figure as double)/1e12) || 'T'
+         WHEN CAST(figure as double) >= 1000000000 THEN currency || format(concat('%1.', CAST(round_to as VARCHAR), 'f'), CAST(figure as double)/1e9) || 'B'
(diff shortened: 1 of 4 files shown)


henrystats:
Hola ser @jeff-dude, I'm trying to create a format_figure function that could be used for counters on dune instead of the current workaround that folks do... It's the same workaround just more simplified.

I'm not sure if this would even be helpful haha, but I'm trying to learn functions/macros so this seemed like a great place to start. 

I tried testing the function I wrote in Macros in one of the models but it didn't work... I followed similar process as get_href to create the function... 

Is building of functions possible / does it require special permissions to make it work, or maybe I'm doing something wrong somewhere.

jeff-dude:
> Hola ser @jeff-dude, I'm trying to create a format_figure function that could be used for counters on dune instead of the current workaround that folks do... It's the same workaround just more simplified.
> 
> I'm not sure if this would even be helpful haha, but I'm trying to learn functions/macros so this seemed like a great place to start.
> 
> I tried testing the function I wrote in Macros in one of the models but it didn't work... I followed similar process as get_href to create the function...
> 
> Is building of functions possible / does it require special permissions to make it work, or maybe I'm doing something wrong somewhere.

user-defined functions are not supported on dunesql at this time. it is on the roadmap for our engine team. we've had a few macros that we turned into functions that can be run on the dune app directly, but it's not a feasible long-term solution and therefore will be redesigned and built out later on.

you can still build a macro in spellbook, then use that macro within other spells, but not directly on the app. i'm not sure how helpful a spellbook-specific macro is for you in this example, but if you wanted to repurpose and do that, all good. otherwise, feel free to close

henrystats:
ah I see, thank you very much ser @jeff-dude... The function is only useful if it's on dunesql soo I'll close this pr!

I'll look for something else I could build with macros that could be useful for other spellbook users. I was interested in learning how the macros work and wanted to build something hence this pr..

Thank you very much for reviewing ser, will close this pr now!

PR #4245, opened 2023-09-04
add major base token prices

grkhr:


jeff-dude (approved):
thank you!

notes for the failure on tokens test:
- API: https://api.coinpaprika.com/v1/coins/axl-axelar
- returns id: "id":"ico-axelar"
- while they are different, the API still returns, so we should be good
- this is likely just a bug in the test script

PR #4246, opened 2023-09-04
Make first activity.to crosschain - Ready for review

henrystats:


jeff-dude (approved):
thank you 🤝 

remind me, why did we land on append strategy for this spell lineage again? rather than merge like most others?

henrystats:
Hmm, I'm building this similarly to the first_funded_by model but I remembered that I followed the conversation with Hildobby & OxRob where append was suggested.. 

As far as I remember, since we're using a left anti join to remove addresses that are already in the model, then there's no need to use merge since only unique entries will be added to the table. 

Whilst using merge would bring the same results, **append** is the best/faster incremental strategy here since it'd only add new entries and not actually check if the new entries it's adding already exist in the model. 

And since we already have a left anti join, we don't need to worry about duplicates.

jeff-dude:
> @jeff-dude Hmm, I'm building this similarly to the first_funded_by model but I remembered that I followed the conversation with Hildobby & OxRob where append was suggested..
> 
> As far as I remember, since we're using a left anti join to remove addresses that are already in the model, then there's no need to use merge since only unique entries will be added to the table.
> 
> Whilst using merge would bring the same results, **append** is the best/faster incremental strategy here since it'd only add new entries and not actually check if the new entries it's adding already exist in the model.
> 
> And since we already have a left anti join, we don't need to worry about duplicates.

agreed on the logic being fine, just curious! thank you

PR #4254, opened 2023-09-05
[Easy] - Fix Element Polygon MATIC Price

henrystats:


henrystats:
@jeff-dude @0xRobin WMATIC amount original & usd value is missing for the trades where the WMATIC token was hardcoded...

This is because the address copy/pasted wasn't set to lower case, I edited and set to lower case and triggered reruns of the downstream models so it gets refreshed.

0xRobin (approved):
@henrystats thanks for catching this!! :pray: 

The comments on downstream models aren't necessary as the backend of Dune will automatically refresh any downstream models. 
The comments would only be necessary if you want to refresh these models within the github CI run.
I'll remove these changes and queue this up for merging :heavy_check_mark:

jeff-dude:
@antonio-mendes if you do some merges, this one can be included

PR #4074, opened 2023-08-14
Migrate balances_ethereum_erc20_day

lajarre:
# Summary of changes

- Migrating `balances_ethereum_erc20_day` and `balances_ethereum_erc20_hour` and related files to DuneSQL and update their related schema
- `tests/balances/ethereum/balances_ethereum_erc20_day_assert_nonnegative.sql`: relax test by rounding to 0 digits. This might be improved if necessary.
- Non-compliant tokens check: relaxed to `< 1100`, from `< 550`. This seems reasonable as the test itself was reporting a number of reported non-compliant tokens slightly above 1000.

PR #4257, opened 2023-09-05
Migrate balances erc20 _daily only_

lajarre:
Same as #4074 but deleted the hourly spell


lajarre:
closing in favor of #4074 which is closer to completion

antonio-mendes:
Thanks for cleaning up @lajarre , appreciate it

PR #4258, opened 2023-09-05
Base ETH & ERC20 Balances - WIP 

henrystats:


henrystats:
@antonio-mendes I can update the remaining models later today (I need to update optimism) and then add the erc20 name for the other evms... I'll be back here in about 1hr to make the changes on other models if it's fine.

For base, deposits / withdrawals of ETH are only trigged onchain (that's where an event is logged) so I'm trying to find a workaround for that.

antonio-mendes:
Awesome @henrystats that sounds perfect! 

Can you give me more context on the ETH balances for Base and what would be missing with the spell in it's current state? From the few validations I've sone it seemed the result is accurate with super small decimal discrepancies, similar to Polygon https://github.com/henrystats/henrystats/blob/86f24f355952805ee153150d277c3f0f45ddc515/models/transfers/base/eth/transfers_base_eth_tfers.sql#L55

henrystats:
I tested using the balances table, here's an example: 

https://basescan.org/tx/0xf8bd7ac4e009c67af22793b246948103f91260c0aed88b5fdb82c4e603c2edf8

https://basescan.org/address/0x05dffc9a8703501442fc1b0a25976523edbde094#deposittxs

https://dune.com/queries/3018374

The query is returning a negative balances because the 2.5 ETH deposited onto the wallet is missing.

It is showing in the base.transactions table, but as a self transaction, so the address is sending the ETH to itself I'm not sure if it's the same with "Withdrawals" so it feels like the best thing to do would be to use the l1 deposit/withdrawal transaction?

henrystats:
@antonio-mendes

henrystats:
@antonio-mendes, as I go deeper into this rabbit hole It becomes more complicated! 

There's a lot of cases where an address is able to send ETH out and there are no traces, no transaction of ETH to the wallet (there are transactions on the transactions table with no traces), no deposits through the bridge and the address transfers ETH out. 

Lots of weird things going on, I've spent the past few hours trying to work around Optimism & Base balances and it's been really wild. 

I'll continue to dig and see if I can find something, there are similar issues with Arbitrum & Avalanche_c as well although those seem very straightforward. 

Looking forward to when we'd eventually get all this figured out!

henrystats:
@antonio-mendes this is a list of the addresses with less than 0 as their eth balance. 

https://dune.com/queries/3023159

I noticed that a bunch of it are contract addresses, the contract creation transaction that creates the contract isn't included when we select data from the traces table. 

A couple are also system accts or got their original tokens at genesis (so we'd have to manually enter that?)

And then there are addresses that start sending ETH out with no original receipt of funds, I'm guessing these were also funded at genesis? since there's no logs/traces/transactions data on the transfer in.

antonio-mendes:
Oh good findings!

So a few thoughts:
- I'm inclined to think we should exclude balances of contracts from these balances tables, so basically exclude addresses we find in the `creation_traces` table
- This could mean in the future, specially with "smart accounts" gaining traction we should have a balances_smart_accounts spell
- Regarding the wallets that start with a balance at genesis yeah we should probably hardcode those, it could be a separate table but included in this PR. Do you have somewhere to fetch this data from?

jeff-dude:

#4521 
#4523 
#4633 
#4698

PR #4259, opened 2023-09-06
Polygon ERC20 & Matic balances - Ready for Review

henrystats:


henrystats:
@jeff-dude @antonio-mendes, This is the first balances model that is ready for review since I need to make some adjustments unique for optimism & base. 

I'll duplicate this for (avalanche, fantom, bnb, & arbitrum)... 

I made all the changes that we discussed in the comments on the optimism pr, and also added partitioning by block_month for the incremental models and switched to  3 day interval for the incremental run...

Since we want to have this in BETA, I'm thinking of dropping the post hook? and then the main table which will be (evm_fungible_balances_beta) would be the model with the post hook and then once we're confident that the data is accurate then we can remove the beta?

I've checked a few addresses on polygonscan and matched their results with what was on the spell and it all looks correct so far!

antonio-mendes (approved):
@henrystats Awesome job!! From my validations this all adds up and we're gonna deploy this is a "beta" stage. So we'll get the tables in production and we will try to get some feedback on them and to find if there are any bugs.

Some notes:
- WMATIC is decoded under the project name of `mahadao` this is wrong, but it's something we need to fix on our end. It's not a trivial change and it doesn't impact the results or data quality so we will fix it later
- I added `erc20_` as a prefix to the erc20 tables so there's no confusion when querying them
- Long-term I think we (or other contributions) should spend more time thinking about tests and the non-compliant part of the spell to make it as robust as possible
- There are very very small discrepancies (less than 0.00001) when comparing the MATIC balances to PolygonScan for some wallets, I think it must be due to some gas spent we're missing in calcs. Could be something to investigate in the long-term as well

PR #4265, opened 2023-09-06
Polygon erc721/1155 transfers

tomfutago:


henrystats:
@tomfutago thank you sooo much for working on these sir.. really excited for when we finally have balances tables across all evm chains and also for NFTs & fungible tokens.

I wanted to tag you as well that I had started working on migration + creating balances models for erc20 tokens & mainnet tokens for all the evm chains to avoid duplication of work.

tomfutago:
> @tomfutago thank you sooo much for working on these sir.. really excited for when we finally have balances tables across all evm chains and also for NFTs & fungible tokens.
> 
> I wanted to tag you as well that I had started working on migration + creating balances models for erc20 tokens & mainnet tokens for all the evm chains to avoid duplication of work.

No worries and thanks for letting me know @henrystats 👍 

I actually also worked on (and already submitted for review) erc20 balances but only for Celo chain. Out of curiosity - with other evm chains are you using previous approach aka views for balances? In my tests they had really poor performance on dunesql, so I went with building incremental load, but even for a small chain like Celo - I had to split my PR into stages as hourly balances build kept timing out.

I will refrain then from looking at erc20 balances for other chains - will look at erc721/1155 balances instead, deal? ;)

henrystats:
😊 yay @tomfutago ! Niceee I saw that too soo I'd only build balances fungible for the other evm chains. So yeah sounds like a good plan! really excited for when this all comes together!

Yes I used a similar approach as before but I switched the unique transfer id to use unique keys and made the transfers table incremental. 

Everything seems to work fast for all the models I built soo far except for bnb which is taking about 30mins on the transfers base table tests (bnb has billions of erc20 transactions).

So I'm thinking of suggesting removing the **tests* * for the transfers table after review and it's confirmed that the spells works correctly...

But the balances and other models run fine and in a few mins.

henrystats:
Also I switched from union to UNION ALL in the base table transfers because the two tables are already distinctive there's no need to use a UNION and check for unique values.

tomfutago:
Good point on `union all` 👍 Yep, transfers incremental - I'm doing that too on Celo, but I'm surprised that balances are working fast on views - both daily and hourly are adding tons of additional records to fill the timing gaps between the transfers. I put some numbers in my comment on Celo balances - it goes into over 200 bln for hourly balances and I'd assume it's way bigger for other chains..

jeff-dude:

#4521 
#4523 
#4633 
#4698

PR #4267, opened 2023-09-06
Enable Vacuum on DuneSQL Spells

a-monteiro:


a-monteiro:

a-monteiro:
recheck

jeff-dude (approved):
sounds good. we will need to coordinate the timing of the merge to turn off `state:modified` step to avoid large full refresh in prod.

we also won't be able to see this run successfully in CI tests due to number of models affected. did you manually test this command on dunesql on one or two spells? or are we confident this is simple change that'll be fine?

a-monteiro:
> sounds good. we will need to coordinate the timing of the merge to turn off `state:modified` step to avoid large full refresh in prod.
> 
> we also won't be able to see this run successfully in CI tests due to number of models affected. did you manually test this command on dunesql on one or two spells? or are we confident this is simple change that'll be fine?

This won't have any effect until the job to vacuum the tables is deployed. I haven't tested it since it's a very simple change, but I'll give it a try in a couple of spells and I'll let you know.

jkylling (approved):

PR #4272, opened 2023-09-06
Avalanche ERC20 & AVAX balances - WIP

henrystats:


tomfutago:
Hey @henrystats - just to add to the conversation we had earlier on and going back to my curiosity around balances view performance - I tried this simple test based on the models you have built for avalanche (I realise this is wip tho):

`-- incremental model, runtime of this query < 1 sec:
select count(*) from test_schema.git_dunesql_2d7466d8_transfers_avalanche_c_erc20_agg_hour

-- view, runtime < 1 sec:
select count(*) from test_schema.git_dunesql_2d7466d8_transfers_avalanche_c_erc20_rolling_hour

-- view, **waited for 15 minutes** and cancelled the query:
select count(*) from test_schema.git_dunesql_2d7466d8_balances_avalanche_c_erc20_hour`

henrystats:
Ah, @tomfutago I see what you mean. My original reply assumed that you were referring to run time in GitHub ci? since you said you needed to seperate the transfers and balances model. 

I think it'd be important for us to add a filter in the balances model that removes balances with amounts less than 0, most of those are just overflows I think and not relevant. There's probably a few million entries in each models that are not useful. 

When I tested the models on dune.com, I filtered for addresses which I assume is what most people would use the balances table for? 

But do you have suggestions on how we can make this run faster? there's way too many entriessssss

henrystats:
I wonder if it's because I'm filtering in my test query on dune.com with VARBINARY casts that's making it run faster? 

I also don't know if an incremental table would run any faster considering that both would still have large volume of data which is why I didn't spend time thinking about how this could be built incrementally.

henrystats:
woah! oke this is taking a longggggg time [query](https://dune.com/queries/3000939) trying to get usdt supply on avalanche

henrystats:
I'm running with "large" though and USDt on avalanche has over 9million transactions so I'm guessing there are a lot of addresses!

henrystats:
about 1million unique addresses received usdt... soo there are probably a lot of rows with balances < 0 that's inflating the table! currently only about 270k holders...

henrystats:
table runs pretty quick for smaller tokens!

henrystats:
just realized that I'm spamming here, apologies... so here's a summary:

But @tomfutago looking forward to hear what you think... The tables are very big, filtering for wallet_address runs fast. 

Filtering for token_address runs fast as well except when the token is really big and has a lot of entries (like usdc, usdt & the likes)

We should also filter for where balances <= 0 to remove billions of rows that are not needed in the balances table.

tomfutago:
Hmm, all great findings and valid questions. I'm wiped out now so will look at it tomorrow in more depth, but one immediate thought: not entirely sure about removing 0 balances - it will definitely get rid of billions of records, but I thought the main idea behind the balances spell was to have this daily/hourly continuity, isn't it? With all 0 records removed it'll create gaps in history. On the other hand they're all 0 anyway, so maybe you'll right 🤷‍♂️

henrystats:
A'ighty! looking forward to tomorrow! Yeah that's my thought process too but I'm guessing when the ethereum balances spell was built, zero values were intentionally not excluded soo I think I agree with you that it's probably best to leave that.

tomfutago:
Some follow up comments:

1. Back to 0 balances:
- cons: lots of them, pros: mentioned before history continuity and easier to query specific-point-in-time balances
- in some cases, 0s will show for the whole history only because user transferred in & out token only once within the same day/hour. I'd say that's good because that token touched that specific wallet once so should be included.
- with potential exclusion of 0s it still need to be done carefully so that at least first 0 stays in (when user sells/transfers out token) to show the final balance
- overall, I'd vote to keep 0s in (if it's up for voting 😉)

2. Unfortunately all Avalanche test models build yesterday are gone by now so can't test performance comparison, so just few thoughts:
- on view vs incremental model having the same amount of data to query topic - agreed, but I've just re-run my Celo incremental daily balances and this runs under 1s (4.8bln records):

`select count(*) from test_schema.git_dunesql_6988471b_balances_celo_erc20_day`

- I think the part where view performance struggles is this join:

`FROM 
daily_balances b
INNER JOIN 
days d 
    ON b.day <= d.day 
    AND d.day < b.next_day`

obv same (and even much more so) for hourly view. My understanding of Trino is in early stages, but I don't think we can add an index on `day` column to help it out in any way, so again - I'd vote for incremental balances rather than views as they seem to perform better - I'm open to be proven wrong tho.

p.s. sorry for the wall of text 😂

tomfutago:
Ooh, good point with window functions.. hmm, I'm actually adding recency_index in the balances, and you're right, with incremental load that won't work properly. I was hoping to use recency_index to pull latest balance.. Will have a think on how to fix this (first thought is to simply remove it really, and tweak how latest balance is retrieved, but maybe there's a better way).

And you're also right on rolling transfer scripts - they should just be set as views. Thanks for checking!

And yep, rolling day/hour transfers is the key to pull balances.

henrystats:
@antonio-mendes there's also an issue here with AVAX (some missing transactions which also involves a deposit (maybe bridging)) will check today and see if I can fix as well.

jeff-dude:

#4521 
#4523 
#4633 
#4698

PR #4274, opened 2023-09-06
Add PNDC to Tokens and Price

timmaycastags:
This request is to add the contract for PNDC to the ETH Tokens and Price tables.


timmaycastags:

models/prices/ethereum/prices_ethereum_tokens_legacy.sql
jeff-dude, line 1: for prices, we no longer use the legacy files / engine. please add this value to the file in the same directory that is non-legacy, dunesql version

you can then remove this change from this file 👍

jeff-dude (changes requested):
no need to apologize for exploring spellbook -- thank you for contributing!
i've left one comment for change below

models/prices/ethereum/prices_ethereum_tokens_legacy.sql
timmaycastags: Thanks Jeff,  The non-legacy price file already has the pndc coin added.  I will remove the legacy update and no change for price is needed.

jeff-dude: sounds good!
looks like your second commit re-added it back. please do look to remove, so there are only two files in the PR (tokens files)

PR #4281, opened 2023-09-07
Fix chainlink ocr incremental error on gas_daily and request_daily

anon-r-7:
Fixes an issue on Chainlink OCR with incremental views for <chain>_gas_daily and <chain>_request_daily where the "AND" clause incorrectly results in fewer results than expected and the "OR" clause corrects this by pulling transactions from fulfilled OR reverted transactions, with a FULL OUTER JOIN.


jeff-dude (approved):
sounds good!

PR #4284, opened 2023-09-07
Fix bug causing wrong token_bought_amount_raw

couralex6:
Fixing a typo that slipped through while migrating shibaswap. I replaced `amount0Out` with `amount1Out`, causing all `token_bought_amount` to be 0 in `shibaswap_ethereum.trades`


jeff-dude (approved):
good catch!

lets save to merge tomorrow morning, as prod has some recent merges from earlier

PR #4296, opened 2023-09-08
Flashloans Migration - #3 - Dydx - Ready for Review 

henrystats:

### DIFF
--- a/models/dydx/ethereum/dydx_ethereum_flashloans.sql
+++ b/models/dydx/ethereum/dydx_ethereum_flashloans.sql
@@ -1,5 +1,7 @@
 {{ config(
-      alias = alias('flashloans')
+    tags = ['dunesql']
+      , alias = alias('flashloans')
+      , partition_by = ['block_month']
       , materialized = 'incremental'
       , file_format = 'delta'
       , incremental_strategy = 'merge'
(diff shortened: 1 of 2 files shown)


models/dydx/ethereum/dydx_ethereum_flashloans.sql
```diff
@@ -14,23 +16,23 @@
 WITH flashloans AS (
     SELECT d.evt_block_time AS block_time
         , d.evt_block_number AS block_number
-        , CAST(get_json_object(get_json_object(d.update, '$.deltaWei'), '$.value') AS DECIMAL(38,0)) AS amount_raw
+        , CAST(json_extract_scalar(JSON_PARSE(json_extract_scalar(d."update", '$.deltaWei')), '$.value') as double) AS amount_raw
```
jeff-dude, line 19: on dunesql, should we cast this as `uint256` instead? similar to `dex.trades` and `nft.trades` lineage?

this would require the casting change on all models in the flashloans lineage

cc @hildobby

jeff-dude: may be worthwhile to quickly scan through each column and double check types. i think leaving `fee` as double makes sense for instance.

jeff-dude: ah, i see, since you are dividing later in the query, it becomes double anyway

jeff-dude: no problem -- we should be good then

jeff-dude (approved):

PR #4318, opened 2023-09-12
[Questions 🤔] Celo nft spells

tomfutago:


tomfutago:
I'm looking at adding nft_celo spells and have few questions:
- `nft_transfers` looks to be migrated for existing chains (?) but `unique_transfer_id` column looks like a hash (?) - in other words doesn't match old logic pre-dunesql. I went with removing `unique_transfer_id` column completely and replaced it with `unique_key` in config + `dbt_utils.unique_combination_of_columns` test in schema.yml as per other migrated spells - is that the way to go?
- also, `nft_transfers` used `anti join` logic, which I replaced with left join + where clause, but is it needed for incremental merge strategy? Might improve performance with pre-filtered records to merge perhaps?
- `nft_approvals` - old logic has: `cast(approved as boolean) as approved`, but approved is varbinary column - should this column be simply left as is?

tomfutago:
ah, there's already `nft_celo_transfers` spell in chains model - so it's actually not needed in nft/celo/ ?

0xRobin:
Hi @tomfutago, thanks for taking a look at this!

> * `nft_transfers` looks to be migrated for existing chains (?) but `unique_transfer_id` column looks like a hash (?) - in other words doesn't match old logic pre-dunesql. I went with removing `unique_transfer_id` column completely and replaced it with `unique_key` in config + `dbt_utils.unique_combination_of_columns` test in schema.yml as per other migrated spells - is that the way to go?
There is indeed already a `nft_celo_transfers.sql` model that is migrated. This uses the macro to make the model so it's consistent across all chains. However the model is missing in `nft_transfers.sql`, we should add celo there to the list.

> * `nft_approvals` - old logic has: `cast(approved as boolean) as approved`, but approved is varbinary column - should this column be simply left as is?
You're right here, seems like this column was misinterpreted when the model was initially made. Side note: none of the other `_approvals` models have been migrated yet to dunesql.

models/nft/celo/nft_celo_aggregators.sql
```diff
@@ -0,0 +1,16 @@
+{{ 
+    config(
+        tags = ['dunesql'],
```
0xRobin, line 3:
```suggestion
        tags = ['dunesql'],
        schema = 'nft_celo',
```

models/nft/celo/nft_celo_approvals.sql
```diff
@@ -0,0 +1,75 @@
...
+  owner as address,
+  'erc721' as token_standard,
+  cast(true as boolean) as approval_for_all,
+  contract_address,
+  cast(null as uint256) as token_id,
+  approved, --cast(approved as boolean) as approved,
+  evt_tx_hash as tx_hash,
+  evt_index
+from {{ source('erc721_celo', 'evt_ApprovalForAll') }}
+from erc721_celo.evt_ApprovalForAll
```
0xRobin, line 51:
```suggestion
from {{ source('erc721_celo', 'evt_ApprovalForAll') }}
```

0xRobin (commented):
made some extra comments here

models/nft/celo/nft_celo_approvals.sql
```diff
@@ -0,0 +1,75 @@
+{{ 
+    config(
+        tags = ['dunesql'],
+        alias = alias('approvals'),
```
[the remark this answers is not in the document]
tomfutago: Cool, will keep this in mind going forward!

models/nft/celo/nft_celo_approvals.sql
tomfutago: Yep, looks like I left it in testing state, thanks for catching it early on 👍 Maybe worth me taking a look at converting this to nft_approvals macro (following similar logic as for nft_transfers)?

tomfutago:
Thanks @0xRobin for your comments and suggestions!

models/nft/celo/nft_celo_approvals.sql
0xRobin: that would be amazing!

tomfutago:
> > * `nft_approvals` - old logic has: `cast(approved as boolean) as approved`, but approved is varbinary column - should this column be simply left as is?
> 
> You're right here, seems like this column was misinterpreted when the model was initially made. Side note: none of the other `_approvals` models have been migrated yet to dunesql.

hmm, I checked this again (hopefully properly this time 🙄) and it looks like it's only `erc721_chain.evt_Approval` where approved is varbinary, but on `erc721/1155_chain.evt_ApprovalForAll` it's boolean.. dunesql doesn't like casting varbinary to boolean tho - so in order to pull all events into `union all` would be casting approved to varchar a way to go?

0xRobin:
> looks like it's only `erc721_chain.evt_Approval` where approved is varbinary, but on `erc721/1155_chain.evt_ApprovalForAll` it's boolean.

I dug op the original EIP and this is actually expected.
So we have a bit of mixed usage of "approved" where it first means the address that is approved and in the other event means if the operator address is approved or revoked for handling the NFTs. 
```solidity
    /// @dev This emits when the approved address for an NFT is changed or
    ///  reaffirmed. The zero address indicates there is no approved address.
    ///  When a Transfer event emits, this also indicates that the approved
    ///  address for that NFT (if any) is reset to none.
    event Approval(address indexed _owner, address indexed _approved, uint256 indexed _tokenId);

    /// @dev This emits when an operator is enabled or disabled for an owner.
    ///  The operator can manage all NFTs of the owner.
    event ApprovalForAll(address indexed _owner, address indexed _operator, bool _approved);
```
I suggest we add a new `operator` field (the address approved to handle the NFT).
on `evt_Approval` we would do:
```sql
select
  boolean 'True' as approved,
  approved as operator,
```
and on `evt_ApprovalForAll` this would be:
```sql
select
  approved,
  operator,
```

What do you think? Also tagging @hildobby as he originally created the approval spells.

tomfutago:
@0xRobin sounds good to me!

@hildobby - I'm going to create `nft_approvals` macro based on your original code, hope that's cool with you?

tomfutago:
Summary:
- added Celo to `nft_aggregators`
- added `nft_approvals` macro and applied to all evm chains (new additions: Celo and Base)


Notes on nft_approvals macro:

Going with what @0xRobin suggested:
- on evt_Approval:
```
  boolean 'True' as approved,
  approved as operator,
```

- on evt_ApprovalForAll:
```
  approved,
  operator,
```

... and my added assumption
- on erc1155 evt_ApprovalForAll:
```
  account as address
```
- on erc721 evt_Approval & evt_ApprovalForAll:
```
  owner as address
```

0xRobin (approved):
:heavy_check_mark: Great use of a macro @tomfutago!

tomfutago:
Thanks @0xRobin! 

One small follow up thought: wouldn't it be better to keep all model scripts in their original folders (per chain) rather than `chains` folder? If more macros get added and all models will go to this folder it might get messy, what do you think?

0xRobin:
> Thanks @0xRobin!
> 
> One small follow up thought: wouldn't it be better to keep all model scripts in their original folders (per chain) rather than `chains` folder? If more macros get added and all models will go to this folder it might get messy, what do you think?

100%, I'm planning to do some cleanup and move models around when all of the NFT migrations are done.

jeff-dude (approved):
nice work!

PR #4319, opened 2023-09-12
Update Creator Logic for smart wallet deployers

MSilb7:
Add method IDs for smart wallet transactions & add handling for deployers.

### DIFF
--- a/macros/models/contracts/contract_creator_project_mapping.sql
+++ b/macros/models/contracts/contract_creator_project_mapping.sql
@@ -69,7 +69,11 @@ SELECT *
         select 
             '{{chain}}' AS blockchain
             ,ct."from" as trace_creator_address
-            ,ct."from" as creator_address
+            ,CASE WHEN ct."from" IN (SELECT creator_address from {{ref('contracts_deterministic_contract_creators')}} ) THEN t."from" --tx sender
+              WHEN aa.project = 'Gnosis Safe' THEN t.to --smart wallet
+              WHEN aa.project = 'ERC4337' THEN bytearray_substring(t.data, 145,20) --smart wallet sender
+              ELSE ct."from"
+            END as creator_address
             ,ct.address as contract_address
             ,ct.block_time as created_time
             ,ct.block_number as created_block_number
@@ -108,6 +112,8 @@ SELECT *
             {% if is_incremental() %}
             and sd.created_time >= date_trunc('day', now() - interval '7' day)
             {% endif %}
+          left join {{ ref('evm_smart_account_method_ids') }} aa 
+            ON aa.method_id = bytearray_substring(t.data,1,4)
           where 
             1=1
             {% if is_incremental() %}
@@ -627,6 +633,8 @@ WHERE contract_order = 1
       ,c.code
       ,c.code_deploy_rank_by_chain
       ,CASE WHEN c.trace_creator_address = c.created_tx_from THEN 1 ELSE 0 END AS is_eoa_deployed
+      ,CASE WHEN c.top_level_tx_method_id in (SELECT method_id FROM {{ ref('evm_smart_account_method_ids') }}) THEN 1 ELSE 0 END AS is_smart_wallet_deployed
+      ,CASE WHEN c.trace_creator_address in (SELECT creator_address from {{ref('contracts_deterministic_contract_creators')}} ) THEN 1 ELSE 0 END AS is_deterministic_deployer_deployed
     from cleanup as c 
     left join {{ ref('contracts_project_name_mappings') }} as dnm -- fix names for decoded contracts
       on lower(c.contract_project) = lower(dnm.dune_name)
--- a/models/contracts/chains/contracts_chains_schema.yml
+++ b/models/contracts/chains/contracts_chains_schema.yml
@@ -104,6 +104,13 @@ models:
       - &is_eoa_deployed
         name: is_eoa_deployed
         description: "Was the contract deployed by an EOA (test if the trace creator = the transaction sender)"
+      - &is_smart_wallet_deployed
+        name: is_smart_wallet_deployed
+        description: "Was the contract deployed by a smart wallet (i.e. Gnosis Safe, ERC4337 wallet)"
+      - &is_deterministic_deployer_deployed
+        name: is_deterministic_deployer_deployed
+        description: "Was the contract deployed by a deterministic deployer factory (i.e. create2)"
+
       - &is_updated_on_last_run
         name: is_updated_on_last_run
         description: "Was this contract updated on the last run (used for incremental table building)"
@@ -151,6 +158,8 @@ models:
       
   #     - *code_deploy_rank_by_chain
   #     - *is_eoa_deployed
+  #     - *is_smart_wallet_deployed
+  #     - *is_deterministic_deployer_deployed
   #     - *is_updated_on_last_run
 
   # - name: contracts_base_contract_creator_project_mapping
@@ -196,6 +205,8 @@ models:
       
   #     - *code_deploy_rank_by_chain
   #     - *is_eoa_deployed
+  #     - *is_smart_wallet_deployed
+  #     - *is_deterministic_deployer_deployed
   #     - *is_updated_on_last_run
 
   # - name: contracts_bnb_contract_creator_project_mapping
@@ -241,6 +252,8 @@ models:
       
   #     - *code_deploy_rank_by_chain
   #     - *is_eoa_deployed
+  #     - *is_smart_wallet_deployed
+  #     - *is_deterministic_deployer_deployed
   #     - *is_updated_on_last_run
 
   # - name: contracts_celo_contract_creator_project_mapping
@@ -286,6 +299,8 @@ models:
       
   #     - *code_deploy_rank_by_chain
   #     - *is_eoa_deployed
+  #     - *is_smart_wallet_deployed
+  #     - *is_deterministic_deployer_deployed
   #     - *is_updated_on_last_run
 
   # - name: contracts_arbitrum_contract_creator_project_mapping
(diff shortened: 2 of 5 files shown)


jeff-dude:
more merge conflicts to help resolve, @MSilb7 

then is this in a spot you'd like to have it reviewed?

MSilb7:
resolved conflicts, should be good now

MSilb7:
resolved conflicts, should be good now @jeff-dude

jeff-dude:
considering this on hold until we can finalize #4582 

we are having trouble in prod on these spells and are proposing ideas there to help.

jeff-dude:
@MSilb7 plz check out the latest gh discussion on migration being completed and things you can stop doing on new/legacy spells:
https://github.com/duneanalytics/spellbook/discussions/4662

jeff-dude:
also, can we close #4582 to focus efforts on this PR for re-enabling these spells?

MSilb7:
> also, can we close #4582 to focus efforts on this PR for re-enabling these spells?

yeah, does it make sense to still use the incremental predicate here?

jeff-dude:
> > also, can we close #4582 to focus efforts on this PR for re-enabling these spells?
> 
> yeah, does it make sense to still use the incremental predicate here?

i think so. let's filter down both the source and target as much as possible, which that helps us do

MSilb7:
> > > also, can we close #4582 to focus efforts on this PR for re-enabling these spells?
> > 
> > 
> > yeah, does it make sense to still use the incremental predicate here?
> 
> i think so. let's filter down both the source and target as much as possible, which that helps us do

Got it, yeah we should apply it to the base level & iterate_creators, but likely not the final mapping one.

MSilb7:
I couldn't find it in dbt docs, but can we do incremental predicates on multiple fields? i.e. block_time from transactions and creation_traces, or can we only pick one?

jeff-dude:
> I couldn't find it in dbt docs, but can we do incremental predicates on multiple fields? i.e. block_time from transactions and creation_traces, or can we only pick one?

good question. we'd likely need to alter the macros built for this to handle array and/or dict type inputs to handle multiple. maybe @aalan3 will have interest to try and help there, since he helped build it

MSilb7:
👀👀👀👀👀 @jeff-dude @aalan3 

image

jeff-dude:
> 👀👀👀👀👀 @jeff-dude @aalan3
> 
> image

great progress here!

first thought, at first glance:
- historical is faster than incremental -- do we have room to optimize or should we potentially consider this a daily table instead?
  - one thing to try: use the new incremental predicate on filters, maybe that'll speed things up from 7 days --> 1 day

MSilb7:
Making some small edits to remove the prod_exclude tag & clean up column names (i.e we used both `creation_` and `created_` so I unified to `created_`)

macros/models/contracts/find_self_destruct_contracts.sql
```diff
@@ -31,14 +31,14 @@ FROM (
           = (CASE WHEN cardinality(sd.trace_address) = 0 then cast(-1 as bigint) else sd.trace_address[1] end)
       and sd.type = 'suicide'
       {% if is_incremental() %}
-      and sd.block_time >= date_trunc('day', now() - interval '7' day)
+      AND {{ incremental_predicate('sd.block_time') }}
       and cr.address NOT IN (SELECT contract_address FROM {{this}} ) --ensure no duplicates
```
jeff-dude, line 35: is this really needed?
`and cr.address NOT IN (SELECT contract_address FROM {{this}} ) --ensure no duplicates`

address is already the unique key on the spell, so the merge statement built will ensure no dupes since it joins on that column and only inserts/updates as needed

i wonder if this extra filter against itself is causing a slowdown

macros/models/contracts/find_self_destruct_contracts.sql
```diff
@@ -63,7 +63,7 @@ FROM (
       AND sds.type = 'suicide'
       AND sds.address IS NOT NULL
       {% if is_incremental() %}
-      and sds.block_time >= date_trunc('day', now() - interval '7' day)
+      AND {{ incremental_predicate('sds.block_time') }}
       and cr.address NOT IN (SELECT contract_address FROM {{this}} th WHERE th.blockchain = '{{chain}}' ) --ensure no duplicates
```
jeff-dude, line 67: same question here, on the same filter condition

models/contracts/chains/contracts_base_contract_creator_project_iterated_creators.sql
```diff
@@ -0,0 +1,15 @@
+ {{
+  config(
+        schema = 'contracts_base',
+        alias = 'contract_creator_project_iterated_creators',
+        materialized ='incremental',
```
jeff-dude, line 5: these spells are incremental, yet the macro has no incremental logic. what is the reason behind that?

jeff-dude:
i'm having a hard time wrapping my head around the runtimes. it still worries me to see the incremental runs take 4x as long as full historical. with the extra complicated logic included to make incremental work, adding runtime, is it even worth having these as incremental? have you considered just regular tables which full refresh each time? or can we continue to find ways to make incremental run faster (i.e. when joining to {{this}}, always filter that down to last day or so to match source)

MSilb7:
> i'm having a hard time wrapping my head around the runtimes. it still worries me to see the incremental runs take 4x as long as full historical. with the extra complicated logic included to make incremental work, adding runtime, is it even worth having these as incremental? have you considered just regular tables which full refresh each time? or can we continue to find ways to make incremental run faster (i.e. when joining to {{this}}, always filter that down to last day or so to match source)

We could make them all not incremental?

Another alternative could be if the final `contracts.contract_mapping` is not incremental, but is instead re-built every time (having this be materialized vs a few is what I was going for). I can try this, it removes the need to join to {{this}}.

macros/models/contracts/find_self_destruct_contracts.sql
MSilb7: Without this, we'd get duplicate contracts in the CTEs, which may screw up some of the joins and ranking columns

models/contracts/chains/contracts_base_contract_creator_project_iterated_creators.sql
MSilb7: copy and paste issue, making a materialized table (unless it should be something else)?

macros/models/contracts/find_self_destruct_contracts.sql
MSilb7: actually, we should be able to remove this one

MSilb7:
**open questions from @jeff-dude :**
- what is the starting point for each blockchain lineage, that ultimately feeds the cross-chain spell? what is second step? third? and so on. expectation is each blockchain follows the same lineage
- `contract_creator_project_base_level` -> `contract_creator_project_iterated_creators` -> `contract_creator_project_mapping`
  - First step is collect new contracts and union this with existing contracts to get our starting set & check which existing contracts need to be iterated upon in addition to new ones, in order to identify their deployer - This happens in `contract_creator_project_base_level` macro.
    - There are refs here to `contracts_deterministic_contract_creators` which stores known deterministic deployers (we could solve this if we can ingest creation traces as create2, create3, etc) & `evm_smart_account_method_ids` which stores smart wallet transaction methods (which ideally don't change once this gets standardized). It's not possible logic-wise to move these in to the last step, since this influences the creator address of a contract.
  - Second step is to take this list, and for contracts where we need to find the deployer, we iterate up to X times (3 now) to check if this contract was deployed by another contract (which requires the whole list) over and over until we get to ~some deployer list. This is only ~5 loops for now, ideally we figure out how to infinitely loop up to ensure accuracy. This happens in the `contract_creator_project_iterated_creators` macro.
    - Similar to the first step, this also has a ref to `contracts_deterministic_contract_creators`
  - Third step is to take our deployed contract list, append any missing pre-deploys, and map the contracts to metadata tables (refs). This includes tokens tables, deployer <> project mapping, project name overrides, etc. One complexity: Since Dune does not yet have a table listing all tokens for all chains, we need to look through events tables to identify whether a contract is a token or not. Once Dune adds tokens tables, we can simplify this part. This happens in `contract_creator_project_mapping`
  - We generate files for these 3 macros for each chain, but we only union the last step `contract_creator_project_mapping` to be cross-chain (first two steps are split to optimize runtime, but aren't useful on their own)
- for each step, what is materialization strategy?
  - Steps 1 & 3 are incremental builds. Since step 2 returns the entire list every time, it's just materialized as a table. Unifying across chain is also materialized as a table, so that we don't have to track which contracts in each chain were updated on each incremental build (we had this logic in before, but it caused incremental builds to be ~10 mins slower than initial builds).
- if materializing into a table, are you reading from upstream spells that are hardcoded metadata? i.e. ref() usage in the macros, but the ref() is a hardcoded metadata spell that is updated relatively frequently. we are trying to avoid that. the goal is: if materializing a step in the process, try to keep it to just base/decoded table data, OR an upstream spell that contains only base/decoded data, so the lineage is not as complex and overlapping 
- the ideal state is to wait until the very end of lineage to join to metadata that is constantly updated, and ideally have it as a view, where the intermediate tables on the way are materialized without it
  - There are infrequently updated metadata refs in steps 1 & 2, but the frequently updated refs (i.e. deployer <> project name mapping) are saved until step 3, after the heavy compute steps have already completed.

MSilb7:
Making mods per https://github.com/duneanalytics/spellbook/issues/4637

MSilb7:
Migrating the whole contracts folder - note that this will make initial and incremental builds in tests take longer since additional spells are impacted.

MSilb7:
Maybe we close this and instead use this to https://github.com/duneanalytics/spellbook/pull/4747 - which also adds OP and removes the complexity of re-running old models.

PR #4324, opened 2023-09-13
Useful macros for Balancer

viniabussafi:
This PR creates a macro (get_balancer_link) and edits two others (get_chain_explorer_address, get_chain_explorer_tx_hash) that would be useful for us at Balancer (e. g. this [query](https://dune.com/queries/2632759/4372295)).

@mendesfabio @thetroyharris 


antonio-mendes (approved):
Lgtm @viniabussafi ! Merging

macros/public/get_chain_explorer_link.sql
```diff
@@ -0,0 +1,22 @@
+{% macro get_chain_explorer_link() %}
```
antonio-mendes, line 1: Actually @viniabussafi I think this exists already - https://github.com/duneanalytics/spellbook/blob/9c360192574f965f7f2e90c618efd259affba8df/macros/public/get_chain_explorer_address.sql#L1 .

Can you take a look and if it's the same thing indeed can you update the PR to just add the `get_balancer_link macro`?

antonio-mendes (changes requested):
Requested change as we already have a macro to `get_chain_explorer` for addresses and transactions

macros/public/get_chain_explorer_link.sql
viniabussafi: hey, @antonio-mendes! The only difference at this point is that get_chain_explorer_address returns an URL but not a proper link to the chain explorer
image

macros/public/get_balancer_link.sql
```diff
@@ -0,0 +1,5 @@
+{% macro get_balancer_link() %}
+      create or replace function get_balancer_link(chain_ STRING, address STRING)
+      returns STRING 
+      return concat('<a href="','https://app.balancer.fi/#/',chain_,'/pool/', address, '"target ="_blank">', 'balancer')
```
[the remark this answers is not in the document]
viniabussafi: nice catch, thanks

macros/public/get_chain_explorer_link.sql
thetroyharris: The returning of a link is nice, my only suggestions is to use `https://etherscan.io/search?f=0&q=` as the base link and build off of that. The link I posted will return the proper page for a block, token/eoa, and tx. Just double check that it works for the chains you have listed.

I would also suggest changing the name to `get_link` or something. The shorter the better.

markusbkoch: Agree the scope is different and having a clickable link is useful; but maybe we can refactor the code to reuse the existing macro @viniabussafi ? Also, I think you're missing the closing `</a>` at the end of each record.

markusbkoch: I would also suggest making the link text an argument of the function for increased flexibility and renaming the arguments for readability

`get_chain_explorer_link(chain_ STRING, address_ STRING, link_text_ STRING)`

macros/public/get_balancer_link.sql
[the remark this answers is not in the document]
mendesfabio: that won't work for optimism too, should redirect to op.beets.fi

macros/public/get_chain_explorer_link.sql
markusbkoch: Actually nvm, just noticed `get_href` exists, so this function's purpose can be achieved with 

`get_href(get_chain_explorer_address(blockchain, address), link_text)`

antonio-mendes: Ah ok I get the point now. My question is more it can be confusing to have:
1. `get_chain_explorer_link`
2. `get_chain_explorer_address` (already in Prod)
3. `get_chain_explorer_tx_hash` (already in Prod)

Knowing that 2 and 3 are a subset of 1. But also if we just remove 2 and 3 to use just 1 we break existing queries.

I like the idea of a general `get_chain_explorer_link` that is clickable and intelligent enough to adapt, so the ideal think would be to have this macro, and maybe rework 2 and 3 so it "redirects" to 1 in the background? What do you think?

viniabussafi: I agree, @antonio-mendes. Will work on them and tag you once they´re done. Thanks a lot

macros/public/get_balancer_link.sql
```diff
@@ -0,0 +1,11 @@
+{% macro get_balancer_link() %}
+      create or replace function get_balancer_link(chain_ STRING, address STRING)
```
thetroyharris, line 2: Forgot to mention: 
replace `address` w `pool_id`.

macros/public/get_chain_explorer_link.sql
markusbkoch: > maybe rework 2 and 3 so it "redirects" to 1 in the background
Maybe I misunderstand, but I don't think we can change 2 and 3 because queries that use them expect them to return the URL and nothing more.

viniabussafi: Markus is right. Maybe we can just stick with `get_href` atm and only have get_balancer_link right now.

macros/public/public_macros_schema.yml
```diff
@@ -22,4 +22,6 @@ macros:
     arguments:
       - name: value
         type: column name or expression
-        description: "Valid base58 string"
\ No newline at end of file
+        description: "Valid base58 string"
+  - name: get_balancer_link
+    description: "Returns a link to access a specific pool on the Balancer app, based on its blockchain and address"
```
thetroyharris, line 27: sub address w/ pool id

macros/public/get_chain_explorer_tx_hash.sql
```diff
@@ -19,4 +19,4 @@
          when 'goerli' = chain_ then 'https://goerli.basescan.org/tx/' || CAST(hash_ AS VARCHAR)
```
thetroyharris, line 19: Is this goerli meant to for ethereum or base?

viniabussafi: not sure, tbh. just kept the original macro

macros/public/get_chain_explorer_link.sql
antonio-mendes: Makes sense to me! We can then think about improving existing macros going forward in individual PRs

antonio-mendes:
Pushed a small commit just adding newlines at the end of some files so no changes are done to files that are supposed to be left untouched.

If CI passes will merge this afterwards

antonio-mendes (approved):

PR #4327, opened 2023-09-13
Update Polygon and Matic balances design

antonio-mendes:
Changes implemented here:
- Renamed some variables:
  - `day` to `block_day`
  - `hour` to `block_hour`
- Added a new variable `block_month` to daily and hourly balances tables

Some notes @jeff-dude :
- Adding `block_month` will allow us to partition some of these spells easily if we decide to materialise them in the future. It also aligns nicely with dex.trades and nft.trades
- About prices, I didn't end up adding any extra CTE. I was checking [this](https://github.com/duneanalytics/spellbook/blob/9c360192574f965f7f2e90c618efd259affba8df/models/_sector/nft/trades/nft_ethereum_trades_beta.sql) example from nft.trades, and the more I thought about it the more I couldn't see how it was justified here. This is because of 2 reasons. First we only add prices as the last join in the balances lineage. Secondly, we join based on a specific minute of `prices.usd` so we already filter the prices there to a super small subset of rows.

### DIFF
--- a/models/balances/polygon/erc20/balances_polygon_erc20_day.sql
+++ b/models/balances/polygon/erc20/balances_polygon_erc20_day.sql
@@ -20,7 +20,7 @@ time_seq AS (
 
 days AS (
     SELECT 
-        time.time AS day 
+        time.time AS block_day 
     FROM time_seq
     CROSS JOIN unnest(time) AS time(time)
 ),
@@ -41,7 +41,8 @@ daily_balances as (
 
 SELECT
     b.blockchain,
-    d.day,
+    cast(date_trunc('month', d.block_day) as date) as block_month,
+    d.block_day,
     b.wallet_address,
     b.token_address,
     b.amount_raw,
@@ -52,12 +53,12 @@ FROM
 daily_balances b
 INNER JOIN 
 days d 
-    ON b.day <= d.day 
-    AND d.day < b.next_day
+    ON b.day <= d.block_day 
+    AND d.block_day < b.next_day
 LEFT JOIN 
 {{ source('prices', 'usd') }} p
     ON p.contract_address = b.token_address
-    AND d.day = p.minute
+    AND d.block_day = p.minute
     AND p.blockchain = 'polygon'
 -- Removes likely non-compliant tokens due to negative balances
 LEFT JOIN 
--- a/models/balances/polygon/erc20/balances_polygon_erc20_hour.sql
+++ b/models/balances/polygon/erc20/balances_polygon_erc20_hour.sql
@@ -19,7 +19,7 @@ years as (
 ),
 
 hours as (
-    select date_add('hour', s.n, y.year) as hour
+    select date_add('hour', s.n, y.year) as block_hour
     from years y
       cross join unnest(sequence(0, 9000)) s(n)
     where s.n <= date_diff('hour', y.year, y.year + interval '1' year)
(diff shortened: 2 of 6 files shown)


antonio-mendes:
FYI @henrystats , mainly about the `block_` suffix I added to the relevant timestamp fields. Don't worry if you don't do it on all your PRs I can do that at review time

models/balances/polygon/erc20/balances_polygon_erc20_day.sql
```diff
@@ -52,12 +53,12 @@ FROM
...
 days d 
-    ON b.day <= d.day 
-    AND d.day < b.next_day
+    ON b.day <= d.block_day 
+    AND d.block_day < b.next_day
 LEFT JOIN 
 {{ source('prices', 'usd') }} p
     ON p.contract_address = b.token_address
-    AND d.day = p.minute
+    AND d.block_day = p.minute
```
jeff-dude, line 61: based on how the sequence of days is built above, would this `block_day` value always have a timestamp of `00:00` since its truncated to day? then if so, we always join to prices on that particular minute? is that better than aggregating minute in prices up to day and averaging?

models/balances/polygon/erc20/balances_polygon_erc20_hour.sql
```diff
@@ -52,12 +53,12 @@ FROM
...
 hours d 
-    ON b.hour <= d.hour 
-    AND d.hour < b.next_hour
+    ON b.hour <= d.block_hour 
+    AND d.block_hour < b.next_hour
 LEFT JOIN 
 {{ source('prices', 'usd') }} p
     ON p.contract_address = b.token_address
-    AND d.hour = p.minute
+    AND d.block_hour = p.minute
```
jeff-dude, line 61: i guess similar question here, but at hourly level?

models/balances/polygon/erc20/balances_polygon_erc20_day.sql
antonio-mendes: Exactly that's how it's working now!

Hmm that's a good question, I'm not sure what's better I think both are fair as long as it's transparent as we describe it in the table description. This join like this is also cheaper computationally probably.

I am temped to leave this as is right now, and if we think the "mean" approach makes sense we should create a separate spell for "mean daily price" and join on that everywhere we use this logic, rather than creating CTEs on each spell.

antonio-mendes: What do you think @jeff-dude ?

models/balances/polygon/erc20/balances_polygon_erc20_hour.sql
antonio-mendes: yep 👍

models/balances/polygon/erc20/balances_polygon_erc20_day.sql
jeff-dude: works for me, just raising awareness 👍 look to call that out in the readme specifically, that'll help be transparent and that we thought of both approaches

jeff-dude (approved):

models/balances/polygon/erc20/balances_polygon_erc20_day.sql
antonio-mendes: Awesome @jeff-dude , thanks for thinking out loud with me on this

PR #4340, opened 2023-09-14
Create `First funded by` Macro  & Add `Celo` - Ready For Review

henrystats:


henrystats:
@tomfutago I wanted to tag you to this because I noticed that you're building a lot of celo spells (incase you had plans to work on this), I'm creating Macros for the addresses_events sector and will include new chains like base and celo.

henrystats:
@jeff-dude I'm not very sure I understand the use of Macros clearly but I was thinking they'd be useful in the addresses_events sector since macros are being used a lot for crosschain evm spells and this is one. 

I created a macro for first_funded_by and added `celo`, I was going to add `base` but I'm still trying to figure out the eth transfers / deposits in the balances spell for base soo when I figure that out, I'd come back to this to update and add base since most of those missing txns in the balances spell are usually the first_funding transaction, the data for base would be off for those addresses.

jeff-dude:
> @jeff-dude I'm not very sure I understand the use of Macros clearly but I was thinking they'd be useful in the addresses_events sector since macros are being used a lot for crosschain evm spells and this is one.
> 
> I created a macro for first_funded_by and added `celo`, I was going to add `base` but I'm still trying to figure out the eth transfers / deposits in the balances spell for base soo when I figure that out, I'd come back to this to update and add base since most of those missing txns in the balances spell are usually the first_funding transaction, the data for base would be off for those addresses.

you're right -- this is a great use case for a macro. you can think of it simply as reusable, parameterized code which centralizes the logic in one spot. that way, when you have various blockchain versions that do the same thing, this simplifies the build and support.

we don't force it upon users, but certainly welcome it! we try our best to lead by example, with some of our incredible resident wizards starting the trends.

jeff-dude:
i'll let @0xRobin give an initial review, as he has assigned himself, but thanks as always for contributing

henrystats:
I seee.. thank you very much ser @jeff-dude

dbt_project.yml
```diff
@@ -542,6 +542,8 @@ models:
         +schema: addresses_events_optimism
       polygon:
         +schema: addresses_events_polygon
+      celo:
+        +schema: addresses_events_celo
```
0xRobin, line 546: small nit:
This can be removed given you've added `schema = addresses_events_celo` in the model config (in `models/addresses_events/celo/addresses_events_celo_first_funded_by.sql`), which is the preferred way to set the schema. :+1: 
Note however that `models/addresses_events/celo/addresses_events_celo_first_funded_by_legacy.sql` is missing the schema in the config.

macros/models/_sector/addresses_events/addresses_events_first_funded_by.sql
```diff
@@ -0,0 +1,23 @@
+{% macro addresses_events_first_funded_by(blockchain, traces) %}
```
0xRobin, line 1:
```suggestion
{% macro addresses_events_first_funded_by(blockchain) %}
```
The traces variable wasn't used.

0xRobin (approved):
some minor comments, but nothing that blocks getting this in.
Great use of a macro and thanks for also updating the seed file. :+1:

dbt_project.yml
jeff-dude: good call out, a step towards cleaning the project out and making schema more obvious in the model itself. we can look to merge when you address this minor change @henrystats

henrystats: yes thank you ser @0xRobin! minor change addressed ser @jeff-dude

jeff-dude (approved):

PR #4345, opened 2023-09-15
OP & BASE Static Updates

MSilb7:
Add correct AERO token to Base
Add Velo V2 token to OP


jeff-dude (approved):

jeff-dude:
note if anyone merges while i'm out, this will trigger a large set of downstream refreshes in prod.

PR #4346, opened 2023-09-15
Fix BNB & Polygon Balances Hourly Sequences

henrystats:

### DIFF
--- a/models/balances/bnb/bep20/balances_bnb_bep20_hour.sql
+++ b/models/balances/bnb/bep20/balances_bnb_bep20_hour.sql
@@ -21,7 +21,7 @@ years as (
 hours as (
     select date_add('hour', s.n, y.year) as block_hour
     from years y
-      cross join unnest(sequence(0, 9000)) s(n)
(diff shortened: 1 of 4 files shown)


henrystats:
@antonio-mendes here's the pr to fix the hourly sequences, it should run fine because there's no unique test on the balances models, we only had tests on the transfer models. 

I was wondering why the bug wasn't picked up during the unique test and just saw that we don't have unique tests on the balances models and we only have on the base transfer tables (agg & transfers). 

The balances tables have a lot more rows than the transfers table since the balances are deduplicated over a period of time, running a tests on the hourly balances schema in my opinion would timeout depending on the limit (especially for models like bnb).

tomfutago:
Hey @henrystats good points. I think it's fairly safe to set uniqueness check for daily balances (on Celo, which granted has much less data) it completed within few minutes, but hourly timed out so had to turn it off. Unfortunately that's where the culprit for the bug was tho..

antonio-mendes (approved):

PR #4354, opened 2023-09-15
Migrate NFT Mints - Ready for review 

henrystats:


henrystats:
@0xRobin @hildobby, I was migrating the nft_wallet_metricss spell and noticed that it relied on nft_mints & nft_wash_trades (which you're already working on @hildobby) so I thought to migrate nft_mints. 

I made a couple of changes and wanted to write here as the pr is now ready for review (there's some weird issue with opensea's test although I feel that's something that would be easily fixed as I didn't make any changes to the logic). 


1. In both native mints models (optimism & ethereum) I needed to make a change to the nft_per_tx subquery as that was returning UINT256 '0' for the nft_minted_in_tx column, there are some nft's transferred (erc1155) with their amount column being zero which is causing the amount column that we're pulling out to break since we're dividing by that. 

I'm guessing due to the change in which the nft_transfers column was written, that changed, although I noticed similar issue when migrating some spells from spark-sql to dune-sql (the division by zero error).

2. I changed the unique key from "unique_trade_id" to using unique_keys, I needed to do some work around in the underlying models in nft_mints, adding evt_index to the models that didn't have it. Also did a tiny workaround in opensea.  I used to think we were moving towards the unique_keys route hence why I made the change but I'm surprised that it wasn't changed during the nft's migration so should I change that back?

3. In the logic of the native mint models (optimism & ethereum), there was a join to the erc20 table that was causing duplicates which caused some nft mints to have a wrong amount. Since SUM was used, it was inflating the amounts for some mints and even causing duplicates. I made a fix to only join with the erc20s table if there was actually no entry in ethereum.traces (since we're using a coalesce, I believe that is safe). 

These were the major changes I made and I'm excited to hear your feedbacks ser.

0xRobin:
> although I noticed similar issue when migrating some spells from spark-sql to dune-sql (the division by zero error).

I have also come across this problem a few times where it did seem like spark was somehow catching the division by zero and dunesql was failing on it. Can be really hard to debug too :(

> I used to think we were moving towards the unique_keys route hence why I made the change but I'm surprised that it wasn't changed during the nft's migration so should I change that back

During the migration we sometimes opted to not include any fixes (and translate the logic 1-to-1) just to reduce the complexity. No need to change this back though, moving away from a one column unique id to a set of unique columns is indeed what we're trying to get to.

> there's some weird issue with opensea's test
 I'll rerun the CI and will take a look.

henrystats:
Thank you very much ser @0xRobin, I think that's a great idea! I just updated the code to use row_number to filter. Not sure if I wrote it cleanly so pls lmk if I should rewrite it. I should probably have wrapped it in a cte.

0xRobin:
seems like the problem was the magiceden solana model containing null token_id values. I've defaulted that to 0.

0xRobin (approved):

hildobby:
> @0xRobin @hildobby, I was migrating the nft_wallet_metricss spell and noticed that it relied on nft_mints & nft_wash_trades (which you're already working on @hildobby) so I thought to migrate nft_mints.

Just to keep you posted, the ported nft.wash_trades was merged @henrystats

0xRobin:
@jeff-dude @antonio-mendes @couralex6 let's merge this in.

PR #4359, opened 2023-09-16
Celo native transfers and aggs

tomfutago:


tomfutago:
Summary:
- added Celo native transfers and aggregations
- `union all` in `transfers_celo_native` includes erc20 transfers for CELO token - afaik Celo doesn’t utilise WCELO the same way as other evm chains
- following on the above, updated erc20 transfers to exclude native CELO transfers to logically separate erc20 and native transfers

Test query to check native balances: https://dune.com/queries/3029589

jeff-dude:

#4521 
#4523 
#4633 
#4698

PR #4366, opened 2023-09-18
Create `erc20` transfers macro

henrystats:

### DIFF
--- a/macros/models/_sector/transfers/erc20/transfers_erc20.sql
+++ b/macros/models/_sector/transfers/erc20/transfers_erc20.sql
@@ -1,0 +1,98 @@
+{% macro transfers_erc721(blockchain, erc20_evt_transfer, wrapped_token_deposit, wrapped_token_withdrawal) %}
+
+WITH
+
+
+erc20_transfers  as (
+        SELECT 
+            'receive' as transfer_type, 
+            evt_tx_hash,
+            evt_index, 
+            evt_block_time,
+            to as wallet_address, 
+            contract_address as token_address,
+            CAST(value as double) as amount_raw
+        FROM 
+        {{ erc20_evt_transfer }}
+        {% if is_incremental() %}
+            WHERE evt_block_time >= date_trunc('day', now() - interval '3' Day)
+        {% endif %}
+
+        UNION ALL 
+
+        SELECT 
+            'send' as transfer_type, 
+            evt_tx_hash,
+            evt_index, 
+            evt_block_time,
+            "from" as wallet_address, 
+            contract_address as token_address,
+            -CAST(value as double) as amount_raw
+        FROM 
+        {{ erc20_evt_transfer }}
+        {% if is_incremental() %}
+            WHERE evt_block_time >= date_trunc('day', now() - interval '3' Day)
+        {% endif %}
+),
+
+wrapped_token_events as (
+        SELECT 
+            'deposit' as transfer_type, 
+            evt_tx_hash, 
+            evt_index, 
+            evt_block_time,
+            dst as wallet_address, 
+            contract_address as token_address, 
+            CAST(wad as double)as amount_raw
+        FROM 
+        {{ wrapped_token_deposit }}
+        {% if is_incremental() %}
+        WHERE evt_block_time >= date_trunc('day', now() - interval '3' Day)
+        {% endif %}
+
+        UNION ALL 
+
+        SELECT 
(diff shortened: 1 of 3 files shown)


henrystats:
@antonio-mendes, @jeff-dude I created macros for erc20 transfers & balances_fungible since the logic is the same across all the evm chains. 

@antonio-mendes can this be reviewed first so I could rewrite all the open erc20 balances spells!

I have some notes:

1. balances_fungible works for both mainnet tokens & erc20 tokens of each evm chain since the logic for both mainnet tokens & erc20 tokens are similar, the only difference is that we're not filtering for non compliant tokens & also mainnet tokens (evms that have erc20 transfers for their native token)

I wrote the logic of the macro so that in that case, it doesn't run the filter for both. 

`-- Removes likely non-compliant tokens due to negative balances
{% if balances_noncompliant %}
LEFT JOIN {{ balances_noncompliant }} nc
    ON rh.token_address = nc.token_address
{% endif %}
WHERE rh.recency_index = 1
{% if balances_noncompliant %}
AND nc.token_address IS NULL
{% endif %}
-- Removes mainnet token for chains which have erc20 transfer event for mainnet tokens (eg optimism)
{% if filter_mainnet_token %}
AND rh.token_address != {{filter_mainnet_token}}
{% endif %}`

2. In the hour balances, the current workaround for getting the hourly sequence doesn't work for when the chain hasn't been active for more than a year (I noticed this while working on the base spells). The previous setup however worked since it produced about 8.5k rows. 

I added an `is_more_than_year_ago` variable right after the `first_transaction_date` variable so that it's easy for anyone setting up a macro for a new chain. 

Also wrote the logic in the hour macro so that according to the variable above, it updates the code

`{% if is_more_than_year_ago %}
    -- This block will execute if first_transaction_date is more than a year ago
    years as (
        select year
        from (values (sequence(CAST('{{first_transaction_date}}' as timestamp), cast(date_trunc('year', now()) as timestamp), interval '1' year))) s(year_array)
        cross join unnest(year_array) as d(year)
    ),
    hours as (
        select date_add('hour', s.n, y.year) as block_hour
        from years y
        cross join unnest(sequence(1, 9000)) s(n)
        where s.n <= date_diff('hour', y.year, y.year + interval '1' year)
    ),
{% else %}
    -- This block will execute if first_transaction_date is within the last year
    time_seq AS (
        SELECT 
            sequence(
            CAST('{{first_transaction_date}}' as timestamp),
            date_trunc('hour', cast(now() as timestamp)),
            interval '1' hour
            ) AS time 
    ),
    hours AS (
        SELECT 
            time.time AS block_hour 
        FROM time_seq
        CROSS JOIN unnest(time) AS time(time)
    ),
{% endif %}`

I plan to reproduce it across all the chains, I've also tested it against the polygon model that's running on dune already and the results tally. I also checked the compiled code using ~dbt compile and everyything looks good!


looking forward to when this gets reviewed/merged so I can reproduce across all the evm chains, macros seem soo fun haha.

henrystats:
sorry that the comment is messed up, not sure how to escape "```" in GitHub, it keeps turning the rest of my comment to a code snippet.

henrystats:
@antonio-mendes I was wondering if this could be reviewed / merged first so I could replicate for the pending erc20 balances?

it also takes into account the new naming columns (block_day & block_hour) and I've tested it against the polygon model.

antonio-mendes:
Will review this next 👍 ! Thanks for going through this it's way more efficient and is gonna be nice so we can quickly have balances for new chains

antonio-mendes:
> sorry that the comment is messed up, not sure how to escape "```" in GitHub, it keeps turning the rest of my comment to a code snippet.

I managed to edit your comment so it's more readable 👍

antonio-mendes (changes requested):
@henrystats The logic of this is seeming overall right to me, however if we merge this PR it'll drop the current Transfers and Balances we have for Polygon and we don't want that as it'll break user queries.

Can you reflect in this PR or already, or a separate PR that we'd merge right after this one creating the Polygon ERC20 Transfers and Balances?

antonio-mendes:
Ah wait, ignore my previous review! I can see you've done what I asked already! Validating some results internally, and we can probably merge this today

antonio-mendes (approved):
Just validates a local run against what we have in prod for Polygon and indeed everything is adding up! Merging this, then will do more validations against Arbitrum and BNB that are live as well, but it looks really good right now 👍 !

PR #4376, opened 2023-09-20
Run status upload script on schedule

couralex6:
Runs the upload script twice a day or on demand to avoid hitting quota


couralex6:
Will merge tomorrow

couralex6:
No longer needed

PR #4377, opened 2023-09-20
Create `transfers_native` macro & add `balances_eth` - WIP

henrystats:


models/transfers/ethereum/eth/transfers_ethereum_eth_tfers.sql
```diff
@@ -0,0 +1,24 @@
+{{ config(
+    tags=['dunesql'],
+    schema = 'transfers_ethereum_eth',
+    materialized = 'incremental',
+    partition_by = ['block_month'],
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['transfer_type', 'tx_hash', 'trace_address', 'wallet_address', 'block_time'], 
+    alias = alias('eth_tfers'),
```
antonio-mendes, line 9: This should be just `eth` right? So we'd have a table show up as `transfers_ethereum.eth` on Dune

models/transfers/ethereum/eth/transfers_ethereum_eth_schema.yml
```diff
@@ -0,0 +1,148 @@
+version: 2
+
+models:
+  - name: transfers_ethereum_eth_tfers
```
antonio-mendes, line 4: Same here, rename this table to just `eth` I'd say

macros/models/_sector/transfers/transfers_native.sql
```diff
@@ -0,0 +1,182 @@
+{% macro transfers_native(blockchain, traces, transactions, native_token_address, genesis_balances=null, native_erc_transfers=null) %}
```
antonio-mendes, line 1: When you `native_token_address` here you are referencing the wrapped version of the currency right? So for example for Ethereum you'd use the address of WETH right?

antonio-mendes (changes requested):
Added a few notes! 

I would also like to ask you if you'd consider writing a README for this macro on the root of the directory where it's defined? Just because this work has to take into account quite a few caveats and it would be great to have that documented close to the code.

Thank you for your service once more!

models/transfers/ethereum/eth/transfers_ethereum_eth_tfers.sql
henrystats: hmm there's a transfers_ethereum_eth.sql model that @MSilb7 built... It's a different set up and has different columns as the table isn't deduped like we have for balances. 

That's why I added tfers to the end of the table.

henrystats:
> Added a few notes!
> 
> I would also like to ask you if you'd consider writing a README for this macro on the root of the directory where it's defined? Just because this work has to take into account quite a few caveats and it would be great to have that documented close to the code.
> 
> Thank you for your service once more!

No worries ser, I can add a readme!

models/transfers/ethereum/eth/transfers_ethereum_eth_tfers.sql
antonio-mendes: Ok we should converge to this new design if it makes more sense, I'll reach out to @MSilb7 to get his thoughts

macros/models/_sector/transfers/fungible/transfers_fungible_rolling_hour.sql
```diff
@@ -0,0 +1,16 @@
+{% macro transfers_fungible_rolling_hour(transfers_fungible_agg_hour) %}
```
MSilb7, line 1: Why is there no `transfers_fungible` without aggregation?

i.e. what if I needed to see each transfer event?

macros/models/_sector/transfers/transfers_native.sql
MSilb7: prices has native tokens (where blockchain = NULL) which may be simpler for this case.

macros/models/_sector/transfers/transfers_native.sql
```diff
@@ -0,0 +1,182 @@
...
+        WHERE (call_type NOT IN ('delegatecall', 'callcode', 'staticcall') OR call_type IS NULL)
+        AND success
+        AND CAST(value as double) > 0
+        AND to IS NOT NULL 
+        AND to != 0x0000000000000000000000000000000000000000
+        {% if is_incremental() %}
+            AND block_time >= date_trunc('day', now() - interval '3' Day)
+        {% endif %}
+
+        UNION ALL
```
MSilb7, line 25: This is missing ETH transfers from the L1<>L2 Bridge for Optimism Pre-Bedrock (June 6, 2023) (see existing OP ETH Transfers).

We'd need to use the placeholder ERC20 here, and ignore it for erc20 transfers.

Example here: https://optimistic.etherscan.io/tx/0xec2b3a5f852c0595dd167510ede630082ff08fa0bb99ad03ea9de81b3b135f40

MSilb7: ^ Not sure if any other L2 bridges had similar behavior

MSilb7: alternatively: does `native_erc_transfers` handle for this? hard to tell

macros/models/_sector/transfers/transfers_native.sql
```diff
@@ -0,0 +1,182 @@
...
+
+        UNION ALL 
+
+        SELECT 
+            'send' as transfer_type, 
+            tx_hash,
+            trace_address, 
+            block_time,
+            "from" as wallet_address, 
+            {{native_token_address}} as token_address,
```
MSilb7, line 33: is this inteded to be `0xeeeeeeee....`, `0xDeadDeAddeAddEAddeadDEaDDEAdDeaDDeAD0000`, or something else?

macros/models/_sector/transfers/transfers_native.sql
```diff
@@ -0,0 +1,182 @@
+{% macro transfers_native(blockchain, traces, transactions, native_token_address, genesis_balances=null, native_erc_transfers=null) %}
+
+WITH
+
+native_transfers  as (
+        SELECT 
+            'receive' as transfer_type,
```
MSilb7, line 7: Why the receive / send model versus "one row per transfer"?

I worry that this will make it tough to avoid double counting versus the existing transfers models (and would break all existing eth transfers models).

MSilb7 (commented):
This should not replace the existing eth transfers tables anywhere, since it would make all queries that reference them become incorrect.

macros/models/_sector/transfers/transfers_native.sql
```diff
@@ -0,0 +1,182 @@
...
+        AND success
+        AND CAST(value as double) > 0
+        AND "from" IS NOT NULL 
+        AND "from" != 0x0000000000000000000000000000000000000000 
+        {% if is_incremental() %}
+            AND block_time >= date_trunc('day', now() - interval '3' Day)
+        {% endif %}
+),
+
+{% if native_erc_transfers %}
```
MSilb7, line 47: is this wrapping and unwrapping, or something else?

MSilb7:
The initial intent of the ETH Transfers tables was to have a table the mirrored the `erc20_evt_transfer`, but for ETH.

macros/models/_sector/transfers/transfers_native.sql
```diff
@@ -0,0 +1,182 @@
...
+WITH
+
+native_transfers  as (
+        SELECT 
+            'receive' as transfer_type, 
+            tx_hash,
+            trace_address, 
+            block_time,
+            to as wallet_address, 
+            {{native_token_address}} as token_address,
```
MSilb7, line 12: why is there no "from" here?

How will I know what addresses the transfer is between?

macros/models/_sector/transfers/transfers_native.sql
```diff
@@ -0,0 +1,182 @@
+{% macro transfers_native(blockchain, traces, transactions, native_token_address, genesis_balances=null, native_erc_transfers=null) %}
+
+WITH
+
+native_transfers  as (
+        SELECT 
+            'receive' as transfer_type, 
+            tx_hash,
+            trace_address, 
+            block_time,
```
MSilb7, line 10: block_number would be useful as well

MSilb7:
Also - it seems odd to have genesis balances included here. Should it be it's own separate table? (i.e. do we consider genesis balance a transfer? Or just state?)

henrystats:
@MSilb7 I agree that the `transfers_native macro` I'm building in this pr shouldn't replace the one that's currently setup.

They have different use cases, this one is meant to be used for the `balances` spell hence why there's a lot of differences as we're only picking the columns & data in a way that makes it easier to build the balances spell. 


Btw @MSilb7 there's now a fungible.transfers spell that @hildobby built, and it's built similarly to how the current transfers_chain_eth is set up. 

It has similar columns as well (although some columns aren't included), I'd be happy to help update spells that reference  transfers_optimism_eth & transfers_ethereum_eth so that they use fungible transfers (and update the macro  if necessary) and then we can depecrate those models after since it's the same data in fungible transfers (although that also includes erc20 transfers).

macros/models/_sector/transfers/transfers_native.sql
henrystats: this for native token transfers that are erc20 token events (optimism & celo have those kind of events)

macros/models/_sector/transfers/transfers_native.sql
henrystats: responded in pr conversation.

macros/models/_sector/transfers/transfers_native.sql
henrystats: yeah, depends on what the wrapped native token address is, it should be the same with the native token transferred in erc20 events (for optimism and celo e.g)

macros/models/_sector/transfers/transfers_native.sql
henrystats: yeah native_erc_transfers handle for that.

macros/models/_sector/transfers/fungible/transfers_fungible_rolling_hour.sql
henrystats: this is meant to be used for the balances spell. for transfers without aggregation, there's a fungible.transfers spell

Apologies for the confusion!

henrystats:
@MSilb7 for genesis balances, it's included because there's no event of the transfer and we need it to accurately get the balance of a wallet. 

So counting it as a deposit (from the first block) works correctly. I don't think this macro should have any relation to the transfers_ethereum_eth spell you built. Its only relation would be the fungible.transfers macro built by @hildobby.

jeff-dude:

#4521 
#4523 
#4633 
#4698

PR #4380, opened 2023-09-20
[WIP] Oneinch misc updates 1

grkhr:
This is the part 1 of our major logic upgrade. We'll leave ar_ tables until we fully migrate to the new logic. Also PR's will be splitted to 2-3 as we'll be merging huge blockchains spearately.

New views:
- oneinch_exchange_contracts
- oneinch_fusion_farms
- oneinch_fusion_resolvers
- oneinch_methods

Changes:
- calls_transfers: new columns, new logic, renaming
- fusion_executors: refactoring


jeff-dude (approved):
thank you for the collaboration over a few PRs and finding a way to break out per blockchain (or at least removing larger ones) 🙌 

i think we should be all set here

PR #4381, opened 2023-09-20
cleanup + shutdown of legacy nft.approvals

0xRobin:
1. moved everything under new `/_sector/nft` directory structure
2. clean op schema.yml files + add column tests for all sub models
3. added a `remove` tag to stop running these models on legacy spark jobs.
4. added a `block_month` column and use that as partitioning to reduce the number of partitions


models/_sector/nft/approvals/legacy/nft_arbitrum_approvals_legacy.sql
```diff
@@ -1,6 +1,6 @@
 {{ config(
-	tags=['legacy'],
-	
+	tags=['legacy', 'remove'],
```
couralex6, line 2: What's the `remove` tag here?

0xRobin: > What's the `remove` tag here?
This tag will disable the legacy model (both on dbt cloud as in the CI)
It was introduced here: https://github.com/duneanalytics/spellbook/pull/4188

couralex6: Ah this was introduced when I was off. Looks great!

couralex6 (approved):

jeff-dude (approved):
nice! looking forward to all of us pushing through more `remove` tags

PR #4388, opened 2023-09-20
Update Bebop - ready for review

B1boid:
Ready for review

Add new smart contract for bebop-trades


B1boid:

Hosuke (approved):
LGTM.✅
Thank you @B1boid

B1boid:
> LGTM.✅ Thank you @B1boid
> 
> By the way do you want to add new seed for testing the trades from new contracts?

Done✅

jeff-dude (approved):
works for me, thank you!

for reference, @B1boid in the future, you could break out spells per version, then have more that union them back together. that could be useful if you ever wanted tables that granular and assign different version values and such. we won't require breaking up by version, but it's something other projects have done (maybe more so when logic is v different)

PR #4390, opened 2023-09-20
lido accounting upstream spell migration

ppclunghe:
lido.accounting migration to DuneSQL.


jeff-dude (approved):
pushed one commit for adding back jinja syntax on a ref() that was removed, otherwise should be good 👍

jeff-dude:
> pushed one commit for adding back jinja syntax on a ref() that was removed, otherwise should be good 👍

had to push another commit with more updates to the `period` column, as it must be `date` data type to be the partition column

PR #4399, opened 2023-09-22
Add block_date for ens set name

sankinyue:
I'm glad to see that dune added block_date as a partition to improve query speed.

When I used this table, I found that the incremental update strategy would cause address or ens to not be unique

### DIFF
--- a/models/ens/ens_set_name_latest.sql
+++ b/models/ens/ens_set_name_latest.sql
@@ -2,9 +2,8 @@
     config(
          alias = alias('set_name_latest')
         ,tags = ['dunesql']
-        ,materialized = 'incremental'
+        ,materialized = 'table'
         ,file_format = 'delta'
-        ,incremental_strategy = 'merge'
         ,unique_key = ['address', 'name']
         ,post_hook='{{ expose_spells(\'["ethereum"]\',
                                     "project",
@@ -13,20 +12,20 @@
     )
 }}
 
--- ENS: Old Reverse Registrar Contract Address Created Block Number
-{% set initial_tx_block_number = 3787060 %}
-
 with
-set_name_detail as (
+  set_name_detail as (
     -- setName
     select
           block_time
+        , block_number
         , "from"                                                as address
         , to                                                    as registrar
         , from_utf8(bytearray_rtrim(substr(input, 5 + 2 * 32))) as name
+        , tx_index                                              as index
         , tx_hash
-    from {{source('ethereum', 'traces')}}
-    where block_number >= {{initial_tx_block_number}}
+    from {{source('ethereum', 'transactions')}}
+    where block_date >= date '2017-05-29'
+        and block_number >= 3787060
         and to in (
               0x9062c0a6dbd6108336bcbe4593a3d1ce05512069 -- ENS: Old Reverse Registrar
             , 0x084b1c3c81545d370f3634392de611caabff8148 -- ENS: Old Reverse Registrar 2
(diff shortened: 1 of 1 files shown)


jeff-dude:
hi @sankinyue 👋 

if you're certain those two fields, `address` and `name` should make each row unique, but the incremental job is still loading duplicates, the most likely culprit is null values in one or both of those columns. the null causes the merge statement used in incremental runs to fail the join condition, therefore thinking it should be inserted again.

before we take the shortcut and just make this a table instead, do you want to investigate that further?

sankinyue:
> hi @sankinyue 👋
> 
> if you're certain those two fields, `address` and `name` should make each row unique, but the incremental job is still loading duplicates, the most likely culprit is null values in one or both of those columns. the null causes the merge statement used in incremental runs to fail the join condition, therefore thinking it should be inserted again.
> 
> before we take the shortcut and just make this a table instead, do you want to investigate that further?

Thank You @jeff-dude 

I think this filtering method has helped me remove those null values. 

 `and substr(from_utf8(bytearray_rtrim(substr(input, 5 + 2 * 32))), -4) = '.eth'` 

However, due to Dune's latest functional changes (query run time within 120s), my Dune account can no longer check this query in batches.

jeff-dude:
> > hi @sankinyue 👋
> > if you're certain those two fields, `address` and `name` should make each row unique, but the incremental job is still loading duplicates, the most likely culprit is null values in one or both of those columns. the null causes the merge statement used in incremental runs to fail the join condition, therefore thinking it should be inserted again.
> > before we take the shortcut and just make this a table instead, do you want to investigate that further?
> 
> Thank You @jeff-dude
> 
> I think this filtering method has helped me remove those null values.
> 
> `and substr(from_utf8(bytearray_rtrim(substr(input, 5 + 2 * 32))), -4) = '.eth'`
> 
> However, due to Dune's latest functional changes (query run time within 120s), my Dune account can no longer check this query in batches.
> 
> I think there is another situation that makes the incremental job unable to get the final state.
> 
> for example:
> 
> block_date	address	name	tx_hash
> 2023-10-01	0x1	abc.eth	0xa
> 2023-10-02	0x2	def.eth	0xb
> 2023-10-03	0x1	ghi.eth	0xc
> 2023-10-04	0x2	abc.eth	0xd
> The result I want to get is:
> 
> block_date	address	name	tx_hash
> 2023-10-03	0x1	ghi.eth	0xc
> 2023-10-04	0x2	abc.eth	0xd
> I guess this may be the reason why incremental job cannot make `address` and `name` unique.

if i'm reading this correctly, then the only unique value is `address`, as you want to update `block_date, name, tx_hash`. with that in mind, the unique key in your spell should only be `address`, then it should work as expected

jeff-dude:
i think it's worth a try here, give it a shot for sure

sankinyue:
Sorry, my previous example didn't account for other special cases.
For example:

| block_date | address | name    | tx_hash |
| ---------- | ------- | ------- | ------- |
| 2023-10-01 | 0x1     | abc.eth | 0xa     |
| 2023-10-02 | 0x2     | def.eth | 0xb     |
| 2023-10-04 | 0x2     | abc.eth | 0xd     |

If only unique value is `address`, result:

| block_date | address | name    | tx_hash |
| ---------- | ------- | ------- | ------- |
| 2023-10-03 | 0x1     | abc.eth | 0xc     |
| 2023-10-04 | 0x2     | abc.eth | 0xd     |

This will allow two addresses to have the `same`.

My hope is that the final result is a unique `address` corresponding to a unique `name`, so that when analyzing user addresses for certain protocols, I can uniquely locate one-to-one cases, avoiding `left join` leading to data duplication.

jeff-dude:
> Sorry, my previous example didn't account for other special cases. For example:
> 
> block_date	address	name	tx_hash
> 2023-10-01	0x1	abc.eth	0xa
> 2023-10-02	0x2	def.eth	0xb
> 2023-10-04	0x2	abc.eth	0xd
> If only unique value is `address`, result:
> 
> block_date	address	name	tx_hash
> 2023-10-01	0x1	abc.eth	0xa
> 2023-10-04	0x2	abc.eth	0xd
> This will allow two addresses to have the `same`.
> 
> My hope is that the final result is a unique `address` corresponding to a unique `name`, so that when analyzing user addresses for certain protocols, I can uniquely locate one-to-one cases, avoiding `left join` leading to data duplication.

let's see if i'm following this time.

one address can only have one name at any given time, but over time, addresses can have different names (or a name could go to a different address). the `tx_hash` helps show which tx had the name change.

if that is the case, then could you make unique key:
`block_date` and `address`

that assumes new tx's only come in if/when a name changes per address, then would insert new row with new block date, new address/name combo and the new tx which changed it

if that is still an incorrect understanding, there has to be a way to find a unique combo of columns that work

sankinyue:
> > Sorry, my previous example didn't account for other special cases. For example:
> > block_date	address	name	tx_hash
> > 2023-10-01	0x1	abc.eth	0xa
> > 2023-10-02	0x2	def.eth	0xb
> > 2023-10-04	0x2	abc.eth	0xd
> > If only unique value is `address`, result:
> > block_date	address	name	tx_hash
> > 2023-10-01	0x1	abc.eth	0xa
> > 2023-10-04	0x2	abc.eth	0xd
> > This will allow two addresses to have the `same`.
> > My hope is that the final result is a unique `address` corresponding to a unique `name`, so that when analyzing user addresses for certain protocols, I can uniquely locate one-to-one cases, avoiding `left join` leading to data duplication.
> 
> let's see if i'm following this time.
> 
> one address can only have one name at any given time, but over time, addresses can have different names (or a name could go to a different address). the `tx_hash` helps show which tx had the name change.
> 
> if that is the case, then could you make unique key: `block_date` and `address`
> 
> that assumes new tx's only come in if/when a name changes per address, then would insert new row with new block date, new address/name combo and the new tx which changed it
> 
> if that is still an incorrect understanding, there has to be a way to find a unique combo of columns that work

Even if you use block_date or tx_hash as unique values, the expired `0x1` ➡️ `abc.eth` will remain in the table.

Let me explain the relationship between address and ens, somewhat similar to the relationship between ip and domain.

Regardless of the number of records resolved for previous settings, there will only be one address and one name. 

i.e. a request for `abc.eth` will only return `0x2` and a request for `0x2` will only return `abc.eth`.

sankinyue:
Hi @jeff-dude , can we continue with this pr process and let me know if there's anything else I need to explain, thanks!

jeff-dude (approved):
okay, if you'd rather just flip it to full table refresh that only runs 1x/day rather than incremental hourly, that's fine.

if you do decide to revisit post-merge, we can try to find a way to make incremental work

PR #4414, opened 2023-09-24
Add first activity macro - ready for review 

henrystats:


henrystats:
excluded optimism from this because it requires joining with the ovm_legacy table and it's the only case like that so best to just leave it as it is without using a macro.

jeff-dude (approved):
love it!

PR #4415, opened 2023-09-25
fix orca aToB missing decoding

andrewhong5297:
We've run into a solana decoding error where some instructions have extra bytes that are invalid. Anchor allows this, but it breaks our decoding. 

For now, I'm doing a manual decoding of "data" for "aToB" which starts at the [42nd byte](https://docs.rs/whirlpools/latest/whirlpools/instruction/struct.Swap.html) (8 + 8 + 8 + 16 + 1 + 1). This will require a full rerun for this table.


andrewhong5297:
I've checked the error cases from before, and don't see any issues/outliers in amount_usd anymore.

andrewhong5297:
thanks antonio! 🙌

PR #4417, opened 2023-09-25
Fix balances schema name on Polygon

antonio-mendes:
This is what was causing the double columns to show up in the ui.

We want a schema calles `balances_{blockchain}` for now and then the `{standard}_{aggregation}` as table name.


henrystats:
ah nice!

PR #4419, opened 2023-09-25
adapt equality test to handle null = null

0xRobin:
adapt equality test to handle null = null


0xRobin:
I would not consider these test failures as blocking for this PR as they would also fail under the older logic. 
(worth looking into, but that's a different task).

I hope we can merge this with admin powers. :slightly_smiling_face:

jeff-dude (approved):
i'm good with that approach

PR #4428, opened 2023-09-26
hack to refresh all seeds by using the PR body

0xRobin:
This allows people to add the following string to the PR body: `CI:refresh-all-seeds`
This removes the need to manually fabricate changes in seeds (by swapping ros) to get them to run in the CI, The `@` operator does overshoot a bit (runs more seeds than necessary). Hence the conditional to only run it with a specific string in the PR body. 
See it in action here: https://github.com/duneanalytics/spellbook/pull/4427

### DIFF
--- a/.github/workflows/dbt_slim_ci.yml
+++ b/.github/workflows/dbt_slim_ci.yml
@@ -87,7 +87,12 @@ jobs:
           [[ -z "$test" ]] && { echo "Success"; exit 0; } || { echo "Found active child models of removed models:"; echo "$test"; exit 1; }
 
       - name: dbt seed
+        if: ${{ !contains(github.event.pull_request.body,'CI:refresh-all-seeds') }}
         run: "dbt seed $PROFILE --select state:modified$TAG --exclude tag:prod_exclude tag:remove --state ."
+
+      - name: dbt seed @
+        if: ${{ contains(github.event.pull_request.body,'CI:refresh-all-seeds') }}
+        run: "dbt seed $PROFILE --select @state:modified$TAG --exclude tag:prod_exclude tag:remove --state ."
 
       - name: dbt run initial model(s)
         run: "dbt -x run $PROFILE --select state:modified$TAG --exclude tag:prod_exclude tag:remove --defer --state ."


.github/workflows/dbt_slim_ci.yml
```diff
@@ -87,8 +87,13 @@ jobs:
           [[ -z "$test" ]] && { echo "Success"; exit 0; } || { echo "Found active child models of removed models:"; echo "$test"; exit 1; }
 
       - name: dbt seed
+        if: ${{ !contains(github.event.pull_request.body,'CI:refresh-all-seeds') }}
```
jeff-dude, line 90: if you're adding a whole new step, why is this conditional needed?

0xRobin: Basically we want to do:
```
if (PR contains "CI:refresh-all-seeds"):
  dbt seed @state:modified
else:
  dbt seed state:modified
  ```
 But since GitHub workflows don't have an if-else pattern, we need 2 steps with inverted conditionals.

jeff-dude: ah, if doesn't contain, got it

jeff-dude (approved):
works for me
fyi @duneanalytics/team-spellbook

aalan3:
>we will look to get working in prod on a daily and/or weekly run so that this isn't needed. currently, we hit access issues on truncate/drop. i will look to raise a linear task.

It's possible to do schedule [cron jobs on Github Actions](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule). Haven't tried it myself though.

jeff-dude:
> > we will look to get working in prod on a daily and/or weekly run so that this isn't needed. currently, we hit access issues on truncate/drop. i will look to raise a linear task.
> 
> It's possible to do schedule [cron jobs on Github Actions](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule). Haven't tried it myself though.

we are looking to get working on dbt cloud, rather than gh actions

jeff-dude:
> I tested something similar a couple months ago here: [157328e](https://github.com/duneanalytics/spellbook/commit/157328e45df214c426efcdee27777688866d5922)
> 
> CI flags would be super useful for a number of use cases (skipping CI, forcing seeds, automatically add labels, ...)

ah nice. we should consider this in future quarter OKR planning

PR #4430, opened 2023-09-26
Clean up tigris models 

henrystats:

### DIFF
--- a/models/tigris/arbitrum/events/tigris_arbitrum_events_add_margin.sql
+++ b/models/tigris/arbitrum/events/tigris_arbitrum_events_add_margin.sql
@@ -49,7 +49,7 @@ add_margin_events AS (
             ap.contract_address as project_contract_address
         FROM {{ source('tigristrade_arbitrum', add_margin_trading_evt) }} ap
         {% if is_incremental() %}
-        WHERE ap.evt_block_time >= date_trunc('day', now() - interval '7' day)
+        WHERE 1 = 0 
         {% endif %}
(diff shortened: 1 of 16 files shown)


models/tigris/arbitrum/events/tigris_arbitrum_events_add_margin.sql
```diff
@@ -49,7 +49,7 @@ add_margin_events AS (
             ap.contract_address as project_contract_address
         FROM {{ source('tigristrade_arbitrum', add_margin_trading_evt) }} ap
         {% if is_incremental() %}
-        WHERE ap.evt_block_time >= date_trunc('day', now() - interval '7' day)
+        WHERE 1 = 0
```
jeff-dude, line 52: is it cleaner to just remove the whole `if is incremental` block of code, if you don't want to use it? so remove all three lines, rather than force a false return?

henrystats: hmm... in that case wouldn't the data from the table be included when it runs incrementally?

henrystats: I've always thought otherwise haha

jeff-dude: ah, yeah, i follow now. sorry for the confusion!

jeff-dude (approved):

PR #4437, opened 2023-09-27
SPE-126 Migrate `zerion_{chain}.trades` tables

hildobby:


models/zerion/arbitrum/zerion_arbitrum_trades.sql
```diff
@@ -1,5 +1,7 @@
 {{ config(
-    alias = alias('trades')
+    tags=['dunesql']
+    ,schema = 'zerion_arbitrum'
+    ,alias = alias('trades')
     ,partition_by = ['block_date']
```
aalan3, line 5: We can actually drop partitioning on all these tables because they are too small to benefit from it.

hildobby: Sg, I will remove the partitioning now.

Btw, any colour on best practices for partitioning? It would be cool to have some ballpark numbers for:
- minimum amount of entries where a dataset would benefit from partitioning
- some range or sweetspot numbers for partition sizes

All I know is that `block_month` has proven to be a lot better for partitioning than `block_date` in some places apparently, but the size of those partitions will vary greatly based on the dataset. If it's a gas-level spell (lots of entries), maybe weekly is better for example 🤷‍♂️

Without much knowledge on how it looks in the backend, it's hard to assess when it's useful/less

Same thing applies for `incremental_predicates` if you have any tips there on when it should/shoudn't be used

models/zerion/arbitrum/zerion_arbitrum_trades.sql
```diff
@@ -1,6 +1,7 @@
 {{ config(
-    alias = alias('trades')
-    ,partition_by = ['block_date']
```
couralex6, line 3: Why remove partitioning here?

couralex6: nvm just saw Alan's comment

couralex6 (approved):

models/zerion/arbitrum/zerion_arbitrum_trades.sql
aalan3: Yeah this is a great question. `block_month` is generally preferable to `block_date` but in most cases you could probably start without it all together. Some examples:

* Big tables:`fees_traces` or solana tables  -> `block_date`  (these tend to be > 1TB)
* Medium tables: dex.trades -> `block_month` (>50GB)
* Small tables: this one -> no need for partitioning

We should use incremental predicates for big and some medium tables but smaller ones shouldn't really need it.

It can also be a bit difficult for you to know how big a table is since we don't really have this data exposed anywhere but trying to give some intuition about this above.

aalan3: Unfortunately not at the moment, I don't think it's easy to dynamically decide on partitioning since sometimes we need to add an extra column for it to work for example.

jeff-dude (approved):

PR #4442, opened 2023-09-28
Always attempt to parse contributors as JSON

jkylling:
~~I have not tested this.~~

### DIFF
--- a/macros/expose_spells.sql
+++ b/macros/expose_spells.sql
@@ -1,4 +1,8 @@
 {% macro expose_spells(blockchains, spell_type, spell_name, contributors) %}
+  {%- set validated_contributors = tojson(fromjson(contributors | as_text)) -%}
+  {%- if ("%s" % validated_contributors) == "null" -%}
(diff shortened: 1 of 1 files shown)


belen-pruvost:
Test: https://github.com/duneanalytics/spellbook/pull/4443

belen-pruvost:
I don't see this check erroring out on the CI on my test (side note: it might not be the best test, I was trying to escape models)

jeff-dude:
> I don't see this check erroring out on the CI on my test (side note: it might not be the best test, I was trying to escape models)

we don't run the `expose_spell` macro in CI, which is why we never see these errors pop up until prod. you can see through line 2:
`if target == prod`

jkylling:
@jeff-dude @aalan3 the expose macro is updated to throw an exception whenever it is unable to parse the contributors string as JSON. This should serve as a basic safeguard against malformed contributors lists.

jeff-dude (approved):
got it -- thanks for the assistance here!

i will look to merge this at a time where it's the only merge, as it'll require some special handling in prod & i'd like to monitor to ensure post-hook queries don't fail with the changes (since CI doesn't run this)

PR #4455, opened 2023-09-28
`dex.info` & `nft.marketplaces_info` column fix

hildobby:
Renaming the columns to be consistant with the respective trades tables' column names.


jeff-dude (approved):
smart, sounds good to me

PR #4458, opened 2023-09-29
Solana nfts

andrewhong5297:
This is going to contain:
- tokens.solana_nfts
- nft.solana_trades
- tensorswap.trades v1 and v2 
- magiceden.trades v2


andrewhong5297:
@0xRobin if you want to take a look and let me know how to keep schemas/folders consistent for solana nft trades, the spells are basically good to go

andrewhong5297:
I've checked that these values match up with everything I expect, you can see a [reference dashboard here](https://dune.com/ilemi/solana-nfts). Tested the new tables against the matviews I was playing with. 

Not sure why the spark job is failing. But after @0xRobin takes a look at schema/organization, this should be good to merge. It's actually not that expensive to run, I could run all three models consecutively in Dune under the 30 minute time limit.

models/_sector/nft/trades/solana/nft_solana_trades_legacy.sql
```diff
@@ -0,0 +1,19 @@
...
+        , tags = ['legacy']
+        , alias = alias('trades', legacy_model=True)
+        ,materialized = 'incremental'
+        ,file_format = 'delta'
+        ,incremental_strategy = 'merge'
+        ,unique_key = ['unique_trade_id','block_slot']
+        ,post_hook='{{ expose_spells(\'["solana"]\',
+                                    "sector",
+                                    "nft",
+                                    \'["ilemi"]\') }}'
```
0xRobin, line 13: You can remove these lines from the config of a dummy model. Then the model will default to a view and the CI error on spark should be resolved. (It was looking for the `unique_key` columns which aren't present in the dummy model).

Same comment applies to the other dummy models.

0xRobin (changes requested):

models/_sector/nft/trades/solana/nft_solana_trades_legacy.sql
andrewhong5297: done

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,231 @@
+{{
+    config(
+        schema = 'magiceden_solana'
+        , tags = ['dunesql']
+        , alias = alias('trades')
+        ,materialized = 'incremental'
+        ,file_format = 'delta'
+        ,incremental_strategy = 'merge'
+        ,unique_key = ['unique_trade_id','block_slot']
```
0xRobin, line 9: We've been pushing to phase out the use of `unique_trade_id` for the merge key. 
Can we make this a minimal set of unique columns?
For reference on evm models this is `['block_number','tx_hash','sub_tx_trade_id']` where the `sub_tx_trade_id` is a unique number for when multiple trades happen in 1 tx, often we use the `evt_index` here or use a `row_number()` partitioned on `tx_hash` and ordered by `trace_address`.

From looking at this model I think the equivalent would be something like `['block_slot', 'tx_id', 'call/log_order']`?
If possible. let's adapt the same `sub_tx_trade_id` column name.

0xRobin: In the overall `trades_solana` model this unique_key would then become `['project', 'block_slot', 'tx_id', 'sub_tx_trade_id']`

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,231 @@
...
+            , cast(null as varchar) as aggregator_name
+            , cast(null as varchar) as aggregator_address
+            , t.call_tx_id as tx_id
+            , t.call_block_slot as block_slot
+            , t.call_tx_signer as tx_signer
+            , t.taker_fee as taker_fee_amount_raw --taker fees = platform fees
+            , t.taker_fee/1e9 as taker_fee_amount
+            , t.taker_fee/1e9 * sol_p.price as taker_fee_amount_usd
+            , t.taker_fee/t.price as taker_fee_percentage
+            , t.maker_fee as maker_fee_amount_raw
```
0xRobin, line 210: curious, who is this maker fee payed to?
If we already have platform (taker) and royalty fees. :thinking:

models/magiceden/solana/magiceden_solana_trades.sql
0xRobin: after that we should add this test to the model schema files to ensure we have a unique set:
```
   tests:
          combination_of_columns:
            - block_slot
            - tx_id
            - sub_tx_trade_id
```

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,231 @@
...
+            , tk.account_mint --token id equivalent
+            , tk.verified_creator --contract address equivalent
+            , tk.collection_mint --contract address equivalent
+            , 'TSWAPaqyCSx2KABk68Shruf4rp7CxcNi8hAsbdwmHbN' as project_program_id
+            , cast(null as varchar) as aggregator_name
+            , cast(null as varchar) as aggregator_address
+            , t.call_tx_id as tx_id
+            , t.call_block_slot as block_slot
+            , t.call_tx_signer as tx_signer
+            , t.taker_fee as taker_fee_amount_raw --taker fees = platform fees
```
0xRobin, line 206: if `taker fees = platform fees`, is there a reason to not adopt the `platform_fee_(..)` column names from nft.trades?

0xRobin (changes requested):

models/magiceden/solana/magiceden_solana_trades.sql
andrewhong5297: It is paid to the user if negative, the platform if positive. Interesting rebate mechanism on Solana

models/magiceden/solana/magiceden_solana_trades.sql
andrewhong5297: Technically positive maker fees count towards platform fees. It is a weird nuance, i can call it out more in the column descriptions

andrewhong5297:
oh wait ignore re-request, I remembered I removed the tests to see what was causing the error lol

andrewhong5297:
so weird, I don't get duplicates in the app. I feel like there is some null thing happening https://dune.com/queries/3065149/5104532

andrewhong5297:
oookay we're good to go here @0xRobin @jeff-dude 👍

jeff-dude:
> oookay we're good to go here @0xRobin @jeff-dude 👍

i'll let rob give it the final pass monday morning

andrewhong5297:
found some intricacies in how they handle fees/prices, pushed a fix

andrewhong5297:
since it's only two models, it's worth also discussing if we want to make this just a view for now instead of an incremental table. I'm sure that this table will still likely change/need fixes as it is in its infancy.

0xRobin:
@andrewhong5297 looking good!

A final addition we need is a seed file and check for each of the marketplace models.
This would mean:
under a `/seeds/_sector/nft/solana` directory
1. adding a `_schema.yml` file containing:
```yml
version: 2

seeds:
  - name: magiceden_solana_trades_seed
    config:
      tags: dunesql

  - name: tensorswap_solana_trades_seed
    config:
      tags: dunesql
```
normally it's only necessary to specify `column_types` for columns that are not correctly inferred (varbinary/uint256)
2. adding the 2 seed files
These need to consist of "match columns" and "check columns", evm example here:
https://github.com/duneanalytics/spellbook/blob/f3b65a25987e92485cb02b4a73531c801e057f73/seeds/_sector/nft/archipelago_ethereum_base_trades_seed.csv#L1
3. adding the seed check to the model configs.
evm example here: 
https://github.com/duneanalytics/spellbook/blob/f3b65a25987e92485cb02b4a73531c801e057f73/models/_sector/nft/trades/ethereum/platforms/_schema.yml#L20-L33

No need for more than ~10 example rows in these seeds, bonus points if they cover multiple tx types.

andrewhong5297:
> @andrewhong5297 looking good!
> 
> A final addition we need is a seed file and check for each of the marketplace models. This would mean: under a `/seeds/_sector/nft/solana` directory
> 
> 1. adding a `_schema.yml` file containing:
> 
> ```yaml
> version: 2
> 
> seeds:
>   - name: magiceden_solana_trades_seed
>     config:
>       tags: dunesql
> 
>   - name: tensorswap_solana_trades_seed
>     config:
>       tags: dunesql
> ```
> 
> normally it's only necessary to specify `column_types` for columns that are not correctly inferred (varbinary/uint256) 2. adding the 2 seed files These need to consist of "match columns" and "check columns", evm example here:
> 
> https://github.com/duneanalytics/spellbook/blob/f3b65a25987e92485cb02b4a73531c801e057f73/seeds/_sector/nft/archipelago_ethereum_base_trades_seed.csv#L1
> 
> 
> 3. adding the seed check to the model configs.
> evm example here:
> https://github.com/duneanalytics/spellbook/blob/f3b65a25987e92485cb02b4a73531c801e057f73/models/_sector/nft/trades/ethereum/platforms/_schema.yml#L20-L33
> 
> No need for more than ~10 example rows in these seeds, bonus points if they cover multiple tx types.

sounds good! added to the overall schema file since nesting the solana schema in a new folder/file was causing issues.

0xRobin (approved):
letsgo :rocket: 
Q: what do we do with [magiceden_solana_events.sql](https://github.com/duneanalytics/spellbook/blob/f3b65a25987e92485cb02b4a73531c801e057f73/models/_sector/nft/trades/old/platforms/magiceden_solana_events.sql) and [opensea_solana_events.sql](https://github.com/duneanalytics/spellbook/blob/f3b65a25987e92485cb02b4a73531c801e057f73/models/opensea/solana/opensea_solana_events.sql)?

0xRobin:
> I'd imagine we can just drop them completely

I would prefer keeping them around. Even if they don't produce much new data, the historical data is still valuable. 
Can you check whether your new magiceden model includes the old data? (then we can at least drop the old ME model).
And then I'd suggest porting the OpenSea model to the new solana schema?

Or do you believe the older models are just of that low quality that it's better to have no data on them instead of wildly inaccurate data?

andrewhong5297:
> > I'd imagine we can just drop them completely
> 
> I would prefer keeping them around. Even if they don't produce much new data, the historical data is still valuable. Can you check whether your new magiceden model includes the old data? (then we can at least drop the old ME model). And then I'd suggest porting the OpenSea model to the new solana schema?
> 
> Or do you believe the older models are just of that low quality that it's better to have no data on them instead of wildly inaccurate data?

They don't capture price or fee data accurately since you can't rely on just pre/post token balances. I likely won't port the opensea one because they have negligible volume on Solana. 

we could keep them in the repo as `_deprecated` for reference, but I would not keep them in nft.trades otherwise people probably will get confused why there is a difference between nft.trades solana and nft_solana.trades

0xRobin:
Ok I've looked into the magiceden data: seems like we have 3 different periods.
Feb 2022 -> Jan 2023: Old model has data, new model is missing
Jan 2023 -> May 2023: Old model has data, new model has data. Results are acceptably equal.
May 2023 -> Current: Old model is missing data, new model has data.

image

Given the amount of volume captured by the old model, and also the accuracy of the old model during the overlap in period 2. I don't think it makes sense to just get rid of the old data.

andrewhong5297:
Yes they have different marketplace version they have not decoded yet. If you really want to leave it in the original trades spell, go ahead. I will be telling people to not use it

0xRobin:
Oh I think we definitely want to get the data out of `nft.trades` and into `nft_solana.trades`. A dual setup is the worst of all options, and should only be temporary. :) 
I was trying to figure out whether the new ME model was a full replacement of the old one, which in the current state it seems not. So we should still keep the model itself around for future reference until someone can port it to the new solana schema.
So I think:
1. merge this
2. port the existing`magiceden_solana_events.sql` model to the new schema
3. remove existing solana models from `nft.trades`
4. point users towards `nft_solana.trades`

Is the best approach here, but I'll let @jeff-dude and/or @aalan3  make the final judgement call on that.

andrewhong5297:
> Oh I think we definitely want to get the data out of `nft.trades` and into `nft_solana.trades`. A dual setup is the worst of all options, and should only be temporary. :) I was trying to figure out whether the new ME model was a full replacement of the old one, which in the current state it seems not. So we should still keep the model itself around for future reference until someone can port it to the new solana schema. So I think:
> 
> 1. merge this
> 2. port the existing`magiceden_solana_events.sql` model to the new schema
> 3. remove existing solana models from `nft.trades`
> 4. point users towards `nft_solana.trades`
> 
> Is the best approach here, but I'll let @jeff-dude and/or @aalan3 make the final judgement call on that.

yeah okay this works for me!

jeff-dude:
thanks for the review @0xRobin!

happy to jump in to try and finalize if you guys are all good. i can look tomorrow morning.

andrewhong5297:
Yup I'm all set, will add the other magic eden model maybe in a week or two

andrewhong5297:
Pushing a small date change fix, no logic changed

models/_sector/nft/trades/solana/nft_solana_trades.sql
```diff
@@ -0,0 +1,83 @@
...
+                                    "sector",
+                                    "nft",
+                                    \'["ilemi"]\') }}'
+    )
+}}
+
+-- ,file_format = 'delta'
+-- ,incremental_strategy = 'merge'
+-- ,partition_by = ['project','block_month']
+-- ,unique_key = ['project','trade_category','outer_instruction_index','inner_instruction_index','account_metadata','tx_id']
```
jeff-dude, line 17: can we remove these to avoid confusion?

if you ever materialize in the future, should be able to copy/paste config options from any underlying project spell, as the goal is to have same unique key throughout

models/_sector/nft/trades/solana/nft_solana_trades.sql
```diff
@@ -0,0 +1,83 @@
...
+] %}
+
+
+{% for marketplace in solana_marketplaces %}
+SELECT
+        blockchain,
+        project,
+        version,
+        cast(date_trunc('day', block_time) as date) as block_date,
+        cast(date_trunc('month', block_time) as date) as block_month,
```
jeff-dude, line 32: i notice these fields are also in the project-level spells, we should just pull the column as-is here and avoid building/casting again

models/_sector/nft/trades/solana/nft_solana_trades.sql
```diff
@@ -0,0 +1,83 @@
...
+        royalty_fee_percentage,
+        amm_fee_amount_raw,
+        amm_fee_amount,
+        amm_fee_amount_usd,
+        amm_fee_percentage,
+        instruction,
+        outer_instruction_index,
+        inner_instruction_index,
+        unique_trade_id,
+        row_number() over (partition by unique_trade_id order by tx_id) as duplicates_rank
```
jeff-dude, line 77: you built the `duplicates_rank` column, i'd imagine to match `nft.trades`, but don't filter down below where value is 1. can you add the filter too, or remove this column if not?

models/_sector/nft/trades/solana/nft_solana_trades_schema.yml
```diff
@@ -0,0 +1,163 @@
...
+       nft solana trades
+    # tests:
+    #  - dbt_utils.unique_combination_of_columns:
+    #       combination_of_columns:
+    #         - project
+    #         - trade_category
+    #         - outer_instruction_index
+    #         - inner_instruction_index
+    #         - account_metadata
+    #         - tx_id
```
jeff-dude, line 21: you can go ahead and remove this. as long as it remains materialized as a view, no need here, only needed in project-level spells/schemas

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+                , cast(json_extract_scalar(json_parse(split(logs, ' ')[3]), '$.royalty') as double) as royalty
+                , logs
+            FROM (
+                SELECT
+                    call_tx_id
+                    , call_block_slot
+                    , call_outer_instruction_index
+                    , call_inner_instruction_index
+                    , call_log_messages
+                FROM {{ source('magic_eden_solana','m2_call_mip1ExecuteSaleV2') }}
```
jeff-dude, line 35: this table is relatively large (>750k rows). since this is an incremental model, can we add incremental logic here?

jeff-dude: alan introduced new incremental macro as well, to keep spellbook clean and ability to alter number of days we look back in one universal location too, rather than each individual model (work to do on our side to get all spells updated, but might as well start clean here).

reference:
https://github.com/duneanalytics/spellbook/pull/4487/files#diff-dc8d4490c16a624e93e14dcb8cae5c6f8f53697c962e2e4fae095432d705eb4cR51-R53

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+                UNION ALL 
+                SELECT
+                    call_tx_id
+                    , call_block_slot
+                    , call_outer_instruction_index
+                    , call_inner_instruction_index
+                    , call_log_messages
+                FROM {{ source('magic_eden_solana','m2_call_executeSaleV2') }}
```
jeff-dude, line 43: same idea here, but this one is even larger (2mil+)

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+                UNION ALL 
+                SELECT
+                    call_tx_id
+                    , call_block_slot
+                    , call_outer_instruction_index
+                    , call_inner_instruction_index
+                    , call_log_messages
+                FROM {{ source('magic_eden_solana','m2_call_ocpExecuteSaleV2') }}
```
jeff-dude, line 51: same idea here. it's best to apply incremental filters directly on read in CTEs, rather than outside of CTEs

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+            ) LEFT JOIN unnest(call_log_messages) as log_messages(logs) ON True
+            WHERE logs LIKE '%Program log:%royalty%price%seller_expiry%' --must log these fields. hopefully no other programs out there log them hahaha
+            AND try(json_parse(split(logs, ' ')[3])) is not null --valid hex
+            -- --QA
+            -- AND call_tx_id = '4RrNYCNkJCLjvj4unGgo6owJeEg5qxWZJLYnfyiJgFVypashPaJFVgQRivEvcxeEt3xrQUiUM9bm7VP76XtPenxw'
+            -- AND call_block_slot = 219668172
```
jeff-dude, line 57: safe to remove this?

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+                , row_number() over (partition by call_tx_id order by call_outer_instruction_index asc, call_inner_instruction_index asc) as call_order
+            FROM {{ source('magic_eden_solana','m2_call_ocpExecuteSaleV2') }}
+        ) trade
+        --this shortcut ONLY works if you know that a log is only emitted ONCE per call.
+        LEFT JOIN royalty_logs rl ON trade.call_tx_id = rl.call_tx_id
+            AND trade.call_block_slot = rl.call_block_slot
+            AND trade.call_order = rl.log_order 
+        {% if is_incremental() %}
+        WHERE trade.call_block_time >= NOW() - interval '7' day
+        {% endif %}
```
jeff-dude, line 177: let's move this into the trade CTE on each read instead. similar to above comments, lets use the new incremental macro

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+        -- --QA
+        -- WHERE trade.call_tx_id = '4RrNYCNkJCLjvj4unGgo6owJeEg5qxWZJLYnfyiJgFVypashPaJFVgQRivEvcxeEt3xrQUiUM9bm7VP76XtPenxw'
+        -- AND trade.call_block_slot = 219668172
```
jeff-dude, line 180: safe to remove?

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+    )
+    
+    , raw_nft_trades as (
+        SELECT
+            'solana' as blockchain
+            , 'magiceden' as project
+            , 'v2' as version
+            , date_trunc('day', t.call_block_time) as block_date
+            , date_trunc('month', t.call_block_time) as block_month
```
jeff-dude, line 189: similar to how you did in cross-chain spell, let's cast these to date, otherwise it'll force timestamp due to column used in the function carrying forward. we also plan to remove the logic in cross-chain, so let's ensure it's clean here

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+            , t.royalty/1e9 as royalty_fee_amount
+            , t.royalty/1e9 * sol_p.price as royalty_fee_amount_usd
+            , case when t.royalty = 0 then 0 else t.royalty/t.price end as royalty_fee_percentage
+            , t.instruction
+            , t.outer_instruction_index
+            , coalesce(t.inner_instruction_index,0) as inner_instruction_index
+        FROM trades t
+        LEFT JOIN {{ ref('tokens_solana_nft') }} tk
+            ON t.account_metadata = tk.account_metadata
+        LEFT JOIN {{ source('prices', 'usd') }} sol_p ON sol_p.blockchain = 'solana' and sol_p.symbol = 'SOL' and sol_p.minute = date_trunc('minute', t.call_block_time) --get sol_price
```
jeff-dude, line 236: what do we think about following the new pattern set in `nft.trades`, where we remove the dependency to metadata tables like tokens/prices from the project-level spells, and focus solely on base/decoded data?

this helps with orchestration, both in prod and future CI tests, as well as when building spells downstream in the future. in the new setup, we have a universal macro that all project-level spells can then join to which does the join to tokens/prices/other metadata, `enrich_trades_macro` or something similar.

@0xRobin do you think we can apply that design here too?

keep in mind, we are going to apply similar design to `dex.trades` soon too. it is a design practice we're trying to implement across all of spellbook

jeff-dude: one callout:
i recognize the token metadata for solana appears to be dynamic (reads from source tables), so it is likely going to be modified less often than the manually tracked evm ones, which was the main reason for our pain point. i do still think it's a good practice to apply here though

models/magiceden/solana/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+    )
+
+SELECT 
+*
+, concat(project,'-',trade_category,'-',cast(outer_instruction_index as varchar),'-',cast(coalesce(inner_instruction_index, 0) as varchar),'-',account_metadata,'-',tx_id) as unique_trade_id
```
jeff-dude, line 241: why is this column needed and/or wanted?

models/magiceden/solana/magiceden_solana_trades_schema.yml
```diff
@@ -0,0 +1,171 @@
...
+       description: "Seller wallet address"
+     - &buyer
+       name: buyer
+       description: "Buyer wallet address"
+     - &amount_usd
+       name: amount_usd
+       description: "USD value of the trade at time of execution"
+       tests:
+         - dbt_utils.accepted_range:
+             max_value: 1000000000 # $1b is an arbitrary number, intended to flag outlier amounts early
```
jeff-dude, line 74: we can likely remove this test at the project levels, and just leave at sector level spell above. that's how we do it for dex/nft, just as a safety check to be proactive on finding outliers before it's found in queries/dashboards

models/tensorswap/solana/tensorswap_solana_trades.sql
jeff-dude, line 1: looks like every comment i left on magiceden will also apply here, so i'll leave them out on this spell. let's ensure we apply changes to both as we go 👍

models/tensorswap/solana/tensorswap_solana_trades_schema.yml
```diff
@@ -0,0 +1,171 @@
...
+       description: "Seller wallet address"
+     - &buyer
+       name: buyer
+       description: "Buyer wallet address"
+     - &amount_usd
+       name: amount_usd
+       description: "USD value of the trade at time of execution"
+       tests:
+         - dbt_utils.accepted_range:
+             max_value: 1000000000 # $1b is an arbitrary number, intended to flag outlier amounts early
```
jeff-dude, line 74: similar idea here, can likely remove, especially if we remove join to prices at project-level too

jeff-dude (changes requested):
left a bunch of comments below, let me know what ya think on them all 🤝

models/magiceden/solana/magiceden_solana_trades.sql
andrewhong5297: fair, I've removed collection_mint which is the only one that can change. the other accounts cannot change.

models/magiceden/solana/magiceden_solana_trades.sql
andrewhong5297: ah it was from testing, I'll remove

models/_sector/nft/trades/solana/nft_solana_trades.sql
andrewhong5297: I'll just cast them here since that was the original ask from Rob

models/_sector/nft/trades/solana/nft_solana_trades.sql
andrewhong5297: ah okay. i'll remove

andrewhong5297:
> left a bunch of comments below, let me know what ya think on them all 🤝

addressed them, thanks! let's see if the CI still runs

andrewhong5297:
Lmao no idea what changed for the seeds to fail, no logic changed. Does the new incremental do some different filtering for CI where the date range is a lot smaller?

andrewhong5297:
yeah okay so the new incremental only goes back 1 day which screws up all seeds. I've deleted seeds for now, this PR is good to go.

0xRobin:
@andrewhong5297 the seeds failing is a sign that something else is off.
Looks like the resulting tables actually only have results from the last day.
I'm taking a look.

0xRobin:
Should be good now. :heavy_check_mark: 
You were missing `{% if is_incremental %}` blocks around the incremental filter so it was also running on the initial full refresh. 
Note that I also moved the models and seeds around a bit.

models/_sector/nft/trades/solana/platforms/magiceden_solana_trades.sql
```diff
@@ -0,0 +1,248 @@
+{{
+    config(
+        schema = 'magiceden_solana'
+        , tags = ['dunesql']
+        , alias = alias('trades')
+        ,materialized = 'incremental'
+        ,file_format = 'delta'
+        ,incremental_strategy = 'merge'
```
aalan3, line 8: can we also add incremental predicates to the materializations? it can really help during incremental runs

andrewhong5297: Idk what this means, where and what do you want me to put

andrewhong5297:
> Should be good now. :heavy_check_mark: 
> 
> You were missing `{% if is_incremental %}` blocks around the incremental filter so it was also running on the initial full refresh. 
> 
> Note that I also moved the models and seeds around a bit.

Oh so it still needs the if even with the macro? Got it

models/_sector/nft/trades/solana/platforms/magiceden_solana_trades.sql
andrewhong5297: not sure what this means/looks like

aalan3: Add `incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')],` in the config here. This will make it so that the merge into also filters on the destination table instead of doing a full scan which improves performance when the destination table is big.

andrewhong5297:
hold on on merging, I have two more models to add

models/_sector/nft/trades/solana/platforms/tensorswap_solana_trades.sql
```diff
@@ -0,0 +1,260 @@
...
+            , coalesce(t.inner_instruction_index, 0) as inner_instruction_index
+        FROM trades t
+        LEFT JOIN {{ ref('tokens_solana_nft') }} tk
+            ON t.account_metadata = tk.account_metadata
+        LEFT JOIN {{ source('prices', 'usd') }} sol_p ON sol_p.blockchain = 'solana' and sol_p.symbol = 'SOL' and sol_p.minute = date_trunc('minute', t.call_block_time) --get sol_price
+    )
+
+SELECT
+*
+, concat(project,'-',trade_category,'-',cast(outer_instruction_index as varchar),'-',cast(coalesce(inner_instruction_index,0) as varchar),'-',account_metadata,'-',tx_id) as unique_trade_id
```
aalan3, line 256: Do we need this for UX purposes or is it added due to habit (we used to have this as the `unique_key` above)? Otherwise we should drop this since these tend to take up quite a lot of space. 

If you think it's makes the table nicer to query though I am not against it. We could also use `{{ dbt_utils.generate_surrogate_key(['field_a', 'field_b'[,...]]) }}` so that it becomes a hash which is smaller.

andrewhong5297: oh meant to drop this

andrewhong5297:
alright this looks good

models/_sector/nft/trades/solana/nft_solana_trades.sql
```diff
@@ -0,0 +1,72 @@
+{{
+    config(
+        schema = 'nft_solana'
+        , tags = ['dunesql']
+        , alias = alias('trades')
+        , materialized = 'view'
```
jeff-dude, line 6: we can explore materializing this, similar to how we do `nft.trades`
unless the ultimate plan is to keep this as a view, then add to `nft.trades` where it's materialized

models/_sector/nft/trades/solana/nft_solana_trades.sql
```diff
@@ -0,0 +1,72 @@
...
+] %}
+
+
+{% for marketplace in solana_marketplaces %}
+SELECT
+        blockchain,
+        project,
+        version,
+        cast(date_trunc('day', block_time) as date) as block_date,
+        cast(date_trunc('month', block_time) as date) as block_month,
```
jeff-dude, line 27: in the future, when we finalize strategy for partitioning, we may end up moving these to project-level spells so they can partition on one or the other, but this is fine for now

models/_sector/nft/trades/solana/platforms/magiceden_v2_solana_trades.sql
```diff
@@ -0,0 +1,243 @@
...
+    config(
+        schema = 'magiceden_v2_solana'
+        , tags = ['dunesql']
+        , alias = alias('trades')
+        ,materialized = 'incremental'
+        ,file_format = 'delta'
+        ,incremental_strategy = 'merge'
+        ,incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')]
+        ,unique_key = ['project','trade_category','outer_instruction_index','inner_instruction_index','account_mint','tx_id']
+        ,post_hook='{{ expose_spells(\'["solana"]\',
```
jeff-dude, line 11: when refactoring, we may want to remove the post-hook to expose project-level, as the intention is to push users to the final downstream spell `nft_solana.trades` to query

jeff-dude: this applies to all project spells, will only comment here

models/_sector/nft/trades/solana/platforms/tensorswap_v2_solana_trades.sql
```diff
@@ -0,0 +1,139 @@
+{{
+    config(
+        schema = 'tensorswap_v2_solana'
+        , tags = ['dunesql']
+        , alias = alias('trades')
+        ,materialized = 'incremental'
+        ,file_format = 'delta'
+        ,incremental_strategy = 'merge'
+        ,incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')]
+        ,unique_key = ['project','trade_category','outer_instruction_index','inner_instruction_index','account_merkle_tree','leaf_id','tx_id']
```
jeff-dude, line 10: in the refactor, we should try to make it so all projects have the same unique keys, even if that means forcing default values on columns in specific models

models/_sector/nft/trades/solana/nft_solana_trades_schema.yml
jeff-dude, line 1: in refactor, we can rename to `_schema.yml` to be consistent with this directory (keeps file at top of directory, hwich is nice)

jeff-dude (commented):
few minor notes below for potential refactor later on, but otherwise this looks good! let's get it going 🚀

jeff-dude (approved):

PR #4462, opened 2023-09-29
Some dex.sandwiches fixes

hildobby:


hildobby:
@jeff-dude It's failing because of a < $1b check on `amount_usd`. `dex.sandwiches` only pulls `amount_usd` from `dex.trades`, so if it made it in `dex.trades` it should be fine in here, right?

models/dex/dex_schema.yml
```diff
@@ -392,6 +392,7 @@ models:
       - *token_bought_amount_raw
       - *token_sold_amount
       - *token_bought_amount
+      - *amount_usd
```
jeff-dude, line 395: if you reuse this field, it'll pick up the test applied in `dex.trades` schema above in same file

you can likely avoid `*` usage here and just rewrite `amount_usd` as its own here again. be sure to not use `&` again though, as it's used above. just apply name/description only

jeff-dude: applying the test on trades is enough, don't need to repeat here

jeff-dude: but for full transparency, the reason the test is failing is due to the `dex_sandwiches` spell not being modified in this PR, therefore it reads from prod, and prod doesn't have `amount_usd` column so it's not actually failing the test, but failing to run the test in the first place.

if you are adding `amount_usd` to the macro, and all blockchains pick it up, you likely want to modify the cross-chain view here as well to pull that new column

hildobby: Alright, I applied the fix for `amount_usd` in the schema and modified all the spell files (added 1 line) to retrigger CI tests that hopefully pass this time

jeff-dude (approved):
lgtm!

PR #4476, opened 2023-09-30
automate fungible token metadata for solana

andrewhong5297:
Solana tokens all have an init when deployed and then get metadata attached at some point later. We may run into some edge cases where the model runs after the init for deploying but before the metadata attachment - should be pretty rare though.


jeff-dude:
nice, it's cool to see that the metadata can be dynamic based on some source tables

this won't have an impact, but to avoid confusion, maybe you should now remove the `static` tag in the config block at the top? since this is a view, it'll still get created when merged into main, then always stay up-to-date at query runtime of course, but it could be misleading to see static there now

andrewhong5297:
ah yeah, removed!

jeff-dude (approved):
sounds good!

@couralex6 can you include this in your recent merge(s)? it's just a view, so shouldn't be an issue on prod

andrewhong5297:
oh wait I meant to make this a table

andrewhong5297:
hold on

andrewhong5297:
lmk if that changes anything

andrewhong5297:
nvm keeping it a view, it only takes 10 seconds to run no reason to materialize it

andrewhong5297:
Ah okay got it

andrewhong5297:
only takes 1 second lol, we're good to go

PR #4481, opened 2023-10-01
Create `arkham_ethereum.airdrop_claims`

hildobby:


Hosuke (approved):
LGTM.✅
Thank you @hildobby 

[Sample arkham_ethereum_airdrop_claims](https://dune.com/queries/3077949)

models/arkham/ethereum/arkham_ethereum_airdrop_claims_legacy.sql
```diff
@@ -0,0 +1,46 @@
...
+    ELSE CAST((SELECT price FROM early_price)*t.amount/POWER(10, 18) AS double)
+    END AS amount_usd
+, '{{arkm_token_address}}' AS token_address
+, 'APE' AS token_symbol
+, t.evt_index
+FROM {{ source('arkham_ethereum', 'Airdrop_evt_Claimed') }} t
+LEFT JOIN {{ ref('prices_usd_forward_fill_legacy') }} pu ON pu.blockchain = 'ethereum'
+    AND pu.contract_address='{{arkm_token_address}}'
+    AND pu.minute=date_trunc('minute', t.evt_block_time)
+WHERE t.evt_block_time BETWEEN '2022-03-17' AND '2022-06-16'
```
jeff-dude, line 46: how come date filters differ between dunesql and legacy? also, do we still need legacy or is this lineage migrated fully?

hildobby: I just copied the spell from another airdrop, just modifying what's needed. Afaik, dates need to be casted from strings to work in dunesql while just strings are find in spark, hence the different syntax.

Also, I've been adding legacy models to all my spells, it's faster than looking for if it's needed (+ still not too sure when it is/isn't needed tbh)

jeff-dude: not a syntax difference, but a difference in date range itself. i assume it means you don't care about data in the legacy file?

hildobby: Ahh yeah my bad, there's a bunch I forgot to change on the legacy file to match the DuneSQL one, just fixed those inconsistencies

jeff-dude: for your awareness, this may be helpful to know when to avoid making legacy changes:
```
comm -23 <( dbt ls -s airdrop_ethereum_claims_legacy+,tag:legacy --resource-type model --output path | sed 's/_legacy.sql/.sql/g'| sort ) <( dbt ls -s airdrop_ethereum_claims+,tag:dunesql --resource-type model --output path | sort )
```

this should output spells which exist on spark, but not yet migrated to dunesql. you can also run the inner commands separate to see the outputs for each `dbt ls`. for example:
image

nothing returns in the overall command, meaning you can safely ignore legacy files since migration is complete (including downstream spells). then you can see each of the inner commands return two spells as expected.

jeff-dude: you should be able to reuse that, and just change spell names as needed

jeff-dude: with that in mind, do you want to remove modifications to legacy here? you can also do a follow-up PR if you want to remove the legacy files for full `airdrop_claims` lineage if you want. if not, we'll eventually get to it.

hildobby: Just removed the legacy file @jeff-dude, lmk if there's more I need to do

hildobby: I'm confused, CI tests are asking me to add the legacy file back in, I thought it would be fine without it, what's going on here @jeff-dude ?

Screenshot 2023-10-13 at 12 13 38

models/arkham/ethereum/arkham_ethereum_airdrop_claims.sql
```diff
@@ -0,0 +1,46 @@
+{{
+    config(
+        tags=['dunesql'],
+        schema = 'arkham_ethereum',
```
Hosuke, line 4: Maybe we do not need to add the new legacy spell only when the schema is already existed in any legacy spells.

models/arkham/ethereum/arkham_ethereum_airdrop_claims_legacy.sql
jeff-dude: grr, okay, we'll add back for now. we're v close to shutting off spark, so we'll be past this soon

jeff-dude (approved):

PR #4487, opened 2023-10-02
Incremental predicate macro

aalan3:
This introduces a `incremental_predicate(column)` macro to be used everywhere. Both in `incremental_predicate` for `merge into` (incremental updates) and in the `WHERE` clause inside `is_incremental()`. 

There are two variables:
* `DBT_ENV_INCREMENTAL_TIME` defaults to `1`
* `DBT_ENV_INCREMENTAL_TIME_UNIT` defaults to `day`

With `DBT_ENV_INCREMENTAL_TIME_UNIT` we can experiment with doing runs that select a few hours instead of days. This could potentially make everything a bit faster.

I've updated `nft.transfers` as an example on how this works.


couralex6 (approved):

jeff-dude:
i think you hinted at this on our 1:1 today, but has it always been the case that incremental takes twice as long as historical for this spell?

in the CI here, it's ~250s --> ~500s

that's expected, but still better than it is now in prod?

jeff-dude:
in general, this approach looks good to me, so if you wanted to merge to test in prod, feel free 👍

aalan3:
@jeff-dude I can't say for certain but I've seen that a few times, I think it's because of the join the merge into creates.

PR #4488, opened 2023-10-02
fix 0x bnb `api_fills` maker & taker token

jeff-dude:
fyi @RantumBits @Hosuke @aalan3 @andrewhong5297 

i compiled the spell and ran it on both dunesql and spark (removed some of the code that didn't focus on uni_v2). only the dune team has access to run directly on spark atm, so it helps for us to step in here.

i noticed this issue exists on spark as well, but we never noticed because the `volume_usd` defaults to `NULL` due to a missing `0x` concat at the front of maker/taker token.
image
notice the maker token address is `WBNB` but the amount is massive, which should actually be flipped to the taker token (imagine it's an altcoin).

it looks like `topic2` and `topic3` need to be flipped (on spark that is, on dunesql it's `topic1` and `topic2`)

when i flipped them, the query on dunesql showed `volume_usd` to be:
`2.6709849819945e-7`
fraction of a cent, which the block explorer shows:
https://bscscan.com/tx/0x4870e8b93cac7cc25dbdd40d0858f6f73518b635a3e32a8e76fc4bdb74711f48

### DIFF
--- a/models/zeroex/bnb/zeroex_bnb_api_fills.sql
+++ b/models/zeroex/bnb/zeroex_bnb_api_fills.sql
@@ -323,8 +323,8 @@ uni_v2_swap as (
 , uni_v2_pair_creation as (
     SELECT
         bytearray_substring(data,13,20) as pair,
-        bytearray_substring(topic2, 13, 20) AS makerToken,
-        bytearray_substring(topic1, 13, 20) AS takerToken,
+        bytearray_substring(topic1, 13, 20) AS makerToken,
+        bytearray_substring(topic2, 13, 20) AS takerToken,
         rank() over (partition by bytearray_substring(data,13,20) order by block_time asc) rnk
     FROM {{ source('bnb', 'logs') }} creation
     WHERE creation.topic0 = 0x0d3648bd0f6ba80134a33ba9275ac585d9d315f0ad8355cddefde31afa28d0e9  -- all the uni v2 pair creation event
--- a/models/zeroex/bnb/zeroex_bnb_api_fills_deduped.sql
+++ b/models/zeroex/bnb/zeroex_bnb_api_fills_deduped.sql
@@ -15,8 +15,9 @@
 WITH fills_with_tx_fill_number
 AS
 (
-    SELECT   row_number() OVER ( partition BY tx_hash ORDER BY evt_index ASC ) AS tx_fill_number
-           , *
+    SELECT
+        row_number() OVER ( partition BY tx_hash ORDER BY evt_index ASC ) AS tx_fill_number
+        , *
     FROM {{ ref('zeroex_bnb_api_fills') }}
     WHERE 1=1
     AND swap_flag = true


Hosuke (approved):
Much simpler than my approach.

jeff-dude:
dunesql results:
image

jeff-dude:
> i noticed this issue exists on spark as well, but we never noticed because the `volume_usd` defaults to `NULL` due to a missing `0x` concat at the front of maker/taker token.

the reason it defaults to null on spark is because those fields are used to join to:
`tokens.erc20`
`prices.usd`
but without the `0x` up front, the joins fails, therefore calculations fail with no values to calculate (like decimals)

jeff-dude:
once the CI test finishes, we can query the test table and ensure no massive outlier `volume_usd` values come through

aalan3:
ah nice catch! that `topic1/2` error is subtle and a thing to keep in mind elsewhere if the same method is used. Ideally we would use the decoded tables for these things and avoid decoding by hand.

jeff-dude:
i don't think we're in the clear just yet, sadly :(

while that particular tx was fixed, it looks like there are still tons of tx's which come through with `volume_usd` blown out of proportion:
```
select *
from test_schema.git_dunesql_cffc8fe5_zeroex_bnb_api_fills
-- where tx_hash = 0x4870e8b93cac7cc25dbdd40d0858f6f73518b635a3e32a8e76fc4bdb74711f48
order by volume_usd desc
limit 100
```

either the "fix" here had a side effect, or it only covered some scenarios but not all.

i think we need to take a few steps:
- exclude bnb from prod for now while debugging further
- remove bnb from cross-chain spell
- refresh 0x spells without bnb
- continue debugging in separate PR (i'll likely close this one)

@RantumBits will you have time to pick this up from here and help fix the bnb spell to ensure maker/taker are used properly, that way `volume_usd` ends up as expected?

RantumBits:
> i don't think we're in the clear just yet, sadly :(
> 
> while that particular tx was fixed, it looks like there are still tons of tx's which come through with `volume_usd` blown out of proportion:
> 
> ```
> select *
> from test_schema.git_dunesql_cffc8fe5_zeroex_bnb_api_fills
> -- where tx_hash = 0x4870e8b93cac7cc25dbdd40d0858f6f73518b635a3e32a8e76fc4bdb74711f48
> order by volume_usd desc
> limit 100
> ```
> 
> either the "fix" here had a side effect, or it only covered some scenarios but not all.
> 
> i think we need to take a few steps:
> 
> * exclude bnb from prod for now while debugging further
> * remove bnb from cross-chain spell
> * refresh 0x spells without bnb
> * continue debugging in separate PR (i'll likely close this one)
> 
> @RantumBits will you have time to pick this up from here and help fix the bnb spell to ensure maker/taker are used properly, that way `volume_usd` ends up as expected?

yes looks like there are still issues with maker/taker swapped. i'll take a look

jeff-dude:
plz do start a new, fresh branch on your side and open a PR then tag me/hosuke when ready for review 👍 
i'll close this for now

PR #4491, opened 2023-10-02
Add base to contracts mapping

jeff-dude:
this reverts commit f3b65a25987e92485cb02b4a73531c801e057f73, which was the original revert when finding issues in prod. now looking to add back when prod is ready.

fyi @MSilb7


jeff-dude:
need to wait to merge this when it's the only new code & the cluster is prepared

PR #4499, opened 2023-10-03
SPE-144 Migrate remaining OP models

couralex6:


aalan3 (approved):

couralex6:
DBT throwing error for model `op_chains_optimism_chain_list`:
```
IN value and list items must be the same type or coercible to a common type. Cannot find common type between integer and varchar(1), all types (without duplicates): [integer, varchar(1)]
```
This is the DBT generic test `accepted_values` for values `[0,1]` and the values for the `is_superchain` field are correct:

```
FROM(values
         ('optimism',   'Optimism Mainnet', 10, '2021-06-23',   1)
        ,('base',       'Base Mainnet',     NULL,   NULL,       1)
) op (blockchain, blockchain_name, chain_id, start_date, is_superchain)
```

I think we should just change that field to Boolean, which will enforce the two value without a test. I'll try that now.

cc @jeff-dude @MSilb7

jeff-dude:
> DBT throwing error for model `op_chains_optimism_chain_list`:
> 
> ```
> IN value and list items must be the same type or coercible to a common type. 
> Cannot find common type between integer and varchar(1), all types (without duplicates): [integer, varchar(1)]
> ```
> 
> This is the DBT generic test `accepted_values` for values `[0,1]` and the values for the `is_superchain` field are correct:
> 
> ```
> FROM(values
>          ('optimism',   'Optimism Mainnet', 10, '2021-06-23',   1)
>         ,('base',       'Base Mainnet',     NULL,   NULL,       1)
> ) op (blockchain, blockchain_name, chain_id, start_date, is_superchain)
> ```
> 
> I think we should just change that field to Boolean, which will enforce the two value without a test. I'll try that now.
> 
> cc @jeff-dude @MSilb7

hmm did you compile the test and run on the dune app?

i'm curious what the query looked like and why that failed. the values should be int for both. i think it's worth a try to quickly compile and run the query to see closer

couralex6:
Just kidding found the issue @jeff-dude 

Needed to add `quote: false` to `accepted_values` test to use int values

https://docs.getdbt.com/reference/resource-properties/tests#accepted_values

PR #4501, opened 2023-10-03
Solana balances performance fix

aalan3:
This PR aims to improve performance for solana balances a bit. 

* Uses the `incremental_predicate` macro
* Uses `incremental_predicate` to limit the merge into scope
* Switches the token accounts to use the spell instead of fetching from account activity.

### DIFF
--- a/models/solana_utils/solana_utils_daily_balances.sql
+++ b/models/solana_utils/solana_utils_daily_balances.sql
@@ -18,7 +18,7 @@ WITH
       tokens_accounts as (
             SELECT
             distinct address
-            FROM {{ ref('solana_utils','token_accounts') }}
+            FROM {{ ref('solana_utils_token_accounts') }}
       )
       , updated_balances as (
             SELECT
--- a/models/solana_utils/solana_utils_daily_balances.sql
+++ b/models/solana_utils/solana_utils_daily_balances.sql
@@ -37,7 +37,7 @@ WITH
                   or (address in (select address from tokens_accounts) AND token_mint_address is not null)
                   )
             {% if is_incremental() %}
-            AND incremental_predicate('block_time')
+            AND {{incremental_predicate('block_time')}}
             {% endif %}
       )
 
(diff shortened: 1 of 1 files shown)


models/solana_utils/solana_utils_daily_balances.sql
```diff
@@ -6,50 +6,48 @@
...
 }}
 
-WITH 
+WITH
       tokens_accounts as (
-            SELECT 
+            SELECT
             distinct address
-            FROM {{ source('solana','account_activity') }}
-            WHERE tx_success
```
aalan3, line 21: @andrewhong5297 can we do this? The diff here is that `tx_success` is not in the where for `solana_utils.token_accounts`

models/solana_utils/solana_utils_daily_balances.sql
```diff
@@ -6,50 +6,48 @@
...
-            WHERE tx_success 
+            WHERE tx_success
             AND (
                   --if the address is a token_mint_account, then mint_address should not be null
                   (address NOT in (select address from tokens_accounts))
                   or (address in (select address from tokens_accounts) AND token_mint_address is not null)
                   )
             {% if is_incremental() %}
-            AND block_time >= date_trunc('day', now() - interval '1' day)
+            AND {{incremental_predicate('block_time')}}
```
aalan3, line 40: Incremental macro might not actually be a good idea here because if we go for hours instead of days it would mess things up here.

models/solana_utils/solana_utils_daily_balances.sql
andrewhong5297: errr idk what this ref file is. is this solana_utils.token_accounts? if so then yeah that is probably fine

aalan3: It's this one https://github.com/duneanalytics/spellbook/blob/main/models/solana_utils/solana_utils_token_accounts.sql

PR #4504, opened 2023-10-03
Drop legacy cow_protocol/tx_hash_labels

aalan3:
I opted to just remove the files here directly because the schema will still exist. So we don't actually need the mock model for these.


jeff-dude (approved):
🫡

PR #4505, opened 2023-10-03
Use different trino cluster

belen-pruvost:
Towards NGN-1225

### DIFF
--- a/.github/workflows/dbt_slim_ci.yml
+++ b/.github/workflows/dbt_slim_ci.yml
@@ -16,7 +16,7 @@ concurrency:
 
 jobs:
   dbt-test:
-    runs-on: [ self-hosted, linux, spellbook-trino ]
+    runs-on: [ self-hosted, linux, spellbook-trino-ci ]
     strategy:
       matrix:
         engine: [ 'dunesql', 'spark' ]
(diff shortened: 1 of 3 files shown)


belen-pruvost:
I changed the `scripts/missing_legacy_model.sh` script so that the triggering model (`trigger_scale`) won't need a legacy equivalent. This change is causing the pytest-automation check to run. The check is failing but from looking at the errors it does not seem related to the changed I've done, @aalan3 / @couralex6 could you help me understand if this is indeed unrelated?

aalan3:
try excluding it from the [dbt ls](https://docs.getdbt.com/reference/commands/list) command instead. otherwise if you just want a quick fix add a legacy model as well. I am having a similar issue and we might actually drop this soon if possible.

belen-pruvost:
We've tested these retries and the routing to the new spellbook-ci cluster and it is working nicely.

@aalan3 / @couralex6 / @jeff-dude - can this be merged so that new PR runs happen on an isolated environment?

jeff-dude (approved):
v nice! this should be a game changer, thank you 🙏

aalan3 (approved):
nice thank you!

PR #4514, opened 2023-10-03
Create `blur_ethereum.blend_events`

hildobby:


jeff-dude:
requesting quick review from @0xRobin, since you tagged it `nft`

hildobby:
> looks good. ✔️ I do question if we should integrate this with `nft.lending` and make sure we use the same schema? What do you think @hildobby? @henrystats built out the other 2 `nft.lending` models I believe.

Oh sure, haven't looked into the structure of nft.lending at all but I can take a look. My fear though is that most lending protocols work differently from one another, but I'll take a look

henrystats:
yeah agreed @hildobby a lot of the lending protocols work differently and for most of them, we'd have to have an "event"  table which has the protocol's data and then a "lending" table which is what we have a general schema for. 

I was working on nftfi queries (will be adding to spellbook as well) and noticed that it was very different from the current tables on dune (Astaria & bend_dao) and I think the events & lending tables approach would work as for most of these protocols, those extra added data are needed for anyone looking to do analysis that's more than just looking at the loan data (loan amounts, borrower, nft etc).

0xRobin:
Ok, let's get this already in in the current state. :+1:

jeff-dude (approved):

PR #4516, opened 2023-10-04
nft.trades fix for `Gem` > `OpenSea Pro` rebranding

hildobby:


0xRobin (approved):
This approach looks good! :+1: 
I'll re-trigger the CI and see if the cluster is more stable at the moment

hildobby:
> This approach looks good! 👍 I'll re-trigger the CI and see if the cluster is more stable at the moment

If there starts to be a lot more rebrands then we can consider having `start_time` & `end_time` in the aggregators list and use that instead but this is definitely way easier for now

jeff-dude (approved):

PR #4517, opened 2023-10-04
Add Goerli to `evms` tables

hildobby:


hildobby:
@MSilb7 u up? new chain about to drop in `evms`

jeff-dude:
> @MSilb7 u up? new chain about to drop in `evms`

something i mentioned to @MSilb7 in dm's:
while we approved (and will continue to allow) the `evms` spells, the intention was to leave as a view and only use on the app/dashboards. we shouldn't use them as a shortcut source in spells downstream of these, i.e. for cross-chain spells. we should still follow the normal spell per blockchain (bonus if macro is used to centralize), then build cross-chain on top. it's how we can keep the processing/load time reasonable within spellbook.

i see that you aren't doing that in this PR, but just an fyi. we'll have to have a larger conversation if that is wanted in the future

models/evms/evms_transfers.sql
```diff
@@ -0,0 +1,49 @@
+{{ config(
+        tags = ['dunesql'],
+        alias = alias('transfers'),
```
jeff-dude, line 3: just double checking here --
you have alias `transfers` but the source tables below are `blocks`? is that correct?

MSilb7:
> > @MSilb7 u up? new chain about to drop in `evms`
> 
> something i mentioned to @MSilb7 in dm's: while we approved (and will continue to allow) the `evms` spells, the intention was to leave as a view and only use on the app/dashboards. we shouldn't use them as a shortcut source in spells downstream of these, i.e. for cross-chain spells. we should still follow the normal spell per blockchain (bonus if macro is used to centralize), then build cross-chain on top. it's how we can keep the processing/load time reasonable within spellbook.
> 
> i see that you aren't doing that in this PR, but just an fyi. we'll have to have a larger conversation if that is wanted in the future

Yeah I'm a hard yes vote for `evms` to be a native table. There's too much overhead thinking of things as distinct chains now. I barely ever do "one chain only" queries anymore.

Especially as more L2s launch, segmenting by chain will make things completely unusable (imo).

MSilb7:
> @MSilb7 u up? new chain about to drop in `evms`

I do wonder if we should add a 'chain_type' field natively in emvs (i.e. mainnet vs testnet) versus having someone need to join to evms_info

jeff-dude:
> Yeah I'm a hard yes vote for `evms` to be a native table. There's too much overhead thinking of things as distinct chains now. I barely ever do "one chain only" queries anymore.
> 
> Especially as more L2s launch, segmenting by chain will make things completely unusable (imo).

raising internally, will see what we can do

hildobby:
> > @MSilb7 u up? new chain about to drop in `evms`
> 
> I do wonder if we should add a 'chain_type' field natively in emvs (i.e. mainnet vs testnet) versus having someone need to join to evms_info

Yeah I thought about it, maybe a column like a boolean `is_mainnet` would be the most efficient usage-wise & storage-wise here?

models/evms/evms_transfers.sql
hildobby: that file isn't supposed to be here, just deleted it

jeff-dude (approved):
happy to proceed here, then you guys can change columns later if needed

PR #4520, opened 2023-10-04
NFT cleanup of transfers and washtrades

0xRobin:
Moving the transfer and wash_trades models + schema's in the new directory.
No removal of spark models yet, will do in a separate later PR.


jeff-dude (approved):
works for me

jeff-dude:
i need to merge #4442 without the modified step in prod, so i think i can group this one in too as it's just moving files around and cleanup. no data needs to be refreshed.

Issue #4521, opened 2023-10-04
Migrate and fix erc20.transfers


aalan3:
## Problem
We have a bunch of issues with `erc20.transfers` at the moment and we have been neglecting the fixes due to the migration. Right now the transfers tables are a bit all over the place with how they work and types, specially around `amount_raw`. We want to clean these up so that we can fix balances.

This issue exists to fix and migrate the spells and discuss the way forward. 

## Goal
The goal is for all evms to have a working `erc20.transfers` using a macro, so that implementation is the same (as much as possible) across chains. We want to fix these so that we can then start to fix erc20 balances and bring it back.

## How
Going forward we want to use macros for these spells since they are all very similar. In #4366 (shout out to @henrystats for this!) macro was introduced for erc20.transfers.

The above macro needs to be updated in the following way:
* `amount_raw` is a double but should be the regular `uint256` type which `evt_Transfer` uses
* Add predicates to the wrapped token section
* Add incremental predicates to the materialization as well

### Breaking changes
Right now the macro adds a few columns and removes `unique_transfer_id`. Adding columns can be breaking but is usually not. But removing  `unique_transfer_id` is breaking. If we agree that this is fine we can remove it.

### `evms.erc20_transfers`
There is currently a table called `evms.erc20_transfers` which exists to make querying `evt_transfer` tables cross chain easier. 

@hildobby @MSilb7 do you see a need for `transfers_<blockchain>.erc20_transfers` to also have a `evms` view and if so should we change the name of `evms.erc20_transfers` to ``evms.erc20_evt_transfers` for example? 

It would be good to try to get naming a bit more consistent. This is a breaking change though so if we do it we can update the related queries automatically with an internal tool we have for these things.

Let me know what you think, how do this table relate to the evms one? Could we even just replace it?

## Related 
Some bug issues:
https://github.com/duneanalytics/spellbook/issues/3431 - "from" is show literally as `from` instead of the actual address
https://github.com/duneanalytics/spellbook/issues/3108  - Types are wrong but I think this one might be fixed?

There are also more than a few [PRs open](https://github.com/duneanalytics/spellbook/pulls?q=is%3Apr+is%3Aopen+transfers) that addresses the transfers tables.

jeff-dude:
fyi @henrystats @tomfutago @primatepngs @lajarre
i believe you all have open PRs related to transfers and/or balances

[...]

PR #4522, opened 2023-10-04
Erc20 transfers macro

aalan3:
This PR is part of #4521.

I've updated the following:
* Uses the `incremental_predicate` macro 
* `amount_raw` is no longer converted to double - it should be `uint256`
* Updated Polygon to comply


henrystats:
@aalan3 it's casted to double because uint256 can't be negative, I think you'd have to cast to INT256? Which I think wasn't working as well but id on't remember why..

aalan3:
@henrystats have a look at the current version, this is now using `try_cast` to get to `int256`.

henrystats:
awesome, looks good to me!

macros/models/_sector/transfers/erc20/transfers_erc20.sql
```diff
@@ -4,99 +4,103 @@ WITH
...
             evt_tx_hash,
-            evt_index, 
+            evt_index,
             evt_block_time,
-            to as wallet_address, 
+            to as wallet_address,
             contract_address as token_address,
-            CAST(value as double) as amount_raw
-        FROM 
+            try_cast(value as int256) as amount_raw
```
lajarre, line 14: It seems to me that this would produce some data loss on some potentially relevant pieces of data. Are we confident these are not relevant cases?

For example, a treasury address receiving more than half the token supply of a token which total supply is `2^256-1` would be filtered out by the `try_cast` here.

See my comment there: https://github.com/duneanalytics/spellbook/issues/4521#issuecomment-1754028324

jeff-dude:
(assuming this is old, in favor of #4633, closing this now)

PR #4524, opened 2023-10-04
Daily EVMs Usage

MSilb7:
Aggregated daily usage data for EVMs (since [this query](https://dune.com/queries/3016144) started failing all of the sudden).

This hopefully will run, but may fail since EVMs is a view versus a native table.


Hosuke (approved):
LGTM.✅
Thank you @MSilb7

models/evms/evms_aggregate_daily_usage.sql
```diff
@@ -0,0 +1,94 @@
+   {{
+  config(
+        tags = ['dunesql'],
+        schema = 'evms',
+        alias = alias('aggregate_daily_usage'),
+        materialized ='incremental',
+        file_format ='delta',
+        incremental_strategy='merge',
+        unique_key=['blockchain', 'block_date'],
+        partition_by=['blockchain', 'block_date'],
```
Hosuke, line 10: Do you have any better partition strategy here?

jeff-dude:
let's use this as a talking point for tomorrow's conversation

models/evms/evms_aggregate_daily_usage.sql
MSilb7: what would a good alternative be? I didn't have a major reason for picking these, other than them being the unique keys. But I would always read multiple chains at once.

jeff-dude:
for now, i'd like to avoid building too many downstream spells on top of the `evms` schema spells. there are ongoing conversations around materializing outside of spellbook and/or exploring cross-chain decoded tables (tbd on all of this)

PR #4530, opened 2023-10-05
Fix excluded spell

ppclunghe:
fix for lido_liquidity_arbitrum_kyberswap_v2_pools excluded due to duplicates


models/lido/liquidity/arbitrum/lido_liquidity_arbitrum_kyberswap_v2_pools.sql
jeff-dude, line 1: looks like your branch was out-of-date from main branch. after i merged in main, there is only one file edit here (i assume as expected). you will need to remove the `prod_exclude` tag from the config block at the top of the model, in order for it to run in CI steps and be re-enabled in prod post-merge.

after you push a commit to remove that tag, let's see how CI runs 👍

PR #4533, opened 2023-10-05
SPE-200 restructuring dex.trades with a macro approach

Hosuke:
Things to discuss:

1. Requirement of Macros/Models: Do we need macros implemented in macros/models/_sector/dex/platforms/uniswap_v2_forked_base_trades.sql?

2. Consistency with Proposed Ideas: I would like to confirm whether our current implementation aligns with the ideas presented in the https://github.com/duneanalytics/spellbook/issues/4637 and https://github.com/duneanalytics/spellbook/discussions/4625


Hosuke:
There are a small number of differences in row counts. 
Other than that, the initial models appear to be fine.
[Sample dex_ethereum_trades_beta](https://dune.com/queries/3081639)

0xRobin:
I'll give this a closer look on monday and share some more insights now that we've been using the `nft.trades` macro setup for some time.

jeff-dude:
pushed some commits with my thoughts:
- use new incremental filter macro across all modified here
- change table aliases in macro to more intuitive names, such as `erc20_bought` rather than `erc20a`
- reorg directory structure, so it matches schema / table name a bit better `.../platforms/<project>/<blockchain>/model.sql`
- share one schema file, one source file across all platforms (i could go either way on this one, but may be nice to have one at root of `platforms/` directory) -- let me know if you guys disagree
- remove partitions at project-level spells
- remove `amount_usd` at project-level, as we don't have prices joined in, so never used
- `cast()` instead of `try_cast()`
- move `v1` into schema name, rather than table name (like it was)
- not sure we need to do chain-specific dex level spells, i made it just `dex_enrich_trades` instead -- again, i'm open to convo here, but may be overkill
  - this change will require a bit more work here, but i had to step away, so i pushed all my local commits. i will comment on the code where i think we need changes and would like to see

models/_sector/dex/trades/dex_trades_beta.sql
```diff
@@ -0,0 +1,31 @@
...
+    ,('ethereum',   'uniswap',    '1',    ref('uniswap_ethereum_v1_base_trades'))
+    ,('ethereum',   'uniswap',    '2',    ref('uniswap_ethereum_v2_base_trades'))
+    ,('ethereum',   'uniswap',    '3',    ref('uniswap_ethereum_v3_base_trades'))
+] %}
+
+
+-- macros/models/sector/dex
+{{
+    dex_enrich_trades(
+        blockchain = 'ethereum'
```
jeff-dude, line 25: how can i set this to:
`blockchain = base_models[0]`?

jeff-dude: then we can control blockchain name in the above variable, keep it model specific, then we can be cross-chain here

models/_sector/dex/trades/dex_trades_beta.sql
```diff
@@ -0,0 +1,31 @@
...
+        ,models = base_models
+        ,transactions_model = source('ethereum', 'transactions')
```
jeff-dude, line 27: same here, need to make it:
`,transactions_model = source('base_models[0]', 'transactions')`
but it failed `dbt compile`, since i'm already in a variable i assume? how to nest them?

models/_sector/dex/trades/dex_trades_beta.sql
```diff
@@ -0,0 +1,31 @@
...
+    alias ='trades_beta',
+    partition_by = ['block_month'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['block_date', 'project', 'version', 'tx_hash', 'evt_index']
+    )
+}}
+
+-- (blockchain, project, project_version, model)
```
jeff-dude, line 13: ideally, we add a fifth element: `project_start_date` and pull from each model as they are in current dex trades

then we can pass into macro, apply else filter on inner join to base tx table

the way the macro is set up now though, it loops through base project models, but then just one join to base tx table, so not feasible. it may also be overkill to do all of this, so just suggesting to look into

jeff-dude: i don't think we'd want to loop around all tables and join base tx each time, that would be too many joins. so not sure my full thoughts here yet, thinking out loud

models/_sector/dex/trades/dex_trades_beta.sql
Hosuke: We can remove `blockchain` parameter in macro dex_enrich_trades() to implement this.
i.e.
```
{% macro dex_enrich_trades(
    models=[],
    transactions_model=null,
    tokens_erc20_model=null,
    prices_model=null
    )
%}

WITH base_union AS (
    {% for dex_model in models %}
    SELECT
        '{{ dex_model[0] }}' as blockchain,
        '{{ dex_model[1] }}' as project,
        '{{ dex_model[2] }}' as version,
...
```

models/_sector/dex/trades/dex_trades_beta.sql
Hosuke: I think the dbt compile failed because as base_models is a list of dex.trades, which is more like a 2D array. It usually needs to work with jinja loop.

models/_sector/dex/trades/dex_trades_beta.sql
Hosuke: I actually use the similar design as `map-reduce` for the joining stage. I divided the dex.trades pipelines into:

1. base trades info (directly read/parse from decoded sources table) from each dex.
2. joining with transaction info for a specific chain, which these trades from different dexs shared. (ideally this *.transactions is only reading for once for each chain).
3. merge trades from different chains into full dex.trades table.

Hosuke:
```
union stage:
[dex1 + dex2 + dex3 + dex4 + ...] on the same chain
join stage 1    || (Reading only once big table *.transactions)
                || Joining *.transactions using (tx_hash, evt_index)
               \\//
[dex.trades on a specific chain]
join stage 2    || (Reading only once big table prices.usd where blockchain = *)
                || Joining a specific chain's prices.usd using (block_time)
               \\//
[detailed dex.trades on a specific chain] + [detailed dex.trades on other chains]
```

models/_sector/dex/trades/platforms/_schema.yml
```diff
@@ -0,0 +1,135 @@
...
+      contributors: jeff-dude, masquot, soispoke, hosuke
+    config:
+      tags: [ 'ethereum', 'dex', 'trades', 'uniswap', 'v1']
+    description: "uniswap ethereum v1 base trades"
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - tx_hash
+            - evt_index
+      - dbt_utils.test_equal_rowcount:
```
Hosuke, line 48: Seems the `equal_rowcount` is not available in this version of dbt, maybe we can manually add this test into generic tests:
https://github.com/dbt-labs/dbt-utils/blob/main/macros/generic_tests/equal_rowcount.sql

Hosuke:
[Sample uniswap_v3_ethereum.trades_beta](https://dune.com/queries/3090559)

0xRobin:
I've put some of my thoughts about the current macro design in here: https://github.com/duneanalytics/spellbook/issues/4580

models/_sector/dex/trades/dex_trades_beta.sql
Hosuke: Also here is another implementation: #4607

Hosuke:
Implementing based on the pipeline mentioned here:
https://github.com/duneanalytics/spellbook/issues/4580#issuecomment-1768665454

models/_sector/dex/trades/platforms/uniswap/ethereum/uniswap_ethereum_v3_base_trades.sql
```diff
@@ -0,0 +1,49 @@
+{{ config(
+    tags=['dunesql'],
+    schema = 'uniswap_v3_ethereum',
+    alias = 'base_trades',
```
Hosuke, line 4: By the way since legacy spells are removed, do we still need the `alias()` here?
e.g.
```suggestion
    alias = alias('base_trades'),
```

macros/models/_sector/dex/platforms/uniswap_v2_forked_base_trades.sql
Hosuke, line 1: Not sure we need this macro, as even though there are many forked uniswap v2 dexes, but the detailed fields may vary.

jeff-dude:
thanks for the continued work here @Hosuke, much appreciated

i will try to prioritize this tomorrow. i keep running into a few other high priority items, but i need to help push this along.

0xRobin:
@Hosuke @jeff-dude There are some changes I made to the NFT setup that I believe we should mimic here after the last sector spell design discussion.

The PR (https://github.com/duneanalytics/spellbook/pull/4668) is still WIP but the main changes I think we should copy are:

1. Changing the enrichment macro to only do the enrichment (left join of all the meta tables), instead of join + enrichment
2. Split the final model in a union (materialized) base model, and the enrichment model.
3. Due to these changes I've moved away from this syntax (which I've found to be confusing looking back on it) 
```
{% set base_models = [
    ('arbitrum',   'uniswap',    '3',    ref('uniswap_arbitrum_v3_base_trades'))
] %}
```
  and back to a normal array of model references in the union model. 
  This does mean including the `blockchain`, `project` and `version` columns in the platform models.
4. Added the necessary transaction columns to the platform models (`tx_from` and `tx_to` , this can be achieved easily by using the `add_tx_from_and_to` macro), which means we can make the enrichment macro independent of the chain and just apply it at the top level union. 

-- 
Let me know if something is unclear or if I can help on anything. 
The relevant models to take a look at in my PR are:
- `models/_sector/nft/trades/ethereum/nft_ethereum_base_trades.sql` (chain specific base union)
- `models/_sector/nft/trades/nft_base_trades.sql` (full base union)
- `models/_sector/nft/trades/nft_trades_beta.sql` (enrichment model)
- `macros/models/_sector/nft/enrich_nft_trades.sql` (the enrichment macro)

0xRobin (changes requested):
@Hosuke Great work already!!!
There are some changes I made to the NFT setup that I believe we should mimic here after the last sector spell design discussion.

The PR (https://github.com/duneanalytics/spellbook/pull/4668) is still WIP but the main changes I think we should copy are:

1. Changing the enrichment macro to only do the enrichment (left join of all the meta tables), instead of join + enrichment
2. Split the final model in a union (materialized) base model, and the enrichment model.
3. Due to these changes I've moved away from this syntax (which I've found to be confusing looking back on it) 
```
{% set base_models = [
    ('arbitrum',   'uniswap',    '3',    ref('uniswap_arbitrum_v3_base_trades'))
] %}
```
  and back to a normal array of model references in the union model. 
  This does mean including the `blockchain`, `project` and `version` columns in the platform models.
4. Added the necessary transaction columns to the platform models (`tx_from` and `tx_to` , this can be achieved easily by using the `add_tx_from_and_to` macro), which means we can make the enrichment macro independent of the chain and just apply it at the top level union. 

-- 
Let me know if something is unclear or if I can help on anything. 
The relevant models to take a look at in my PR are:
- `models/_sector/nft/trades/ethereum/nft_ethereum_base_trades.sql` (chain specific base union)
- `models/_sector/nft/trades/nft_base_trades.sql` (full base union)
- `models/_sector/nft/trades/nft_trades_beta.sql` (enrichment model)
- `macros/models/_sector/nft/enrich_nft_trades.sql` (the enrichment macro)

Hosuke:
Based on the above comment, I will modify the current PR one step at a time.
1. Moving `blockchain, project and version` columns into the platform models.

Hosuke:
> 4. Added the necessary transaction columns to the platform models (`tx_from` and `tx_to` , this can be achieved easily by using the `add_tx_from_and_to` macro), which means we can make the enrichment macro independent of the chain and just apply it at the top level union.

Due to `block_number` may not be included in the decoded tables for the most dex_base_trades. I may use other approaches.

Hosuke:
3. Add `block_number` in base_trades, which is not existed before.

Hosuke:
[Sample uniswap_v3_ethereum.trades_beta](https://dune.com/queries/3145292)

hildobby:
@Hosuke are you planning on including tx_index (index from `.transactions` tables) in dex.trades? It would be awesome to have as tx ordering is very important to play back pool history and to look into mev trades

jeff-dude:
> @Hosuke are you planning on including tx_index (index from `.transactions` tables) in dex.trades? It would be awesome to have as tx ordering is very important to play back pool history and to look into mev trades

once we get the foundation in place here, we can revisit and add new columns

jeff-dude:
@Hosuke i pushed a bunch of commits to align my thoughts / feedback. i tried to separate commits to make it easy to follow, but one of the last commits i pushed a lot into one 😅 

i think we are much closer to the structure @0xRobin is doing for nft in his linked PR.

full disclosure: i haven't run this end-to-end, so highly likely it fails somewhere & i need to run soon to catch my flight to lisbon

macros/models/_sector/dex/platforms/uniswap_v2_forked_base_trades.sql
jeff-dude, line 1: i can't help but wonder if we should rename these two platform-specific macros to something more generic. is uniswap fork too specific? can we find a more appropriate name for both versions?

models/_sector/dex/trades/_schema.yml
```diff
@@ -0,0 +1,100 @@
...
+        name: tx_from
+        description: "Address which initiated the trade"
+      - &tx_to
+        name: tx_to
+        description: "Address which received the trade"
+      - &evt_index
+        name: evt_index
+        description: "Index of the corresponding trade event"
+
+  - name: dex_base_trades
```
jeff-dude, line 100: todo: finalize this section

macros/models/_sector/dex/platforms/uniswap_v2_forked_base_trades.sql
Hosuke: I am also thinking if we can have better naming for these macros.

Hosuke:
> @Hosuke are you planning on including tx_index (index from `.transactions` tables) in dex.trades? It would be awesome to have as tx ordering is very important to play back pool history and to look into mev trades

It will be very simple to add in future, as currently we use this macro for `transactions` related fields:
https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/add_tx_from_and_to.sql

You can discuss with Rob as well.

macros/models/_sector/dex/platforms/uniswap_v2_forked_base_trades.sql
```diff
@@ -0,0 +1,51 @@
...
+    )
+%}
+
+WITH dexs AS
+(
+    SELECT
+        t.evt_block_number AS block_number
+        ,t.evt_block_time AS block_time
+        ,t.to AS taker
+        ,t.contract_address AS maker
```
Hosuke, line 16: Also this line we may need to discuss, which is better for default `maker`? 
`t.contract_address` or `CAST(NULL AS VARBINARY)`?

tests/generic/compare_dex_trades_sample.sql
Hosuke, line 1: I tried to design a new test that helps ensure `dex.trades` data consistency for the new approach.

macros/test-helpers/check_seed_macro.sql
```diff
@@ -2,9 +2,19 @@
...
+    , filter=None
+    , with_needed=True
+) %}
 
-    with matched_records as (
+    {% if with_needed %}
+    with
+    {% endif %}
+    matched_records as (
         select
```
Hosuke, line 18: I also modified `check_seed_macro` slightly, which triggered ~400 check_* tests

Hosuke: Recovoered.

Hosuke:
The [dbt test failure](https://github.com/duneanalytics/spellbook/actions/runs/6690149058/job/18174924668?pr=4533#step:12:549) may be false positive:
https://dune.com/queries/3157499

Currently the new models keeps the data consistency with the original ones.

macros/models/_sector/dex/enrich_dex_trades.sql
```diff
@@ -0,0 +1,70 @@
...
+            ) AS amount_usd,
+        base.token_bought_address,
+        base.token_sold_address,
+        coalesce(base.taker, base.tx_from) AS taker,
+        base.maker,
+        base.project_contract_address,
+        base.tx_hash,
+        base.tx_from,
+        base.tx_to,
+        base.evt_index
```
grkhr, line 51: some of the projects just don't emit events. i think it'd be great to add `trace_address` or `call_trace_address` as in current **dex_aggregator.trades** implementation. or just merge it in one column, but it should be named clearly.

models/_sector/dex/trades/base/_schema.yml
```diff
@@ -0,0 +1,24 @@
+version: 2
+
+models:
+  - name: uniswap_v3_base_base_trades
```
grkhr, line 4: this naming standard is very confusing as we have **base** blockchain 😬

models/_sector/dex/trades/bnb/dex_bnb_base_trades.sql
```diff
@@ -0,0 +1,47 @@
...
+-- (blockchain, project, project_version, model)
+{% set base_models = [
+    ref('uniswap_v3_bnb_base_trades')
+] %}
+
+WITH base_union AS (
+    SELECT * FROM (
+        {% for base_model in base_models %}
+        SELECT
+            blockchain,
```
grkhr, line 18: why not just `select *` as schema already defined and determined in macro?

models/_sector/dex/trades/base/_schema.yml
Hosuke: Yes, do we have any better naming options?

models/_sector/dex/trades/bnb/dex_bnb_base_trades.sql
Hosuke: We try to list all columns in all roll-up spells, to avoid column mismatching. There were bugs happened before and they were very hard to identify.

models/_sector/dex/trades/base/_schema.yml
hildobby: I agree it's a bit confusing, but it should remain consistent with the nft structure imo to avoid having too different structures. Probably something like `raw`?

grkhr: `core_` / `root_` / `master_` / `anchor_` / `initial_` / `raw_`

models/_sector/dex/trades/bnb/dex_bnb_base_trades.sql
grkhr: as far as they use one macro and there aren't if-else conditions on columns, it should be fine. but yes, i agree that explicit is better then implicit :)

models/_sector/dex/trades/base/_schema.yml
Hosuke: Those are good ideas, I will wait for other teammates' decision. I think `raw_` is a good choice.

grkhr: actually i kind of don't like `raw_`. it's just underlying fundamental macro which has some transformations. and raw is raw :)

Hosuke: or `fundamental_`?

macros/models/_sector/dex/platforms/uniswap_v2_forked_base_trades.sql
jeff-dude: i'll push a commit shortly that cleans the names up a bit

macros/models/_sector/dex/platforms/uniswap_v2_forked_base_trades.sql
jeff-dude: for now, let's leave as-is, since this is how it was in existing dex trades

macros/models/_sector/dex/enrich_dex_trades.sql
jeff-dude: we just removed `trace_address` in the migration to dunesql, since not a single project within the `dex.trades` lineage provided a value for this column. they all populated events instead. i'd be curious to see examples of dex projects in the existing lineage that would use this field?

grkhr: i can't find examples :)

thanks for sharing the context!

models/_sector/dex/trades/base/_schema.yml
tomfutago: How about staging_ or stg_ ?

jeff-dude: i had also started typing out `staging_` or `stg_` then took it back. maybe staging is the right approach.
`dex.trades`
`dex.stg_trades`
`dex_<blockchain>.stg_trades`
`uniswap_<version>_<blockchain>.stg_trades`

grkhr: i think staging is the best option from of all the above. I also like just simple `stage_` prefix

grkhr: but.. shouldn't it be postfix instead of prefix?

jeff-dude: i just pushed with `stg` 😅 but open to `stage` too

tomfutago: I like `stg_` - short and sweet

Also, if there's more staging models in one folder, the "main" one can be pushed to the top of the list by adding extra underscore (so `stg__`) - that way it'll stay together with _schema and _sources files.

macros/models/_sector/dex/dex_fork_trades.sql
```diff
@@ -0,0 +1,108 @@
+{% macro dex_fork_v2_trades(
```
grkhr, line 1: I think this marco is applicable only for uniswap (or their forks) as you rely on `Pair_evt_Swap` schema and `Factory_evt_PairCreated` table. Check `openocean_optimism_trades` for example.

jeff-dude: yeah, most likely all uniswap forks, but we have a ton in the dex lineage. it felt weird naming the macro `uniswap_fork_` instead of something generic. happy to revisit though, if others feel like it's more important to call out that these contracts were forked from uniswap. thought maybe leaving out project specific names would be better

jeff-dude:
@Hosuke (or anyone else interested) this should be ready for review again. i'm comfortable with the current format, with a few minor caveats:
- do we need to revisit the tests and see why counts are off? we should compare counts of these CI table build to the actual `dex.trades` with filters as needed
- naming standards on macros
- adding new columns (though i'd like to do that in a follow up PR i think)

macros/models/_sector/dex/dex_fork_trades.sql
grkhr: My vote is for "uniswap_fork..."

macros/models/_sector/dex/enrich_dex_trades.sql
```diff
@@ -0,0 +1,79 @@
+{% macro enrich_dex_trades(
+    base_trades = null
```
grkhr, line 2: should we get rid of `base_` here too and replace to `stg_`?

grkhr:
@jeff-dude as far as I understood you are materializing 3 steps:

1. every {project}{version}{blockchain}.stg_trades
2. dex.stg_trades (which contains all of the tables above with duplicates_rank filter)
3. dex.trades (_beta) (which is enriched with metadata)

Is this intended? As step 1 and step 2 are the same data which will be stored twice

macros/models/_sector/dex/enrich_dex_trades.sql
jeff-dude: yep, good catch, pushing shortly

jeff-dude:
> @jeff-dude as far as I understood you are materializing 3 steps:
> 
> 1. every `{project}{version}{blockchain}.stg_trades`
> 2. `dex.stg_trades` (which contains all of the tables above with `duplicates_rank` filter)
> 3. `dex.trades` (`_beta`) (which is enriched with metadata)
> 
> Is this intended? As step 1 and step 2 are the same data which will be stored twice

let's give it a shot with step 2 as a view, see how performance compares (will push in a sec)

macros/models/_sector/dex/dex_fork_trades.sql
jeff-dude: @Hosuke @0xRobin @hildobby @aalan3 
anyone have an opinion on naming standard for dex macros which are more than likely uniswap forks (i don't know how to quickly validate that)? a generic name like above, or just `uniswap_fork_<version>_trades`?

in this PR, we have one example dex fork, but there'll be plenty more to come

models/_sector/dex/trades/dex_stg_trades.sql
```diff
@@ -0,0 +1,54 @@
+{{ config(
+    schema = 'dex'
+    , alias = 'stg_trades'
+    , materialized = 'view'
```
jeff-dude, line 4: @grkhr i think one reason to materialize this level is because each blockchain-specific spell below in the list are views that join to `<blockchain>.transactions` base table to grab `tx_from`/`tx_to` columns. rather than materialize each one of those, we can materialize here once. then the final stage will materialize with further logic, such as metadata joins and column metrics.

here is the line to join to base table per blockchain:
https://github.com/duneanalytics/spellbook/pull/4533/files#diff-9c7ba2ea0d536632d3c6b2c9a7483c8dbf03d497288f35620ac02991e5b946e5R42

grkhr: Makes sense! Just looked again through all of the logic. Let's leave step 2 materialized.

grkhr: I was thinking that project-chain-specific spells should materialize tx_from/to, but realized that it will be inconvenient to the end contributors. Always better to keep requirements to the end users simple.

Hosuke:
> @Hosuke (or anyone else interested) this should be ready for review again. i'm comfortable with the current format, with a few minor caveats:
> 
> * do we need to revisit the tests and see why counts are off? we should compare counts of these CI table build to the actual `dex.trades` with filters as needed

The diff in counts may due to the latest refresh time. If we add an ending time filter, like `where block_time < TIMESTAMP '2023-11-01'`, we may have closer results.

> * naming standards on macros
> * adding new columns (though i'd like to do that in a follow up PR i think)

We may consider if we can add `tx_index` in dex.trades and `add_tx_from_and_to` macro.

macros/models/_sector/dex/dex_fork_trades.sql
markusbkoch: +1 for making it clear in the macro name what the original project being forked is. Also, anticipating there'll be multiple projects to be forked, wouldn't it make more sense to silo each macro in its own sql file?

jeff-dude: works for me!
pushing update soon.

jeff-dude:
> > @Hosuke (or anyone else interested) this should be ready for review again. i'm comfortable with the current format, with a few minor caveats:
> > 
> > * do we need to revisit the tests and see why counts are off? we should compare counts of these CI table build to the actual `dex.trades` with filters as needed
> 
> The diff in counts may due to the latest refresh time. If we add an ending time filter, like `where block_time < TIMESTAMP '2023-11-01'`, we may have closer results.
> 
> > * naming standards on macros
> > * adding new columns (though i'd like to do that in a follow up PR i think)
> 
> We may consider if we can add `tx_index` in dex.trades and `add_tx_from_and_to` macro.

i think we can feel confident in row counts / raw amount totals:
image


as for `tx_index`, i'd like to add in follow up PR that is cleaner, since we will need to change that existing macro and how it's used a bit. i may even rename it, since it's now pulling more columns. i don't know the full impact of that, so don't want to introduce here.

jeff-dude:
CI tests are looking good too 🔥 

run times were improved by adding incremental predicates on the target & then also adding in a missing incremental filter in the `enrich_dex_trades` macro, now incremental looks much more reasonable

macros/models/_sector/dex/dex_fork_trades.sql
hildobby: I just want names to not be confusing, `uniswap_fork` makes it seem like Uniswap is behind it with, similar to how `nft_base_ethereum.trades` seems to be related to the blockchain base although it isn't (even for forks imo). My suggestion would be to avoid using any already established names unless that name is directly linked/behind that data, that could also mean establishing some standardised way to name forks. Maybe something like `unifork`, `uniswapfork`, `dex_unifork` or whatever else doesn't contain the full `_uniswap_` name

Btw, @0xBoxer made a PR for all Uniswap v2 forks at ETHLisbon this weekend, this probably is the most scalable method as all that's needed is name mapping on top (which is part of the PR). https://github.com/duneanalytics/spellbook/pull/4730

Also curious about @0xRobin & @aalan3's thoughts here

macros/models/_sector/dex/enrich_dex_trades.sql
```diff
@@ -0,0 +1,83 @@
...
+        AND p_bought.contract_address = base.token_bought_address
+        AND p_bought.blockchain = base.blockchain
+    LEFT JOIN
+        prices as p_sold
+        ON p_sold.minute = date_trunc('minute', base.block_time)
+        AND p_sold.contract_address = base.token_sold_address
+        AND p_sold.blockchain = base.blockchain
+    {% if is_incremental() %}
+    WHERE
+        {{ incremental_predicate('base.block_time') }}
```
grkhr, line 75: if you'd move this to the upper level it should improve incremental performance more. i mean:

```
with 

...

stg_cte as (
	select * from {{ stg_trades }}
	{% if is_incremental() %}
		where {{ incremental_predicate('base.block_time') }}
	{% endif %}
)

...

from stg_cte
left join 

...
```

filtering data as early as possible before joins almost always helps.

P.S: also — `base` :/

jeff-dude: i can give it a shot, but not sure we'll gain much from what i've seen before.

ugh, why did they have to name their blockchain `base`, that's such a useful word for aliases and such 🥲

grkhr: i hope Stargate won't release their STG chain haha

grkhr: > not sure we'll gain much from what i've seen before

`dex_trades_beta  [SUCCESS in 14.06s]`→ `dex_trades_beta  [SUCCESS in 9.90s]`
not much, but still around 1.5x which can easily become x2-x3 on a larger dataset i think

grkhr: @jeff-dude yep it's just bias. Trino's optimizer is pretty smart. 

image
image

Anyway good to know!

Hosuke: Wow which tool are you using here, seems pretty helpful.

macros/models/_sector/dex/dex_fork_trades.sql
aalan3: When it comes to `base` I agree that it can be confusing. The way to "tell" the difference is that `base` in the namespace (`nft_base....`) would be the chain, whereas in the table name (`....trades_base`) it refers to what we are doing here. Alternative could be maybe `core` that was suggested by someone? Another option could be to do some sort of `parent` or `main` naming. (btw Zora being a project and a chain also adds some more confusion to our taxonomy). 

Personally I actually associate `uniswap_fork` with non official uniswap things (pancakeswap etc), I think because usually in open source projects a fork means that someone broke out of the original project and created a separate org (but ofc not always). Curious if the uniswap contracts have a different name that could be used, like `seaport` vs `opensea`. Only thing I could find was `uniswap protocol Vx`.

grkhr: > Personally I actually associate uniswap_fork with non official uniswap things (pancakeswap etc),
good point. i can suggest a `_compatible` postfix instead of `_fork`

> When it comes to base I agree that it can be confusing
seems that we had a consesnsus in the other thread and decided to name these things as `_stg`
https://github.com/duneanalytics/spellbook/pull/4533/files/bd6ff597cd299203cfd03d498cc65b6783cfaeb2#diff-60520bb14954436eba8e725c491507a5165a3f6b8a2a0fcff85af99cf6f7cb1b

aalan3: >good point. i can suggest a _compatible postfix instead of _fork

I like `_compatible` because it works for both the forks and uniswap itself.

I'll look at the link, thanks!

jeff-dude:
thank you again to all who provided feedback and helped push this beta version of the new `dex.trades` design 🙏 

while everything looked good on the logic within, there was some back and forth on naming standards across PR comments, dm's & internal convo. i've pushed some final commits with what we're sticking with for now:
- back to `base_` prefix on upstream building block spells. while having a `base` blockchain causes some confusion, this still seems most logical between all the options (stage can be confusing, either different env altogether or a staging table which usually means truncate/load to prep for an incremental load downstream)
- stick to `uniswap_` name in generic dex macros, but change `fork` to `compatible`

i plan to merge this soon, then i can revisit the linked gh issue and summarize our findings within this PR. we can also use that issue to continue conversation as we test out the beta and obtain more feedback.

PR #4535, opened 2023-10-05
Drop airdrop claims

aalan3:
Dropped airdrops claims, the blockchain files and one more level down for the project files.


aalan3:
The CI check here for schemas is not happy here. Not sure I should add dummy models or if there's another way

jeff-dude (approved):

aalan3:
@jeff-dude don't merge this just yet, if we do it will fail CI for all subsequent PRs. We need to update the check or script first.

jeff-dude:
> @jeff-dude don't merge this just yet, if we do it will fail CI for all subsequent PRs. We need to update the check or script first.

yeah, holding off, just marking for my awareness to find later

jeff-dude:
closing due to latest approach discussed in our grooming session, where we don't need lineage drops

PR #4537, opened 2023-10-05
Drop dao votes, proposals, transactions and balances legacy

aalan3:
There's still addresses left.


jeff-dude (approved):

jeff-dude:
closing due to latest approach discussed in our grooming session, where we don't need lineage drops

PR #4540, opened 2023-10-05
`labels` remove airdrop until upload source is fixed

jeff-dude:
fyi @MSilb7 
the name of the data upload source has changed and now prod is failing. i'm not sure how to find the new names to update, so i'm excluding.


Hosuke:
I think I know the new source name, which is:
`dune.oplabspbc.dataset_op_airdrop1_addresses_detailed_list`
Seems the path has been changed.

aalan3:
@Hosuke yes that's probably right, those files were uploaded to two namespaces (`dune_upload` was the old way)

PR #4541, opened 2023-10-06
Re-enable airdrop labels

aalan3:
This PR fixes https://github.com/duneanalytics/spellbook/pull/4540


macros/dune/source.sql
```diff
@@ -1,6 +1,6 @@
 {% macro source(source_name, table_name) %}
   {% set rel = builtins.source(source_name, table_name) %}
-  {%- if target.type == 'trino' -%}
+  {%- if target.type == 'trino' and source_name != 'dune' -%}
```
jeff-dude, line 3: i guess this macro change breaks the legacy check step?

aalan3: I think everything becomes `modified` which then triggers the legacy check for all unmigrated models?

jeff-dude:
i believe this is the PR that @couralex6 mentioned he could take a look at, in terms of changing state modified to only body in the CI test so that the macro doesn't trigger everything?

we promised alan notes, so just wanted to follow up here, even if we don't make any progress since we're focused on finalizing migrating spells

jeff-dude:
closing for now, until we find need for resolving dune file uploads

PR #4416, opened 2023-09-25
RW-5: cleanup block_date in base trades models

0xRobin:
Clean up partitioning and `block_date` columns in the lower levels of nft.trades (beta).

To recap:
- no partitioning and `block_date` in platform specific models
- partitioning on `block_month` in `nft_ethereum_trades_beta` and `nft_events`

PR #4542, opened 2023-10-06
Revert "Revert "RW-5: cleanup block_date in base trades models (#4416)""

0xRobin:
Reverts duneanalytics/spellbook#4539


0xRobin:
@jeff-dude thanks for catching the issue in prod. Here's the fix.
I checked all the base_trades models and only sudoswap has any downstream dependencies, so we should be good.

jeff-dude (approved):

PR #4543, opened 2023-10-06
new CI cluster setup: add is_legacy in alias macro

jeff-dude:
i'm not 100% sure if it'll break prod or not without the `legacy_model=True` param in the alias macro, so i'm playing it safe here

also added a quick comment in the test model for clarity to anyone poking around

### DIFF
--- a/models/test/trigger_scale.sql
+++ b/models/test/trigger_scale.sql
@@ -5,4 +5,9 @@
         )
 }}
 
+/*
+this dummy model is intended only to be used in the gh action for PR CI tests
+the goal is to run this on loop until the cluster is spun up, as a model run helps turn the cluster back on after auto-shutdown
+*/
+
 SELECT 1 as test
--- a/models/test/trigger_scale_legacy.sql
+++ b/models/test/trigger_scale_legacy.sql
@@ -1,6 +1,6 @@
 {{ config(
         schema='test',
-        alias = alias('trigger_scale_legacy'),
+        alias = alias('trigger_scale_legacy', legacy_model=True),
         tags= ['legacy','prod_exclude']
         )
 }}


aalan3:
I think this will add `_legacy` to the name? https://github.com/duneanalytics/spellbook/blob/main/macros/dune/alias.sql

jeff-dude:
> I think this will add `_legacy` to the name? https://github.com/duneanalytics/spellbook/blob/main/macros/dune/alias.sql

i think it does the opposite? helps to avoid adding the suffix? otherwise, all of our spark models would have that in the name?

models/test/trigger_scale.sql
```diff
@@ -5,4 +5,9 @@
         )
 }}
 
+/*
+this dummy model is intended only to be used in the gh action for PR CI tests
+the goal is to run this on loop until the cluster is spun up, as a model run helps turn the cluster back on after auto-shutdown
```
belen-pruvost, line 10: 👍

models/test/trigger_scale_legacy.sql
```diff
@@ -1,6 +1,6 @@
 {{ config(
         schema='test',
-        alias = alias('trigger_scale_legacy'),
+        alias = alias('trigger_scale_legacy', legacy_model=True),
         tags= ['legacy','prod_exclude']
```
belen-pruvost, line 4: Considering it has the prod_exclude tag, is this enough for not building this model on spark? (and therefore, if that's the case, then there wouldn't be a need to remove legacy from the model name)

jeff-dude: it is, but then again, that could cause issues if `test` schema doesn't exist on spark 🤔 i will monitor to see how it goes

PR #4529, opened 2023-10-05
Add zkSync to `nft.transfers`

hildobby:

PR #4548, opened 2023-10-06
Add zkSync to `sectors.nft.transfers`

lgingerich:


jeff-dude:
there appears to be another open PR (#4529) which does the same thing. since that was opened first, and has started the review, i will close this for now 👍

PR #4549, opened 2023-10-06
SPE-143 Opensea Views

couralex6:
Not in the `nft_trades` lineage, but some people might still use these views. Migrating them since they're easy to translate


0xRobin (approved):
:heavy_check_mark:

jeff-dude (approved):

PR #4563, opened 2023-10-09
[Easy] Add more missing erc20 tokens

harisang:
Easy PR that adds a few more missing ethereum erc20 tokens


harisang:
@jeff-dude Whenever you have time, could you have a look to let me know why some checks are failing? I skipped updating the legacy file, as was suggested to me, but it might be that this is causing some checks to fail (but maybe I am missing something here). Thanks!

jeff-dude (approved):
looks like you're past the issue, based on current state. sorry for missing it earlier, not sure what it was 🤷‍♂️

PR #4566, opened 2023-10-09
Materialise `fungible_ethereum.transfers`

hildobby:
With the `fungible.transfers` tables acting as the equivalent of `nft.transfers` for native+erc20 tokens, it would be super useful to materialise so I and others can start using it more (it saves a lot of work I otherwise have to do in almost all queries that look into erc20/native transfers) since it's very slow to use in its current view state

Here is a first PR that materialises it for ethereum so we can talk about it and iron out potential bugs before doing the same for other chains too


pandajackson42:
I want to express my strong support for this PR. Making `fungible.transfers` materalized are valuable and would greatly benefit wizards' lifes.

MSilb7:
Strong +1 on this & any "query UX improvement/simplification" tables like this being materialized.

jeff-dude:
seems `dbt compile` is failing here? can you run locally to ensure compile works?

then we can see performance on the CI run

hildobby:
> seems `dbt compile` is failing here? can you run locally to ensure compile works?
> 
> then we can see performance on the CI run

Ah, I accidently added a character which bricked the spell, just fixed it and it compiles fine locally now!

MSilb7:
https://dune.com/queries/3106967

^ Another case, this query would save a few hundred lines (and analyst time) if `fungible.transfers` was materialized. The view was too slow to use.

MSilb7:
> > https://dune.com/queries/3106967
> 
> > ^ Another case, this query would save a few hundred lines (and analyst time) if fungible.transfers was materialized. The view was too slow to use.
> 
> Can you explain which part of the query was replaced with a manual approach here?

Lines 28 - 394 (except the BTC part) could be replaced by `fungible.transfers`

This is union all'ing native token transfers and ERC20 transfers across ethereum, base, and polygon. Other chains are excluded for now for simplicity. But a `fungible.transfers` table would condense all of this to one query.

Current pain is the copy and paste for chains, amount of lines of code, & the requirement to make edits in every instance. This makes it incredibly more error prone and a worse query UX.

The current `fungible.transfers` is unusable as a view, because this query takes too long (hits free tier timeout errors).

jeff-dude:
closing in favor of the work @aalan3 is doing on the transfers spells. we can consider reopening this after that is done, to see if it makes sense to have more spells built on this (or maybe they could be repurposed to read from the new spells)

PR #4572, opened 2023-10-10
change CI cluster trigger to use dbt debug

0xRobin:
~~change CI steps order.
Compile needs to happen before the cluster trigger step, otherwise the cluster trigger step fails on compile errors and repeats for an hour.~~

Change the CI cluster trigger step to use `dbt debug`.
This saves us the additional compile time and also resolves any stuck-in-loop scenario's when having compile errors.


aalan3:
doesn't `dbt run` also compile? not sure I understand why this is failing

0xRobin:
> EDIT: Ahh I think I get it, if there is a compile error you'd be stuck on the loop.

indeed. The compile error on the trigger run gets treated the same as 'cluster is not yet available, let's try again'

0xRobin:
@jeff-dude to validate this we should run the CI on this PR when the cluster is down. to make sure the compile step doesn't fail when it can't reach the cluster. I'm not sure anymore if the CI compile is dependent on the connection (as I can compile locally without one)

0xRobin:
Ok after slack convo, I'll adapt the step to use `dbt debug` instead of running the trigger model. 
dbt debug does a connection test which runs `select 1 as id` under the hood.
This will save us a compile (and thus ~2min) on every CI run.

.github/workflows/dbt_slim_ci.yml
```diff
@@ -56,7 +56,7 @@ jobs:
 
       - name: dbt dependencies
         run: "dbt deps"
-      
+
       - name: Run dbt model on Trino if dunesql engine is used, and retry until cluster is available
```
belen-pruvost, line 60:
```suggestion
      - name: Run dbt debug on Trino if dunesql engine is used, and retry until cluster is available
```

belen-pruvost (approved):
I've re run the CI checks of this PR and I can see that the `select 1 as id` query is being correctly routed on the spellbook-ci UI. So this works 🎉

jeff-dude:
thanks @0xRobin for simplifying!

i will close this, as #4574 does this change and also adds to other workflows in the repo.

PR #4573, opened 2023-10-10
RW-3: Fix literal casts in NFT sector

0xRobin:
Post migration todo: 
replace any `cast(123 as double)` to `double '123'` which is the preferred approach. 

Regex
find: `(cast\(|CAST\()(\d+)(\s*as\s*)([^\)\s]*)\)` 
replcae: `$4 \'$2\'`


0xRobin:
@jeff-dude only missing seed failures here. Don't really want to go through all of them for this trivial change.
Hope we can merge with the errors. :pray:

jeff-dude (approved):
yeah, that's cool with me, i'll save to merge when it's a solo merge so it's easier to track / revert if needed

PR #4579, opened 2023-10-11
Add WardenSwap DEX

chain-l:
Adding wardenswap trades data


Hosuke (approved):
LGTM.✅
Thank you @chain-l 
[Sample wardenswap_optimism_trades]()

dbt_project.yml
```diff
@@ -1425,4 +1425,9 @@ models:
...
       +schema: openxswap
       optimism:
-        +schema: openxswap_optimism
\ No newline at end of file
+        +schema: openxswap_optimism
+
+    wardenswap:
+      +schema: wardenswap
+      optimism:
+        +schema: wardenswap_optimism
```
jeff-dude, line 1433: we're looking to eventually clean up the project file a bit by removing schema definitions there, rather put the schema name in the model config block itself.

do you mind removing modifications to this file and ensuring each new spell has the schema config properly applied?

jeff-dude (changes requested):

dbt_project.yml
chain-l: Done. 
Removed from dbt_project.yml and have added schema in config.

Let me know if there is anything else.

jeff-dude (approved):
thank you!

PR #4582, opened 2023-10-11
Use incremental predicate in contract mappings

aalan3:
This table causes is seemingly causing some stability issues in the DBT run.  This PR contains two changes to help with that:

1. I added `incremental_predicate` everywhere which makes the models go back 1 day instead of 7 (this is something that all incremental models will have eventually) @MSilb7 I couldn't figure out if it's possible to also add `incremental_predicates` in the config to also filter the destination table, could you please assist on this?
2. There's a `daily` tag on the contract mapping models to run them once a day instead of hourly.


jeff-dude (approved):
this is a good step towards making these spells more efficient. @MSilb7 has a few more PRs that look to continue enhancing these macros in order to add more blockchains efficiently

i need to prep the prod env for this prior to merging:
- new job that mimics hourly, but make it daily and add `tag:daily` to the select
- edit existing hourly job to exclude `tag:daily` in all steps

MSilb7:
> This table causes is seemingly causing some stability issues in the DBT run. This PR contains two changes to help with that:
> 
> 1. I added `incremental_predicate` everywhere which makes the models go back 1 day instead of 7 (this is something that all incremental models will have eventually) @MSilb7 I couldn't figure out if it's possible to also add `incremental_predicates` in the config to also filter the destination table, could you please assist on this?
> 2. There's a `daily` tag on the contract mapping models to run them once a day instead of hourly.

1 day vs 7 day is fine. What do you mean by adding to the destination table?

aalan3:
>What do you mean by adding to the destination table?

On the incremental run we do a `merge into` which in practice becomes a join between the source (the running query) and the destination (the target where it's saved).  I can't tell if it's applicable here but on other tables since we know we are only selecting 1 day from the source we could also select 1 day from the destination. This is what `incremental_predicates` sets in the config.

jeff-dude:
pushed a couple commits:
- rename macro files to match macro name within, got confused for a second trying to find them
- update tag to `prod_daily`, as there are other models using `daily` for other purposes. this is more obvious anyway for us in orchestration
- try to add incremental predicates, see if this works

jeff-dude:
hitting worker node crashes :(

jeff-dude:
past the worker node crashes, but lots of failures on duplicates after the incremental run. that's not a good sign 🤔

jeff-dude:
closing in favor of #4319

PR #4584, opened 2023-10-12
Solana balances fix again

andrewhong5297:
When accounts on solana get closed, they don't emit any post token balance. So, we have to do this odd join to get the token_mint_address of a closed account (detected by post_balance of sol = 0) to get this, so that we don't have a bunch of wrong balances where the state of 0 is not captured. 

annoying but no other workaround that I could find

### DIFF
--- a/models/solana_utils/solana_utils_daily_balances.sql
+++ b/models/solana_utils/solana_utils_daily_balances.sql
@@ -6,7 +6,7 @@
         materialized='incremental',
         file_format = 'delta',
         incremental_strategy='merge',
-        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.day')],
+        incremental_predicates = [DBT_INTERNAL_DEST.block_time >= date_trunc('day', now() - interval '1' day)],
         unique_key = ['token_mint_address', 'address','day'],
         post_hook='{{ expose_spells(\'["solana"]\',
                                     "sector",
@@ -41,7 +41,7 @@ WITH
                   WHERE tx_success 
                   AND block_time > now() - interval '1' day
                   {% if is_incremental() %}
-                  AND {{incremental_predicate('block_time')}}
+                  AND block_time >= date_trunc('day', now() - interval '1' day)
                   {% endif %}
             ) WHERE latest_tk = 1
       )
--- a/models/solana_utils/solana_utils_daily_balances.sql
+++ b/models/solana_utils/solana_utils_daily_balances.sql
@@ -46,7 +46,6 @@ WITH
                         AND aa.token_mint_address is null --only join on the empty token mints
                         AND tk.created_at <= aa.block_time --only get token mints that were created at or before this account activity
                   WHERE tx_success
-                  AND block_time > now() - interval '1' day
                   {% if is_incremental() %}
                   AND block_time >= date_trunc('day', now() - interval '1' day)
                   {% endif %}
--- a/models/solana_utils/solana_utils_daily_balances.sql
+++ b/models/solana_utils/solana_utils_daily_balances.sql
@@ -38,6 +38,7 @@ WITH
                         AND aa.token_mint_address is null --only join on the empty token mints
                         AND tk.created_at <= aa.block_time --only get token mints that were created at or before this account activity
                   WHERE tx_success 
+                  AND block_time > now() - interval '30' day
                   {% if is_incremental() %}
                   AND block_time >= date_trunc('day', now() - interval '1' day)
                   {% endif %}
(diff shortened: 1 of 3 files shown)


andrewhong5297:
okay so it is not lost on me that this is an annoying table to have to edit/refresh. It goes over the memory limit so it needs a special endpoint for the first run. 

However, this is a data breaking issue that I've added a fix for. I've limited to just 1 day in the PR, we will need to take that out when we are ready to merge. I'm off next week but feel free to go ahead and merge without me here.

aalan3:
Can we incorporate some of my fixes here? https://github.com/duneanalytics/spellbook/pull/4501

Mainly using `{{ ref('solana_utils_token_accounts') }}` and adding `[DBT_INTERNAL_DEST.block_time >= date_trunc('day', now() - interval '1' day)],` in the config. I'm hoping we can improve performance a bit.

models/solana_utils/solana_utils_daily_balances.sql
```diff
@@ -14,33 +14,35 @@
...
-                  )
-            {% if is_incremental() %}
-            AND block_time >= date_trunc('day', now() - interval '1' day)
-            {% endif %}
+                  , coalesce(post_token_balance,0) as token_balance --tokens are already correct decimals in this table
+                  , coalesce(token_balance_owner,original_token_balance_owner) as token_balance_owner
+                  , tx_id
+                  , block_time
+                  --could be one token mint closing and another opening on the same day. edge case but it exists.
+                  , row_number() OVER (partition by address, coalesce(token_mint_address,original_token_mint_address), date_trunc('day', block_time) order by block_slot desc) as latest_balance
```
aalan3, line 28: Can't we do `max_by` block_slot and tx id here instead and get rid of the window function? I think that should be better for performance.

models/solana_utils/solana_utils_token_accounts.sql
```diff
@@ -3,10 +3,8 @@
         schema = 'solana_utils',
         tags = ['dunesql'],
         alias = alias('token_accounts'),
-        materialized='incremental',
+        materialized='table',
```
aalan3, line 6: why can this no longer be incremental?

aalan3:
Also curious how did you find out that this is an issue? Would be good to know how we should debug potential Solana balance issues in the future.

andrewhong5297:
> Also curious how did you find out that this is an issue? Would be good to know how we should debug potential Solana balance issues in the future.

I was calculating NFT balances and noticed that some balances were not up to date. Then checked account_activity and saw that the values were null, and figured out that when Solana closes accounts there is no postTokenBalance posted so the logic was not capturing closed accounts as 0 balance.

andrewhong5297:
> Can we incorporate some of my fixes here? #4501
> 
> ~Mainly using `{{ ref('solana_utils_token_accounts') }}`~ and adding `[DBT_INTERNAL_DEST.block_time >= date_trunc('day', now() - interval '1' day)],` in the config. I'm hoping we can improve performance a bit.
> 
> EDIT: I saw that you already added `solana_utils_token_accounts`, thank you!

I see you use `incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.day')],` in your PR, should I use that too? or should I use `incremental_predicates = [DBT_INTERNAL_DEST.block_time >= date_trunc('day', now() - interval '1' day)],` instead?

models/solana_utils/solana_utils_daily_balances.sql
```diff
@@ -40,7 +41,7 @@ WITH
                   WHERE tx_success 
                   AND block_time > now() - interval '1' day
                   {% if is_incremental() %}
-                  AND block_time >= date_trunc('day', now() - interval '1' day)
+                  AND {{incremental_predicate('block_time')}}
```
aalan3, line 44: we shouldn't use the macro here as well, sorry for the confusion

andrewhong5297: oh okay got it

andrewhong5297:
> Use `incremental_predicates = [DBT_INTERNAL_DEST.block_time >= date_trunc('day', now() - interval '1' day)]` instead. I don't think we should use the macro in this specific case because this table is daily and we want to bring the macro down to hours eventually.

seems to be giving a syntax error

models/solana_utils/solana_utils_daily_balances.sql
andrewhong5297: I did not know this function existed! I'm added this in now.

aalan3 (approved):

PR #4585, opened 2023-10-12
Attempting to add Glacier CL implementation to dex.trades

discochuck:
I don't see this going particularly well right out of the gate.

For context I used quickswap's v3 as a basis as they're using Algebra CL pool tech as well. 

I'm thinking that directory structure issues might come into play as they have a polygon/ dir in their
project folder.  So again, I'm hoping for the best and expecting something more likely.


discochuck:

PR #4589, opened 2023-10-13
Fill in `dex.info` entries

hildobby:


models/dex/chains/dex_schema.yml
```diff
@@ -24,6 +24,10 @@ models:
       - &project
         name: project
         description: "Project name of the DEX"
+        tests:
+          - relationships:
+              to: ref('dex_info')
+              field: project
```
Hosuke, line 30: Great use of `relationships` test.

Hosuke (approved):
LGTM.✅
Thank you @hildobby

models/dex/dex_schema.yml
```diff
@@ -26,6 +26,10 @@ models:
       - &project
         name: project
         description: "Project name of the DEX"
+        tests:
+          - relationships:
+              to: ref('dex_info')
+              field: project
```
Hosuke, line 32: Also does this mean when a new dex is being added to spellbook, we need to add it into dex.info as well? Like newly added `wardenswap`?

hildobby: Wanted to spark the conversation with this PR. My goal is for those to hold high level information. Currently if I look into stuff like dex.trades and see a bunch of protocols I have no idea what each of those are, dex.info () and info tables as a whole in each sector would help get user some added context information. For now its just a few columns but more can always be added.

A good example is airdrop.info which has stuff like eligible addresses, claimable tokens, total token supply etc. When looking at airdrop data its hard to contextualise without this data. And now that I need to fill in info for 30+ diff airdrops, its a long process to find all of it (hence why its mostly incomplete). Of all people I expect the original contributor to be the most knowledgeable on those protocol so would be cool to ask just one extra line in dex.info. What do you think?

jeff-dude: now that this has come up across multiple PRs, for different sectors, can we centralize the convo? can you open a gh issue, with title something along the lines of:
`request: new universal test for sector-level spells`

then give an overview of all the details?

we haven't had a chance to clean out the issues page yet while finalizing the migration, but i think it's still worthwhile to get started there.

jeff-dude (approved):

PR #4591, opened 2023-10-13
Lowercase `dex.flashloans` project names

hildobby:
Making all project values lowercase for consistency (+ can now easily get proper name from `dex.info`). Looked into queries using dex.flashloans and there seem to be none that would be affected by this change as they are either broken temporary queries or don't use that column (other than mine but I'm fine with rugging myself if it's for a cleaner spellbook code across the board): https://dune.com/browse/queries?q=dex.flashloans

### DIFF
--- a/models/aave/arbitrum/aave_v3_arbitrum_flashloans.sql
+++ b/models/aave/arbitrum/aave_v3_arbitrum_flashloans.sql
@@ -40,7 +40,7 @@ WITH flashloans AS (
     )
     
 SELECT 'arbitrum'                                                       AS blockchain
-     , 'Aave'                                                           AS project
+     , 'aave'                                                           AS project
      , '3'                                                              AS version
      , CAST(date_trunc('Month', flash.block_time) as date)              AS block_month
      , flash.block_time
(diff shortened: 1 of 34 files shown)


Hosuke (approved):
LGTM.✅
Thank you @hildobby

jeff-dude (approved):
this works for me. thanks for looking into potential queries that could break, even if it's just your own 🤷‍♂️ 

there's a common theme across a few of these PRs and how the `info` spells link to major sector spells. might be worth another up front conversation about full usage and plan there, to give us all more confidence and ability to set up the repo to handle it (maybe good candidate for a gh issue when we launch those templates)

hildobby:
> this works for me. thanks for looking into potential queries that could break, even if it's just your own 🤷‍♂️
> 
> there's a common theme across a few of these PRs and how the `info` spells link to major sector spells. might be worth another up front conversation about full usage and plan there, to give us all more confidence and ability to set up the repo to handle it (maybe good candidate for a gh issue when we launch those templates)

Just checked, the only spell affected is nft.wash_trades which doesn't use the changed columns. For the .info spells discussion, I explained my goals here: https://github.com/duneanalytics/spellbook/pull/4594

PR #4556, opened 2023-10-07
More ETH Stakers + allow for more tagging methods

hildobby:
This PR does:
- reworks `staking_ethereum.deposits` to fix some bugs + allow entity tagging using `tx_from` and `pubkey` (instead of just `depositor_address`)
- creates `staking_ethereum.flows` view containing staking deposits & withdrawals data
- creates `staking_ethereum.info`, used to tie entities to their twitter account so I can redirect ppl to it more info
- tags more entities that I shared [here](https://x.com/hildobby_/status/1710326919898579023?s=20) and [here](https://x.com/hildobby_/status/1710317783894958083?s=20) (those are currently hardcoded in queries)
- rename some entities that had a rebrand (`Cake DeFi` > `Bake`, `KingHash` > `HashKing`, `PieDAO`  > `Auxo`)
- removed tag for `Kiln`'s batch staking contract
- restructures sector to keep it organised (all entities tables in the entities folder, which all culminiate in `staking_ethereum.entities` view)

PR #4593, opened 2023-10-13
Create `unique_or_null` generic test

hildobby:
Will be useful to use on the staking_ethereum_entities spell's `depositor_address`, `tx_from` & `pubkey` columns once this PR is merged: https://github.com/duneanalytics/spellbook/pull/4556

### DIFF
--- a/tests/generic/is_unique_or_null.sql
+++ b/tests/generic/is_unique_or_null.sql
@@ -1,0 +1,10 @@
+{% test is_unique_or_null(model, column_name) %}
+
+select {{ column_name }} AS unique_field
+, COUNT(*) AS n_records
+FROM {{ model }}
+WHERE {{ column_name }} IS NOT NULL
+GROUP BY {{ column_name }}
+HAVING COUNT(*) > 1
+
+{% endtest %}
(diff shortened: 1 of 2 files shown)


Hosuke (approved):
We may have one more demo usage in this PR.

jeff-dude:
> We may have one more demo usage in this PR.

let's revisit this PR after the linked PR is merged, so examples are a bit easier to do. i agree that it'd be nice to see how it's applied to a spell, the output test query from `dbt compile`, expected results vs. actual results, etc etc

adding new generic tests is relatively new concept within spellbook, so would be good to be calculated about the approach

0xRobin:
@hildobby dbt has support for filtering on existing generic tests.

So this can also be achieved with:
```
tests:
  - unique:
      config:
        where: "project is not null"
   ```
   https://docs.getdbt.com/reference/resource-properties/tests#unique

hildobby:
> @hildobby dbt has support for filtering on existing generic tests.
> 
> So this can also be achieved with:
> 
> ```
> tests:
>   - unique:
>       config:
>         where: "project is not null"
> ```
> 
> https://docs.getdbt.com/reference/resource-properties/tests#unique

Oooh I'll test that instead in https://github.com/duneanalytics/spellbook/pull/4621 and close this PR if it works as intended, didn't know you could add a where clause like that, thanks for the heads up!

hildobby:
Update, it works flawlessly, you're a wizard @0xRobin ty 🙌

closing this PR

PR #4594, opened 2023-10-13
Add relationships tests to ensure all nft spells project values are in `nft.marketplaces_info`

hildobby:
…info

### DIFF
--- a/models/_sector/nft/trades/schema.yml
+++ b/models/_sector/nft/trades/schema.yml
@@ -34,7 +34,7 @@ models:
      - &project
        name: project
        description: "Project"
-        tests:
+       tests:
           - relationships:
               to: ref('nft_marketplaces_info')
               field: project
--- a/models/_sector/nft/trades/schema.yml
+++ b/models/_sector/nft/trades/schema.yml
@@ -34,6 +34,10 @@ models:
      - &project
        name: project
        description: "Project"
+        tests:
+          - relationships:
+              to: ref('nft_marketplaces_info')
+              field: project
      - &version
        name: version
        description: "Project version"


jeff-dude (commented):
if i'm following, the intention is for any new nft marketplace added to `nft.trades` lineage needs to also add that project to [this](https://github.com/duneanalytics/spellbook/blob/main/models/nft/nft_marketplaces_info.sql) file?

what's the ultimate end goal here between these spells and the test?

@0xRobin what do you think about this test?

hildobby:
@jeff-dude @0xRobin I thought PRs would be the easiest to get the conversation going with issues still completely inactive

The goal with those would be to hold high level info for its sector. Those can be joined on other spells using the project column to get extra info. For now I just added some columns but more can always be added. They provide extra context that isn't necessarily available when looking at the other spells on their own. I definitely don't know the 50+ projects in there so it's nice to easily check.

Other columns could be added such as:
- has_token (& token_contract)
- documentation_link

If you disagree with the need for projects to be added here when added to other spells we can remove the test ofc but imo the person adding the project should be the most knowledgeable on those high level easy question and it only takes a minute. What do you guys think?

0xRobin (approved):
thanks for the write up in https://github.com/duneanalytics/spellbook/issues/4631!

I'm a fan of this as it ensure we don't have metadata-tables that become out of sync (and thus not useful) when the reference tables change. 
The test is also really fast so there's no real CI overhead added. :+1:

jeff-dude:
i would suggest for now, we leave out the test, but maintain `<sector>.info` spells to prepare. when we finalize the `nft.trades` redesign, we can add this test

i've added it in the appendix section of issue #4637

hildobby:
Closing, will reopen once the nft sector redesign is done

PR #4598, opened 2023-10-13
Balancer Pool Labels fix

viniabussafi:
This PR fixes weighted pool labels on arbitrum, ethereum, and polygon, given that the weighted pool factories for these chains allow for both `weights` and `normalizedWeights`. Previously, only the  `weights` field was being considered, which was causing some pools to be missing.

@mendesfabio @thetroyharris


jeff-dude (approved):
works for me!

PR #4601, opened 2023-10-13
Change zkSync Revenue measurement method

lgingerich:
L2 revenue for zkSync Era was previously being measured from the fee collecting addresses on L1.

Change to directly measure L1 revenue from L2 tx fees.

### DIFF
--- a/models/rollup_economics/ethereum/l2_revenue.sql
+++ b/models/rollup_economics/ethereum/l2_revenue.sql
@@ -14,25 +14,18 @@
 
 
 SELECT
-date_trunc('day', tr.block_time) as day
+date_trunc('day', t.block_time) AS day
 , 'zksync era' AS name
-, SUM(cast(tr.value as double)/1e18) as l2_rev
-, SUM(p.price * cast(tr.value as double)/1e18) as l2_rev_usd
-FROM {{ source('ethereum','traces') }} tr
+, SUM((t.gas_used * t.gas_price)/POWER(10,18)) AS l2_rev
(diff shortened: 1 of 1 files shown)


jeff-dude (commented):
@lgingerich i'm curious the motivation here. the thing that sticks out to me is moving from `ethereum` source table to `zksync` source table. is this spell intended to be only for `zksync` or all L2's? did you coordinate with the original developer @Jam516 ?

lgingerich:
This spell is for all L2s. 

The current version of the code I'm changing calculates revenue for zkSync based on ETH transferred to the fee collecting address on L1. My PR is to instead calculate revenue for zkSync directly based on tx fees paid on L2. This is how all other L2s that Dune supports are calculated. 

As has happened before, the L1 fee collecting address for zkSync can change, and then the revenue is incorrect and we have to make a PR to add the new wallet. 

I consulted with Kofi on the original implementation for zkSync, and I notified him that I submitted this PR as well.

jeff-dude:
> This spell is for all L2s.
> 
> The current version of the code I'm changing calculates revenue for zkSync based on ETH transferred to the fee collecting address on L1. My PR is to instead calculate revenue for zkSync directly based on tx fees paid on L2. This is how all other L2s that Dune supports are calculated.
> 
> As has happened before, the L1 fee collecting address for zkSync can change, and then the revenue is incorrect and we have to make a PR to add the new wallet.
> 
> I consulted with Kofi on the original implementation for zkSync, and I notified him that I submitted this PR as well.

ah, i see, i glanced too quickly and didn't notice the others union'ed below. thank you!

jeff-dude (approved):

PR #4602, opened 2023-10-13
Fix OP Stack L1 Data Fees in `rollup_economics_ethereum`

lgingerich:
OP Stack chains in `rollup_economics_ethereum.l1_data_fees` relied on a manual data upload. This PR is to eliminate reliance on a manual data upload.

@MSilb7 @Jam516 I'd greatly appreciate both of you reviewing this. Thanks!


lgingerich:
The last data in `l1_data_fees` is from 2023-10-04, while both `l1_verification_fees` and `l2_revenue` are up to date.

`l1_data_fees.sql` uses UNION ALL so I'm not sure why or if OP Stack data being out-of-date would cause other chains to be out-of-date.

MSilb7:
> OP Stack chains in `rollup_economics_ethereum.l1_data_fees` relied on a manual data upload. This PR is to eliminate reliance on a manual data upload.
> 
> @MSilb7 @Jam516 I'd greatly appreciate both of you reviewing this. Thanks!

Yeah modifying to use a Dune spell table makes sense! The upload table is in flux.

MSilb7:
^ Seems like the issue may be due to the upload namespace change: `dune.oplabspbc.dataset_op_stack_chain_metadata`

But either way, likely more future-proof to not pull from the upload.

MSilb7:
Related, I made an issue for idea discussion on and op stack metadata log: https://github.com/ethereum-optimism/op-analytics/issues/249 I imagine that this is something that could be pulled from at some point in the future for identifying more OP Stack L2 contracts on L1. Feel free to add ideas/thoughts!

lgingerich:
@MSilb7 Thanks for giving the stamp of approval, appreciate it!

And good idea on starting up that discussion. This is something that I'll definitely deal with re: ZK Stack. I need to think about it more but I'll try to add in my thoughts.

jeff-dude (commented):
@lgingerich thanks for looking into an improvement here!

we're currently thinking through how to best use the data upload feature in spellbook, when the data is truly only available via manual maintenance. we've only allowed a select few spells into `main` that use these datasets until we finalize that to avoid downstream orchestration issues.

while the spell joined in this change is a manual dataset via spellbook, ideally we find a way to move away from that in the future too, as we've seen those files grow v large in certain use cases. this goes back to why we're looking to solve the point above.

with that said, let's move forward here for now.

please note, in order for CI test to properly run, you need to remove the `prod_exclude` tag from the config block above. this spell has actually been inactive due to the upload source name changing :)

lgingerich:
@jeff-dude thanks for the callout, I've removed the `prod_exclude` tag. However the CI test is now failing and I don't quite understand what the error is. Any ideas?

Jam516:
> OP Stack chains in `rollup_economics_ethereum.l1_data_fees` relied on a manual data upload. This PR is to eliminate reliance on a manual data upload.
> 
> @MSilb7 @Jam516 I'd greatly appreciate both of you reviewing this. Thanks!

I agree with Michael, looks like a good fix 👍

jeff-dude:
> @jeff-dude thanks for the callout, I've removed the `prod_exclude` tag. However the CI test is now failing and I don't quite understand what the error is. Any ideas?

looks like the unique keys [assigned in config block](https://github.com/duneanalytics/spellbook/pull/4602/files#diff-b1d69d1cd99e3619aa41bd77863e9cfd5eb6a3eacaf07aa5f383215ce1353648R8) are not actually unique, as the test applied [here](https://github.com/duneanalytics/spellbook/blob/main/models/rollup_economics/ethereum/rollup_economics_ethereum_schema.yml#L13-L17) is returning duplicates on those columns.

the test essentially writes a quick group by / having clause to look for dupes. you can run `dbt compile` and navigate through target directory to find exact test query and run on dune, if you want.

either the latest changes introduced duplicates or somehow duplicates got produced over time, since this passed the original PR test on creation

lgingerich:
I started diagnosing and redoing this but haven't fully worked through it yet. Still trying to finish it up this week

jeff-dude (approved):

PR #4606, opened 2023-10-15
Celo nft sector additions

tomfutago:
Adding Celo to `nft.transfers` and `nft.wash_trades`


jeff-dude (commented):
nice & clean!

can you add the new spells to their respective schema yml files?

jeff-dude (approved):

PR #4608, opened 2023-10-15
`nft_mints` macro

tomfutago:


tomfutago:
Goal: create `nft.mints` macro initially based on existing [nft_ethereum_native_mints](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/nft/mints/native/nft_ethereum_native_mints.sql) spell and then refactored to avoid long list of left joins + aggs (with another long list of group by columns) all in one final query - and then window function on top of it..

Notes:
* there are a number of columns with hardcoded defaults - left them as they were, as they seem to replicate the same column list as `nft.trades` - not sure tho if there’s a plan to create macro/spell to combine the two?
* similarly, `seller` column always defaults to burn address atm..
* `amount_raw`, `amount_original` and `amount_usd` aggregated in 2 separate cte's per native or erc20 cost for the mint, which is meant to improve performance - still would be great if it was possible to replace `traces` with more performant table/spell - ideas welcome 🙏
* extended uniqueness by adding `currency_contract` and the reason for it are cases with more than 1 erc20 transfer on nft mint. maybe not the best example, but that’s what happens when Uniswap V3 Positions NFT-V1 are minted and liquidity added to the pool
* temporarily removed `nft_optimism_native_mints` spell from global `nft_mints` as it uses modified version of original `nft_ethereum_native_mints` spell, so needs to be looked at separately. and also, back to the note on the `amounts..` : it sums them up, which for different erc20 tokens is incorrect imo. And then that row_number on the final query only takes highest `amount_usd` per mint 🤔

@jeff-dude linking this back to #4637

jeff-dude:
asking @0xRobin to take a quick look, as i know he's deep in the `nft` redesign efforts and this may be on the roadmap

0xRobin (commented):
I'll take a closer look at this this week!
Can you explain a bit more why you've chosen to exclude optimism and magiceden from the overall `nft.mints` model?

tomfutago:
Thanks @0xRobin for looking into this! I should have properly clarify: this is still WIP and my initial attempt at adding this macro - just wanted to get some initial feedback before I go too deep while potentially going in completely wrong direction. Both exclusions are only temporary - my plan is to revisit them after getting the feedback 🙏

Hope that makes sense 🤞 can expand if/when needed.

jeff-dude:
closing due to #4819

Issue #4523, opened 2023-10-04
EVM Balances main issue


aalan3:
This is the main issue for brining balances back alive.

EDIT: please see this [comment](https://github.com/duneanalytics/spellbook/issues/4523#issuecomment-1827582865) on the updated approach on this.

## Problem
We currently have several different implementations of transfers and a broken balances implementation. 

For example we have: 
* [Fungible transfers](https://github.com/duneanalytics/spellbook/blob/main/models/fungible/fungible_transfers.sql) which include both native and erc20 transfers
* [ERC20 transfers](https://github.com/duneanalytics/spellbook/tree/main/models/transfers) which calculates wallet balance changes. It also has different implementations including aggregations
* [Balances](https://github.com/duneanalytics/spellbook/tree/main/models/balances/ethereum) which builds on transfers but is currently broken (except for Polygon).
   * Balances in the current implementation is aggregated to day or hour

Additionally we have a bunch of PRs trying to fill in the gaps in different places.

## Goal
We want to have balances on Dune and have a one unified implementation that includes both erc20 tokens and native tokens (ETH, BNB etc). We currently have a bunch of different implementations or suggested implementations when it comes to either balances or transfers. The aim is to have one implementation.

The goals are:
* Implement a model for erc20 transfers and native transfers
* Implement balances that include wrapped tokens, native tokens and erc20 tokens
    * No longer aggregate on time, instead on blocks or transactions
* Clean up naming and use bring everything under proper namespaces

We consider NFT balances out of scope for this issue but it could be future work for both transfers (`nft_transfers` also already exists) and balances.

### Fungible transfers model
We can base our transfers table off `fungible` I think it does pretty much what we want when it comes to native and ERC20 transfers but the name and structure should change. 

We should have one table for both native and erc20 transfers

### Structure
Following the discussion in https://github.com/duneanalytics/spellbook/issues/4580#issuecomment-1767204966. We want the structure to be the following on new models and sector level spells
`base_model` (materialized) -> `enrichment_model` (view)

So for balances we would use the transfers base model to create the balances base model since balances will also be enriched in a view.

There might be cases where doing the enrichment in the view might not perform well, in those cases we can still materialize but we should try without it at first.

### Naming
This is a bit tricky. We currently have `transfers_erc20` and `fungible_transfers`, we should consider that new users might not know what `fungible` means or even `erc20`.  IMO we should use either `tokens`, `transfers` and/or `balances` namespaces for this. Naming suggestions are welcome
[...]

PR #4610, opened 2023-10-16
Add a More Thorough MATIC Transfers for Polygon

MSilb7:
The existing `transfers_polygon_matic` is fairly unusable since it doesn't expose the "to" and "from" address of each transfer. Initiating a version that is more usable and is equivalent to `transfers_ethereum_eth` (which mirrors erc20 transfer tables with a few extra valuable columns)

I didn't want to overwrite the existing table in case that screwed up any existing queries.


jeff-dude:

#4521 
#4523 
#4633 
#4698

PR #4614, opened 2023-10-16
prod hotfix: fix incorrect column dependency on eth staking

jeff-dude:
fyi @hildobby 
another reminder that we need to check downstream dependencies when changing column names. we will have to consider an automated test addition in the CI test gh action.


hildobby:
> fyi @hildobby another reminder that we need to check downstream dependencies when changing column names. we will have to consider an automated test addition in the CI test gh action.

Ah thank you for the hotfix! And yeah would love a CI test for this, it's not something I would remember to check everytime

PR #4627, opened 2023-10-18
Oneinch: optimize top views

grkhr:


jeff-dude (approved):
niiice -- a new use case example in macros. this could be helpful in the future, thanks!

fyi @aalan3 @couralex6 @0xRobin @Hosuke @antonio-mendes

PR #4629, opened 2023-10-18
Integration tests

couralex6:
Removed two steps from CI which are no longer needed:
* `Check no common models between DuneSQL and Spark`
* `Check for missing legacy models`

And added translated integration test models


jeff-dude:
can close this now? or have a plan for them?

PR #4633, opened 2023-10-19
Token transfers for ethereum

aalan3:

This PR implements a transfers_base macro which materializes ERC20 transfers and native transfers into a single table and a transfers_enrichment macro. The design is based on the `fungible` transfers. 

They are meant to be used together where the base is for materialization and the enrichment is a view on top of it. I'm not yet sure the enrichment step will work as a view performance-wise. I think `prices_usd_forward_fill` is a bit inefficient but we'll see.


macros/models/_sector/tokens/transfers_base.sql
```diff
@@ -0,0 +1,65 @@
...
+{%- set token_standard_20 = 'bep20' if blockchain == 'bnb' else 'erc20' -%}
+{# denormalized tables are not yet in use #}
+{%- set denormalized = True if blockchain in ['base'] else False -%}
+
+WITH transfers AS (
+    SELECT block_time
+    , block_number
+    , tx_hash
+    , value AS amount_raw
+    , CAST(NULL AS varbinary) AS contract_address
```
0xRobin, line 11: can we use `{{VAR("ETH_ERC20_ADDRESS"}}` here instead of `NULL`. :thinking:

macros/models/_sector/tokens/transfers_base.sql
```diff
@@ -0,0 +1,65 @@
...
+    , t.evt_block_number AS block_number
+    , t.evt_tx_hash AS tx_hash
+    , t.value AS amount_raw
+    , t.contract_address
+    , '{{token_standard_20}}' AS token_standard
+    , t."from"
+    , t.to
+    , t.evt_index
+    , NULL AS trace_address
+    FROM {{ erc20_transfers }} t
```
0xRobin, line 37: if we filter native transfers with `value > 0`, should we do the same for erc20?

models/tokens/ethereum/tokens_ethereum_transfers.sql
```diff
@@ -0,0 +1,17 @@
...
+                                    "fungible",
+                                    \'["hildobby", "aalan3"]\') }}'
+        )
+}}
+
+
+{{transfers_enrich(
+    blockchain='ethereum',
+    transfers_base = ref('tokens_ethereum_transfers_base'),
+    tokens_erc20 = ref('tokens_erc20'),
```
0xRobin, line 16: this should be static across chains, so we can remove it from the variables and hardcode the ref in the macro?

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
```diff
@@ -0,0 +1,17 @@
+{{ config(
+        alias = alias('transfers_base'),
+        tags=['dunesql'],
+        partition_by = ['block_date'],
```
0xRobin, line 4: using `block_date` instead of `block_month` here makes sense because we will have a lot of rows. :+1:

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
```diff
@@ -0,0 +1,17 @@
...
+        unique_key = ['block_number', 'tx_index', 'evt_index', 'trace_address'],
+        )
+}}
+
+{{transfers_base(
+    blockchain='ethereum',
+    traces = source('ethereum','traces'),
+    transactions = source('ethereum','transactions'),
+    erc20_transfers = source('erc20_ethereum','evt_Transfer'),
+)}}
```
0xRobin, line 17: I think we'd want to include the WETH deposit and withdraw events here. 
It's a bit of an ethereum specific edge case so I don't think we should add it to the macro. We could union it in here?

macros/models/_sector/tokens/transfers_enrich.sql
```diff
@@ -0,0 +1,36 @@
...
+    ELSE cast(t.amount_raw as double)
+ END AS amount
+, prices.price AS usd_price
+, CASE WHEN t.token_standard = 'native' THEN (t.amount_raw / power(10, tokens_erc20.decimals)) * prices.price
+    WHEN t.token_standard = 'erc20' OR t.token_standard = 'bep20' THEN (t.amount_raw / power(10, 18)) * prices.price
+    ELSE NULL
+ END AS usd_amount
+FROM {{transfers_base}} t
+LEFT JOIN {{tokens_erc20}} tokens_erc20 on tokens_erc20.contract_address = t.contract_address
+LEFT JOIN {{ ref('prices_usd_forward_fill') }} prices ON prices.blockchain = '{{blockchain}}'
```
0xRobin, line 30: let's use `prices_usd` for now for performance reasons.
I can take a look at materializing the forward fill in an incremental model so we can query it more efficiently.

0xRobin (changes requested):
great! Some initial comments:

macros/models/_sector/tokens/transfers_base.sql
aalan3: I checked and there are about 25M rows that have `value = 0` (can't be negative due to the type). No idea why this happens but here's a few example events https://etherscan.io/tx/0x347f678c54e205ed54d0b7c7a77f19e8447e239ce4908123eec6f7059c6b2f4e#eventlog

macros/models/_sector/tokens/transfers_enrich.sql
aalan3: I'll try it

macros/models/_sector/tokens/transfers_base.sql
aalan3: this is a bit confusing to me tbh, sometimes I see `0xeeee...` used as this as well. and would `0x0000...` be consistent across chains?

macros/models/_sector/tokens/transfers_enrich.sql
```diff
@@ -1,3 +1,9 @@
+{%- macro case_when_token_standard(native_column, erc20_column, else_column) -%}
```
aalan3, line 1: not sure if this is more readable or not but trying it

macros/models/_sector/tokens/transfers_base.sql
0xRobin: Yes it's due to devs not checking their logic before calling a `transferFrom()`

macros/models/_sector/tokens/transfers_base.sql
0xRobin: yes that's why we need to start using the variable, we can probably rename it to `NATIVE_ERC20_ADDRESS` and use it across chains. 
We can change the variable to `0xeeee..` or `0xee` (or even `null` I guess) if we want.

0xRobin: relevant: this [draft EIP-7528](https://eips.ethereum.org/EIPS/eip-7528)

macros/models/_sector/tokens/transfers_enrich.sql
0xRobin: I've recently learned we can also do
```
CASE token_standard
  WHEN 'native' THEN column1
  WHEN 'erc20' THEN column2
  ELSE column3
END
```
which is a bit more readable then all the inline comparisons

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
aalan3: I've included it the macro now. It exists for polygon as well at least. It's optional in the macro though.

macros/models/_sector/tokens/transfers_base.sql
aalan3: Updated this to use a native contract address. I will update this to use a variable in a separate PR since we need to change the name there and edit a few models after.

macros/models/_sector/tokens/transfers_base.sql
```diff
@@ -0,0 +1,108 @@
...
+    , to
+    , value AS amount_raw
+    FROM {{ traces }}
+    WHERE success
+    AND (call_type NOT IN ('delegatecall', 'callcode', 'staticcall') OR call_type IS null)
+    AND value > UINT256 '0'
+    {% if is_incremental() %}
+    AND {{incremental_predicate('block_time')}}
+    {% endif %}
+
```
grkhr, line 28: Are you planning to add gas fees further in spells? It's also sort of transfer which affects balances

aalan3: Do you have any more insight on this? gas fees makes balances a bit tricky but aren't a lot of the fees in the traces?

grkhr: you can do just:

`select "from" as spender, gas_price * gas_used as amount from {blockchain}.transactions`

but it's better to doublecheck if this is applicable for L2 as well

macros/models/_sector/tokens/transfers_base.sql
```diff
@@ -0,0 +1,108 @@
...
+    UNION ALL
+
+    SELECT t.evt_block_time AS block_time
+    , t.evt_block_number AS block_number
+    , t.evt_tx_hash AS tx_hash
+    , t.evt_index
+    , NULL AS trace_address
```
jeff-dude, line 35: similar to `evt_index` above, maybe cast this null?

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
```diff
@@ -0,0 +1,22 @@
+{{ config(
+        alias = 'transfers_base',
+        tags=['dunesql'],
```
jeff-dude, line 3: we can remove this tag now

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
```diff
@@ -0,0 +1,22 @@
+{{ config(
+        alias = 'transfers_base',
```
jeff-dude, line 2: please also add the schema property here

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
```diff
@@ -0,0 +1,22 @@
...
+        tags=['dunesql'],
+        partition_by = ['token_standard', 'block_date'],
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'merge',
+        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')],
+        unique_key = ['block_number', 'tx_index', 'evt_index', 'trace_address'],
```
jeff-dude, line 9: are we missing `tx_hash` as a unique key? or does introducing `tx_index` make it unique without the hash?

models/tokens/ethereum/tokens_ethereum_transfers.sql
```diff
@@ -0,0 +1,15 @@
+{{ config(
+        alias = 'transfers',
```
jeff-dude, line 2: please also add the schema property here

models/tokens/ethereum/tokens_ethereum_transfers.sql
```diff
@@ -0,0 +1,15 @@
...
+        materialized = 'view',
+        post_hook='{{ expose_spells(\'["ethereum"]\',
```
jeff-dude, line 4: should we remove this post hook for now? until everything is final? then maybe only expose the cross-chain transfers (if that is the final ultimate plan)?

models/tokens/ethereum/tokens_ethereum_transfers.sql
jeff-dude, line 1: similar to the macros directory structure above, maybe we should move this to:
`models/_sector/tokens/transfers/<blockchain>/`
this would be in line with the `nft.trades` new structure, as well as `dex.trades` in PR phase

jeff-dude: this would apply to both spells, of course

jeff-dude: in addition, can you add a schema yml file too? that will be the baseline for dbt docs (when / if we revamp them), then also you can apply some tests. one test in particular, a unique combo of columns, on the incremental one below. that would also help us determine quickly if `tx_hash` is needed

macros/models/_sector/tokens/transfers_enrich.sql
```diff
@@ -0,0 +1,39 @@
...
+, t.tx_index
+, t.evt_index
+, t.trace_address
+, t.token_standard
+, t.tx_from
+, t.tx_to
+, t."from"
+, t.to
+, t.contract_address
+, {{case_when_token_standard('\'SYMBOL\'', 'tokens_erc20.symbol', 'NULL')}} AS symbol
```
jeff-dude, line 24: am i reading this right, that the `'SYMBOL'` will be used directly? if so, what is the intention?

jeff-dude (changes requested):

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
aalan3: `tx_index` is equivalent to `tx_hash` here (combined with `block_number`) so it's not needed, this should be more efficient.

models/tokens/ethereum/tokens_ethereum_transfers.sql
aalan3: Not sure if we should only do cross chain here because of the size but will still remove it until we are fully ready.

macros/models/_sector/tokens/transfers_enrich.sql
aalan3: This was just a placeholder, I'll update.

models/tokens/ethereum/tokens_ethereum_transfers.sql
aalan3: Let's discuss structure in https://github.com/duneanalytics/spellbook/pull/4698 instead. 

I'll add the schema and a test.

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
```diff
@@ -0,0 +1,22 @@
...
+}}
+
+-- TODO: use variable for native_contract_address
+{{transfers_base(
+    blockchain='ethereum',
+    traces = source('ethereum','traces'),
+    transactions = source('ethereum','transactions'),
+    erc20_transfers = source('erc20_ethereum','evt_Transfer'),
+    native_contract_address = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee',
+    wrapped_token_deposit = source('zeroex_ethereum', 'weth9_evt_deposit'),
```
grkhr, line 20: also — weth isn't connected to zeroex. it'd be better to resubmit weth contract to Dune and change the source.

0xRobin: fun fact: The first WETH contract was actually first developed and deployed by some people from 0x labs.
Agree though it should be renamed in the future, but it has built up a lot of downstream dependencies so name changes are not trivial.

grkhr:
@aalan3 genesis balances should be included as well!

https://github.com/duneanalytics/spellbook/blob/main/models/balances/ethereum/genesis/genesis_balances.sql

aalan3:
> @aalan3 genesis balances should be included as well!
> 
> https://github.com/duneanalytics/spellbook/blob/main/models/balances/ethereum/genesis/genesis_balances.sql

Good point. We have this issue for [optimism as well ](https://github.com/duneanalytics/spellbook/issues/4523#issuecomment-1796007802). The question is, should we consider these transfers or just include them in the balances later? We would have to put dummy values for block_number, trace call etc.

grkhr:
> > @aalan3 genesis balances should be included as well!
> > https://github.com/duneanalytics/spellbook/blob/main/models/balances/ethereum/genesis/genesis_balances.sql
> 
> Good point. We have this issue for [optimism as well ](https://github.com/duneanalytics/spellbook/issues/4523#issuecomment-1796007802). The question is, should we consider these transfers or just include them in the balances later? We would have to put dummy values for block_number, trace call etc.

I'd say both options are ok. Let's see what others think?

macros/models/_sector/tokens/transfers_base.sql
aalan3: Thank you. I will look into this in a follow up PR.

models/tokens/ethereum/tokens_ethereum_transfers_base.sql
jeff-dude: looks like maybe we need an additional column in unique keys, based on test failure?

aalan3: Yeah I am adding it at the moment. I think the nulls are causing issues with the merge into.

aalan3: No duplicates now. I'm not sure native transfers are 100% correct though (aside from gas) so will need to test this further later.

macros/models/_sector/tokens/transfers_base.sql
```diff
@@ -0,0 +1,110 @@
...
+    )
+
+SELECT '{{blockchain}}' as blockchain
+, cast(date_trunc('day', t.block_time) as date) as block_date
+, t.block_time
+, t.block_number
+, t.tx_hash
+-- method_id => first 4bytes of data
+-- We have to create this because evt_index and trace_address can be null
+, {{dbt_utils.generate_surrogate_key(['t.block_number', 'tx.index', 't.evt_index', "array_join(trace_address, ',')"])}} as unique_key
```
jeff-dude, line 92: do we prefer this approach of adding surrogate key, rather than `coalesce()` each unique key to a default value?

this could be a decision that has an impact across spellbook. of course, we don't have to standardize it all right away, but will be nice to be consistent moving forward at minimum (and on larger sector spells)

aalan3: TBH I don't like having `unique_key` in most cases but here I think it works because not sure what we should coalesce to. I think it can be confusing for users to have a value in `evt_index/trace_address` that doesn't lead anywhere. 

I think if we do have a unique key column we should always try to use `generate_surrogate_key` because 1) it hashes the output so that it takes less space and 2) it works the same everywhere.

jeff-dude: i'm okay with the rationale behind the first part. i will look to consider the approach in new `dex.trades` design, if applicable.

agree fully with the second part.

jeff-dude (approved):
i'm comfortable with this current state. thank you for taking in all the input from the community and helping get this moving 🚀 

for those following along, keep in mind that this is in somewhat of a "beta" state and we welcome all feedback once live on how to improve. in the meantime, other blockchain versions will look to follow suit of ethereum here.

macros/models/_sector/tokens/transfers_base.sql
0xRobin: I still believe we should adapt the incremental logic to use `IS NOT DISTINCT FROM`  as equality operation so that the merge works as we would want with respect to `null` values.

PR #4640, opened 2023-10-20
Token balances ethereum

aalan3:
This should be considered a beta version of token balances so this is not yet exposed. The tables are split by `native` and `erc20` tokens because doing everything together uses too much memory in Trino, these are considered intermediary tables and are unioned together in the main table `balances_base`. 

Some notes:
- Gas fees and block rewards are not in transfers yet so they will not be counted here
- We have to cast to double before the `sum()` because there's no way of catching overflows that happen inside of a aggregation at the moment
- For other bigger chains we likely have to split up the tables to create even more intermediary tables


jeff-dude:
are you good with my commits?

it timed out, as expected. we can still look to merge tomorrow, if you are good with the current state.

aalan3:
Yeah thank you! I was going to do transfers in a separate PR but might as well do it here. Tests are also a good idea.

jeff-dude (approved):

PR #4641, opened 2023-10-20
make smaller

andrewhong5297:
making the daily balances calc in solana smaller. let's just bump up the cluster to force this through, I don't think it makes sense to have a double scan of account_activity by splitting it into multiple models.


andrewhong5297:
for @aalan3

aalan3 (approved):

jeff-dude (approved):

PR #4643, opened 2023-10-20
Create `dex.crossdomain_mev` [WIP]

hildobby:


jeff-dude:
similar question here as other older PRs, what's the status?

if we want to proceed, we'd need the timeout increase to be removed

hildobby:
> similar question here as other older PRs, what's the status?
> 
> if we want to proceed, we'd need the timeout increase to be removed

will close for now but might reopen later

PR #4649, opened 2023-10-20
update vrf fulfilled tx query

AnonJon:
Update table joins on vrf fulfilled transactions tables for all available networks.


jeff-dude (approved):
looks good!

jeff-dude:
@AnonJon have you & team been made aware of the macro approach in spellbook? i could see immense value in using that for your chainlink spells, since they repeat across many blockchains. you can centralize the logic in a macro and then just have blockchain-specific models calling the macro.

here is example from 1inch team:
macro [here](https://github.com/duneanalytics/spellbook/blob/main/macros/models/_project/oneinch/oneinch_calls_transfers_macro.sql)
spell example [here](https://github.com/duneanalytics/spellbook/blob/main/models/oneinch/ethereum/oneinch_ethereum_calls_transfers.sql)

there are plenty of other examples as well & cool stuff being built. let me know what ya think!

PR #4653, opened 2023-10-22
Celo curvefi

tomfutago:


Hosuke (approved):
LGTM.✅
Thank you @tomfutago 

By the way you can add curvefi_celo_trades to the roll-up spell:
https://github.com/duneanalytics/spellbook/blob/main/models/curvefi/curvefi_trades.sql

[Sample curvefi_celo_trades](https://dune.com/queries/3139769)
[Sample curvefi_celo_pools](https://dune.com/queries/3139800)

tomfutago:
That was the plan, thanks for the reminder @Hosuke ! it's added now.

models/curvefi/celo/curvefi_celo_pools.sql
```diff
@@ -0,0 +1,22 @@
+{{
+    config(
+        tags = ['dunesql'],
+        alias = alias('pools'),
```
jeff-dude, line 4: can you add schema property here too?

i believe right now, without it assigned, it'll default to `curvefi.pools` since it's under the `curvefi` directory, but no entry exists for `curvefi/celo/` in the dbt_project file

we also want to keep schema in the models themsevles moving forward, for more transparency (we'll clean up project-wide when time permits)

models/curvefi/celo/curvefi_celo_pools.sql
```diff
@@ -0,0 +1,22 @@
...
+        materialized = 'table',
+        unique_key = ['version', 'tokenid', 'token', 'pool'],
+        post_hook = '{{ expose_spells(\'["celo"]\',
```
jeff-dude, line 7: another thing we're cleaning up a bit:
does this spell need to be exposed on data explorer?

rule of thumb on this:
is the intention to query directly on the app often (then expose), or is it just a stepping stone to be used in downstream spells (then don't expose)?

models/curvefi/celo/curvefi_celo_trades.sql
```diff
@@ -0,0 +1,105 @@
+{{
+    config(
+        tags = ['dunesql'],
+        alias = alias('trades'),
```
jeff-dude, line 4: schema property here too plz

models/curvefi/celo/curvefi_celo_trades.sql
```diff
@@ -0,0 +1,105 @@
...
+        partition_by = ['block_month'],
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'merge',
+        unique_key = ['block_date', 'blockchain', 'project', 'version', 'tx_hash', 'evt_index'],
+        post_hook = '{{ expose_spells(\'["celo"]\',
```
jeff-dude, line 10: same question here on exposing the spell (we're leaning towards not exposing many of these upstream spells, but rather just the end result of `dex.trades`)

models/curvefi/curvefi_trades.sql
```diff
@@ -2,10 +2,10 @@
         tags = ['dunesql'],
         partition_by = ['block_month'],
         alias = alias('trades'),
-        post_hook='{{ expose_spells(\'["ethereum","avalanche_c","optimism","fantom"]\',
+        post_hook='{{ expose_spells(\'["ethereum","avalanche_c","optimism","fantom","celo"]\',
```
jeff-dude, line 5: you can leave this for now, as it's existing spell, we can consider on our future cleanup

jeff-dude (changes requested):
a few minor config comments below 🙏

models/curvefi/celo/curvefi_celo_pools.sql
tomfutago: makes sense, will add `curvefi_celo` schema

models/curvefi/celo/curvefi_celo_pools.sql
tomfutago: yep, this spell actually doesn't have to be exposed - will remove

models/curvefi/celo/curvefi_celo_trades.sql
tomfutago: agreed, will remove

PR #4657, opened 2023-10-22
Add more zksync erc20s

hildobby:
Source: https://explorer.zksync.io/tokenlist


hildobby:
@lgingerich are USDC.e & USDT.e different from USDC & USDT? If it's just the bridged tokens, it might be easier to keep the names consistent across chains, what do you think?

Screenshot 2023-10-23 at 00 30 10

lgingerich:
These are just bridged versions of USDC and USDT, but I'd like to keep the `.e` notation to distinguish from the native versions which we'll eventually have

hildobby:
> These are just bridged versions of USDC and USDT, but I'd like to keep the `.e` notation to distinguish from the native versions which we'll eventually have

Ah alright sg!

Hosuke (approved):

jeff-dude (approved):

PR #4659, opened 2023-10-23
prices.usd -> change forward fill lookback and add a monitor

0xRobin:
Given that `prices.usd` has gotten a lot more stable (less latency)
we can decrease the lookback (and thus compute cost) of the forward fill model. 

This PR also adds a lightweight monitor table so we can go back and inspect quantile metrics on the latency of the different price tokens. 
more info on qdigest [here](https://trino.io/docs/current/functions/qdigest.html)

### DIFF
--- a/models/prices/prices_usd_monitor.sql
+++ b/models/prices/prices_usd_monitor.sql
@@ -1,0 +1,20 @@
+{{ config(
+        tags = ['dunesql'],
+        schema='prices',
+        alias = alias('usd_monitor'),
+        materialized = 'incremental',
+        incremental_strategy = 'append'
+        )
+}}
+
+-- monitor to gather quantile information about prices latency
+select
+    now() as recorded_at
+    ,cast(qdigest(latency)as varbinary) latency_digest_varbinary
+from (
+    select
+    blockchain, contract_address, date_diff('second',now(),max(minute))) as latency
+    from {{source('prices.usd')}}
+    where minute >= now() - interval '7' day    -- we'll consider anything that's more then 7 days late as stale tokens
+    group by 1,2
+)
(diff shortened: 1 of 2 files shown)


models/prices/prices_usd_monitor.sql
```diff
@@ -0,0 +1,21 @@
...
+        )
+}}
+
+-- monitor to gather quantile information about prices latency
+select
+    now() as recorded_at
+    ,cast(qdigest_agg(latency_seconds) as varbinary) as latency_seconds_digest
+from (
+    select
+    blockchain, contract_address, date_diff('second',max(minute),now()) as latency_seconds
```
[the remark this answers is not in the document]
0xRobin: Yes, this is just to keep it as lightweight as possible.
So we will be storing the latency distribution on each incremental update, but not which token had which latency. 
I'm most of all interested in seeing the avg latency (and other percentiles) over time, as currently we can only inspect the current latency and don't really know how stable that is.

models/prices/prices_usd_monitor.sql
```diff
@@ -0,0 +1,21 @@
+{{ config(
+        tags = ['dunesql'],
```
aalan3, line 2: Can we tag this as monitoring or something?

0xRobin: sure, what would we use if for?

aalan3: potentially running this separate from the hourly incremental job.

jeff-dude: good thought. doesn't hurt to have in case we split out in prod jobs.

jeff-dude (approved):

Issue #4637, opened 2023-10-19
[proposal] sector-level spells redesign


jeff-dude:
### overview
**_note_**: for the purposes of simplicity, most of this will focus on `dex.trades` as it's actively starting the redesign, but all sectors apply!

`dex.trades` has more or less followed the same table design since the early days of abstractions on dune. when spellbook was built, a bit more structure was put in place around how `dex.trades` is built & how it is orchestrated in production to ensure fresh data for all users. over time, as more and more projects have been added & the tech stack which we use to build the spells has improved, we've found various opportunities for improvement. this redesign effort will look to implement all of these opportunities for an improved spell experience for all.

### what's changing (and why)?
- base project-level spells will only contain data from raw & decoded source tables
  - examples: `uniswap_v1_ethereum.base_trades`, `uniswap_v3_optimism.base_trades`
  - to keep dbt lineages & both CI/prod orchestration clean and easy to support
  - less full refreshes of historical data in prod, as data is much less likely to change and/or be modified when reading only raw/decoded data
  - these spells will still be materialized incrementally, same as before
  - the intention isn't for end users to query these on the dune app, but rather as a building block to reach ultimate destination in `dex.trades`
  - if project-level spells want to be exposed on the dune app to be queried, the recommendation will be to create a spell view that sits on top of `dex.trades` and filters on blockchain/project/version/whatever is needed
- chain union spells will call a dbt macro which standardizes a `union all` of input spells, i.e. spells from the previous step on the same blockchain
    - `dex_<blockchain>.base_trades` — this will allow for another building block towards `dex.trades`
    - ~~todo: check if these will be exposed or not on the app, likely not~~ will not expose for now
    - ~~these chain union spells will be materialized incrementally (or try a view, if not intended to be queried on the app?)~~ the chain union spells will be materialized, as they join to base tables to obtain more columns and the read could be slow on a view
- sector union spell will call a dbt macro which enriches the raw/decoded data with all the necessary metadata
  - `dex.trades`
  - metadata: `tokens.erc20`, `prices.usd`, etc.
  - having the metadata limited to the final sector-level spell allows for easier contributions of new projects, simplified prod orchestration & one dbt macro to centralize the logic

### who is leading this effort?
dune team & resident wizards, along with any helpful ideas/contributions/feedback from the community 🙏 

### where can the code be found?
there'll be various PRs over time, but two examples:
[...]

PR #4668, opened 2023-10-24
Revamp NFT to new sector spell design structure

0xRobin:
Adapt the NFT sector top level models after the discussion in https://github.com/duneanalytics/spellbook/issues/4637

The structure:

- platform models (`blur_ethereum_base_trades.sql`) - base schema, materialized
- chain union (`nft_ethereum_base_trades.sql`) - base schema, view
- full union (`nft_base_trades.sql`) - base schema, materialized
- enrichment (`nft_trades_beta.sql`) - full schema, view

specific to NFT, a final layer on top to keep compatibility to the currently exposed column names: 
- backwards port (`nft_trades.sql`) - full schema (old), view, exposed on Dune


0xRobin:
Hmm. can't figure our where this error is coming from..
image

0xRobin:
Ok I think this is ready.

macros/models/_sector/nft/add_nft_tx_data.sql
jeff-dude, line 1: does it make sense to adapt the existing one for this use case?
https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/add_tx_columns.sql

the more we add variations of it, there becomes a certain point where it's better to just write the code directly in models instead? i guess sector-specific macro for each keeps it pretty clean, but just thinking it could scale quicker than we realize

.github/workflows/dbt_slim_ci.yml
```diff
@@ -49,7 +49,7 @@ jobs:
           [[ -z "$test" ]] && { echo "Success: All models have a custom schema"; exit 0; } || { echo "Found models without custom schema:"; echo "$test"; exit 1; }
 
       - name: dbt seed
-        run: "dbt seed $PROFILE --select state:modified --exclude tag:prod_exclude tag:remove --state ."
+        run: "dbt seed $PROFILE --select @state:modified --exclude tag:prod_exclude tag:remove --state ."
```
jeff-dude, line 52: remind me the impact of `@` being added?

models/_sector/nft/trades/compatibility_views/nft_events.sql
```diff
@@ -0,0 +1,11 @@
+{{ config(
+        schema = 'nft',
+        alias = 'events',
+        tags = ['dunesql'],
```
jeff-dude, line 4: can remove dunesql tag

jeff-dude: will only comment here, but applies universally

models/_sector/nft/trades/ethereum/nft_ethereum_base_trades.sql
```diff
@@ -0,0 +1,73 @@
...
+    FROM {{ nft_model }}
+    {% if is_incremental() %}
+    WHERE {{incremental_predicate('block_time')}}
+    {% endif %}
+    {% if not loop.last %}
+    UNION ALL
+    {% endif %}
+    {% endfor %}
+    )
+where duplicates_rank = 1
```
jeff-dude, line 69: with this being a view, should we remove incremental logic / duplicates check?

macros/models/_sector/nft/add_nft_tx_data.sql
0xRobin: We can't use the existing one here due to the specific logic that's needed to truncate the data column.

.github/workflows/dbt_slim_ci.yml
0xRobin: it will run all the upstream seeds of models impacted by the modified changes.

0xRobin: Feel free to remove this change if we don't want to push it on `main` ( I think we should, dealing with the fabricated seed changes to trigger them in the CI has become more of a hassle than the extra CI time (<30sec) introduced by this.

models/_sector/nft/trades/ethereum/nft_ethereum_base_trades.sql
0xRobin: hmm, good question. We should not remove the duplicate check, but good idea to clean up the incremental blocks.

.github/workflows/dbt_slim_ci.yml
jeff-dude: let's give it a chance & monitor it on main branch

macros/models/_sector/nft/enrich_nft_trades.sql
```diff
@@ -0,0 +1,108 @@
...
+    case when base.price_raw > uint256 '0' then cast(100*base.royalty_fee_amount_raw/base.price_raw as double) else double '0' end as royalty_fee_percentage,
+    coalesce(agg1.contract_address,agg2.contract_address) as aggregator_address,
+    {% if aggregator_markers != null %}
+    CASE WHEN coalesce(agg_mark.aggregator_name, agg1.name, agg2.name)='Gem' AND base.block_number >= 16971894 THEN 'OpenSea Pro' -- 16971894 is the first block of 2023-04-04 which is when Gem rebranded to OpenSea Pro
+        ELSE coalesce(agg_mark.aggregator_name, agg1.name, agg2.name)
+        END as aggregator_name
+    {% else %}
+    coalesce(agg1.name,agg2.name) as aggregator_name
+    {% endif %}
+FROM {{base_trades}} base
```
jeff-dude, line 84: i was wondering why the final step took so long in incremental run -- looks like you're missing incremental logic / filter on this base table?

models/_sector/nft/trades/nft_trades_beta.sql
```diff
@@ -0,0 +1,13 @@
+{{ config(
+    schema = 'nft',
+    alias = 'trades_beta',
+    partition_by = ['blockchain','project','block_month'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
```
jeff-dude, line 7: can you add incremental predicates here too? to refine the target read, similar to addition to macro for source read?

jeff-dude (changes requested):

macros/models/_sector/nft/enrich_nft_trades.sql
0xRobin: Thank you, I was looking for the reason of those ~15min incremental steps, but wasn't able to find it.

jeff-dude:
i merged in #4848 since we've had warnings pop through often post version upgrade. now you can see CI fail on that step.

0xRobin:
it's weird that I don't get this compile warning locally.. :thinking:

jeff-dude:
> it's weird that I don't get this compile warning locally.. 🤔

yeah, i've noticed that too sometimes. usually the warnings i see in CI end up being true gaps though. let me know if that isn't the case here and something else odd is going on

0xRobin:
I was able to get them locally now, might be a partial of full parse thing.
Downside, the current CI setup is "fail-fast" meaning you only get to see the first warning..
There doesn't seem to be an option to make it error but not exit on warnings though.

jeff-dude:
> I was able to get them locally now, might be a partial of full parse thing. Downside, the current CI setup is "fail-fast" meaning you only get to see the first warning.. There doesn't seem to be an option to make it error but not exit on warnings though.

yep, i had to do:
`dbt clean`
`dbt deps`
`dbt compile`

i should probably alias that locally to run all three every time i do `dbt compile`



> Downside, the current CI setup is "fail-fast" meaning you only get to see the first warning..
> There doesn't seem to be an option to make it error but not exit on warnings though.

agree on this, but at the same time, this *should* be relatively rare. your PR is unique in terms of moving / restructuring, ideally that is less often and/or us internally doing the heavy lifting on that.

jeff-dude (approved):

PR #4669, opened 2023-10-24
EVMs - Test Partition by Chain

MSilb7:
Queries using `evms` are slow when they're joined to each other. Since joins should often joined on chain, I'm testing to see if we can partition a view (in lieu of it being a native table).


aalan3:
Partitioning here will not do anything since these are not materialized. But if it was, partitioning this by blockchain should in practice be the same as doing a view union on the underlying tables (like we already are), except for maybe a marginal gain due the number of stages Trino creates. 

Generally I have found that joining big tables with other big tables becomes an issue specially if it's not easily sortable (for example joining only on hash). You could try reversing the order of the joins if they differ in size a lot, you want the smaller table to be the one in the join.

PR #4674, opened 2023-10-24
fix indexing and add swap v2

andrewhong5297:
adding raydium swapv2 in, and also fixing a few issues with raydium v4. 

Basically raydium v4 (which actually came out before v3 so idk why it's v4) has many pools from 2021 not initialized using the idl methods. Because of this, I decided to generate all the necessary data from the token transfers themselves. 

Note that doing so means that getting pool fees will likely be harder in the future. 

There are also roughly 100k trades (out of 129 million) that are missing right now because I have not figured out the logic for detecting the [number of serum cancel orders preceding the transfers](https://solscan.io/tx/3vmrLviVasgSnLcjj8pv685DKuumgrK6hxsZ3EvxeRYU9p4VuYgbfSJQbDiU3HFcccMGNxU4Rc6BALAU9buU2dr2). Will dig into this more in the future.


andrewhong5297:
should be a pretty quick/easy fix, have already QAed.

jeff-dude (approved):

PR #4678, opened 2023-10-24
zeroex - op taker update

RantumBits:
fix op taker field


RantumBits:
@Hosuke update to taker_field on OP fills

jeff-dude:
> @Hosuke update to taker_field on OP fills.
> 
> I ran into an error when testing seed file so removed for this

removing the seed test makes me a bit nervous, as we hit errors on bnb model previously where token bought/sold was flipped and caused amounts to blow up in value.

can we readd and try to work through the issue? what was the issue exactly? did you run `dbt compile`, grab the test query built, run on dune app?

PR #4679, opened 2023-10-24
oneinch: exclude another troubled transaction

jeff-dude:
fyi @grkhr 
this spell has required us to exclude a few transactions, i have another to exclude. these tx's are misrepresented downstream and `amount_usd` value is exponentially higher than it should be

### DIFF
--- a/models/oneinch/ethereum/uniswap/oneinch_uniswap_v3_ethereum_trades.sql
+++ b/models/oneinch/ethereum/uniswap/oneinch_uniswap_v3_ethereum_trades.sql
@@ -376,3 +376,7 @@ LEFT JOIN {{ source('prices', 'usd') }} as prices_eth
     {% else %}
     AND prices_eth.minute >= TIMESTAMP '{{project_start_date}}'
     {% endif %}
+WHERE 
+    src.tx_hash not in (
+        0x14613a8f439f5f65b3235fbae95b67bc5f6aa7b0937af73e8df593e00d8abb12 -- broken tx with wrong amount and decimals, causing $8,439,415,024,975,841 vol
+    )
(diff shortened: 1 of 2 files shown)


couralex6 (approved):

jeff-dude:
need to wait to merge until we re-enable `state:modified` step in prod

grkhr:
@jeff-dude no problem!

you can also exclude this one 
https://github.com/duneanalytics/spellbook/blob/c8761af35612d0d3f43b75cf635f3a947048f183/models/oneinch/ethereum/uniswap/oneinch_uniswap_v3_ethereum_trades.sql#L382

I made this change in another PR already, just in case if you want to align it in this cleanup

jeff-dude:
> @jeff-dude no problem!
> 
> you can also exclude this one
> 
> https://github.com/duneanalytics/spellbook/blob/c8761af35612d0d3f43b75cf635f3a947048f183/models/oneinch/ethereum/uniswap/oneinch_uniswap_v3_ethereum_trades.sql#L382
> 
> I made this change in another PR already, just in case if you want to align it in this cleanup

glad we're on the same page -- i've already opened a ticket internally with our decode team as to why that tx has an incorrect amount in the decoded source table.

```
select *
from oneinch_ethereum.AggregationRouterV5_call_uniswapV3Swap
where call_tx_hash = 0x14613a8f439f5f65b3235fbae95b67bc5f6aa7b0937af73e8df593e00d8abb12
```

image

jeff-dude:
i'll handle outside of this PR for now. ideally we get resolved at the source instead.

grkhr:
It's not decoding issue actually. It's wrong amount in input data (idk how it can be possible), but that "amount" is causing the problem here with abnormal volume

image

grkhr:
So I'd suggest just filter this and wait till we refactor all 1inch spells included in dex_aggregator.trades's spells

jeff-dude:
ah, good catch. thank you for clarification. i will add filter to the PR here then.

PR #4689, opened 2023-10-25
add Zora on zora to the referral rewards sector

0xRobin:
add Zora on zora to the referals rewards sector.
Also includes some logic fixes to the `is_referral` column


macros/models/_sector/referral/platforms/rabbithole_referral_rewards.sql
```diff
@@ -4,28 +4,34 @@
...
     {% if is_incremental() %}
-    where {{incremental_predicate('evt_block_time')}}
+    and {{incremental_predicate('tx.block_time')}}
     {% endif %}
-)
+{% if is_incremental() %}
+where {{incremental_predicate('evt_block_time')}}
+{% endif %}
 
-{{ add_tx_from_and_to('model', blockchain) }}
```
jeff-dude, line 30: any reason in particular you're moving away from this macro here?

models/_sector/referral/rewards/platforms/zora_zora_rewards.sql
```diff
@@ -0,0 +1,16 @@
+{{ config(
+    schema = 'zora_zora',
```
jeff-dude, line 2: i assume it's to match decoded naming standard, but still a bit odd to see `zora_zora` instead of just `zora`

is it worth simplifying that naming standard universally, or better to indicate it's zora protocol on zora blockchain to differentiate?

macros/models/_sector/referral/platforms/rabbithole_referral_rewards.sql
0xRobin: Yes I needed the tx_from column to fill the referee_address

models/_sector/referral/rewards/platforms/zora_zora_rewards.sql
0xRobin: I don't like it either.. But still think it's the best we can do for the reason you mentioned above.

jeff-dude (approved):

PR #4690, opened 2023-10-25
Staking typo fix

hildobby:
Btw @jeff-dude, you mentioned query csv uploads if there starts to be too many entries, would updating the csv also trigger a rerun on the spells? How would that work?


jeff-dude:
> Btw @jeff-dude, you mentioned query csv uploads if there starts to be too many entries, would updating the csv also trigger a rerun on the spells? How would that work?

it would be a source in dbt, so no, it wouldn't auto trigger downstream refreshes. this is part of the risk of using uploads as sources right now, as we don't fully understand how often they may change and affect downstream usage. we need to figure out as much detail there as we can before we use heavily in spellbook and have some proactive guardrails to help us

jeff-dude (approved):

PR #4695, opened 2023-10-26
Bug under nft aggregators schema

antonio-mendes:
Was trying to run a dbt compile locally and ran into a problem with what I believe is a bug. 

Pushed a fix here but I might be missing some context. Feel free to ignore this PR if that's the case


aalan3:
I ran into the same issue when running 1.6 locally. We are planning on upgrading the DBT version soon too.

PR #4699, opened 2023-10-27
fix uniswap trades

markusbkoch:
Swap events contain amounts in and out for both tokens 0 and 1, so we actually have two "swaps" in each event:
* Token 0 in, token 1 out
* Token 1 in, token 0 out

Most of the time, one of these swaps has empty/zero amounts, ie the trader is only swapping one token for another.  However, there are some edge cases where **both** amounts for one of the tokens is non-zero, and both amounts for the other token are zero. eg [0x2643cc80871c4134704863b17f10901ea6d11646fc10162ff1fb504e7137c193](https://etherscan.io/tx/0x2643cc80871c4134704863b17f10901ea6d11646fc10162ff1fb504e7137c193#eventlog), evt index 75

In other cases, both amounts for both tokens are non-zero eg [0x5afc81f74fbfca62c7d083cccbadd01158f65bfdb1f2c7fb24e4866c1d96e346](https://etherscan.io/tx/0x5afc81f74fbfca62c7d083cccbadd01158f65bfdb1f2c7fb24e4866c1d96e346#eventlog), evt index 27

As [this query](https://dune.com/queries/3150762) shows, dex.trades only partially reflects the information present in the swap events mentioned above. This happens because the current implementation of `uniswap_v2_ethereum_trades.sql` assumes there is only one swap contained in the Swap event: if `amount0Out == 0` the code assumes token 1 is being bought, else token 0 is being bought.

This PR addresses this limitation by extracting both swaps from the Swap event and discarding the "empty swap" that is present in the vast majority of cases.

While events like the ones in the transactions mentioned here are very much the exception, they can have significant impact in some analyses. For example, the data regarding the swap txn `0x2643cc...` would lead to believe that the pool sold 50M USDC for 0 WETH when in reality the pool sold 50M USDC for 50.1M USDC.

cc @mendesfabio @viniabussafi @thetroyharris for comments/review


jeff-dude:
```
14:52:44  Database Error in test check_dex_seed_uniswap_v2_ethereum_trades_ethereum__uniswap__2 (models/uniswap/ethereum/uniswap_ethereum_schema.yml)
14:52:44    TrinoExternalError(type=EXTERNAL, name=DELTA_LAKE_INVALID_SCHEMA, message="Metadata not found in transaction log for test_data.dex_trades_seed", query_id=20[23](https://github.com/duneanalytics/spellbook/actions/runs/6668740709/job/18124984341?pr=4699#step:12:24)1027_145240_03630_p3atd)
```
this error means the seed file in dbt doesn't exist to run the test against. here are some helpful things to know in the setup:
- the test is defined [in the schema file](https://github.com/duneanalytics/spellbook/blob/main/models/uniswap/ethereum/uniswap_ethereum_schema.yml#L117-L120), which is edited in this PR
- the test is [here](https://github.com/duneanalytics/spellbook/blob/main/tests/generic/check_dex_seed.sql), which calls a nested macro [here](https://github.com/duneanalytics/spellbook/blob/main/macros/test-helpers/check_seed_macro.sql). if you run `dbt compile` locally, that is the query which compiles based on params in the schema file passed in
- the referenced seed file is [here](https://github.com/duneanalytics/spellbook/blob/main/seeds/dex/trades/dex_trades_seed.csv), which contains rows manually added in dev process to prove out the query outputs expected data

in order to force the seed to build, you can make a minor modification to the seed file in this PR for CI test to build it out for downstream usage. you can add a newline or some other similar dummy change, to force into PR. can you give that a shot, see if it gets past that error?

jeff-dude:
as for the change being made in the PR, that's an interesting find. my initial thought is to be a bit careful, as there are quite a few other spells which are forks of uniswap v2, that use the same logic. i'd imagine we want to universally update the logic in order to have `dex.trades` accurate downstream for all dex's that follow a similar pattern.

we are actually in the process of a `dex.trades` redesign, while using uniswap as the example starting point project. within that redesign, we are looking to build a macro for all spells which use the same logic as uniswap v2, that way code is handled in one spot for future maintenance like this.

i wonder if we wait to make changes there? [here](https://github.com/duneanalytics/spellbook/pull/4533) is the PR, for reference

mendesfabio:

mendesfabio:
@jeff-dude thoughts on `Error getting snapshot for tokens.erc20`? I've seen someone on Discord saying it happens when a transaction from the seeds has a token that's not in `tokens.erc20` - but I didn't add any, simply merged with main.

jeff-dude:
> @jeff-dude thoughts on `Error getting snapshot for tokens.erc20`? I've seen someone on Discord saying it happens when a transaction from the seeds has a token that's not in `tokens.erc20` - but I didn't add any, simply merged with main.

this may have been just bad timing, as prod is going through a refresh and you may have run into when that spell was being refreshed. lets try to simply rerun it first

jeff-dude:
turns out this was an issue on our side (access control change), but it should be good now. sorry for the confusion! please give it another go when you are ready

markusbkoch:
Thanks for the input @jeff-dude. Although other projects are probably even less affected by this than uniswap is, I agree the suggestions here should be incorporated in #4533 (and/or #4730). I can submit PRs to those branches with the relevant changes, or we can wait until those are merged and only then incorporate the suggestions from this PR? Or did you have something else in mind?

jeff-dude:
> Thanks for the input @jeff-dude. Although other projects are probably even less affected by this than uniswap is, I agree the suggestions here should be incorporated in #4533 (and/or #4730). I can submit PRs to those branches with the relevant changes, or we can wait until those are merged and only then incorporate the suggestions from this PR? Or did you have something else in mind?

the changes in #4533 will be in "beta" until the entire `dex.trades` lineage fits into this new format. the original uniswap models will stay as-is until overwritten with the new ones post-beta phase. is this issue pressing to fix, time-wise? if so, we can consider moving forward here in this PR.

markusbkoch:
>the changes in 4533 will be in "beta" [...] is this issue pressing to fix, time-wise?

Yea, in that case I'd suggest going ahead with the fix as it impacts analyses like LVR computation which require knowledge of all tokens bought/sold by the pool

> we should also make the same changes to the new uniswap macro(s) as needed

Agreed. And to #4730 too (cc @0xBoxer). Let me know how you guys would like to coordinate on that front

jeff-dude:
> @jeff-dude just wanted to confirm when this is merged will the changes propagate to all queries/spells that depend on this, like dex.trades? Or will the changes by default only be applied to new data as it comes in?

yes, anything downstream will be refreshed along with these

models/uniswap/ethereum/uniswap_v2_ethereum_trades.sql
```diff
@@ -5,7 +5,7 @@
     materialized = 'incremental',
     file_format = 'delta',
     incremental_strategy = 'merge',
-    unique_key = ['block_date', 'blockchain', 'project', 'version', 'tx_hash', 'evt_index'],
+    unique_key = ['block_date', 'blockchain', 'project', 'version', 'tx_hash', 'evt_index', 'token_bought_address', 'token_sold_address'],
```
jeff-dude, line 8: adding two new unique keys is concerning, as we use the exact same set of unique keys in all projects that feed into `dex.trades` and it has worked well universally. why are we adding two more here?

jeff-dude:
can you rebase with main branch & pull in the new uniswap macro for dex redesign, make same changes there?

models/uniswap/ethereum/uniswap_v2_ethereum_trades.sql
markusbkoch: This is needed because this new implementation potentially decodes the swap event into two records - token0 for token1 and token1 for token0. See failed checks right before this commit for more context.

An alternative approach would be to aggregate those two records into a single one, but it would make the code quite a bit more complex/verbose (we'd have to figure out what is the token sold/bought based on the net amounts in/out of tokens 0/1) and potentially cause issues with negative values (still need to confirm this), so that's the tradeoff

markusbkoch: It does seem that the alternative approach I mentioned would result in negative values. Here's an example of an unusual [swap event](https://etherscan.io/tx/0x2bc47e6c0781485058328e32b6edb37520bb23a50d9b968bde4b04471e50436b#eventlog) (index 128) : 

```
amount0In : 659774361155999
amount1In : 250390
amount0Out : 100000355
amount1Out : 0
```

The net amounts are such that the pool's balance of **both** tokens increases. An heuristic that tried to infer the tokens sold/bought would result in both tokens as `token_sold` (ie sold by the trader to the pool). Still unclear to me why this happens, but I assume it's something along the lines of someone having transferred token0 to the pool before the actual trade occurred, and the Swap event just catching up with the state.

markusbkoch:
>can you rebase with main branch & pull in the new uniswap macro for dex redesign, make same changes there?

8559930, should be ready for review

models/uniswap/ethereum/uniswap_v2_ethereum_trades.sql
markusbkoch: I've added the two new keys to the dex.trades schema in 1b9c1fb

jeff-dude:
@markusbkoch thanks again for raising this issue. i think this has a larger impact across spellbook than anticipated, so i've linked a gh issue which outlines the bug found. let's focus conversation there until we come to an agreement, then we can revisit this PR and finalize.

PR #4519, opened 2023-10-04
Init: Referrals sector

0xRobin:
Initial setup of a sector spell to track and analyze referral rewards payed out by protocols.
Currently contains data for:
- zora
- rabbithole
- sound_xyz

PR #4704, opened 2023-10-28
Consolidate lending sector - borrow macro

tomfutago:
Aiming at consolidating lending sector with macro based approach in line with sector-level spells redesign #4637

As a first step - adding `lending.borrow` model that pulls together data for Aave and Moola (for now - will add more projects and spells (supply, flashloans etc) once this is finalised).

Inspiration: #4533 & #4519 🙏


tomfutago:
@jeff-dude any chance this one can have a look?

jeff-dude:
thanks for initiating a new sector 🙏 

thought it may be easier to summarize my initial feedback in one post here:
- can we follow directory structure from dex?
  - `models/_sector/lending/borrow/<blockchain>/platforms/` should be the full directory, with spells at different levels
  - platforms organized in their blockchain directory
  - blockchain specific spell at root of blockchain directory `lending_ethereum_base_borrow`
- i know your vote was for `stage`/`stg` terminology, but since we landed on `base_`, can we rename everything to this standard?
  - remove `staging/` directory
  - remove `stg__lending_borrow` in favor of the blockchain-specific ones from previous step
  - `aave_v1_ethereum_base_borrow` -- this would apply to all platforms
  - sticking to `base_` prefix will allow for downstream views on top of `lending.borrow` for project specific views and no name clash
- consolidate project-specific macros into one file, then have multilpe macros within the file
  - rename macro file to `lending_aave_compatible_borrow`
  - each macro within to be `lending_aave_v1_compatible_borrow` and so on
- remove `evt_` prefix on all column names prior to loading into spells, we typically don't carry that past decoded tables
- completely remove entries from source/schema yaml files in old directories, once confident in new entries
- i think it's important to keep spells that contain the same name as the ones you're deleting, so those exposed views are nice to keep stability of the query experience on the app, in case they are used in dashboards

tomfutago:
@jeff-dude I think I have core of the changes all in - can you have a look when you have a chance pls?

Thing worth adding would be tests or seeds. There are 2 already existing tests (for ethereum and optimism) so I can either replicate that kind of approach or go with seeds - as implemented for new `dex.trades_beta` design. Thoughts?

jeff-dude (approved):
fantastic work here 🪄 thank you for helping initiate a new sector in the new format, add another great example to spellbook!

you can go ahead and remove all those commented out schema/source file in older directories, i.e. everything with `#` preface in older directory yml files. once that is committed, we can look to merge 🤝

jeff-dude:
> @jeff-dude I think I have core of the changes all in - can you have a look when you have a chance pls?
> 
> Thing worth adding would be tests or seeds. There are 2 already existing tests (for ethereum and optimism) so I can either replicate that kind of approach or go with seeds - as implemented for new `dex.trades_beta` design. Thoughts?

seeds are always nice to have, as a sanity check in the dev process. if you want to go ahead and add them, similar to your celo dex PR, feel free!

PR #4730, opened 2023-11-06
Add generic uniswap forks

0xBoxer:
Building models to recognize Uniswap V2 forks across chains. Decoding these in a scaleable manner should alleviate some work for us.

I do this by looking for both Swap events and Pair_Created events in the logs tables and manually decoding the same in the same way we usually decode Uniswap V2 contracts.


jeff-dude:
closing this for now as i clean out the PR queue. if we think it's something to explore further, maybe we throw into linear triage and talk through how this would fully fit into new dex structure / how it impacts existing lineage

PR #4743, opened 2023-11-08
Transfers: Break out wrapped token into its own macro

aalan3:
After a discussion with @0xRobin in https://github.com/duneanalytics/spellbook/pull/4698 we decided to try to break out certain edge cases as their own macros in transfers.


0xRobin (approved):
looks good if tests are passing. :heavy_check_mark:

aalan3:

PR #4746, opened 2023-11-08
rename Zora in evms.info

MSilb7:
Renaming Zora in evms.info to match [Chainlist](https://chainlist.org/?search=zora)

"Remove 'Network' its cleaner"


jeff-dude (approved):
the change to this file itself is fine, but we have downstream dependencies to worry about with a merge. i think we should revisit those before more & more spells use this spell as a reference and cause a huge web of dependencies like erc20 tokens (which we're actively working on decreasing that lineage)

for now, we can label ready for merge and wait for a good time

jeff-dude:
> Got it, yeah this was a case where the "official" name of a chain changed. The unique key is the dune db name, which I imagine won't change.

yeah, all good, no issue with the change.

fyi, we're starting an effort to segment out the dbt project and break out lineages more. spells like this, or tokens/prices, would fit into it's own "project", then end the lineage at a certain point in there. other "projects" would then read that as a source, instead of ref, to help break these out a bit. more to come on this, but we are reaching a point at scale in the project where we need to get creative.

PR #4758, opened 2023-11-10
base chain dex & dex-agg routers

RantumBits:
labels for dex & dex-agg addresses


models/addresses/base/addresses_base_dex.sql
```diff
@@ -0,0 +1,30 @@
+{{config(alias = 'dex',
```
jeff-dude, line 1: please add schema config in the model directly

models/addresses/base/addresses_base_dex.sql
```diff
@@ -0,0 +1,30 @@
...
+        tags=['static'],
+        post_hook='{{ expose_spells(\'["base"]\',
```
jeff-dude, line 3: does this need to be exposed on the app, or is it more useful for downstream spells? we're trying to clean up the usage of the data explorer a bit

jeff-dude (changes requested):

models/addresses/base/addresses_base_dex.sql
RantumBits: I'm good without exposing this and only using for downstream spells

jeff-dude: sounds good. once you resolve these two comments, we can look to proceed

models/addresses/base/addresses_base_dex.sql
RantumBits: i added a separate schema file again here. lmk if any changes needed

models/addresses/base/addresses_base_dex.sql
```diff
@@ -0,0 +1,29 @@
+{{ config(
+    schema = 'addresses_base',
```
jeff-dude, line 2: i communicated that poorly, but i meant to add `schema` config here (i pushed a commit)

that actually worked out though, cause adding schema yml file was good too 🙂

jeff-dude (approved):
looks good to me now

PR #4763, opened 2023-11-10
Oneinch updates 4

grkhr:
- get rid of `_ar_` spellbooks family (will need to be removed on prod too!)
- add `lop_own_trades` to `dex.trades`
- hide **zksync** from `oneinch.lop` until the data is fine


jeff-dude (approved):
lgtm!

PR #4765, opened 2023-11-10
SPE-231 Table rename for incremental full refreshes

couralex6:
The dbt-trino incremental macro [drops a table](https://github.com/starburstdata/dbt-trino/blob/master/dbt/include/trino/macros/materializations/incremental.sql#L61-L65) before executing a full refresh. On every full refresh, the debug logs show this line:
```
Applying DROP to: "hive"."uniswap_v1_ethereum"."trades"
```
This means tables would be unavailable during full refreshes and tables would disappear when a full refresh fails for any reason. This has cause multiple issues in the past since we rely pretty heavily on full refreshes in our current job orchestration.

This issue had already been addressed for spark with the file currently in `macros/dune/incremental.sql`, however, this macro override was not getting picked up by the `dbt-trino` adapter (most likely because the override didn't pass the parameter `adapter = "trino"`). We didn't catch this issue during the migration, because queries reading from missing `dunesql` tables, would automatically fallback to reading from spark tables. 

To fix this downtime issue, we apply the [table materialization logic](https://github.com/starburstdata/dbt-trino/blob/master/dbt/include/trino/macros/materializations/table.sql#L34-L48). Instead of dropping the target table before recreating it, we create a new `intermediary` table, and rename it to the target after creation. 

With these changes, the debug logs show the following statements:
```
# Creating the intermediary table
create or replace table hive.uniswap_v1_ethereum.trades__dbt_tmp with ( ... )

# Backing up the table prior to executing full refresh
alter table "hive"."uniswap_v1_ethereum"."trades" rename to hive.uniswap_v1_ethereum.trades__dbt_backup

# Renaming the intermediary table to the target table name
alter table hive.uniswap_v1_ethereum.trades__dbt_tmp rename to hive.uniswap_v1_ethereum.trades

# Dropping the backup relation
drop table if exists hive.uniswap_v1_ethereum.trades__dbt_backup
```


jeff-dude:
will you repurpose this PR for the fixes to macros? or should we close?

jeff-dude:
makes sense to me!

i assume steps 2 & 3 run so quickly back-to-back that we don't notice any downtime on the table? in theory, its the same as dropping if you rename to `_backup` in terms of end user query experience, but if those two steps are so quick consecutively, then we'd be all good

couralex6:
> makes sense to me!
> 
> i assume steps 2 & 3 run so quickly back-to-back that we don't notice any downtime on the table? in theory, its the same as dropping if you rename to `_backup` in terms of end user query experience, but if those two steps are so quick consecutively, then we'd be all good

The logs show `SQL status: SUCCESS in 0.0 seconds` for both operations. I also see `On master: COMMIT` before and after those 4 operations, which makes me think they're being executed in an atomic transaction. I don't see anything in the macro that would execute that commit so not 100% sure. But regardless, they should be quick enough to not cause any issues.

jeff-dude:
let's find a good time to try to plan/merge tomorrow?

jeff-dude (approved):

PR #4774, opened 2023-11-13
Add `check_seed` tests and seed files with new schema to `base_trades` models

Hosuke:


jeff-dude:
thanks @Hosuke!

i agree to add, my question is:
should we add in the old format, or repurpose to use the generic approach?
https://github.com/duneanalytics/spellbook/blob/main/models/_sector/nft/trades/ethereum/platforms/_schema.yml#L20-L33

if we move to that format, i think i still like having one seed for all of dex. maybe we just move the seed to:
https://github.com/duneanalytics/spellbook/tree/main/seeds/_sector/dex

Hosuke:
> thanks @Hosuke!
> 
> i agree to add, my question is: should we add in the old format, or repurpose to use the generic approach? https://github.com/duneanalytics/spellbook/blob/main/models/_sector/nft/trades/ethereum/platforms/_schema.yml#L20-L33
> 
> if we move to that format, i think i still like having one seed for all of dex. maybe we just move the seed to: https://github.com/duneanalytics/spellbook/tree/main/seeds/_sector/dex

We can recycle the old/existing `dex_trades_seed` records, and for the test we may repurpose to use the generic approach.

The main purpose to use `check_dex_seed` here is to recycle the seed records.
We can just make a copy of the current one into the new intended place, or just move the current one to there.

Hosuke:
I will generate a new seed file with `block_number` field first.

Hosuke:
The new `dex_trades_seed` with `block_number` is generated based on the query here:
https://dune.com/queries/3200929

Hosuke:
There are a lot of files generated from the existing dex_trades_seed, I put them under `seeds/_sector/dex/generated/`. We can see if we want the flat file structure for the seed first, also we can review the schema and file namings.

Hosuke:
I redesign the `dex_base_trades_seed` schema to have columns:
```
        blockchain: varchar
        project: varchar
        version: varchar
        tx_hash: varbinary
        evt_index: uint256
        block_number: uint256
        token_bought_address: varbinary
        token_sold_address: varbinary
        token_bought_amount_raw: uint256
        token_sold_amount_raw: uint256
        block_date: timestamp
```

The data is generated from this query:
https://dune.com/queries/3202536

Hosuke:
Hi @jeff-dude, I generate the new dex_base_trades_seed files with new schema design.
Please review and see if these files are expected.

jeff-dude:
> Hi @jeff-dude, I generate the new dex_base_trades_seed files with new schema design. Please review and see if these files are expected.

a few questions:
- what's the intention of `dex_base_trades_seed` and `dex_base_trades_seed_extended`?
  - i thought the intention was to break those up into project-specific seeds, like you've done here, then remove those?
- in our new dex structure, each project base level spell in the schema file should call its individual seed -- from what i see, only three blockchain schema files are modified in this PR and they call `dex_base_trades_seed` or nothing at all?
  - i'd expect each uniswap blockchain schema file to call its own seed file, same with defiswap

other feedback:
- i removed the `generated` directory, i think it's fine to have all project seeds at the root of `seeds/_sector/dex/`
- can we rename `1inch` to `oneinch` in seeds / seed schemas / wherever else?

Hosuke:
> > Hi @jeff-dude, I generate the new dex_base_trades_seed files with new schema design. Please review and see if these files are expected.
> 
> a few questions:
> 
> * what's the intention of `dex_base_trades_seed` and `dex_base_trades_seed_extended`?
>   
>   * i thought the intention was to break those up into project-specific seeds, like you've done here, then remove those?

The intention of these two seed files is for verification. After verification these two files can be removed.

I am going to use these seed files for `check_seed` tests, as the verification is done.

Hosuke:
I just finalized this PR in the last few commits. Should be ready to go now.

models/_sector/dex/trades/arbitrum/_schema.yml
```diff
@@ -15,5 +15,16 @@ models:
           combination_of_columns:
             - tx_hash
             - evt_index
+      - check_seed:
+          seed_file: ref('uniswap_arbitrum_base_trades')
+          match_columns:
+            - tx_hash
+            - evt_index
+          check_columns:
+            - block_number
```
jeff-dude, line 24: i believe `block_number` should be in the `match_columns` list, not `check_columns`, as we'd intend for it to be part of the join condition instead, right? that's my read on it. i guess technically either way works, as the two current match columns are unique, so adding `block_number` there doesn't add much value. i guess i'm good with how you have it now.

jeff-dude: sorry that was confusing, i started off saying to change, then decided it's fine 🤣

jeff-dude (approved):
looks much better -- thank you!

PR #4788, opened 2023-11-14
Add zkSync and Modify Combined Project Mappings

lgingerich:
1. Add `contracts_project_name_mappings` for zkSync
2. Modify the combined project mappings to reference Optimism and zkSync models rather than rewrite the full list

### DIFF
--- a/models/contracts/contracts_project_name_mappings.sql
+++ b/models/contracts/contracts_project_name_mappings.sql
@@ -4,119 +4,27 @@
     schema = 'contracts',
     alias = 'project_name_mappings',
     unique_key='dune_name',
-    post_hook='{{ expose_spells(\'["ethereum", "optimism", "arbitrum", "avalanche_c", "polygon", "bnb", "gnosis", "fantom", "base", "goerli"]\',
+    post_hook='{{ expose_spells(\'["ethereum", "optimism", "arbitrum", "avalanche_c", "polygon", "bnb", "gnosis", "fantom", "base", "goerli", "zksync"]\',
                               "sector",
                               "contracts",
-                              \'["msilb7", "chuxin"]\') }}'
+                              \'["msilb7", "chuxin", "lgingerich"]\') }}'
     )  
 }}
 
-select 
-  dune_name
-  , mapped_name
-from (
-    values
-     ('lyra_v1',	'Lyra Finance')
-    ,('Lyra V1', 'Lyra Finance')
-    ,('lyra_avalon', 'Lyra Finance')
-    ,('Avalon Lyra', 'Lyra Finance')
-    ,('avalon_lyra', 'Lyra Finance')
-    ,('lyra', 'Lyra Finance')
-    ,('aave_v3', 'Aave')
-    ,('perp_v2', 'Perpetual Protocol')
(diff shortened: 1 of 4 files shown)


lgingerich:
recheck

models/contracts/contracts_project_name_mappings.sql
```diff
@@ -4,119 +4,27 @@
...
-    ,('niftykit_v3','NiftyKit')
-    ,('cozy_v2_beta','Cozy Finance')
-    ,('cozy_v2_prod','Cozy Finance')
-    ,('highlight_xyz','Highlight.xyz')
-    ,('mint_fun','mint.fun')
-    ,('mintfun','mint.fun')
-    ,('allo_protocol', 'allo protocol (gitcoin)')
-    ,('union_protocol', 'Union Finance')
+{% set contracts_project_mapping_models = [
+  ref('contracts_optimism_project_name_mappings')
```
MSilb7, line 15: This will be deprecated. Ideally all project names are mapped here so there's consistent naming across all chains.

MSilb7: If needed, we can try to add logic for an 'exclude_chain' list. But the real ideal solve is for Dune decoding to enforce/override bad namespaces.

lgingerich: >project names are mapped natively

What do you mean by this?

jeff-dude:
how are we proceeding here @lgingerich @MSilb7?

is this outdated with the new dynamic contract mapping spells?

models/contracts/contracts_project_name_mappings.sql
MSilb7: sorry, I meant in the cross-chain version https://github.com/duneanalytics/spellbook/blob/main/models/_sector/contracts/contracts_project_name_mappings.sql

MSilb7:
> how are we proceeding here @lgingerich @MSilb7?
> 
> is this outdated with the new dynamic contract mapping spells?

Hi, yeah [chain-level versions of dynamic mapping are here](https://github.com/duneanalytics/spellbook/tree/main/models/_sector/contracts) (iterates through contract creation trees to get to the top-level deployer [i.e. contract A, created by contract B, created by contract C, created by EOA D] & uses deployer addresses in addition to Dune decodings to map contracts <> projects, so we don't have to track each contract one-by-one)

For zksync, I think we'd need to add some extra "if zksync" logic to the [starting-level table](https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/contracts/contracts_base_starting_level.sql) (pulls contract creation data) to handle for contracts that show an event for being created, but not a creation trace (and any other intricacies).

If we want to preserve the contract-level mappings, they can go in to the [overrides table](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/contracts/contracts_contract_overrides.sql).

jeff-dude:
apologies if i'm getting ahead of myself, but i'm going to close this PR since @MSilb7 is planning to open PRs with code for zksync in contract mappings. if you could tag @lgingerich in that PR to coordinate together, that'd be great 🙏

PR #4812, opened 2023-11-16
ENS source.yml updated, ens_registrations, ens_node_names updated 

nyssarex:
fixed spell, so union works with same amount of columns


jeff-dude (approved):
thank you @nyssarex 🙏

PR #4147, opened 2023-08-23
Create `zora.mints` incl. Ethereum, Zora Network, Optimism, Base & Goerli

hildobby:

PR #4819, opened 2023-11-16
Create macro for `nft.mints` and add `nft_base_native_mints`

chuxinh:
## Description
Create a macro for `nft.mints`, specifically `nft_native_mints` (let me know if we should change the macro name to `nft_native_mints` instead.

Would like to get this one PR reviewed and merged before touching existing queries for `nft_native_mints`

### DIFF
--- a/macros/models/_sector/nft/nft_mints.sql
+++ b/macros/models/_sector/nft/nft_mints.sql
@@ -167,9 +167,9 @@ FROM
         WHERE 1=1
         {% endif %}
         AND nft_mints."from"= 0x0000000000000000000000000000000000000000
-        {% if blockchain == 'ethereum' %}
+        {%- if blockchain == 'ethereum' %}
         AND nft_mints.contract_address NOT IN (SELECT address FROM {{ ref('addresses_ethereum_defi') }})
-        {% endif %}
+        {%- endif -%}
         AND nft_mints.blockchain = '{{blockchain}}'
         {% if is_incremental() %}
         {{incremental_predicate('nft_mints.block_time')}}
--- a/macros/models/_sector/nft/nft_mints.sql
+++ b/macros/models/_sector/nft/nft_mints.sql
@@ -117,7 +117,7 @@ FROM
         , nft_mints.evt_index
         FROM {{ ref('nft_transfers') }} nft_mints
         LEFT JOIN nfts_per_tx nft_count ON nft_count.tx_hash=nft_mints.tx_hash
-        LEFT JOIN{{ base_traces }} et ON et.block_time=nft_mints.block_time
+        LEFT JOIN {{ base_traces }} et ON et.block_time=nft_mints.block_time
             AND et.tx_hash=nft_mints.tx_hash
             AND et."from"=nft_mints.to
             AND (et.call_type NOT IN ('delegatecall', 'callcode', 'staticcall') OR et.call_type IS NULL)
@@ -170,7 +170,7 @@ FROM
         {%- if blockchain == 'ethereum' %}
         AND nft_mints.contract_address NOT IN (SELECT address FROM {{ ref('addresses_ethereum_defi') }})
         {%- endif -%}
-        AND nft_mints.blockchain = '{{blockchain}}'
+         AND nft_mints.blockchain = '{{blockchain}}'
         {% if is_incremental() %}
         {{incremental_predicate('nft_mints.block_time')}}
         {% endif %}
--- a/macros/models/_sector/nft/nft_mints.sql
+++ b/macros/models/_sector/nft/nft_mints.sql
@@ -1,0 +1,185 @@
+{% macro nft_mints(blockchain, base_contracts, base_traces, erc20_transfer, base_transactions ) %}
+
+WITH namespaces AS (
+    SELECT 
+        address
+        , min_by(namespace, created_at) AS namespace
+	FROM {{ base_contracts }}
+	GROUP BY address
+	)
+
+, nfts_per_tx_tmp AS (
+    SELECT 
+        tx_hash
+        , sum(amount) AS nfts_minted_in_tx
+    FROM {{ ref('nft_transfers') }}
+    WHERE 
+        blockchain = '{{blockchain}}'
+        {% if is_incremental() %}
+        AND {{incremental_predicate('block_time')}}
+        {% endif %}
+    GROUP BY tx_hash
+    )
+
+, nfts_per_tx as (
+    SELECT
+        tx_hash
+        , case when nfts_minted_in_tx = UINT256 '0' THEN UINT256 '1' ELSE nfts_minted_in_tx END as nfts_minted_in_tx
+    FROM nfts_per_tx_tmp
+)
+SELECT
+    blockchain
+    , project
+    , version
+    , block_time
+    , block_date
+    , block_month
+    , block_number
+    , token_id
+    , collection
+    , token_standard
+    , trade_type
+    , number_of_items
+    , trade_category
+    , evt_type
+    , seller
+    , buyer
+    , amount_raw
+    , amount_original
+    , amount_usd
+    , currency_symbol
+    , currency_contract
+    , nft_contract_address
+    , project_contract_address
+    , aggregator_name
+    , aggregator_address
+    , tx_hash
+    , tx_from
+    , tx_to
+    , platform_fee_amount_raw
+    , platform_fee_amount
+    , platform_fee_amount_usd
+    , platform_fee_percentage
+    , royalty_fee_receive_address
+    , royalty_fee_currency_symbol
+    , royalty_fee_amount_raw
+    , royalty_fee_amount
+    , royalty_fee_amount_usd
+    , royalty_fee_percentage
+    , evt_index
+FROM
+(
+SELECT
+    *,
+    ROW_NUMBER() OVER (PARTITION BY tx_hash, evt_index, token_id, number_of_items ORDER BY amount_usd DESC NULLS LAST) as rank_index
+FROM
+    (
+        SELECT distinct '{{blockchain}}' AS blockchain
+        , COALESCE(ec.namespace, 'Unknown') AS project
+        , '' AS version
+        , nft_mints.block_time AS block_time
+        , CAST(date_trunc('day', nft_mints.block_time) as date) AS block_date
+        , CAST(date_trunc('month', nft_mints.block_time) as date) AS block_month
+        , nft_mints.block_number AS block_number
+        , nft_mints.token_id AS token_id
+        , tok.name AS collection
+        , nft_mints.token_standard
+        , CASE WHEN nft_mints.amount= UINT256 '1' THEN 'Single Item Mint'
+            ELSE 'Bundle Mint'
+            END AS trade_type
+        , nft_mints.amount AS number_of_items
+        , 'Mint' AS trade_category
+        , 'Mint' AS evt_type
+        , nft_mints."from" AS seller
+        , nft_mints.to AS buyer
+        , CAST(COALESCE(SUM(CAST(et.value as DOUBLE)), SUM(CAST(erc20s.value as DOUBLE)), 0)*(nft_mints.amount/nft_count.nfts_minted_in_tx) AS UINT256) AS amount_raw
+        , COALESCE(SUM(CAST(et.value as DOUBLE))/POWER(10, 18), SUM(CAST(erc20s.value as DOUBLE))/POWER(10, pu_erc20s.decimals))*(nft_mints.amount/nft_count.nfts_minted_in_tx) AS amount_original
+        , COALESCE(pu_eth.price*SUM(CAST(et.value as DOUBLE))/POWER(10, 18), pu_erc20s.price*SUM(CAST(erc20s.value as DOUBLE))/POWER(10, pu_erc20s.decimals))*(nft_mints.amount/nft_count.nfts_minted_in_tx) AS amount_usd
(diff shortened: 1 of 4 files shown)


0xRobin:
This is great @chuxinh! I'll give it a closer look on monday.
Is there a specific reason you want to do the existing chains in a follow up PR?
I think having them already in here and being able to compare the PR output to the existing models would be a benefit.

chuxinh:
> This is great @chuxinh! I'll give it a closer look on monday. Is there a specific reason you want to do the existing chains in a follow up PR? I think having them already in here and being able to compare the PR output to the existing models would be a benefit. Not a must though.

Thought it'd be easier to test with smaller PR first, but I can totally make modifications to existing chains using this macro too when you come back to take a closer look. Appreciate it!

macros/models/_sector/nft/nft_mints.sql
```diff
@@ -0,0 +1,185 @@
...
+        AND {{incremental_predicate('block_time')}}
+        {% endif %}
+    GROUP BY tx_hash
+    )
+
+, nfts_per_tx as (
+    SELECT
+        tx_hash
+        , case when nfts_minted_in_tx = UINT256 '0' THEN UINT256 '1' ELSE nfts_minted_in_tx END as nfts_minted_in_tx
+    FROM nfts_per_tx_tmp
```
0xRobin, line 28: looked into these
```
    select * from (
    SELECT 
        tx_hash
        , sum(amount) AS nfts_minted_in_tx
    FROM nft_ethereum.transfers
    group by tx_hash
    ) where nfts_minted_in_tx = 0
```
they basically all are "mints" that had an ERC1155 transfer with 0 amount (thus not actually minting anything). I assume mostly due to bad contract design that emits the event even if no token was actually minted. 
I think we should filter out these cases instead of adjusting them to 1.

macros/models/_sector/nft/nft_mints.sql
```diff
@@ -0,0 +1,185 @@
...
+        , UINT256 '0' AS platform_fee_amount_raw
+        , DOUBLE '0' AS platform_fee_amount
+        , DOUBLE '0' AS platform_fee_amount_usd
+        , DOUBLE '0' AS platform_fee_percentage
+        , CAST(NULL as VARBINARY) AS royalty_fee_receive_address
+        , '0' AS royalty_fee_currency_symbol
+        , UINT256 '0' AS royalty_fee_amount_raw
+        , DOUBLE '0' AS royalty_fee_amount
+        , DOUBLE '0' AS royalty_fee_amount_usd
+        , DOUBLE '0' AS royalty_fee_percentage
```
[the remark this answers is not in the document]
chuxinh: oh yea, doesn't look like it's even showing up in the `nft.mints` maybe just for individual protocols. Yea I'm happy to have them removed in your cleanup

macros/models/_sector/nft/nft_mints.sql
```diff
@@ -0,0 +1,185 @@
...
+        {%- if blockchain == 'ethereum' %}
+        AND nft_mints.contract_address NOT IN (SELECT address FROM {{ ref('addresses_ethereum_defi') }})
+        {%- endif -%}
+        {% if is_incremental() %}
+        {{incremental_predicate('nft_mints.block_time')}}
+        {% endif %}
+        GROUP BY nft_mints.block_time, nft_mints.block_number, nft_mints.token_id, nft_mints.token_standard
+        , nft_mints.amount, nft_mints."from", nft_mints.to, nft_mints.contract_address, etxs.to, nft_mints.evt_index
+        , nft_mints.tx_hash, etxs."from", ec.namespace, tok.name, pu_erc20s.decimals, pu_eth.price, pu_erc20s.price
+        , agg.name, agg.contract_address, nft_count.nfts_minted_in_tx, pu_erc20s.symbol, erc20s.contract_address, et.success
```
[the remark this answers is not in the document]
chuxinh: Yea i think the group by is used to remove duplicates from joining those erc20 transfer tables (if i recall correctly)

chuxinh:
Possible to know what's causing the error for incrementally build `models/_sector/nft/mints/native/nft_ethereum_native_mints.sql`

macros/models/_sector/nft/nft_mints.sql
```diff
@@ -0,0 +1,181 @@
...
+            AND {{incremental_predicate('etxs.block_time')}}
+            {% endif %}
+        LEFT JOIN {{ ref('nft_aggregators') }} agg 
+            ON etxs.to=agg.contract_address
+            AND agg.blockchain = '{{blockchain}}'
+        LEFT JOIN {{ ref('tokens_nft') }} tok 
+            ON tok.contract_address=nft_mints.contract_address
+            and tok.blockchain = '{{blockchain}}'
+        LEFT JOIN namespaces ec ON etxs.to=ec.address
+        {% if is_incremental() %}
```
0xRobin, line 157: @chuxinh I think this block is the cause of the incremental error.
I have not seen a usage of `{{this}}` in a macro before, and it might resolve to the the macro name instead of the model name. :thinking: 
I'll try running this without this block.

chuxinh: > I have not seen a usage of {{this}} in a macro before, and it might resolve to the the macro name instead of the model name. 🤔
Ah this is like joining to the existing table, I think i've used it a few times in spellbook

0xRobin (changes requested):
seems like the macro logic is not matching up with the current `nft.mints` table. 
From the stats it does look like we are overinflating the volume, so I think we have a bug.
It's not directly clear where the volume is coming from as it's not mapped to a project. :thinking: 
image

https://dune.com/resident_wizards/nft-mints-comparison

0xRobin:
@tomfutago also has a [WIP PR](https://github.com/duneanalytics/spellbook/pull/4608) for nft.mints macro. 
Maybe he can help take a look at this and we can combine forces to make this work?

tomfutago:
Hey @chuxinh and @0xRobin. Nice work here! Yes, I also looked at this (#4608) and attempted to rewrite the main query to avoid long list of columns in group by and row_number on top of it, by splitting base query in (hopefully) more manageable cte's. Thing is - traces table slogs down performance quite a bit (try for example adding polygon to your current version of the macro - I'll bet it'll time out). There are other comments on my PR as well. If anything I've done there is salvageable here - please free to use it.

I know all of the above is not really responding to the volume discrepancies question, but what I'm trying to say is that main query might need reworking first.

chuxinh:
Think I figured out what the problem is for optimism mint amount discrepancy. 

The main differences are due to a bug in [this line](https://github.com/duneanalytics/spellbook/blob/004eacd15f6a2109cecfe9acd4622dfac2a1ec10/models/_sector/nft/mints/native/nft_optimism_native_mints.sql#L151), as a result of which, almost anything with erc20 transfer on OP Mainnet (mostly DeFi NFT positions) would be missing as it's joining to erc20 transfer on Ethereum instead. Examples such as [this](https://optimistic.etherscan.io/tx/0x6ddb646834a5b7b51ede2f455cd5df78439578a3c19732d5caf50f51a2a93cdb), where the `amount_raw` would be null, but if we change that to erc20 transfer optimism, then the number would show up in the example query [here](https://dune.com/queries/3229497/5401361).

So if you can re-run the test result and filter for `currency_symbol` = `ETH` on Optimism, the results should look fairly similar to that in the existing `nft.mints`.

Still need to figure out what's the issue for that on Ethereum

macros/models/_sector/nft/nft_mints.sql
```diff
@@ -0,0 +1,189 @@
...
+	)
+
+, nfts_per_tx AS (
+    SELECT
+        tx_hash
+        , sum(amount) AS nfts_minted_in_tx
+    FROM {{ ref('nft_transfers') }}
+    WHERE
+        blockchain = '{{blockchain}}'
+        AND "from"= 0x0000000000000000000000000000000000000000
```
chuxinh, line 18: think this line is impacting the differences in nft mint amount too as there are mints like [this](https://etherscan.io/tx/0x62c2f6f4a515c7ff3be033b833639fa0cfdac913323a14c7c58972a04a82217d) one where there are multiple transfers. if we only include from `0x0000000000000000000000000000000000000000`
then we are missing out other transfers 
cc: @0xRobin

chuxinh:
> Hey @chuxinh and @0xRobin. Nice work here! Yes, I also looked at this (#4608) and attempted to rewrite the main query to avoid long list of columns in group by and row_number on top of it, by splitting base query in (hopefully) more manageable cte's. Thing is - traces table slogs down performance quite a bit (try for example adding polygon to your current version of the macro - I'll bet it'll time out). There are other comments on my PR as well. If anything I've done there is salvageable here - please free to use it.
> 
> I know all of the above is not really responding to the volume discrepancies question, but what I'm trying to say is that main query might need reworking first.

the previous optimism nft mints uses transfer_eth, which is a separate eth transfer table that probably helps with avoiding using traces instead

macros/models/_sector/nft/nft_mints.sql
0xRobin: I don't think we want to double count this transfer though in your example?

0xRobin:
update:
seems like on `ethereum` we fixed the discrepancy, but on OP it's still present. 
image


From your comment here

> The main differences are due to a bug in [this line](https://github.com/duneanalytics/spellbook/blob/004eacd15f6a2109cecfe9acd4622dfac2a1ec10/models/_sector/nft/mints/native/nft_optimism_native_mints.sql#L151), as a result of which, almost anything with erc20 transfer on OP Mainnet (mostly DeFi NFT positions) would be missing as it's joining to erc20 transfer on Ethereum instead. Examples such as [this](https://optimistic.etherscan.io/tx/0x6ddb646834a5b7b51ede2f455cd5df78439578a3c19732d5caf50f51a2a93cdb), where the `amount_raw` would be null, but if we change that to erc20 transfer optimism, then the number would show up in the example query [here](https://dune.com/queries/3229497/5401361).

I understand that you believe the recent spike is not a bug, but the older model is missing this data?
I reran the results and filtering on `currency_symbol = 'ETH'` confirms this. :+1:

0xRobin (approved):
:heavy_check_mark: I'm happy merging this in in the current state.
There is still a bigger discussion to be had whether `nft.mints` should or should not include defi mints (LP tokens), but that's out of the scope of this PR.

chuxinh:
> ✔️ I'm happy merging this in in the current state. There is still a bigger discussion to be had whether `nft.mints` should or should not include defi mints (LP tokens), but that's out of the scope of this PR.

Thanks for the review Rob! 

I agree that the DeFi LP positions can be skewing the data and make it less relevant especially when we talk about NFTs, we mostly talk about the art category. However, imo instead of removing them from `nft.mints` (which they are really technically part of erc1155 or erc721), an alternative solution is to include another column that flags if that's a DeFi NFT or not. 

Currently, only Ethereum has DeFi LP positions removed (according to the logic and list of DeFi addresses curated by Hildobby). However, it's not perfect since the list is not exhaustive and is ever growing. Would like to see how we can improve on excluding those and I've been trying some other methods which are not perfect either 🤷‍♀️

jeff-dude:
hey guys -- thanks for getting this prepped for merge!

i wanted to note that i merged #4147 which added zora mints. i hope i didn't mess with structure there. if so, please do bring into this PR and clean as needed. if all is good as-is, just comment back and i'll merge 👍

chuxinh:
Hi @jeff-dude I think hildobby's one is separate from the Zora mints I did here, which is zora mints for Zora platform, while here is for the blockchain protocol level. Let me know if that makes sense.

I will update the branch to make sure there's no merge conflict

kotte34723:
This is also a huge unlock for zora team @jeff-dude :)

jeff-dude:
> This is also a huge unlock for zora team @jeff-dude :)

this PR is separate from the other `zora.mints` created, correct?
we'll end up with:
`zora.mints` from prior PR from @hildobby 
`zora.native_mints` in this PR, which feeds into `nft.mints`

is this expected? or do they need to merge together into one?

kotte34723:
AFAICT - @hildobby addresses zora contracts (ie all zora contracts across all chains), whereas this PR gets us the minting activity on the zora platform itself @jeff-dude @chuxinh

models/_sector/nft/mints/native/nft_base_native_mints.sql
```diff
@@ -0,0 +1,19 @@
+{{ config(
+        
+        schema = 'nft_base',
+        alias = 'native_mints',
+        partition_by = ['block_month'],
+		file_format = 'delta',
+		incremental_strategy = 'merge',
```
jeff-dude, line 7: we're missing `materialized = 'incremental',` on this one

jeff-dude: you can see in CI logs, it doesn't run incrementally

chuxinh:
> AFAICT - @hildobby addresses zora contracts (ie all zora contracts across all chains), whereas this PR gets us the minting activity on the zora platform itself @jeff-dude @chuxinh

Yes, that's my understanding too. So if we look at `nft_native_mints` on Optimism and the zora app contracts on Optimism, the native_mints > zora app activities. Depending on if Zora has other nft apps on the chain in the future, i think currently zora mints from zora contracts should be roughly the same as native mints on zora.

Hildobby's work definitely have more context in terms of mint details, which is usually joined in the overall nft mints as my understanding

chuxinh:
@jeff-dude just updated accordingly 🙏

jeff-dude (approved):
thank you all for the patience & helping me understand 🙏

PR #4831, opened 2023-11-18
Add WETH price for Zora Network

hildobby:


scripts/token_checker.py
```diff
@@ -23,7 +23,8 @@ def __init__(self):
                             "fantom": "ftm-fantom",
                             "celo": "celo-celo",
                             "base": "base-base",
-                            "zksync": "zksync-zksync"
+                            "zksync": "zksync-zksync",
+                            "zora": "zora-zora"
```
Hosuke, line 27: `zora` may not be supported in the coinspaprika list yet:
https://api.coinpaprika.com/v1/contracts/

```suggestion
                            "zksync": "zksync-zksync"
```

models/prices/zora/prices_zora_tokens.sql
```diff
@@ -0,0 +1,19 @@
...
+SELECT
+    token_id
+    , blockchain
+    , symbol
+    , contract_address
+    , decimals
+FROM
+(
+    VALUES
+    ('weth-weth', 'zora', 'WETH', 0x4200000000000000000000000000000000000006, 18)
```
Hosuke, line 18: I will try to add weth without modify to `scripts/token_checker.py` and see what will happen.

Hosuke:
Do we need to merge this PR first or we can wait for `zora` being supported in coinspaprika?

Hosuke (approved):
LGTM.✅
Thank you @hildobby

scripts/token_checker.py
jeff-dude: hm, good catch, let's commit this suggestion and see the logs output

jeff-dude:
while i think the spell will build fine and be able to be used downstream in other spells, i'm worried if we merge this prices test will continually fail on all prices PRs until its added to the pricing provider.

@Hosuke are you aware of a way to make this test pass, with the current state of code in PR here?

Hosuke:
> while i think the spell will build fine and be able to be used downstream in other spells, i'm worried if we merge this prices test will continually fail on all prices PRs until its added to the pricing provider.
> 
> @Hosuke are you aware of a way to make this test pass, with the current state of code in PR here?

We may cheat the code `"zora":"eth-ethereum"` to pass the tests before `zora` is officially supported by coinpaprika.

hildobby:
> Do we need to merge this PR first or we can wait for `zora` being supported in coinspaprika?

Is there a known ETA on when coipaprika supports zora? Unless it's within the next few days, I would rather get this merged with the "hack" proposed above since weth is quite important imo

jeff-dude:
> We may cheat the code `"zora":"eth-ethereum"` to pass the tests before `zora` is officially supported by coinpaprika.

nice one!


> Is there a known ETA on when coipaprika supports zora? Unless it's within the next few days, I would rather get this merged with the "hack" proposed above since weth is quite important imo

no known timeline. i know we can add prices through their site, but i'm not sure about blockchains 🤔 i'm comfortable with this workaround though, to workaround our own CI tests. you will only be able to add tokens which exist on ethereum too (assuming price is the same on both chains), then use the zora address in the spell of course

scripts/token_checker.py
```diff
@@ -23,7 +23,8 @@ def __init__(self):
                             "fantom": "ftm-fantom",
                             "celo": "celo-celo",
                             "base": "base-base",
-                            "zksync": "zksync-zksync"
+                            "zksync": "zksync-zksync",
+                            "zora": "eth-ethereum"
```
jeff-dude, line 27: fyi @aalan3 @couralex6 @0xRobin 
we are forcing zora prices to look on ethereum blockchain in the prices CI test. users should be good to add coinpaprika ID values which exist on other blockchains, then just use the zora address value in the spell to join other spells downstream.

jeff-dude: not sure how to communicate with their team on if / when zora will be included in API

jeff-dude (approved):

PR #4832, opened 2023-11-18
Add sushiswap to `dex.trades_beta`

tomfutago:


macros/models/_sector/dex/uniswap_compatible_trades.sql
```diff
@@ -8,6 +8,27 @@
...
+        t.amountIn AS token_sold_amount_raw,
+        null AS amount_usd,
+        t.tokenOut AS token_bought_address,
+        t.tokenIn AS token_sold_address,
+        t.contract_address AS project_contract_address,
+        t.evt_tx_hash AS tx_hash,
+        t.evt_index
+    FROM {{ Pair_evt_Swap }} t
+    {% if is_incremental() %}
+    WHERE {{ incremental_predicate('t.evt_block_time') }}
```
Hosuke, line 27: Seems pretty clean, but not sure if optimism is an edge case.

As the macro may effect many spell, if the case is not general, we may implement the specific case in its base_trades spell.

tomfutago: good point. will update in optimism spell.

Hosuke (approved):
LGTM.✅ and clean.
Thank you @tomfutago !

Hosuke:
Not sure how to refresh and load the existing seed files

jeff-dude:
> Also we can continue submitting **1 PR <--> 1 project** next time, to keep it clean!

i think this fits into one project. it's just sushiswap spans so many blockchains. this is likely easiest from a developer perspective and works for me. easier to coordinate merging too.



> Not sure how to refresh and load the existing seed files.

i was able to refresh 9/10 seeds by flipping rows, but base blockchain only has one row so it didn't come through. that's annoying. maybe we can add another row? two should ideally be the minimum anyway.

edit: it ran fine as-is 🤔 

although it's not the highest priority, we do have a backlog item to work on running seeds in prod so they don't need refreshed like this every time. we're hitting a bug where the job is failing on a few seeds (we run them all in the project), then when the job fails, something is happening where the data is dropped from data lake yet the external table persists. hence the error for metadata missing. we need to investigate further when we have time, as this affects other spells occasisionally too.

jeff-dude:
> i was able to refresh 9/10 seeds by flipping rows, but base blockchain only has one row so it didn't come through. that's annoying. maybe we can add another row? two should ideally be the minimum anyway.
> 
> edit: it ran fine as-is 🤔

looks like base blockchain schema file doesn't include the seed test. maybe we add the seed test in the schema file, add one more row, let it all run again?

models/_sector/dex/trades/base/_schema.yml
```diff
@@ -26,5 +26,20 @@ models:
...
+      contributors: wuligy, tomfutago
+    config:
+      tags: [ 'base', 'dex', 'trades', 'sushiswap', 'v2' ]
+    description: "sushiswap base v2 base trades"
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - tx_hash
+            - evt_index
+
```
Hosuke, line 44:
```suggestion
      - dbt_utils.unique_combination_of_columns:
          combination_of_columns:
            - tx_hash
            - evt_index
      - check_seed:
          seed_file: ref('sushiswap_base_base_trades_seed')
          filter:
            version: 1
          match_columns:
            - tx_hash
            - evt_index
          check_columns:
            - block_number
            - token_bought_address
            - token_sold_address
            - token_bought_amount_raw
            - token_sold_amount_raw

```

models/_sector/dex/trades/base/platforms/sushiswap_v2_base_base_trades.sql
```diff
@@ -0,0 +1,21 @@
...
+        unique_key = ['tx_hash', 'evt_index'],
+        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')]
+    )
+}}
+
+{{
+    uniswap_compatible_v3_trades(
+        blockchain = 'base',
+        project = 'sushiswap',
+        version = '2',
```
Hosuke, line 17: The version is ambiguous for `sushiswap_v2_base.base_trades`,
as the source using here is `sushiswap_v3_base.UniswapV3Pool_evt_Swap`, and `sushiswap_base.trades` uses version `1`.

If we want to keep consistent with `dex.trades`, we can use version `1` here, otherwise we may change the corresponding seed's version.

Hosuke (changes requested):

models/_sector/dex/trades/base/platforms/sushiswap_v2_base_base_trades.sql
tomfutago: Existing `sushiswap_base_trades` is incorrect imo - it should show as version **2**, but I agree that source naming is a mess. In general: sushi v1 is based on uni v2 and sushi v2 on uni v3, so in the case of base chain we have the following sources:

sushi v1 - not added to models yet:
`sushiswap_base.Pair_evt_Swap`
`sushiswap_base.Factory_evt_PairCreated` (contract decoding: Pending as I accidently set it as created by a factory contract..)

sushi v2 - added here, but with confusing naming as you noticed:
`sushiswap_v3_base.UniswapV3Pool_evt_Swap`
`sushiswap_base.UniswapV3Factory_evt_PoolCreated`

Hosuke: Noted, then we may modify the corresponding base_trades seed to version `'2'`.

seeds/_sector/dex/sushiswap_base_base_trades_seed.csv
```diff
@@ -1,2 +1,5 @@
 blockchain,project,version,block_date,tx_hash,evt_index,token_bought_address,token_sold_address,block_number,token_bought_amount_raw,token_sold_amount_raw
 base,sushiswap,1,2023-08-11,0x9a0de02d064c2093c18eb1fcf221c5d2f8ddbb959f152a50953ff7305d7487d3,33,0x4200000000000000000000000000000000000006,0xeb466342c4d449bc9f53a865d5cb90586f405215,2470544,549506070553956765,1014922053
```
Hosuke, line 2: Updated version based on the comment above.
```suggestion
base,sushiswap,2,2023-08-11,0x9a0de02d064c2093c18eb1fcf221c5d2f8ddbb959f152a50953ff7305d7487d3,33,0x4200000000000000000000000000000000000006,0xeb466342c4d449bc9f53a865d5cb90586f405215,2470544,549506070553956765,1014922053
```

models/_sector/dex/trades/base/_schema.yml
```diff
@@ -1,6 +1,8 @@
 version: 2
 
 models:
+  - name: dex_base_base_trades
```
Hosuke, line 4: I just rearranged the schema to make conflicts resolving easier.

Hosuke (approved):
LGTM.✅
Thank you @tomfutago

jeff-dude (approved):

Issue #4759, opened 2023-11-10
migrate all dex models to the new structure / design


Hosuke:
Updated at Apr 23, 2024:
Since all dexes have been migrated to the new dex.trades lineage.
We continue to remove/modify the legacy code project by project.

-----------------------

Hello Team,

I'm opening this issue to keep track of all the Pull Requests (PRs) that implement the new macro approach in dex.trades, as introduced in PR [#4533](https://github.com/duneanalytics/spellbook/pull/4533).

### Purpose
- To monitor and discuss the integration of new dexs into the `_beta` dex lineage.
- Share insights and updates on how these integrations are progressing.

### Call for Contributions
- If you are working on a dex PR that incorporates this new macro approach, please link it here.
- We can submit **_1 PR <--> 1 project_** to keep it clean.
- Share any thoughts or feedback on the macro implementation.

Looking forward to seeing the innovative ways this new approach is being utilized and enhanced!

### Migration tracking table
Dune Query by @tomfutago: https://dune.com/queries/3240358
| Project       | dex.trades | dex.base_trades | dex.trades_beta | Contributors |
|---------------|------------|-----------------|-----------------|--------------|
| oneinch       | ✅          | ❌  |  ✅ #5303  |              |
| aerodrome     | ✅          | ✅ #4837   |  ✅ | @tomfutago  |
| airswap       | ✅          | ✅ #4864  |   ✅  |   @tomfutago   |
| apeswap       | ✅          |  ✅ #4866  |   ✅  |  @Hosuke  |
| arbswap       | ✅          |  ✅ #4869  |  ✅ | @tomfutago  |
| babyswap      | ✅          |  ✅ #4923   |  ✅  |  @tomfutago   |
| balancer      | ✅          |  ✅ #4940   |  ✅ |  @tomfutago  |
| bancor        | ✅          |  ✅ #4972   |   ✅   | @tomfutago  |
| beethoven_x   | ✅          |   ✅ #5047   |   ✅  |  @tomfutago     |
| biswap        | ✅          |  ✅ #4924   |   ✅  |  @tomfutago    |
| camelot       | ✅          |  ✅ #4867   |   ✅  |  @tomfutago   |
| carbon_defi   | ✅          | ✅ #4868  |   ✅ | @tomfutago    |
| carbonhood    | ❌    |   ✅ #4842    |  ✅  |  @tomfutago   |
| clipper       | ✅          |  ✅ #5043   |   ✅  |  @tomfutago   |
[...]

PR #4837, opened 2023-11-20
Add aerodrome to `dex.trades_beta`

tomfutago:
linked to #4759


tomfutago:
Hey @Hosuke, this might be another edge case - it looks like aerodrome is a uniswap V2 fork (all columns match on `Pair_evt_Swap`), but factory pool event has 'pool' rather than 'pair' column (as in uni V3). I updated macro to add `pair_column_name` with default 'pair' so it's backward compatible and shouldn't affect existing spells. Only for cases like aerodrome I can pass 'pool' for that column name and still be able to use macro rather than copy the whole spell. If you disagree let me know.

On another note, I'm getting this error msg now - any hints what's that about?

`TrinoExternalError(type=EXTERNAL, name=DELTA_LAKE_INVALID_SCHEMA, message="Metadata not found in transaction log for test_data.sushiswap_celo_base_trades_seed"`

Hosuke:
> Hey @Hosuke, this might be another edge case - it looks like aerodrome is a uniswap V2 fork (all columns match on `Pair_evt_Swap`), but factory pool event has 'pool' rather than 'pair' column (as in uni V3). I updated macro to add `pair_column_name` with default 'pair' so it's backward compatible and shouldn't affect existing spells. Only for cases like aerodrome I can pass 'pool' for that column name and still be able to use macro rather than copy the whole spell. If you disagree let me know.

I think we can extend the macro's compatibility. cc. @jeff-dude 


> On another note, I'm getting this error msg now - any hints what's that about?
> 
> `TrinoExternalError(type=EXTERNAL, name=DELTA_LAKE_INVALID_SCHEMA, message="Metadata not found in transaction log for test_data.sushiswap_celo_base_trades_seed"`

We may need to find a way to load the existing seed files when needed.

jeff-dude:
> I think we can extend the macro's compatibility. cc. @jeff-dude

yes, i'm open to extending the macro to accept word switches on column / table / whatever is needed 👍 



> We may need to find a way to load the existing seed files when needed.

commented in sushiswap PR about why it's happening. for now, just flip a row with another to push into this PR and have CI build it fresh.

Hosuke:
> Hey @Hosuke, this might be another edge case - it looks like aerodrome is a uniswap V2 fork (all columns match on `Pair_evt_Swap`), but factory pool event has 'pool' rather than 'pair' column (as in uni V3). I updated macro to add `pair_column_name` with default 'pair' so it's backward compatible and shouldn't affect existing spells. Only for cases like aerodrome I can pass 'pool' for that column name and still be able to use macro rather than copy the whole spell. If you disagree let me know.
> 
> On another note, I'm getting this error msg now - any hints what's that about?
> 
> `TrinoExternalError(type=EXTERNAL, name=DELTA_LAKE_INVALID_SCHEMA, message="Metadata not found in transaction log for test_data.sushiswap_celo_base_trades_seed"`

Solved. @tomfutago

macros/models/_sector/dex/uniswap_compatible_trades.sql
```diff
@@ -4,6 +4,7 @@
     , version = null
     , Pair_evt_Swap = null
     , Factory_evt_PairCreated = null
+    , pair_column_name = 'pair'
```
Hosuke, line 7: I think this implementation is simple and clean.

jeff-dude (approved):
thanks for keeping it clean with the macro enhancement 🤝

PR #4844, opened 2023-11-20
Add combined spells to safes

alicecha:


jeff-dude:
hi @alicecha -- thank you for the PR 🙏 

i have a suggestion that should make maintaining the safe spell lineage easier and more efficient (assuming only main change is adding `amount_usd`, git is making it hard to fully see in the compare). rather than joining to prices in each individual blockchain spell, i'd look to save the join for the final step downstream and within a spell materialized as a view.
- this assumes each blockchain version spell is a building block to reach the ultimate end goal, `safe.transactions` and `safe.native_transfers` to be used on the dune app.
- new tokens are added to prices relatively often, so historical data could be missing for `amount_usd` over time
- save each "building block" in the dbt lineage, i.e. the blockchain spells, to just raw/decoded tables so full refreshes are required less often, as that historical data remains static
- to keep solid performance on querying the end result spells, you can introduce a staging layer between the blockchain-specific spells and the final view, which materializes incrementally just like the blockchain ones, but still doesn't join to prices quite yet. just unions the blockchains together, like you do now. you could add a prefix on these final two spells, like `stg_` or `base_` (we use `base_` in other lineages) to indicate it's not final state
- the final spell would be a view and left join to prices and introduce `amount_usd`

i'd also recommend a minor name cleanup on those final two spells (that would be turned into views from above steps, if agreed):
`safe.native_transfers`
`safe.transactions`
where both remove the `_all` suffix. we have a ton of cross-chain spells, but typically keep the names a bit cleaner and leave out the `_all`.

let me know if you think this is something you want to rework here and fit into the structure 🙏 

we're building out a similar structure for new `dex.trades` design, found in [this directory](https://github.com/duneanalytics/spellbook/tree/main/models/_sector/dex/trades), if you wanted examples. those also use macros for reusable code, an added bonus if you want to explore in the future.

alicecha:
Thanks @jeff-dude, this sounds sensible. Would adding usd prices at the end mean that we don't get the usd prices on the blockchain-specific spells?

@danielpartida - let me know if you have any thoughts on this.

jeff-dude:
> Thanks @jeff-dude, this sounds sensible. Would adding usd prices at the end mean that we don't get the usd prices on the blockchain-specific spells?
> 
> @danielpartida - let me know if you have any thoughts on this.

yes, that is the case. my assumption was your downstream queries / dashboards would always be on the final cross-chain spell with filters on the column for blockchain as needed.

another use case we're doing on dex/nft trades is leveraging that prefix in table name to indicate it's a data pipeline table first and foremost (so `base_` prefix), then having a view on top of the cross-chain spell for any chain-specific spells that are required for downstream usage. that view is simply select all from cross-chain spell with filter on blockchain. then you'd have prices there.

jeff-dude:
> Thanks @jeff-dude, that makes sense both in terms of efficiency and maintenance. If we go down this route, I will need a bit of support if that's okay. @danielpartida, do you have any reservations against migrating to cross-chain instead of chain-specific spells?

happy to help & we have multiple good examples to follow now as well (all in that `models/_sector/` directory.

i'm not sure "migrating to cross-chain instead of chain-specific spells" is the right way to frame it. i consider [this spell](https://github.com/duneanalytics/spellbook/pull/4844/files#diff-c16afc1c452fe79b1513c61a047185005ad3d076a65c132642aa703573343673) already to be cross-chain. is it not the one you use downstream in queries / dashboards?

alicecha:
> > Thanks @jeff-dude, that makes sense both in terms of efficiency and maintenance. If we go down this route, I will need a bit of support if that's okay. @danielpartida, do you have any reservations against migrating to cross-chain instead of chain-specific spells?
> 
> happy to help & we have multiple good examples to follow now as well (all in that `models/_sector/` directory)
> 
> i'm not sure "migrating to cross-chain instead of chain-specific spells" is the right way to frame it. i consider [this spell](https://github.com/duneanalytics/spellbook/pull/4844/files#diff-c16afc1c452fe79b1513c61a047185005ad3d076a65c132642aa703573343673) already to be cross-chain. is it not the one you use downstream in queries / dashboards?

No, I just created this spell. We use chain-specific spells in our queries and dashboards at the moment.

danielpartida:
Hi @jeff-dude

If the logic of the spells is correct, any chance that we could push the changes now? We would like to run a couple of analyses with those spells. As a compromise, we would evaluate internally our best spell strategy going forward taking into consideration your input.

jeff-dude (approved):
sounds good. i'd recommend considering sooner rather than later, so not too much downstream on the dune app is built when things could change here. we will continue to move existing & new spells into a similar structure. when you're ready, plz do open an issue with your planned approach and tag us to approve, to make the dev process more efficient 🙏

PR #4847, opened 2023-11-21
Add scroll and linea to rollup_economics

lorenz234:
Added scroll and linea to the rollup economics dune spellbook. 

Only small thing to note here is that Linea sends the verification proofs in the same transaction as the trx data. Therefore at current time it is not possible to distinguish between these two types of costs. I have therefore preliminary added all L1 costs for Linea to the l1_data_fees table.


mseidlx:
@dune-eng is there anything we can do to get this PR processed soon? Would be super helpful for us to get it in

jeff-dude:
please see the failures in the logs. the new unions don't have the same number of columns. be sure the columns match exactly

lorenz234:
ahhh yes, I can quickly fix this, I think a recent update of the spell change the overall structure of the query

jeff-dude (approved):
thanks for the quick fix!

PR #4855, opened 2023-11-22
Add Zora to contracts_mapping

MSilb7:
Adds Zora chain to contracts_mapping spells.

Also contains schema unique column test modifications (to match post_hook)

Per https://github.com/duneanalytics/spellbook/pull/4747, we can observe the existing contracts_mapping for a few days before continuing to add chains.


models/_sector/contracts/zora/contracts_zora_base_iterated_creators.sql
```diff
@@ -0,0 +1,15 @@
+ {{
+  config(
+        schema = 'contracts_zora',
+        alias = 'base_iterated_creators',
+        materialized ='incremental',
+        file_format ='delta',
+        incremental_strategy='merge',
+        unique_key = ['blockchain', 'contract_address'],
```
jeff-dude, line 8: unique keys don't match unique combo of columns test in schema file. plz update one or the other to get them to match and assign the proper keys

jeff-dude: this applies to all zora spells in here

jeff-dude (changes requested):

models/_sector/contracts/zora/contracts_zora_base_iterated_creators.sql
MSilb7: modding schema - created_tx_hash is leftover from legacy versions

jeff-dude (approved):

PR #4856, opened 2023-11-22
Make CEX address list crosschain

hildobby:


models/cex/cex_evms_addresses.sql
Hosuke, line 1: So the original addresses in each chain are now put into this one single file.

Hosuke (approved):

jeff-dude:
maybe i'm misunderstanding, but building this backwards (all evms first, then blockchain views after) doesn't seem to accomplish the same end result.

if `cex_evms_addresses` has all cex addresses, across all blockchains, but no blockchain column, how do we differentiate? when the blockchain-specific views are built on top, they simply add the blockchain column, but don't filter any addresses out. aren't the addresses blockchain specific? or does one address repeat across all evm blockchains?

hildobby:
> maybe i'm misunderstanding, but building this backwards (all evms first, then blockchain views after) doesn't seem to accomplish the same end result.
> 
> if `cex_evms_addresses` has all cex addresses, across all blockchains, but no blockchain column, how do we differentiate? when the blockchain-specific views are built on top, they simply add the blockchain column, but don't filter any addresses out. aren't the addresses blockchain specific? or does one address repeat across all evm blockchains?

We don't need to differentiate, those are EOA addresses which are thus shared across chains. It is much easier to maintain 1 file than multiple. For backwards compatibility I made chain specific spells views of that spell, but also added for other chains so that in case someone wants cex_zora.addresses it's there (might not be obvious to everyone that those are crosschain). The goal however is to only have cex_evms.addresses show up in the table search UI as I believe this is the spell that people should care about and train themselves to use over the chain specific ones (not. jsut about naming but imo the spell name is also explanatory in what it contains).

jeff-dude (approved):

jeff-dude:
> We don't need to differentiate, those are EOA addresses which are thus shared across chains. It is much easier to maintain 1 file than multiple. For backwards compatibility I made chain specific spells views of that spell, but also added for other chains so that in case someone wants cex_zora.addresses it's there (since it might not be obvious to everyone that those addresses are crosschain). The goal however is to only have cex_evms.addresses show up in the table search UI as I believe this is the spell that people should care about and train themselves to use over the chain specific ones (not just about naming but imo the spell name is also explanatory in what it contains, thus being a good learning opportunity for felow analysts).

gotcha. please note, we don't have automation around removing table properties (i.e. what the post hook sets in config). if the chain specific ones still appear in data explorer, that's why. i think because we're rebuilding as views, maybe it'll all reset, but not 100% sure.

PR #4858, opened 2023-11-22
Make OP Contracts Backwards Compatible

MSilb7:
(this accidentally got removed from https://github.com/duneanalytics/spellbook/pull/4747)

Create a view for the OP-Specific contracts mapping that re-aliases column names, such that it is backwards compatible with any queries based on the previous version of the table.


models/_sector/contracts/optimism/contracts_optimism_contract_mapping_backwards_compat.sql
jeff-dude, line 1: would it be better to rename [this one](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/contracts/optimism/contracts_optimism_contract_mapping.sql) to it's schema+alias, then this file can be just `contracts_optimism_contract_mapping.sql`?

MSilb7: We could, but some column names are different, which is why I have this copy.

jeff-dude: what i meant to say:
- `contracts_optimism_contract_mapping.sql` --> `contracts_optimism_contract_creator_project_mapping.sql` since that is the new schema/alias on that file
- `contracts_optimism_contract_mapping_backwards_compat.sql` --> `contracts_optimism_contract_mapping.sql` to match schema/alias in this file

helps clean up the navigation a bit. the downside is that you may need to update downstream refs with the name changes.

what do ya think?

MSilb7: Yeah, I think if I do that, I'll rename the other chains' names as well, since they similarly don't match 1:1.

I'll make a note to hold off on this one until:
- https://github.com/duneanalytics/spellbook/pull/5026
- https://github.com/duneanalytics/spellbook/pull/5046

Are in, to avoid collisions

jeff-dude: are we waiting on this PR again until:
- #5162 
- #5159 

are both merged?

jeff-dude:
closing for now, feel free to reopen in future as needed

PR #4863, opened 2023-11-23
Add airswap_ethereum_base_trades

Hosuke:


tomfutago:
Ah haha, just noticed we're adding the same dex 😂 I'm adding it as a macro in case you want to use it

Hosuke:
> Ah haha, just noticed we're adding the same dex 😂 I'm adding it as a macro in case you want to use it

Ok, I will close this PR. I think your macro could be more useful.

PR #4867, opened 2023-11-23
Add camelot to `dex.trades_beta`

tomfutago:
Linked to #4759.

Notes:
* I updated camelot_v1 to camelot_v2 as that's in line with [Camelot Docs](https://docs.camelot.exchange/contracts/amm-v2/factory)
* `uniswap_compatible_v3_trades` macro includes `fee` column in the base cte, which is not subsequently exposed in the final query (or in the current `dex_trades` model). Camelot v3 doesn't have this column, so I commented it out in the macro. If it ever gets exposed maybe we can revisit it then?


macros/models/_sector/dex/uniswap_compatible_trades.sql
```diff
@@ -73,7 +73,7 @@ WITH dexs AS
         , CASE WHEN amount0 < INT256 '0' THEN f.token0 ELSE f.token1 END AS token_bought_address
         , CASE WHEN amount0 < INT256 '0' THEN f.token1 ELSE f.token0 END AS token_sold_address
         , t.contract_address as project_contract_address
-        , f.fee
+        --, f.fee
```
Hosuke, line 76: Do you think we may further extend the macro's compatibility by adding optional columns in the signature? Like default `optional_columns = ['f.fee']`.

tomfutago: Sure, good idea. My note of fee column not being used down the line still stands though, but worth having a safety net.

macros/models/_sector/dex/uniswap_compatible_trades.sql
```diff
@@ -73,7 +74,11 @@ WITH dexs AS
         , CASE WHEN amount0 < INT256 '0' THEN f.token0 ELSE f.token1 END AS token_bought_address
         , CASE WHEN amount0 < INT256 '0' THEN f.token1 ELSE f.token0 END AS token_sold_address
         , t.contract_address as project_contract_address
-        , f.fee
+        {% if optional_columns %}
+            {% for optional_column in optional_columns %}
+            , {{ optional_column }}
+            {% endfor %}
+        {% endif %}
```
Hosuke, line 81: Updated to a loop.

jeff-dude: can you guys help clarify my understanding of this?
- current `main` branch implementation has `f.fee` being pulled in CTE, but then left out of select downstream
- camelot decoded table(s) doesn't have this column, so it would fail on read
- the `{% if optional_columns %}` checks if the column exists, if yes, includes in CTE, if not, it leaves it out (i can confirm `dbt compile` returns no column)
  - is the if check essentially: try to pull it, if it will fail, then remove?

tomfutago: Correct. For camelot optional_columns is passed as an empty list and if detects it as such and skips this section.

jeff-dude (approved):

macros/models/_sector/dex/uniswap_compatible_trades.sql
tomfutago: No worries!

PR #4869, opened 2023-11-23
Add arbswap to `dex.trades_beta`

tomfutago:


Hosuke (approved):
LGTM.✅
Thank you @tomfutago 

[Sample arbswap_arbitrum_base_trades]()

models/_sector/dex/trades/arbitrum/platforms/arbswap_arbitrum_base_trades.sql
```diff
@@ -0,0 +1,76 @@
...
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'merge',
+        unique_key = ['tx_hash', 'evt_index'],
+        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')]
+    )
+}}
+
+WITH dexs AS (
+    -- Arbswap AMM
```
jeff-dude, line 14: this dex is a good example of open questions as we move forward. from what i can tell, the first source query here is [uniswap v2 compatible](https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/dex/uniswap_compatible_trades.sql#L1)

i wonder if we explore mixing a combo of macro and union of non-macro code?

models/_sector/dex/trades/arbitrum/platforms/arbswap_arbitrum_base_trades.sql
```diff
@@ -0,0 +1,76 @@
...
+    FROM
+        {{ source('arbswap_arbitrum', 'SwapPair_evt_Swap') }} t
+    INNER JOIN {{ source('arbswap_arbitrum', 'SwapFactory_evt_PairCreated') }} f
+        ON f.pair = t.contract_address
+    {% if is_incremental() %}
+    WHERE {{incremental_predicate('t.evt_block_time')}}
+    {% endif %}
+    
+    UNION ALL 
+
```
Hosuke, line 36: We may start trying to replace the part before `UNION ALL` with the macro.

tomfutago:
Combo of macro + straight select applied, but as macro already contains cte - it can't be a direct replacement of current select, but needs to go as a separate nested cte and then union outputs together. If there's better way to go about it - lmk.

Hosuke (approved):
LGTM.✅ for the macro-hybrid spell

jeff-dude (approved):
i'm good with this approach -- thank you for finding a solution!

@Hosuke let's keep in mind for other projects moving forward.

PR #4877, opened 2023-11-26
Celo update coinpaprika prices setup

tomfutago:


tomfutago:
just test.. prices not available

PR #4900, opened 2023-11-29
1inch: 1inch swaps view

max-morrow:
- created 1inch swaps spellbook (oneinch.swaps view): all swaps of 1inch protocols
- added remains & receiver to LOP (Limit Order Protocol) macro


jeff-dude (approved):
looks good!

PR #4911, opened 2023-11-30
Add gas to transfers

aalan3:
This is a start but not sure what the to address here should be. Part of the gas is sometimes burned? But otherwise I guess this would go the validator?


jeff-dude:
closing this for now, until we revisit the need to add

PR #4914, opened 2023-11-30
add post_hook, follow-up to #4715

fmarrr:

### DIFF
--- a/models/gyroscope/gyroscope_trades.sql
+++ b/models/gyroscope/gyroscope_trades.sql
@@ -1,7 +1,11 @@
 {{ config(
     schema = 'gyroscope',
     alias = 'trades',
-    materialized = 'view'
+    materialized = 'view',
+    post_hook='{{ expose_spells(\'["arbitrum", "ethereum","optimism", "polygon"]\',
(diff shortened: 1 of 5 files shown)


models/gyroscope/gyroscope_trades.sql
```diff
@@ -1,7 +1,11 @@
 {{ config(
     schema = 'gyroscope',
     alias = 'trades',
-    materialized = 'view'
+    materialized = 'view',
+    post_hook='{{ expose_spells(\'["arbitrum", "ethereum","optimism", "polygon"]\',
```
jeff-dude, line 5: we typically only expose the cross-chain spell, assuming this is the one you'd query most on the app?

if so, i'd remove the post hook from the chain-specific spells

fmarrr: sounds good! just expose the cross-chain spell should be enough

thanks!

jeff-dude: please do go ahead and push a commit to remove the chain-specific ones 🙏

PR #4939, opened 2023-12-04
Add zksync transactions to safes

alicecha:
Something went wrong with the previous PR. Submitting this one with all previous comments addressed @jeff-dude


jeff-dude:
gm @alicecha -- plz see the failure in CI test and push commit(s) to fix 🙏

models/safe/zksync/safe_zksync_transactions.sql
Hosuke, line 1: Field `execution_gas_used` is missing in this spell.

alicecha (commented):
Added gas variables

models/safe/zksync/safe_zksync_transactions.sql
alicecha: Thanks @Hosuke, I corrected this now.

alicecha:
@Hosuke @jeff-dude, the uniqueness test is failing, not sure why because tested it in Dune query.

models/safe/zksync/safe_zksync_transactions.sql
```diff
@@ -0,0 +1,112 @@
+{{ 
+    config(
+        materialized='incremental',
+        schema = 'safe_zksync',
+        alias = 'transactions',
+        partition_by = ['block_month'],
+        unique_key = ['block_date', 'tx_hash', 'trace_address'],
```
jeff-dude, line 7: it looks like this spell is failing on duplicates after the incremental run, which typically means the unique keys are incorrect and/or contain nulls, so the generated merge statement fails to find a match when it should.

can you validate these are the correct unique keys? are we missing a field? if not, then check if any of these contain null values. if there are nulls, you can look to introduce a surrogate key instead of using fields individually, which will never output a null.

here is example of surrogate key, if you get to the point of needing it:
assign in config [like this](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/tokens/arbitrum/tokens_arbitrum_base_transfers.sql#L9)
build [like this](https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/tokens/transfers_base.sql#L54)

jeff-dude (changes requested):

models/safe/zksync/safe_zksync_transactions.sql
alicecha: Super helpful, thanks! It's all running fine and the tests are passing now :)

models/safe/zksync/safe_zksync_transactions.sql
```diff
@@ -0,0 +1,106 @@
...
+    tr.sub_traces,
+    tr.trace_address,
+    t.success,
+    tr.error,
+    tr.code,
+    t.input, --get input from transactions (because contains the methodID)
+    cast(t.output as varbinary) as "output",
+    t.method,
+    tr.tx_hash as trace_tx_hash, --save the trace_tx_hash to match back on
+    {{dbt_utils.generate_surrogate_key(['t.tx_hash', "array_join(tr.trace_address, ',')"])}} as unique_key
```
jeff-dude, line 101: last step:
can you add `unique_key` to the schema yml file and then add a unique test on that column?

jeff-dude (changes requested):

alicecha:
Should be all good now, thanks!

jeff-dude (approved):
thank you!

PR #4940, opened 2023-12-04
Add balancer to `dex.trades_beta`

tomfutago:
linked to #4759


tomfutago:
@jeff-dude @Hosuke I made a start on adding balancer to `dex.trades_beta` starting with macro that can be useful for balancer's forks (like beethoven_x etc), but there are a couple of things I want to check first before I get to far:
* current `balancer_v1_ethereum_trades` and `balancer_v2_ethereum_trades` spells have `pool_id` and `swap_fee` columns added to the output, which are not added to the current `dex.trades` but for backward compatibility they need adding to exposed balancer spells - I'm assuming that's the way to go?
* even if `pool_id` and `swap_fee` columns are not exposed in `dex.trades_beta` there is a swap logic dependent on swap fees so I'd need to add it anyway (kind of thinking "aloud" here, but let me know if you're disagree)
* `balancer_v2` model uses additional `bpt_prices` spell, which if I understand it correctly - backfills usd token prices calculated based on swaps (if token not present in `prices.usd` feed). I'd assume this additional logic adds quite a bit of an overhead and might even not be necessary if all token prices are available in `prices.usd` table (needs checking). Thoughts?

jeff-dude:
> * current `balancer_v1_ethereum_trades` and `balancer_v2_ethereum_trades` spells have `pool_id` and `swap_fee` columns added to the output, which are not added to the current `dex.trades` but for backward compatibility they need adding to exposed balancer spells - I'm assuming that's the way to go?

let's add those column to the base trades macro(s), then leave out of `dex.trades_beta` for now. we can then build the view on top of `dex.trades_beta` which could join back to base spell and output those columns. does that seem like a reasonable approach?

> * even if `pool_id` and `swap_fee` columns are not exposed in `dex.trades_beta` there is a swap logic dependent on swap fees so I'd need to add it anyway (kind of thinking "aloud" here, but let me know if you're disagree)

i believe the above answers this too?

> * `balancer_v2` model uses additional `bpt_prices` spell, which if I understand it correctly - backfills usd token prices calculated based on swaps (if token not present in `prices.usd` feed). I'd assume this additional logic adds quite a bit of an overhead and might even not be necessary if all token prices are available in `prices.usd` table (needs checking). Thoughts?

i think we should leave this join to `ref(bpt_prices)` in the view downstream as well, similar to above

tomfutago:
> > * current `balancer_v1_ethereum_trades` and `balancer_v2_ethereum_trades` spells have `pool_id` and `swap_fee` columns added to the output, which are not added to the current `dex.trades` but for backward compatibility they need adding to exposed balancer spells - I'm assuming that's the way to go?
> 
> let's add those column to the base trades macro(s), then leave out of `dex.trades_beta` for now. we can then build the view on top of `dex.trades_beta` which could join back to base spell and output those columns. does that seem like a reasonable approach?

> > * `balancer_v2` model uses additional `bpt_prices` spell, which if I understand it correctly - backfills usd token prices calculated based on swaps (if token not present in `prices.usd` feed). I'd assume this additional logic adds quite a bit of an overhead and might even not be necessary if all token prices are available in `prices.usd` table (needs checking). Thoughts?
> 
> i think we should leave this join to `ref(bpt_prices)` in the view downstream as well, similar to above


I went with your suggestions and it’s all ready for testing.

Notes:
* Location of exposed views: for now I added “exposed” folder to follow: `<chain>/<platforms>/exposed` structure
* Macro links to `pools_fees` models, and those together with `bpt_prices` linked in exposed views are still in old “balancer” folders (so maybe exposed views should go to old location as well?)
* Exposed views link back to `base_trades` to pull `pool_id` and `swap_fee` columns and also to `tokens_erc20` and `bpt_prices` to follow the existing spells logic that extends `amount_usd` calculation. I wanted to check how many records are benefiting from this additional logic and attempted to force run main `dex_base_trades` and `dex_trades_beta` but seems like adding line breaks is not enough to force them run..

jeff-dude:
> Notes:
> 
> * Location of exposed views: for now I added “exposed” folder to follow: `<chain>/<platforms>/exposed` structure
> * Macro links to `pools_fees` models, and those together with `bpt_prices` linked in exposed views are still in old “balancer” folders (so maybe exposed views should go to old location as well?)
> * Exposed views link back to `base_trades` to pull `pool_id` and `swap_fee` columns and also to `tokens_erc20` and `bpt_prices` to follow the existing spells logic that extends `amount_usd` calculation. I wanted to check how many records are benefiting from this additional logic and attempted to force run main `dex_base_trades` and `dex_trades_beta` but seems like adding line breaks is not enough to force them run..

i wonder if we should simplify this PR
- only include the standard steps to get into `dex_trades_beta`, i.e. macro(s), base trades spell(s), etc.
- remove all the "exposed" balancer spells, follow up PR which will handle those (we can tag balancer team for help too)

are you good with moving some code around to do that in separate PRs? or will that break something i'm not thinking of?

tomfutago:
> i wonder if we should simplify this PR
> 
> * only include the standard steps to get into `dex_trades_beta`, i.e. macro(s), base trades spell(s), etc.
> * remove all the "exposed" balancer spells, follow up PR which will handle those (we can tag balancer team for help too)
> 
> are you good with moving some code around to do that in separate PRs? or will that break something i'm not thinking of?

Exposed views can be moved to a separate follow up PR - no problem. Macro will still contain `pools_fees` model and `base_trades` will have additional `pool_id` and `swap_fee` columns, which seems unavoidable for backward compatibility. 

Will go ahead and remove exposed views from this PR.

jeff-dude (approved):
lgtm!

when you fix merge conflicts, be sure to update to latest seed test approach from @Hosuke here:
https://github.com/duneanalytics/spellbook/pull/4966/files

PR #4638, opened 2023-10-19
Create `dex.atomic_arbitrages`

hildobby:
# Thank you for contributing to Spellbook!
Please refer to the top of the `readme` in the root of Spellbook to learn how to contribute to Spellbook on DuneSQL.

PR #4943, opened 2023-12-04
`dex.sandwiches` optimisations & updates

hildobby:


jeff-dude (approved):
love it -- thank you!

approving this with the only caveat being to wait until you find out on #4638 if having `dex.trades` with the filters in simple first CTE really helps performance, rather than reading from it multiple times directly. i'll wait to see that outcome and we can either edit this PR too or if no gains, we'll merge as-is

PR #4951, opened 2023-12-05
Contracts Fix: Iterated Contracts Check

MSilb7:
Fixing issue raise in https://github.com/duneanalytics/spellbook/pull/4950

- Add a check that the incremental contracts don't pull in contracts that were reinitialized (likely self-destruct)
  - **I think this is the big fix**

- As an additional check, added a row_number filter at the end to make sure we're only pulling each contract once.


MSilb7:
closing, messed up brances

PR #4961, opened 2023-12-06
Add Zora to ETH Transfers & make it a Macro

MSilb7:
Adding Zora support for the eth transfers spell (intends to mimic erc20_evt_transfer).

Also turns the spell in to a macro (which handles for legacy OP Mainnet bridged eth transfers). This makes it easier to extend to every chain.


aalan3:
Zora was added to the new token transfers table here https://github.com/duneanalytics/spellbook/pull/4916, but might be missing some of you logic, could you please contribute to that one instead?

jeff-dude:
closing due to alan comment. we need to make these new transfers spells public / communicate out their existence to raise awareness.

PR #4962, opened 2023-12-06
Add Calldata to Rollup Economics L1 Data (and create function)

MSilb7:
1. Creates a function for calculating calldata gas given a data bytestring
2. Adds calldata gas used for L1 transactions to the Rollup Economics spell (used to determine how much gas comes from user transaction data versus protocol execution/data).
3. Adds block number as another useful field.

cc @Jam516 for vis.


jeff-dude (approved):
nice wizardry 🪄

PR #4966, opened 2023-12-08
Add check_base_trades_seed generic test

Hosuke:
This test may simplify the repeated contents in `.yml` files.


tests/generic/check_base_trades_seed.sql
```diff
@@ -0,0 +1,10 @@
+-- this tests checks a dex base_trades model for every row in a seed file.
+-- actual implementation in macros/test-helpers/check_seed.sql
+{% test check_base_trades_seed(model, seed_file, filter=None) %}
```
jeff-dude, line 3: i like the idea here

can we be more specific on naming?
`check_dex_base_trades_seed` (add `dex` in there to help differentiate future sectors that follow)

jeff-dude (changes requested):

tests/generic/check_base_trades_seed.sql
Hosuke: I just renamed the test to `check_dex_base_trades_seed`

jeff-dude (approved):
works for me

PR #4971, opened 2023-12-10
Fix "interval" syntax in SQL query example in Contract Mapping

cainky:
Background
- Coming from the Advent of Crypto Data page, I am directed to this README. I ran the example but an error was encountered related to the usage of the interval keyword.

Changes
- Updated the README to reflect the correct syntax for using interval in SQL queries

Referenced the Dune docs for the correct interval syntax: [Dune SQL Documentation on Date and Time Operators](https://dune.com/docs/query/DuneSQL-reference/Functions-and-operators/datetime/#date-and-time-operators).

After updating the syntax, I reran the query with the fix, and it executed successfully


cainky:

jeff-dude (approved):
good catch!
and welcome to spellbook, glad to see ya make it here.

fyi @MSilb7

PR #4976, opened 2023-12-11
1inch: LOP + AR macros updates (1/2; all except ethereum&polygon)

grkhr:
first PR of our lineage refactoring. main goal is simplify tables and optimize workload.

- LOP add new columns
- AR add new columns


jeff-dude (approved):
works for me!

PR #4978, opened 2023-12-12
Upgrade check_seed_macro to support multiple values filters.

Hosuke:
This new macro will compile
```
          filter:
            version:
              - 2
              - mmpool
              - stableswap
```
into
```
    AND seed.version IN ('2', 'mmpool', 'stableswap')
```


jeff-dude (approved):
thank you!

PR #4986, opened 2023-12-13
Contracts - Test Incremental Initial Build & Add Polygon PoS

MSilb7:
Add Polygon PoS to contracts mapping

Edit:
- Testing incremental initial builds, where we just run forward 1 year at a time on the initial build.
- This means it should take 1 hour * # of years to catch up (i.e. ~8 for Ethereum), but this should only matter when we edit the base and incremental tables, which ~should be rare.

Also incorporates https://github.com/duneanalytics/spellbook/pull/5026

---

Merging https://github.com/duneanalytics/spellbook/pull/5128 should be a fast follow after merging this


jeff-dude:
polygon is going to be tough to fit into CI test window :(

MSilb7:
> polygon is going to be tough to fit into CI test window :(

Maybe we can try the similar "look forward up to this many blocks" at the base level to try to chunk it.

MSilb7:
The yearly incremental predicate here now matches what's in https://github.com/duneanalytics/spellbook/pull/4917

(It should look unchanged to whatever is merged second).

Should be ready for review @jeff-dude

MSilb7:
Modifying the forward facing window for runtime

jeff-dude (approved):
excited to see how this goes in prod

assuming it runs as expected, we should check back tomorrow to ensure run times in prod decrease after the number of hourly runs needed to catchup to present time

PR #4987, opened 2023-12-13
add crosschain `tokens.transfers`

jeff-dude:
looking to make the new token transfers spells crosschain & expose on the data explorer. we should start to consider communicating these out as new transfers spells and ask for feedback.


jeff-dude:
testing CI with a smaller date range, we can revert back when/if completed

Hosuke (approved):

models/_sector/tokens/tokens_transfers.sql
```diff
@@ -0,0 +1,75 @@
...
+    unique_key = ['unique_key', 'blockchain'],
+    post_hook='{{ expose_spells(\'["arbitrum", "avalanche_c", "base", "bnb", "celo", "ethereum", "fantom", "gnosis", "optimism", "polygon", "zksync", "zora"]\',
+                                "sector",
+                                "tokens",
+                                \'["aalan3", "jeff-dude"]\') }}'
+)
+}}
+
+{% 
+    set blockchains = [
```
aalan3, line 18: we should refer to the actual models here it makes jumping to the model easier.

models/_sector/tokens/tokens_transfers.sql
```diff
@@ -0,0 +1,75 @@
+{{config(
+    schema = 'tokens',
+    alias = 'transfers',
+    partition_by = ['token_standard', 'block_date', 'blockchain'],
```
aalan3, line 4: I think the order here matters and block_date should be last. I also think maybe we should drop `blockchain` here. If you are querying this table I'm assuming you'd want all or most chains.

models/_sector/tokens/tokens_transfers.sql
jeff-dude: can you clarify on "jumping to model easier"?

aalan3: in vs code for example you can command click the model and go directly to the definition, it's quite nice.

models/_sector/tokens/tokens_transfers.sql
jeff-dude: interesting on the order, didn't realize that. i can reorder.

i think i'd push back on removing `blockchain`, as i think most queries will either filter or group by on `blockchain` on this crosschain spell. isn't that the best use case for adding a partition?

models/_sector/tokens/tokens_transfers.sql
jeff-dude: TIL 🤯 

i can go ahead and make that change. i like the look of this current approach, but can see the benefit of changing outweighing this

models/_sector/tokens/tokens_transfers.sql
aalan3: Yeah maybe you are right, we can keep it then

PR #5002, opened 2023-12-15
1inch: `call_transfers` refactoring, `calls` migration

grkhr:
Hey!

Second part of our lineage refactoring.

- `oneinch_<blockchain>.call_transfers`. old `calls_transfers` will be removed soon.
- got rid of `evms.info`, migrated on our own metadata table
- `oneinch.calls` migration from `calls_transfers` to `ar`+`lop` 
- actualized `all_evm_chains()`
- `oneinch.parsed_transfers_from_calls` macro + view for internal usage


models/oneinch/_meta/oneinch_blockchains.sql
jeff-dude, line 1: i notice this isn't used downstream anywhere. do you plan to use in queries on the app?

jeff-dude (approved):
only one minor comment below, not really a blocker.

the runtimes are looking great too, even with so many chains! great stuff 🙏

models/oneinch/_meta/oneinch_blockchains.sql
grkhr: yes, it's for the app, just in case. not sure we'll use it much, but we'll see

PR #5009, opened 2023-12-16
lido.liquidity adding new pools

ppclunghe:
Adding of Balancer wstETH:wUSDM and Solidly pools to lido.liquidity spell


jeff-dude (approved):
sounds good to me!

PR #5016, opened 2023-12-17
Add arbitrum to aragon dao spells and update dao.transactions - Ready for review

henrystats:


henrystats:
Hola ser @jeff-dude - I saw you added the WIP label so just wanted to tag you that this was already ready for review. I forgot to update the naming.

It's a pretty straight forward update and no changes to code structure, I just added a new chain and removed the highlighted comment where dex prices was commented out as it hadn't been migrated.

models/aragon/arbitrum/aragon_arbitrum_schema.yml
```diff
@@ -0,0 +1,68 @@
...
+    meta:
+      blockchain: arbitrum
+      project: aragon
+      contributors: Henrystats 
+    config:
+      tags: [ 'arbitrum', 'dao', 'aragon', 'dao creator tools', 'Henrystats', 'Aragon app']
+    description: >
+      wallet addresses of daos created on aragon app
+    tests:
+      - dbt_utils.unique_combination_of_columns:
```
jeff-dude, line 14: the test columns don't match the model columns, plz update one or the other to match and ensure uniquness

models/dao/addresses/arbitrum/dao_addresses_arbitrum_schema.yml
```diff
@@ -0,0 +1,34 @@
...
+    description: >
+      wallet addresses of daos created on arbitrum across all dao creator tools
+    columns:
+      - &blockchain
+        name: blockchain
+        description: "blockchain which the DAO is deployed"
+      - &dao_creator_tool
+        name: dao_creator_tool
+        description: "Name of creator tool used to create DAO"
+        tests:
```
jeff-dude, line 20: i don't think we need the seed test here, right?
i believe this seed test is already running in the upstream spell

jeff-dude (changes requested):
in addition to comments below, here are some general feedback thoughts. considering there are other spells already live in this lineage, i won't consider this list blockers for this PR. it would be nice over time to rework the entire lineage though to clean it up:
- add incremental predicate in config block to filter target rows to match source rows, example [here](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/trades/arbitrum/platforms/airswap_arbitrum_base_trades.sql#L9), where you can replace column name as needed to a date/time field in the target
- update query incremental filters to use incremental predicate, this keeps source and target filters in sync and controlled in one location, example [here](https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/dex/uniswap_compatible_trades.sql#L31)
- instead of using `end if`, `if` we can simplify to just `else`
- move schema property directly into model config block, keep project file clean (universal cleanup in the future) and easier to find schema/name when searching repo
- potentially remove joins to tokens/prices/dex_prices/etc to the cross-chain spells only, to minimize impact on how often we run modified full refreshes on this lineage. this change would assume it's okay that chain-specific spells have raw amounts from the chain only, then the final cross-chain can perform transformations to build additional columns

henrystats:
Thank you ser @jeff-dude, I've fixed the seed test and the unique combination of columns.

I'll submit a separate pr later to clean up/rework the entire lineage. So this can be reviewed. 

Seems there's a big spell running dbt test right now hence why the test is failing, will re trigger in a bit.

jeff-dude:
> Thank you ser @jeff-dude, I've fixed the seed test and the unique combination of columns.
> 
> I'll submit a separate pr later to clean up/rework the entire lineage. So this can be reviewed.
> 
> Seems there's a big spell running dbt test right now hence why the test is failing, will re trigger in a bit.

yes, that's likely me, my fault :) should be good now. i won't be running again tonight.

jeff-dude (approved):

PR #5023, opened 2023-12-18
[Contracts] Materialize Chain-Level Tables

MSilb7:
Queries hitting `contracts.contract_mapping` tables are often slow, because each chain-level table is a view (and the contracts table is a view unioning these views together). It takes ~4 mins to query from these tables.

Attempting to speed this up by materializing at the chain-level, so then `contracts.contract_mapping` is a view that unions materialized models. (does it make sense to also materialize at the cross-chain level?)

Follow-up:
- If materializing makes sense, then we can also add partitions


jeff-dude:
> (does it make sense to also materialize at the cross-chain level?)

let's try to keep as a view for now, then see how it performs. we materialize dex/nft trades for performance on the app, so we're willing to take that approach elsewhere, but we should test it out first

PR #5025, opened 2023-12-18
initiate cross chain transfers

jeff-dude:
approach summary:
- reorder columns in `transfers_base` macro
- update `transfers_enrich` macro to mimic dex enrich macro
- create a cross-chain `base_transfers` spell which unions all chains incrementally
- remove enrich from all chain-specific spells, move to enrich only cross-chain `base_transfers`
- add incremental filters in enrich macro to ideally increase performance
- update chain-specific transfers to views with logic to read from final transfers table with filter on `blockchain`
  - plz note the base chain difference in naming standard, to avoid clash with `transfers_base_transfers` cross-chain spell


jeff-dude:
it's still giving CI a hard time. might need to try in prod again, like last time, if we think this is the approach we want to take

models/_sector/tokens/tokens_base_transfers.sql
```diff
@@ -0,0 +1,65 @@
...
+    , ref('tokens_ethereum_base_transfers')
+    , ref('tokens_fantom_base_transfers')
+    , ref('tokens_gnosis_base_transfers')
+    , ref('tokens_optimism_base_transfers')
+    , ref('tokens_polygon_base_transfers')
+    , ref('tokens_zksync_base_transfers')
+    , ref('tokens_zora_base_transfers')
+] %}
+
+with base_union as (
```
aalan3, line 28: why do we need the base union here, can't we just have the union loop below?

jeff-dude: we can, yes, but i was matching the dex/nft design pattern. the difference being we apply a duplicate check on those two sectors, to help prevent loading in duplicates if a project upstream writes dupes then fails later. i left out the duplicates check here to ensure we can get it to run, but it'd be nice to have the structure in place in case we add later

aalan3: We should drop the duplicate check there as well, that should be handled by merge intos and in each model.

jeff-dude: the issue is on historical runs, whether full refresh due to token adds/other upstream changes or edit to model itself. sometimes we've hit duplicates on historical run when joining to prices, we get dupes. or any other bug that could have been introduced over time. this won't be handled in merge statements, as its simply an insert

jeff-dude: it's relatively rare to happen, but it's happened more than once in both dex/nft, so rob had ended up adding the dupe check so we don't have to fix the main sector spell

PR #5026, opened 2023-12-18
[Contracts] Rename deployer_address to trace_deployer_address

MSilb7:
Rename deployer_address to trace_deployer_address to avoid confusion over which column should be used to get the contract creator: `creator_address`


jeff-dude (approved):
works for me.

should we wait to merge this when we're ready for the polygon one at same time? we could merge this, merge main back into polygon branch, then merge that one after?

this all assumes we get to where we need to be on the polygon one too. i took a peek at it, but noticed incremental was taking like 4-5x the time of initial history. didn't dig much into the why yet.

PR #5034, opened 2023-12-20
update ccip requests

AnonJon:
Updating the ccip fulfilled/reverted tx queries and the daily request query to properly collect requests.

### DIFF
--- a/models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
+++ b/models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
@@ -20,11 +20,12 @@ WITH
       ccip_send_traces."from" as "node_address"
     FROM
       {{ ref('chainlink_arbitrum_ccip_send_traces') }} ccip_send_traces
+      WHERE
+        ccip_send_traces.tx_success = true
       {% if is_incremental() %}
         AND ccip_send_traces.block_time >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
       {% endif %}
-      WHERE
-        ccip_send_traces.tx_success = true  
+        
   )
 SELECT
  'arbitrum' as blockchain,
--- a/models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
+++ b/models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
@@ -14,21 +14,17 @@
 WITH
   ccip_fulfilled_transactions AS (
     SELECT
-      tx.hash as tx_hash,
-      tx.block_time as block_time,
-      cast(date_trunc('day', tx.block_time) as date) as date_start,
-      tx."from" as "node_address"
+      ccip_send_traces.tx_hash as tx_hash,
+      ccip_send_traces.block_time as block_time,
+      cast(date_trunc('day', ccip_send_traces.block_time) as date) as date_start,
+      ccip_send_traces."from" as "node_address"
     FROM
       {{ ref('chainlink_arbitrum_ccip_send_traces') }} ccip_send_traces
-      INNER JOIN {{ source('arbitrum', 'transactions') }} tx ON tx.hash = ccip_send_traces.tx_hash
       {% if is_incremental() %}
         AND ccip_send_traces.block_time >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
       {% endif %}
       WHERE
-        ccip_send_traces.tx_success = true
-        {% if is_incremental() %}
-          AND tx.block_time >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
-        {% endif %}     
+        ccip_send_traces.tx_success = true  
   )
 SELECT
  'arbitrum' as blockchain,
--- a/models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
+++ b/models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
@@ -2,61 +2,39 @@
   config(
     
     alias='ccip_fulfilled_transactions',
-    partition_by=['date_month'],
+    partition_by=['date_start'],
     materialized='incremental',
     file_format='delta',
-    incremental_strategy='merge',
-    unique_key=['tx_hash', 'tx_index', 'node_address']
+    incremental_strategy='merge'
   )
 }}
 
 {% set incremental_interval = '7' %}
 
 WITH
-  arbitrum_usd AS (
-    SELECT
-      minute as block_time,
-      price as usd_amount
-    FROM
-      {{ source('prices', 'usd') }} price
-    WHERE
-      symbol = 'ETH'
-      {% if is_incremental() %}
-        AND minute >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
-      {% endif %}      
-  ),
   ccip_fulfilled_transactions AS (
     SELECT
       tx.hash as tx_hash,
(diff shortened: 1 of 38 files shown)


models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
```diff
@@ -2,61 +2,36 @@
   config(
     
     alias='ccip_fulfilled_transactions',
-    partition_by=['date_month'],
+    partition_by=['date_start'],
     materialized='incremental',
     file_format='delta',
-    incremental_strategy='merge',
-    unique_key=['tx_hash', 'tx_index', 'node_address']
```
jeff-dude, line 9: why are we removing the unique keys and the corresponding test in the schema file?

if a model is materialized incrementally and set to merge, we need the unique keys to understand how the merge statement is written. without these key columns, wouldn't it just insert over and over? did you query the output CI tables to test the data?

models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
```diff
@@ -2,61 +2,36 @@
   config(
     
     alias='ccip_fulfilled_transactions',
-    partition_by=['date_month'],
+    partition_by=['date_start'],
```
jeff-dude, line 5: is the dataset large enough to justify making partitions much more granular (day vs. month)?

models/chainlink/arbitrum/chainlink_arbitrum_ccip_request_daily.sql
```diff
@@ -10,44 +10,53 @@
...
+                WHERE
+                    rev.date_start = date_series.date_start
+            ) AS reverted_requests
+        FROM
+            (
+                SELECT
+                    date_start
+                FROM
+            (
+                SELECT
```
jeff-dude, line 39: since this model is built incrementally on `date_start`, can't we apply incremental filters on both the sequence and the refs above?

models/chainlink/arbitrum/chainlink_arbitrum_ccip_reverted_transactions.sql
```diff
@@ -2,60 +2,35 @@
   config(
     
     alias='ccip_reverted_transactions',
-    partition_by=['date_month'],
+    partition_by=['date_start'],
```
jeff-dude, line 5: same question here as fulfilled

models/chainlink/arbitrum/chainlink_arbitrum_ccip_reverted_transactions.sql
```diff
@@ -2,60 +2,35 @@
...
     materialized='incremental',
     file_format='delta',
-    incremental_strategy='merge',
-    unique_key=['tx_hash', 'node_address']
```
jeff-dude, line 9: same question here as fulfilled

jeff-dude (changes requested):
left a few comments on arbitrum, but as always, should apply to all chains

models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
AnonJon: The reasoning for the deletion of unique keys and tests is looking into how we are doing ccip fulfilled tx's there was a lot of un needed data being joined and passed. Unlike the other fulfilled transaction spells for the other chainlink products this one is now taking the traces instead of logs which then causes issues for unique tests since there could/will be rows that have the same tx_hash, node_address etc since there could be multiple of the same trace in the same tx. This was found after testing and bringing back incorrect data in how it was set before. I have done some testing on this PR and didnt find it inserting over and over. I suppose a question to you would be in this case is this the correct way to label it?

models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
AnonJon: this was done only since I have taken out date_month because it was relevant anymore but if it makes sense to add it back then I can do so. Its not terribly large.

models/chainlink/arbitrum/chainlink_arbitrum_ccip_request_daily.sql
AnonJon: yeah ill add the incremental on those

models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
jeff-dude: removing partition altogether is fine too, if you have no need for monthly column and dataset isn't massive

models/chainlink/arbitrum/chainlink_arbitrum_ccip_fulfilled_transactions.sql
jeff-dude: i think you have a few options:
1. add more keys to unique key list (can you add `trace_address` to help differentiate within tx?)
2. change to append-only strategy, but then we need to consider how to avoid writing data multiple times (for instance, incremental runs reading back 7 days would load those 7 days over and over)
3. change to table materialization, where a full refresh happens each time, rather than looking to keep building incrementally

AnonJon: just looked into the first option and I think this may be the best way. Ill add this in and add the tests back for this as well. Thanks for the recommendations.

AnonJon:
Everything seems to be correct outside of the `ccip_request_daily` queries now are showing 0 for everything. Was this what you had mentioned to do on the increment?
`WITH
ethereum_agg AS (
        SELECT
            date_start,
            (
                SELECT
                    COUNT(*)
                FROM
                    {{ ref('chainlink_arbitrum_ccip_fulfilled_transactions') }} eth
                WHERE
                    eth.date_start = date_series.date_start
                {% if is_incremental() %}
                    AND eth.block_time >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
                {% endif %}
            ) AS fulfilled_requests,
            (
                SELECT
                    COUNT(*)
                FROM
                    {{ ref('chainlink_arbitrum_ccip_reverted_transactions') }} rev
                WHERE
                    rev.date_start = date_series.date_start
                {% if is_incremental() %}
                    AND rev.block_time >= date_trunc('day', now() - interval '{{incremental_interval}}' day)
                {% endif %}
            ) AS reverted_requests
        FROM
            (
                SELECT
                    date_start
                FROM
            (
                SELECT
                    sequence (
                        CAST('2023-07-06' AS date),
                        date_trunc ('day', cast(now () AS date)),
                        INTERVAL '1' day
                    ) AS date_sequence
            ) AS seq
            CROSS JOIN UNNEST (seq.date_sequence) AS t (date_start)
            ) date_series
    )`
    
    or are you saying to add it into the `date_series`?

jeff-dude:
> or are you saying to add it into the `date_series`?

you would need to filter `date_series` too, so they match the days filtered. which of these is the end goal?
- time series list of days with total count for that particular day only
- time series list of days with total count for all history each day to see trends over time

if first option, then filter `date_series` in addition to refs *should* work.
if second option, then filter `date_series` only should work?

incremental model design gets tough when we work off a time series. i know we have some examples somewhere in spellbook.

AnonJon:
Just ran some queries on the daily requests and everything seems to be working with adding an additional increment on the `date_series`. Thanks again.

jeff-dude (approved):
awesome, glad that worked!

PR #5036, opened 2023-12-21
polygon NFT trades refactor

0xRobin:
polygon NFT trades refactor


jeff-dude (approved):
sgtm -- nice to see some more chains come through 🙏 

side note: i noticed in a few spots, you've updated to use the incremental predicate macro, but a few other spots in here still use the hardcoded 7-day window. i don't think it's a blocker here, as i'd imagine a few older merged PRs also still use the old incremental filter too. we could explore a clean-up PR later.

PR #5039, opened 2023-12-22
Adds child gauges to balancer gauge labels

viniabussafi:
This PR adds the child gauges logic to the balancer gauge spells. It also creates the spells for avalanche, base and gnosis, containing the previously mentioned logic.

@mendesfabio


jeff-dude (approved):
sounds good to me

PR #5047, opened 2023-12-23
Add beethoven_x to `dex.trades_beta`

tomfutago:
linked to #4759


tomfutago:
I have a few questions on that one, maybe original contributor @MSilb7 can help? 🙏

Beethoven_x afaik is Balancer v2 fork, which seems to be reflected here in the setup for Optimism spell: 
`source('balancer_v2_optimism', 'Vault_evt_Swap')`. Possible copy-paste mistake? The same source is used for actual Balancer v2 spell.

There is also this condition:
```
    WHERE t.tokenIn != bytearray_substring(t.poolId, 1, 10)
        AND t.tokenOut != bytearray_substring(t.poolId, 1, 10)
```
which when reversed into:
```
    WHERE t.tokenIn = bytearray_substring(t.poolId, 1, 10)
        OR t.tokenOut = bytearray_substring(t.poolId, 1, 10)
```
doesn't return any records (in other words, current `where` condition doesn't seem to exclude anything).

Also, `poolId` is aliased as `project_contract_address` - not sure why not `contract_address`?

Spell setup for Fantom is identical but uses `beethoven_x_fantom` as source.

Any insights on the above would be greatly appreciated. Any objections to replicating here similar setup as for Balancer with equivalent spell to `balancer_v2_optimism_pools_fees` as part of Beethoven_x spell?

jeff-dude:
> I have a few questions on that one, maybe original contributor @MSilb7 can help? 🙏
> 
> Beethoven_x afaik is Balancer v2 fork, which seems to be reflected here in the setup for Optimism spell: `source('balancer_v2_optimism', 'Vault_evt_Swap')`. Possible copy-paste mistake? The same source is used for actual Balancer v2 spell.
> 
> There is also this condition:
> 
> ```
>     WHERE t.tokenIn != bytearray_substring(t.poolId, 1, 10)
>         AND t.tokenOut != bytearray_substring(t.poolId, 1, 10)
> ```
> 
> which when reversed into:
> 
> ```
>     WHERE t.tokenIn = bytearray_substring(t.poolId, 1, 10)
>         OR t.tokenOut = bytearray_substring(t.poolId, 1, 10)
> ```
> 
> doesn't return any records (in other words, current `where` condition doesn't seem to exclude anything).
> 
> Also, `poolId` is aliased as `project_contract_address` - not sure why not `contract_address`?
> 
> Spell setup for Fantom is identical but uses `beethoven_x_fantom` as source.
> 
> Any insights on the above would be greatly appreciated. Any objections to replicating here similar setup as for Balancer with equivalent spell to `balancer_v2_optimism_pools_fees` as part of Beethoven_x spell?

thank you for the details. i also recall working with balancer team on beethoven question recently. maybe @viniabussafi @thetroyharris can help here?

also tagging @MSilb7 one more time :)

MSilb7:
> > I have a few questions on that one, maybe original contributor @MSilb7 can help? 🙏
> > Beethoven_x afaik is Balancer v2 fork, which seems to be reflected here in the setup for Optimism spell: `source('balancer_v2_optimism', 'Vault_evt_Swap')`. Possible copy-paste mistake? The same source is used for actual Balancer v2 spell.
> > There is also this condition:
> > ```
> >     WHERE t.tokenIn != bytearray_substring(t.poolId, 1, 10)
> >         AND t.tokenOut != bytearray_substring(t.poolId, 1, 10)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which when reversed into:
> > ```
> >     WHERE t.tokenIn = bytearray_substring(t.poolId, 1, 10)
> >         OR t.tokenOut = bytearray_substring(t.poolId, 1, 10)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > doesn't return any records (in other words, current `where` condition doesn't seem to exclude anything).
> > Also, `poolId` is aliased as `project_contract_address` - not sure why not `contract_address`?
> > Spell setup for Fantom is identical but uses `beethoven_x_fantom` as source.
> > Any insights on the above would be greatly appreciated. Any objections to replicating here similar setup as for Balancer with equivalent spell to `balancer_v2_optimism_pools_fees` as part of Beethoven_x spell?
> 
> thank you for the details. i also recall working with balancer team on beethoven question recently. maybe @viniabussafi @thetroyharris can help here?
> 
> also tagging @MSilb7 one more time :)

Hey sorry, I don't have any memory on = vs != conditions here (I'd recommend looking at sample trade transactions).

Beethoven X is the user interface that managed Balancer's deployment on OP Mainnet. So if there is both Beethoven and Balancer, is it duplicated. Open to naming whatever makes sense here, I seem to remember we had this conversation a year ago in some other PR.

tomfutago:
> Hey sorry, I don't have any memory on = vs != conditions here (I'd recommend looking at sample trade transactions).
> 
> Beethoven X is the user interface that managed Balancer's deployment on OP Mainnet. "Balancer" as an app interface does not exist for OP Mainnet, it is Beethoven X. https://medium.com/balancer-protocol/balancer-is-now-on-optimism-671d55e489b
> 
> So if there is both Beethoven and Balancer in OP dex trades, is it duplicated. Open to naming whatever makes sense here, I seem to remember we had this conversation a year ago in some other PR.

Thanks @MSilb7 that helps! So in my mind the cleanest solution would be to move all `balancer_v2_optimism_<...>` spells to `beethoven_x_optimism_<...>`, as keeping both is definitely confusing while technically it is Beethoven X. Would that work for all of you guys? @MSilb7 @viniabussafi @thetroyharris

viniabussafi:
tagging @mendesfabio for review

mendesfabio:
hey guys! we had this discussion before and concluded that volume should be reported as BeethovenX on Fantom and Balancer on Optimism - folks from BeethovenX agree with this. FYI Balancer also has its own Optimism interface now and would be impossible to distinguish which trades are coming from https://app.balancer.fi/#/optimism or https://op.beets.fi/ since everything interacts with the same contract (the canonical Balancer Vault) and no additional metadata is passed through transactions to allow such distinction.

mendesfabio:
@tomfutago the condition you mentioned that's not returning any results looks wrong. FYI on Balancer the `poolId` is composed of pool address + some random bytes. 

What that line is trying to remove is `Swap` events where either the `tokenIn` or the `tokenOut` is the pool LP token - those are not actually swaps but joins/exits and thus should be removed from `dex.trades` to avoid inflating volume metrics. 

Given the BPT (how we call the LP token) is the first 20 bytes of the `poolId` the clause should look like this:

```
    WHERE t.tokenIn != bytearray_substring(t.poolId, 1, 20)
        AND t.tokenOut != bytearray_substring(t.poolId, 1, 20)
```

tomfutago:
> hey guys! we had this discussion before and concluded that volume should be reported as BeethovenX on Fantom and Balancer on Optimism - folks from BeethovenX agree with this. FYI Balancer also has its own Optimism interface now and would be impossible to distinguish which trades are coming from https://app.balancer.fi/#/optimism or https://op.beets.fi/ since everything interacts with the same contract (the canonical Balancer Vault) and no additional metadata is passed through transactions to allow such distinction.

Great, thanks @mendesfabio for clarification! So even easier solution under this PR then:
- Balancer on Optimism is already migrated to `dex.trades_beta` so it stays as is
- BeethovenX to be removed for Optimism and only migrated for Fantom

If the above is not correct, plmk - in the meantime I'll go with that assumption and go ahead with these changes.

tomfutago:
> > hey guys! we had this discussion before and concluded that volume should be reported as BeethovenX on Fantom and Balancer on Optimism - folks from BeethovenX agree with this. FYI Balancer also has its own Optimism interface now and would be impossible to distinguish which trades are coming from https://app.balancer.fi/#/optimism or https://op.beets.fi/ since everything interacts with the same contract (the canonical Balancer Vault) and no additional metadata is passed through transactions to allow such distinction.
> 
> Great, thanks @mendesfabio for clarification! So even easier solution under this PR then:
> 
> * Balancer on Optimism is already migrated to `dex.trades_beta` so it stays as is
> * BeethovenX to be removed for Optimism and only migrated for Fantom
> 
> If the above is not correct, plmk - in the meantime I'll go with that assumption and go ahead with these changes.

Done as per above.

+additional note: current `beethoven_x_fantom_trades` spell contains this logic:
```
        ,t.amountIn AS token_bought_amount_raw
        ,t.amountOut AS token_sold_amount_raw
        ,t.tokenIn AS token_bought_address
        ,t.tokenOut AS token_sold_address
```
while `balancer_v2_optimism_trades`
```
            swap.amountOut AS token_bought_amount_raw,
            swap.amountIn AS token_sold_amount_raw,
            swap.tokenOut AS token_bought_address,
            swap.tokenIn AS token_sold_address,
```
so (again as in elsewhere PRs) it looks like amount/token In/Out are flipped around - I assume `balancer_v2_optimism_trades` is correct - @mendesfabio @viniabussafi can you guys confirm plz?

Hosuke (approved):
LGTM.✅ for beethoven_x_fantom_base_trades
Thank you @tomfutago 

[Sample beethoven_x_fantom_base_trades](https://dune.com/queries/3347134)

jeff-dude:
> Done as per above.
> 
> +additional note: current `beethoven_x_fantom_trades` spell contains this logic:
> 
> ```
>         ,t.amountIn AS token_bought_amount_raw
>         ,t.amountOut AS token_sold_amount_raw
>         ,t.tokenIn AS token_bought_address
>         ,t.tokenOut AS token_sold_address
> ```
> 
> while `balancer_v2_optimism_trades`
> 
> ```
>             swap.amountOut AS token_bought_amount_raw,
>             swap.amountIn AS token_sold_amount_raw,
>             swap.tokenOut AS token_bought_address,
>             swap.tokenIn AS token_sold_address,
> ```
> 
> so (again as in elsewhere PRs) it looks like amount/token In/Out are flipped around - I assume `balancer_v2_optimism_trades` is correct - @mendesfabio @viniabussafi can you guys confirm plz?

will leave this open for a bit, just to be safe on this question

mendesfabio:
yep the `balancer_v2_optimism_trades` logic looks correct @jeff-dude @tomfutago

jeff-dude (approved):
i will assume since both spells in new setup use the same balancer macro, we're good, and the reference to the difference is in the old dex trades lineage.

edit: this may make comparing the two dex spells difficult, if we don't fix the old lineage. we can either override the test query to use case logic, or fix the old lineage. unless we aren't matching amounts exactly, rather row counts, then we should be fine.

tomfutago:
> edit: this may make comparing the two dex spells difficult, if we don't fix the old lineage. we can either override the test query to use case logic, or fix the old lineage. unless we aren't matching amounts exactly, rather row counts, then we should be fine.

Will fix old lineage.

PR #5048, opened 2023-12-23
Revert adding ImmortalX to `dex.trades_beta`

tomfutago:


tomfutago:
My attempt to squeeze ImmortalX to `dex.trades_beta` as vanilla dex wasn't the greatest idea simply because it's actually perpetuals exchange, so I'm working on correcting this mistake now. This PR is to revert addition, will create a separate PR to add ImmortalX to `perpetual_trades`.

Side effect observation: looks like removing dex from `dex.info` model is "not allowed" even with all related spells removed, It fails on relationship check. 

Also, I added here wardenswap to `dex.info`, which should be done in a separate PR, but I hope it can be forgiven this time..

Hosuke (approved):
LGTM.✅
Thank you @tomfutago

PR #5055, opened 2023-12-27
Added spell for Alienswap base nft marketplace

ARDev097:
Hello !
Added trades data for Alienswap nft marketplace on base.


models/_sector/nft/trades/chains/base/nft_base_base_trades.sql
```diff
@@ -0,0 +1,53 @@
...
+        nft_amount,                -- always 1 for erc721
+        price_raw,
+        currency_contract,
+        platform_fee_amount_raw,
+        royalty_fee_amount_raw,
+        platform_fee_address,   -- optional
+        royalty_fee_address,    -- optional
+        sub_tx_trade_id,
+--        tx_from,              -- not yet available in the base event tables
+--        tx_to,                -- not yet available in the base event tables
```
jeff-dude, line 41: don't we have to supply some sort of default value for these fields, if not available at source, in order for downstream models to pull from? otherwise union cross-chain will fail, since other chains supply values?

jeff-dude (changes requested):

models/_sector/nft/trades/chains/base/nft_base_base_trades.sql
```diff
@@ -0,0 +1,53 @@
...
+    {% if not loop.last %}
+    UNION ALL
+    {% endif %}
+    {% endfor %}
+    )
+where duplicates_rank = 1
+)
+
+-- this will be removed once tx_from and tx_to are available in the base event tables
+{{ add_nft_tx_data('base_union', 'base') }}
```
0xRobin, line 53: @jeff-dude 
these columns are provided here with this macro.
But you are right, it's better to do it in the platform models.

I'll quickly adapt it

models/_sector/nft/trades/chains/base/nft_base_base_trades.sql
0xRobin: see: https://github.com/duneanalytics/spellbook/pull/5055/files#r1446493114

models/_sector/nft/trades/chains/base/nft_base_base_trades.sql
0xRobin: I should have expanded more, the Ethereum version is indeed the same, but all other chains follow this structure (provide the tx_from and tx_to columns in the platform model by using the macro). 
So this is the desired setup. (It's on the todo's to do the same setup on ethereum, but CI was not taking it well the last time I tried)

jeff-dude: gotcha. i'm for sure confusing myself. in the new dex lineage, we're still running that macro at the chain level, like [here](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/trades/arbitrum/dex_arbitrum_base_trades.sql). ideally we're doing the same across sectors. now that i'm thinking about it, it ultimately doesn't really matter right now, as the long term goal (not so long term i guess, happening now) is to remove completely and have available in decoded, which would mean pulling at platform level.

tl;dr
we're good :)

0xRobin:
@jeff-dude Should be good now!

@ARDev097  thanks for the contribution, and sorry about hijacking the PR a bit to push this through. 
Note that I had to remove the 'superchain' folder and the model now just lives under a 'base' chain folder.

jeff-dude (approved):

PR #5064, opened 2023-12-27
Celo ImmortalX - add to `perpetual_trades`

tomfutago:


models/dex/dex_info.sql
```diff
@@ -92,6 +92,5 @@ FROM (VALUES
     , ('mento', 'Mento', 'Direct', 'mento')
     , ('mummy_finance', 'Mummy Finance', 'Direct', 'mummyftm')
     , ('carbonhood', 'Carbonhood', 'Direct', 'carbonhood')
-    , ('immortalx', 'ImmortalX', 'Direct', 'immortalx')
```
Hosuke, line 95: Do you think `ImmortalX` can fit into other market type in this table?

Hosuke (approved):
LGTM.✅
Thank you @tomfutago 

[Sample immortalx_celo_perpetual_trades](https://dune.com/queries/3319312)

models/dex/dex_info.sql
tomfutago: There are 3 main market types in this table atm: Direct, Aggregator and Flashloans - I don't think ImmortalX qualifies as any of these.

jeff-dude: i'd imagine @hildobby added `ImmortalX` -- do you have any opinion on it being in here? or since we're adding it to `perpetual.trades`, maybe it doesn't make sense in `dex.info`?

models/immortalx/celo/immortalx_celo_perpetual_trades.sql
```diff
@@ -0,0 +1,94 @@
...
+  pe.trader,
+  pe.volume_raw,
+  pe.leverage,
+  pe.pnl,
+  tx."from" as tx_from,
+  tx.to as tx_to,
+  pe.tx_hash,
+  pe.evt_index
+from perp_events pe 
+  join {{ source('celo', 'transactions') }} tx on pe.tx_hash = tx.hash and pe.block_time = tx.block_time
```
jeff-dude, line 94: even if performance is fine (assuming data volume on celo is much smaller), we should continue the habit of filtering joins to base tables with an incremental filter in the join conditions, to match the filter on decoded

jeff-dude: side note, i haven't looked at `perpertual.trades` lineage in a bit, so maybe it's missing elsewhere too. i think we should still apply moving forward

jeff-dude (changes requested):

models/dex/dex_info.sql
hildobby: we can add new types ofc, maybe changing it to `Derivatives` or `Perpetual` would make more sense?

models/immortalx/celo/immortalx_celo_perpetual_trades.sql
tomfutago: good idea, will add it in

jeff-dude (approved):
thanks for the quick fixes & all for the opinions 🤝

PR #5082, opened 2024-01-01
Added new spell prices_unsupported_chains_non_native_tokens.sql for P…

cryptokoryo:
…rices spell

Context:
https://discord.com/channels/757637422384283659/1187733580676333578/1191001595870183504

### DIFF
--- a/models/prices/prices_unsupported_chains_non_native_tokens.sql
+++ b/models/prices/prices_unsupported_chains_non_native_tokens.sql
@@ -1,0 +1,25 @@
+{{ config(
+        schema='prices_others',
+        alias = 'tokens',
+        materialized='table',
+        file_format = 'delta',
+        tags = ['static']
+        )
+}}
+SELECT 
+    token_id
+    , blockchain
+    , symbol
+    , contract_address
+    , decimals
+FROM
+(
+    VALUES
+    ('btcs-ordinals', 'brc-20', 'BTCS', edc052335f914ee47a758cff988494fbb569d820e66ac8581008e44b26dcdb43i0, null),
+    ('csas-ordinals', 'brc-20', 'CSAS', 08fc8bc813521635dd1471917a8d4a91749573df5454bac3b6f868e506347156i0, null),
+    ('mice-ordinals', 'brc-20', 'MICE', 42dd980ad18bc5b57bb6900377b65e27cb2d7a9d5c1b993347d84c62db0dd80ei0, null),
+    ('ordi-ordinals', 'brc-20', 'ORDI', b61b0172d95e266c18aea0c624db987e971a5d6d4ebc2aaed85da4642d635735i0, null),
+    ('rats-ordinals', 'brc-20', 'RATS', 77df24c9f1bd1c6a606eb12eeae3e2a2db40774d54b839b5ae11f438353ddf47i0, null),
+    ('sats-ordinals', 'brc-20', 'SATS', 9b664bdd6f5ed80d8d88957b63364c41f3ad4efb8eee11366aa16435974d9333i0, null),
+    ('snek-snek', 'cardano', 'SNEK', 279c909f348e533da5808898f87f9a14bb2c3dfbbacccd631d927a3f, null)
+) as temp (token_id, blockchain, symbol, contract_address, decimals)


models/prices/prices_unsupported_chains_non_native_tokens.sql
```diff
@@ -0,0 +1,25 @@
...
+FROM
+(
+    VALUES
+    ('btcs-ordinals', 'brc-20', 'BTCS', edc052335f914ee47a758cff988494fbb569d820e66ac8581008e44b26dcdb43i0, null),
+    ('csas-ordinals', 'brc-20', 'CSAS', 08fc8bc813521635dd1471917a8d4a91749573df5454bac3b6f868e506347156i0, null),
+    ('mice-ordinals', 'brc-20', 'MICE', 42dd980ad18bc5b57bb6900377b65e27cb2d7a9d5c1b993347d84c62db0dd80ei0, null),
+    ('ordi-ordinals', 'brc-20', 'ORDI', b61b0172d95e266c18aea0c624db987e971a5d6d4ebc2aaed85da4642d635735i0, null),
+    ('rats-ordinals', 'brc-20', 'RATS', 77df24c9f1bd1c6a606eb12eeae3e2a2db40774d54b839b5ae11f438353ddf47i0, null),
+    ('sats-ordinals', 'brc-20', 'SATS', 9b664bdd6f5ed80d8d88957b63364c41f3ad4efb8eee11366aa16435974d9333i0, null),
+    ('snek-snek', 'cardano', 'SNEK', 279c909f348e533da5808898f87f9a14bb2c3dfbbacccd631d927a3f, null)
```
Hosuke, line 24: Happy new year @cryptokoryo !

Since the `brc-20` tokens' address is not varbinary typed (not starting with `0x`), we may still use the varchar strings to represent them.

Also for these new spells we may need to add their corresponding model definition in `schema.yml` files.

cryptokoryo:
GM @Hosuke 
I updated prices_unsupported_chains_non_native_tokens.sql as you insctructed to store contract addresses in varchar. 
Also I added a yml file for the model definition. This is my first one, so I might have missed stuff. 
Please let me know.
Thanks for your assistance.

models/prices/prices_unsupported_chains_non_native_tokens.sql
```diff
@@ -0,0 +1,25 @@
...
+SELECT 
+    token_id
+    , blockchain
+    , symbol
+    , contract_address
+    , decimals
+FROM
+(
+    VALUES
+    ('btcs-ordinals', 'brc-20', 'BTCS', 'edc052335f914ee47a758cff988494fbb569d820e66ac8581008e44b26dcdb43i0', null),
```
Hosuke, line 18: If the `decimals` value is unknown, you can cast null to a certain type.
```suggestion
    ('btcs-ordinals', 'brc-20', 'BTCS', 'edc052335f914ee47a758cff988494fbb569d820e66ac8581008e44b26dcdb43i0', CAST(NULL AS INT256)),
```

cryptokoryo: GM,
Inspired by prices_native_tokens.sql, I did the decimals cast at line 14 with ", CAST(decimals as int) as decimals".

Hosuke (approved):
LGTM.✅
Thank you @cryptokoryo 

[prices_unsupported_chains_non_native_tokens preview](https://dune.com/queries/3339063)

jeff-dude:
@aalan3 do you foresee any issues with how prices are handled behind-the-scenes for this to be added?

models/prices/prices_unsupported_chains_non_native_tokens.sql
```diff
@@ -0,0 +1,25 @@
...
+SELECT 
+    token_id
+    , blockchain
+    , symbol
+    , contract_address
+    , CAST(decimals as int) as decimals
+FROM
+(
+    VALUES
+    ('btcs-ordinals', 'brc-20', 'BTCS', 'edc052335f914ee47a758cff988494fbb569d820e66ac8581008e44b26dcdb43i0', null),
```
aalan3, line 18: My understanding of ordinals aren't great but aren't they techincally on bitcoin? If so we should just add these in bitcoin.

models/prices/prices_unsupported_chains_non_native_tokens.sql
```diff
@@ -0,0 +1,25 @@
...
+    ('csas-ordinals', 'brc-20', 'CSAS', '08fc8bc813521635dd1471917a8d4a91749573df5454bac3b6f868e506347156i0', null),
+    ('mice-ordinals', 'brc-20', 'MICE', '42dd980ad18bc5b57bb6900377b65e27cb2d7a9d5c1b993347d84c62db0dd80ei0', null),
+    ('ordi-ordinals', 'brc-20', 'ORDI', 'b61b0172d95e266c18aea0c624db987e971a5d6d4ebc2aaed85da4642d635735i0', null),
+    ('rats-ordinals', 'brc-20', 'RATS', '77df24c9f1bd1c6a606eb12eeae3e2a2db40774d54b839b5ae11f438353ddf47i0', null),
+    ('sats-ordinals', 'brc-20', 'SATS', '9b664bdd6f5ed80d8d88957b63364c41f3ad4efb8eee11366aa16435974d9333i0', null),
+    ('snek-snek', 'cardano', 'SNEK', '279c909f348e533da5808898f87f9a14bb2c3dfbbacccd631d927a3f', null)
```
aalan3, line 24: Would rather have this in a new cardano file, even if it's just one.

aalan3 (changes requested):
Had some comments and questions. 

In order for this to work this file (or new files) need to also be included in the main prices.tokens view, otherwise they will not be ingested.

models/prices/prices_unsupported_chains_non_native_tokens.sql
jeff-dude: fwiw, my opinion is that we should follow option 1 here. continue to build directory in `models/prices/<chain>/` for each, as needed. then copy/paste the sql/yaml files from other chains and edit down to the necessary rows, update configs, etc.

aalan3: I agree let's do option 1.

I'll get back to you on the issue as well later this week!

cryptokoryo:
@jeff-dude @aalan3 
Changes made. 
Please check.

models/prices/bitcoin/prices_bitcoin_tokens.sql
```diff
@@ -0,0 +1,30 @@
...
+        file_format = 'delta',
+        tags = ['static']
+        )
+}}
+SELECT 
+    token_id
+    , blockchain
+    , symbol
+    , contract_address
+    , CAST(decimals as int) as decimals
```
jeff-dude, line 14: if `decimals` is always null on bitcoin, you can just do `CAST(null as int) as decimals` here and avoid writing it out each time in list

models/prices/bitcoin/prices_bitcoin_tokens.sql
```diff
@@ -0,0 +1,30 @@
...
+        materialized='table',
+        file_format = 'delta',
+        tags = ['static']
+        )
+}}
+SELECT 
+    token_id
+    , blockchain
+    , symbol
+    , contract_address
```
jeff-dude, line 13: if you plan to union this into `prices_tokens`, which it appears you are, then this field will need casted to varbinary to match (hence the current CI failure in the gh workflow).
`from_base58(contract_address) as contract_address`
similar to the solana one [here](https://github.com/duneanalytics/spellbook/blob/main/models/prices/solana/prices_solana_tokens.sql#L13). the downstream implication means you'll need to cast it back to varchar on queries in order to join to bitcoin base tables

jeff-dude (changes requested):
left comments for bitcoin, but appears it will apply to cardano as well

models/prices/bitcoin/prices_bitcoin_tokens.sql
cryptokoryo: @jeff-dude 
There are few complications here. Let me elaborate. 
------------------
Bitcoin:
Bitcoin doesn't have smart contracts. So we can't really talk about contract_address, what is represented here is not even a bitcoin address but is rather known as 'inscription id'. As an example , here are two explorers showing info about 'SATS' ordinal:

https://ordinalswallet.com/inscription/9b664bdd6f5ed80d8d88957b63364c41f3ad4efb8eee11366aa16435974d9333i0

https://ordiscan.com/inscription/357097

The owner (deployer) of SATS is the bitcoin address 'bc1prtawdt82wfgrujx6d0heu0smxt4yykq440t447wan88csf3mc7csm3ulcn' but this address could be the deployer of many other inscriptions. Therefore, it can't uniquely represent SATS. The best way of representing SATS is by using its inscription ID '9b664bdd6f5ed80d8d88957b63364c41f3ad4efb8eee11366aa16435974d9333i0'. But this id is not a base_58 address, so using from_base58 would fail. 

I'm really not an expert here but I see two possibilities:
1/ Similar to prices_native_tokens spell, we could use null as value for contract_address. This is my preferred solution as these are not contract addresses, and we are only interested in the price of the ticker anyway.

2/ The only from_base.. Dune function  that doesn't fail when using the inscription IDs is from_base64. Although this is not a base 64, it's the only non-breaking available function that could be used. If strictly required to have something, we could use from_base_64 of either the inscrption id or the inscription number (ex: 357097 for SATS). I personally don't like this solution since it's confusing. 

------------------
Cardano:
Cardano uses Base58 for Byron addresses and [Bech32](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki) for Shelley. Bech32 is at most 90 characters long . 
Source: 
https://cips.cardano.org/cip/CIP-0019

Solutions for cardano addresses are similar to bitcoin but with also a third option:
1/ Dune implementing a from_bech32 function 

---------
Given the scope of this spell (getting the price for a ticker), and the complications and nuances brought by current and possible future non-evm standards, it seems to me that just using a null value for the contract address (similar to prices_native_tokens) seems like the easiest option. 
I'm adding a tentative commit accordingly but please let me know if it needs to be updated or we need to take a different approach.

jeff-dude: got it, thanks for the context.

i agree, forcing null address on both works for me. you will need to update each schema file test to be like the native one:
https://github.com/duneanalytics/spellbook/blob/main/models/prices/prices_schema.yml#L38-L42

models/prices/cardano/prices_cardano_tokens.sql
```diff
@@ -0,0 +1,20 @@
...
+        materialized='table',
+        file_format = 'delta',
+        tags = ['static']
+        )
+}}
+SELECT 
+    token_id
+    , blockchain
+    , symbol
+    , contract_address
```
jeff-dude, line 13: i believe you accidentally left this column in, which is the current CI failure. removing this should help proceed.

jeff-dude (changes requested):

jeff-dude (approved):
looks good to me now, thanks for the patience and quick fixes 🤝

PR #4960, opened 2023-12-06
Balancer Liquidity Update

viniabussafi:
# Context
This PR comprehends multiple changes across different files with the intention of shielding our liquidity spells from outlier data on token prices. 
What motivated these changes was a discrepancy on the price for OATH, for the [OATH/WETH 80/20 pool on Optimism](https://op.beets.fi/pool/0xd20f6f1d8a675cdca155cb07b5dc9042c467153f0002000000000000000000bc). Balancer frontend had a TVL of approx. 800k while dune displayed a TVL of more than 12 million on 2023-11-30, as shown on this [query](https://dune.com/queries/3259661). With the changes on this PR, it seems fixed, per this [query](https://dune.com/queries/3260199).
From my understanding, this comes from a difference in prices shown in [coinpaprika](https://coinpaprika.com/coin/oath-oath/) and other providers, like [coinmarketcap](https://coinmarketcap.com/currencies/oath/) on that date. I think this is something that the dune team could communicate to coinpaprika.
To avoid that in the future, I have created a whitelist of pricing assets, due to their higher volume and trustability of data. With this whitelist, we can calculate the liquidity on weighted pools based on the price of these pricing assets and their respective weight.

Also, I have created the `protocol_liquidity_eth` and `pool_liquidity_eth` columns, which will be extremely useful now that ethereum prices have been volatile and this can skew our usd data.

# Main changes made to each file

### balancer_token_whitelist
- Created file with selected tokens, their addresses and respective chains

### balancer_v2_{{chain}}_pools_tokens_weights:
- Added a blockchain column
- removed post_hook to clean the data explorer

### balancer_{{chain}}_pools_tokens_weights:
- Added a blockchain column

### balancer_pools_tokens_weights:
- Created multichain spell, which will be useful with balancer_v3 coming next year

### balancer_v2_{{chain}}_liquidity
- Created macro
- Added `eth_prices` CTE, to be used on final select statement
- Created `weighted_pool_liquidity_estimates` and `weighted_pool_liquidity_estimates_2` CTEs, to calculate the liquidity of pricing assets on weighted pools on the first CTE, while the second CTE avoids double count on pools with multiple pricing assets.
- Added coalesce operands to `pool_liquidity_usd` and `protocol_liquidity_usd`, to also account for the cases where a pool is not weighted or doesn't have a pricing asset
- Added `protocol_liquidity_eth` and `pool_liquidity_eth`

### balancer_liquidity
[...]

PR #5084, opened 2024-01-02
Fix blockchains property for Balancer spells

vegarsti:
The `blockchains` property is currently being set as `["{{blockchain}}"]` instead of the concrete blockchain for these `Balancer` spells:

- balancer_v2_arbitrum.liquidity
- balancer_v2_avalanche_c.liquidity
- balancer_v2_base.liquidity
- balancer_v2_ethereum.liquidity
- balancer_v2_gnosis.liquidity
- balancer_v2_optimism.liquidity
- balancer_v2_polygon.liquidity

This was a regression from #4960.


jeff-dude (approved):
nice find!

were you able to test this? when we run `dbt compile` to build a model into sql code, typically this post-hook doesn't output (especially in dev, it's set to only run in prod)

vegarsti:
I've verified the output of this to be correct (`dbt compile` and looking at the `manifest.json`)

jeff-dude:
> Okay, now it's properly fixed. I've verified the output of this to be correct (`dbt compile` and looking at the `manifest.json`).
> 
> The clue was that we're inside a macro, and we already have access to the `blockchain` variable, so we don't need to escape it with curlies, we just need to use it as a variable. That way the variable is inserted correctly before the macro is expanded.
> 
> To verify, check out this branch and do this (find the area around the Optimism spell and then the final lines with the macro)
> 
> ```
> dbt compile && jq . target/manifest.json > manifest.json && rg --context 65 '"model.spellbook.balancer_v2_optimism_liquidity": \{' manifest.json | tail -n 5
> ```
> 
> outputs
> 
> ```
>           {
>             "sql": "{{ expose_spells('[\"optimism\"]',\n                        \"project\",\n                        \"balancer_v2\",\n                        '[\"stefenon\", \"viniabussafi\"]') }}",
>             "transaction": true,
>             "index": null
>           }
> ```

thank you for the extra effort here -- this is actually v helpful for all of spellbook!

fyi @aalan3 @couralex6 @Hosuke @0xRobin @hildobby @tomfutago @grkhr @MSilb7 -- if we ever use blockchain variable in the post-hook (or any really), this is a viable approach

jeff-dude:
> > Okay, now it's properly fixed. I've verified the output of this to be correct (`dbt compile` and looking at the `manifest.json`).
> > The clue was that we're inside a macro, and we already have access to the `blockchain` variable, so we don't need to escape it with curlies, we just need to use it as a variable. That way the variable is inserted correctly before the macro is expanded.
> > To verify, check out this branch and do this (find the area around the Optimism spell and then the final lines with the macro)
> > ```
> > dbt compile && jq . target/manifest.json > manifest.json && rg --context 65 '"model.spellbook.balancer_v2_optimism_liquidity": \{' manifest.json | tail -n 5
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > outputs
> > ```
> >           {
> >             "sql": "{{ expose_spells('[\"optimism\"]',\n                        \"project\",\n                        \"balancer_v2\",\n                        '[\"stefenon\", \"viniabussafi\"]') }}",
> >             "transaction": true,
> >             "index": null
> >           }
> > ```
> 
> thank you for the extra effort here -- this is actually v helpful for all of spellbook!
> 
> fyi @aalan3 @couralex6 @Hosuke @0xRobin @hildobby @tomfutago @grkhr @MSilb7 -- if we ever use blockchain variable in the post-hook (or any variable), this is a viable approach

PR #5091, opened 2024-01-03
Add polygon chain to nft.mints

covendev:
Adding polygon chain to nft.mints - modified schema in nft native section


models/_sector/nft/mints/native/nft_polygon_native_mints.sql
```diff
@@ -0,0 +1,20 @@
+{{ config(
+        
+        schema = 'nft_polygon',
+        alias = 'native_mints',
+        partition_by = ['block_month'],
+		file_format = 'delta',
+        materialized = 'incremental',
+		incremental_strategy = 'merge',
+        unique_key = ['tx_hash','evt_index','token_id','number_of_items']
```
Hosuke, line 9:
```suggestion
        partition_by = ['block_month'],
        file_format = 'delta',
        materialized = 'incremental',
        incremental_strategy = 'merge',
        unique_key = ['tx_hash','evt_index','token_id','number_of_items']
```

Hosuke:
Seems the current error is due to data overflow:
https://github.com/duneanalytics/spellbook/actions/runs/7413026041/job/20171111513?pr=5091#step:11:28

covendev:
> Seems the current error is due to data overflow: https://github.com/duneanalytics/spellbook/actions/runs/7413026041/job/20171111513?pr=5091#step:11:28

ok what might be the best way to solve it. Sorry it's my first time contributing to this repo so a little confused

Hosuke:
> If you find the location of the overflow errors we might be able to replace the definition with:
> 
> ```
> select 
>    TRY(SUM(..)) as overflow column
> from  ...
> where overflow_column is not null
> ```
> 
> which should filter out the rows that produce the errors (if `uint256` is overflowing, we don't have a way around it as that is the biggest data size we have) .

Syncing the thread here:
https://discord.com/channels/757637422384283659/1192420117481541753/1194148093910859816

covendev:
> > If you find the location of the overflow errors we might be able to replace the definition with:
> > ```
> > select 
> >    TRY(SUM(..)) as overflow column
> > from  ...
> > where overflow_column is not null
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which should filter out the rows that produce the errors (if `uint256` is overflowing, we don't have a way around it as that is the biggest data size we have) .
> 
> Syncing the thread here: https://discord.com/channels/757637422384283659/1192420117481541753/1194148093910859816

The issue is to do with addition overflow and has same overflow error with this snippet as well:
SELECT TRY(SUM(amount)) 
FROM nft.transfers
WHERE blockchain = 'polygon'
HAVING TRY(SUM(amount)) IS NOT NULL
LIMIT 100;
error:"Error: UINT256 addition overflow: 1107083904399100582112 + 115792089237316195423570985008687907853269984665640564039457584007913129639935 [Execution ID: 01HME58799Q3Q17KWQ61HQ2DHN]."
https://dune.com/queries/3360510

0xRobin:
@covendev looks like `try(aggregate_function(..))` does not work the way we would expect as it doesn't catch the errors.
One possible way around this is to write our own `reduce_agg(..)` with a lambda function using `try(..)`
```
select 
    reduce_agg(
        val
        ,0
        ,(acc, val) -> try(acc + val)
        ,(acc1, acc2) -> try(acc1 + acc2)
    ) as safe_sum
```
more info: https://trino.io/docs/current/functions/aggregate.html#lambda-aggregate-functions
Unfortunately `uint256` is not supported out of the box here as  a state type.

@belen-pruvost do you think we could get support for this in the query engine?

0xRobin:
@covendev after some internal discussion it doesn't seem like we have a working solution for this at the moment. 
The engine team will look into implementing a `try_sum(...)` that we can use, but it won't be available in the near future..

I'm sorry about that.

covendev:
> @covendev after some internal discussion it doesn't seem like we have a working solution for this at the moment. The engine team will look into implementing a `try_sum(...)` that we can use, but it won't be available in the near future..
> 
> I'm sorry about that.

ok thanks for the support till now. Please let me know once the fix is done.Alternatively, is it possible to obtain data for nft.mints from any other raw data, I wanted to calculate unique holders of an nft collection using nft.mints. Is it possible to query this directly from erc721_polygon.evt_Transfer?

0xRobin:
> ok thanks for the support till now. Please let me know once the fix is done.Alternatively, is it possible to obtain data for nft.mints from any other raw data, I wanted to calculate unique holders of an nft collection using nft.mints. Is it possible to query this directly from erc721_polygon.evt_Transfer?

Yes you can do something like:
```
select distinct to
from erc721_polygon.evt_Transfer
where evt_contract_address = 0xYOUR_NFT_CONTRACT
and from = 0x000000000000...
```
to get all the minters

OR
```
select 
token_id
,max_by(to, evt_block_number) as holder
from erc721_polygon.evt_Transfer
where evt_contract_address = 0xYOUR_NFT_CONTRACT
group by token_id
```
to get all the current holders.

I will close this PR for now then.

PR #5096, opened 2024-01-04
PRO-140 Rename the view from reservoir.tokens_beta to reservoir.tokens

helanto:
The view `reservoir.tokens` already exists and has a different definition. The definition is defined in the tablelog. We need to override the view using the definition in the spellbook. I will remove the definition in tablelog once this PR is merged.

Is it ok to rename the view from `reservoir.tokens_beta` to `reservoir.tokens`, given that `reservoir.tokens` already exists ?? cc @jeff-dude

### DIFF
--- a/models/datasets/reservoir/ethereum/tokens.sql
+++ b/models/datasets/reservoir/ethereum/tokens.sql
@@ -1,6 +1,6 @@
 {{ config(    
     schema = 'reservoir',
-    alias = 'tokens_beta',
+    alias = 'tokens',
     post_hook = '{{ expose_dataset(\'["ethereum"]\',
                 \'[""]\') }}'
     )


hawkaa (approved):
A small rubberstamp from me. I don't have a full understanding of what's going on here but I trust your judgement and that rolling back is easy.

I know there was an attempt to manage the view in spellbook, but it never exited `_beta` stage? With this we make it official and can modify the view in the spellbook?

models/datasets/reservoir/ethereum/tokens.sql
```diff
@@ -1,6 +1,6 @@
 {{ config(    
     schema = 'reservoir',
-    alias = 'tokens_beta',
+    alias = 'tokens',
```
kasiafi, line 3: Maybe rename the old one to avoid collision?

aalan3: the old one shouldn't be in use, was just for testing

aalan3 (approved):
Yes this is ok, I just added beta to test definition in the beginning.

PR #5050, opened 2023-12-25
1inch: lineage refactoring part 4

max-morrow:
4/4 PR of our lineage refactoring.

- cleanup legacy spellbooks & macros
- migrated `ar_trades` & `lop_own_trades` to the new flow (`oneinch.swaps`)
- new design for schemas
- redesigned `swaps`
- introducing fusion_accounts (EOAs of resolvers)
- `fusion_resolvers`, `fusion_executors`, `fusion_farms`, `exchange_contracts`: optimized performance & simplified code

# Thank you for contributing to Spellbook!
Please refer to the top of the `readme` in the root of Spellbook to learn how to contribute to Spellbook on DuneSQL.

PR #5100, opened 2024-01-04
add back 1inch: lineage refactoring part 4 (#5050)

jeff-dude:
This reverts commit 49fd7f90b308101eeb211b9d175b42422d13b575.

fyi @grkhr @max-morrow 

the dbt-trino macro for table materialization defaults to `rename` (reference [here](https://docs.getdbt.com/reference/resource-configs/trino-configs#table)), which is good as we can build temp tables, build backup, then rename temp to final and drop backup. this avoids downtime on the table. however, it appears when switching from view to table, the macro tries to rename the view to backup and apparently our connector to the metastore doesn't support rename on views 🤷‍♂️ 

for now, i've tested locally and overriding this config to `drop` will run successfully. this isn't ideal long term, as i assume it introduces downtime on the table each day when it rebuilds, even if it is quick. i think we should push through with `drop`, then remove that property in a follow-up PR.

we can look into overriding the dbt-trino macro for table materializations, but this use case seems rare and not sure it's worth the immediate effort.


models/oneinch/_meta/oneinch_blockchains.sql
```diff
@@ -2,7 +2,8 @@
     config(
         schema = 'oneinch',
         alias = 'blockchains',
-        materialized = 'view',
+        materialized = 'table',
+        on_table_exists = 'drop',
```
jeff-dude, line 6: for reference, here is config

PR #5101, opened 2024-01-04
Balances view for raw data

aalan3:
This is an example of how the balances could work. We will decide later wether to materialize this or not. For now let's discuss the metadata fields that should be added to this. I've added prices and metadata for erc20 and ~~native but not for nfts~~. I've added nft collection names as metadata as well.


macros/models/_sector/tokens/balances_enrich.sql
```diff
@@ -1,27 +1,47 @@
...
+    CASE
+        WHEN balances.type = 'erc20' THEN balances.amount / power(10, erc20_tokens.decimals) * prices.price
+        WHEN balances.type = 'native' THEN balances.amount / power(10, 18) * prices.price
+        ELSE NULL
+    END as balance_usd,
+    prices.price as price_rate,
+    erc20_tokens.symbol,
+    erc20_tokens.decimals,
+    nft_tokens.name as collection_name
+from {{ source('tokens_ethereum', 'balances_ethereum_0002') }} balances
```
jeff-dude, line 23:
```suggestion
from {{ balances_base }} balances
```

jeff-dude (changes requested):

jeff-dude (approved):

PR #5107, opened 2024-01-05
exclude eth balances

jeff-dude:
since we're moving balances out of spellbook, remove the beta approach here


aalan3:
We should drop these instead.

jeff-dude:
> We should drop these instead.

let's drop them in your PR which builds the new view. i'll close this.

PR #5108, opened 2024-01-05
Update 4x `dex.trades_beta` spells to use `generic_spot_compatible_trades` macro

tomfutago:
linked to #4759

updating:
* mummy_finance_optimism
* sushiswap_v1_optimism
* carbonhood_celo
* mento_v2_celo


tomfutago:
Note (fyi @ARDev097): I updated mummy_finance spell as imo Token/Amount In/Out were other way around. My understanding is that In = sold, Out = bought.

jeff-dude (approved):
thanks for coming back to update these

i'll tag this as open question and wait, as it's quite important we're sure on the token in/out assignment

tomfutago:
> thanks for coming back to update these
> 
> i'll tag this as open question and wait, as it's quite important we're sure on the token in/out assignment

agreed.

not sure if that's evident enough, but here's what I found so far on mummy finance swap. so looking at this trail:
https://github.com/mummy-finance/mummy-contracts -> points to: https://gmxio.gitbook.io/gmx/contracts#swap :

> Call Router.swap with parameters:
> _path: [tokenIn, tokenOut]
> _amountIn: amount of tokenIn to swap
> _minOut: minimum expected output amount
> _receiver: address of the receiver of tokenOut

-> amount **of tokenIn to swap**

jeff-dude (approved):
i agree with your approach here again, similar to other generic PR 👍

PR #5032, opened 2023-12-19
SPE-266 Initiate cross-chain transfers

jeff-dude:
# Thank you for contributing to Spellbook!
Please refer to the top of the `readme` in the root of Spellbook to learn how to contribute to Spellbook on DuneSQL.

PR #5109, opened 2024-01-05
Add `transfers_eth` macro and update ETH transfer spells

chuxinh:
Simplifying and consolidating `transfers_eth` spells for Ethereum, Base, Zora and OP Mainnet
cc: @MSilb7


jeff-dude:
hi @chuxinh 👋 

we recently rebuilt transfers across various chains too. can you look and see if they have what you need? if not, maybe we edit those instead? this is just ethereum, but other chain directories have same pattern.
- step 1: [raw data transfers](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/tokens/ethereum/tokens_ethereum_base_transfers.sql)
- step 2: [enriched transfers](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/tokens/ethereum/tokens_ethereum_transfers.sql)

chuxinh:
> hi @chuxinh 👋
> 
> we recently rebuilt transfers across various chains too. can you look and see if they have what you need? if not, maybe we edit those instead? this is just ethereum, but other chain directories have same pattern.
> 
> * step 1: [raw data transfers](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/tokens/ethereum/tokens_ethereum_base_transfers.sql)
> * step 2: [enriched transfers](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/tokens/ethereum/tokens_ethereum_transfers.sql)

Let me take a look. It seems like they're fundamentally doing something similar. How can I find those spells in the UI? I don't get to see it under tokens. Also wonder if there's a combined view across all chains?

jeff-dude:
> Let me take a look. It seems like they're fundamentally doing something similar. How can I find those spells in the UI? I don't get to see it under tokens. Also wonder if there's a combined view across all chains?

working on final stage cross-chain table [here](https://github.com/duneanalytics/spellbook/pull/5032), with post-hook used to expose on data explorer [here](https://github.com/duneanalytics/spellbook/pull/5032/files#diff-638157f88b3ee253bab967f3783b2b2ff2d185d4f3e753bfe09b2771e7850e69R10). having issues getting it to run on spellbook prod cluster due to large historical data volume for all transfers on all these chains, but will get it working asap

macros/models/_sector/transfers/eth/transfers_eth.sql
```diff
@@ -0,0 +1,71 @@
+{% macro transfers_eth(blockchain, eth_placeholder_contract, base_traces, base_transactions, erc20_transfer ) %}
```
MSilb7, line 1: erc20_transfer should maybe have a true/false flag on whether to look at ti?

Legacy OVM1 should be the only chain listed here that had ETH as an erc20 at some point, but other L2s might in the future.

I don't want to accidently pull in weird ERC20 things for other chains.

macros/models/_sector/transfers/eth/transfers_eth.sql
```diff
@@ -0,0 +1,71 @@
...
+        ,r.to
+        --Using the ETH deposit placeholder address to match with prices tables
+        ,{{eth_placeholder_contract}} as contract_address
+        ,cast(r.value as double) AS value
+        ,cast(r.value as double)/1e18 as value_decimal
+        ,r.tx_hash
+        ,r.trace_address
+        ,r.block_time as tx_block_time
+        ,r.block_number as tx_block_number
+        ,substring(to_hex(t.data), 1, 10) as tx_method_id
```
MSilb7, line 15: this should be bytearray_substring(t.data, 1, 4) - DuneSQL migrated

macros/models/_sector/transfers/eth/transfers_eth.sql
```diff
@@ -0,0 +1,71 @@
...
+        ,cast(r.tx_hash as varchar) || '-' || array_join(r.trace_address,',') as unique_transfer_id
```
MSilb7, line 16: Do we still want this field, or rather use the actual fields as our unique keys?

macros/models/_sector/transfers/eth/transfers_eth.sql
```diff
@@ -0,0 +1,71 @@
...
+        ,t."from" AS tx_from
+    from {{ base_traces }} as r
+    join {{ base_transactions}} as t
+        on r.tx_hash = t.hash
+        and r.block_number = t.block_number
+    where
+        (r.call_type not in ('delegatecall', 'callcode', 'staticcall') or r.call_type is null)
+        and r.tx_success
+        and r.success
+        and r.value > uint256 '0'
```
MSilb7, line 27: Do we need to cast still? I believe there was some query engine update that made us not have to any more.

I think I remember some chains not using uint256

macros/models/_sector/transfers/eth/transfers_eth.sql
```diff
@@ -0,0 +1,71 @@
...
+        ,r.to
+        --Using the ETH deposit placeholder address to match with prices tables
+        ,{{eth_placeholder_contract}} as contract_address
+        ,cast(r.value as double) AS value
+        ,cast(r.value as double)/1e18 as value_decimal
+        ,r.evt_tx_hash as tx_hash
+        ,array[r.evt_index] as trace_address
+        ,r.evt_block_time as tx_block_time
+        ,r.evt_block_number as tx_block_number
+        ,substring(to_hex(t.data), 1, 10) as tx_method_id
```
MSilb7, line 49: Same comment on bytearray_substring

macros/models/_sector/transfers/eth/transfers_eth.sql
```diff
@@ -0,0 +1,71 @@
...
+        ,cast(r.evt_tx_hash as varchar) || '-' || cast(r.evt_index as varchar) as unique_transfer_id
```
MSilb7, line 50: Same comment on if we want this field

MSilb7 (changes requested):
Left some comments on further cleanup things.

macros/models/_sector/transfers/eth/transfers_eth.sql
chuxinh: Do you mean this part only applies to OVM one but not the rest of the OP chains?

```
--ETH Transfers from deposits and withdrawals are ERC20 transfers of the 'deadeadead' ETH token. These do not appear in traces.
```

chuxinh: I'm intending to close this PR of creating macro or maybe just modify the originals for some cleanup. Think the combined `tokens` transfer that the Dune team is been working on will have both ETH transfers and erc20s combined cross chain onwards

PR #5118, opened 2024-01-07
Add platypus_finance to `dex.trades_beta`

tomfutago:
linked to #4759


tomfutago:
Note: added another generic macro - `generic_spot_v2_compatible_trades`, prob not the best name as it's not really v2 of anything but not sure what else to call it. There are a few already migrated dexs that are using similar pattern, so like with the "v1" - once this PR is finalised, I will open a separate one to update dexs that are following the same pattern.

jeff-dude (approved):
noted on the new macro. i don't have a great name for now either. we can revisit when finalizing the tracking gh issue prior to closing.

PR #5122, opened 2024-01-08
Add SmarDEX & DackieSwap to dex.trades on Base

chain-l:
Added trades file for SmarDEX & DackieSwap. 
Seed files are WIP.


macros/models/_sector/dex/smardex_compatible_trades.sql
Hosuke, line 1: This PR may be after #5169 .

PR #5130, opened 2024-01-09
Add static price stability rank to enhance amount_usd calculation

Hosuke:


Hosuke:
I am not sure if it is possible to shrink the new amount_usd calculation into a macro or a function like udf.

models/prices/prices_stability_rank.sql
```diff
@@ -0,0 +1,129 @@
...
+             ('bnb', 'CAKE', 0x0e09fabb73bd3ade0a17ecc321fd13a19e81ce82, 7),
+             ('bnb', 'DAI', 0x1af3f329e8be154074d8769d1ffa4ee058b1dbc3, 8),
+             ('bnb', 'HAY', 0x0782b6d8c4551b9760e74c0545a9bcd90bdc41e5, 9),
+             ('bnb', 'XRP', 0x1d2f0da169ceb9fc7b3144628db156f3f6c60dbe, 10),
+             ('celo', 'cUSD', 0x765de816845861e75a25fca122bb6898b8b1282a, 1),
+             ('celo', 'cEUR', 0xd8763cba276a3738e6de85b4b3bf5fded6d6ca73, 2),
+             ('celo', 'CELO', 0x471ece3750da237f93b8e339c536989b8978a438, 3),
+             ('celo', 'BTC', 0xd629eb00deced2a080b7ec630ef6ac117e614f1b, 4),
+             ('celo', 'BIFI', 0x639a647fbe20b6c8ac19e48e2de44ea792c62c5c, 5),
+             ('celo', '', 0x23ee2343b892b1bb63503a4fabc840e0e2c6810f, 6),
```
aalan3, line 65: Symbol is empty here

models/prices/prices_stability_rank.sql
```diff
@@ -0,0 +1,129 @@
...
+             ('polygon', 'agEUR', 0xe0b52e49357fd4daf2c15e02058dce6bc0057db4, 8),
+             ('polygon', 'PolyDoge', 0x8a953cfe442c5e8855cc6c61b1293fa648bae472, 9),
+             ('polygon', 'TEL', 0xdf7837de1f2fa4631d716cf2502f8b230f1dcc32, 10),
+             ('zksync', 'USDC.e', 0x3355df6d4c9c3035724fd0e3914de96a5a83aaf4, 1),
+             ('zksync', 'BUSD', 0x2039bb4116b4efc145ec4f0e2ea75012d6c0f181, 2),
+             ('zksync', 'WETH', 0x5aea5775959fbc2557cc8789bc1bf90a239d9a91, 3),
+             ('zksync', 'USDT.e', 0x493257fd37edb34451f62edf8d2a0c418852ba4c, 4),
+             ('zksync', 'MAV', 0x787c09494ec8bcb24dcaf8659e7d5d69979ee508, 4),
+             ('zksync', 'iZi', 0x16a9494e257703797d747540f01683952547ee5b, 4),
+             ('zksync', '', 0xed4040fd47629e7c8fbb7da76bb50b3e7695f0f2, 4),
```
aalan3, line 122: Symbol is empty here. If we don't track the price we can't use it here.

Hosuke, line 122: https://explorer.zksync.io/address/0xed4040fD47629e7c8FBB7DA76bb50B3e7695F0f2

models/prices/prices_stability_rank.sql
Hosuke, line 65: https://celoscan.io/token/0x23ee2343B892b1BB63503a4FAbc840E0e2C6810f

models/prices/prices_stability_rank.sql
```diff
@@ -0,0 +1,129 @@
...
+             ('celo', 'CELO', 0x471ece3750da237f93b8e339c536989b8978a438, 3),
+             ('celo', 'BTC', 0xd629eb00deced2a080b7ec630ef6ac117e614f1b, 4),
+             ('celo', 'BIFI', 0x639a647fbe20b6c8ac19e48e2de44ea792c62c5c, 5),
+             ('celo', '', 0x23ee2343b892b1bb63503a4fabc840e0e2c6810f, 6),
+             ('ethereum', 'WETH', 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2, 1),
+             ('ethereum', 'USDT', 0xdac17f958d2ee523a2206206994597c13d831ec7, 2),
+             ('ethereum', 'USDC', 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48, 3),
+             ('ethereum', 'DAI', 0x6b175474e89094c44da98b954eedeac495271d0f, 4),
+             ('ethereum', 'WBTC', 0x2260fac5e5542a773aa44fbcfedf7c193bc2c599, 5),
+             ('ethereum', 'PEPE', 0x6982508145454ce325ddbe47a25d4ec3d2311933, 6),
```
aalan3, line 71: Let's make these lists a bit smaller to maybe 5 coins and not include meme coins and less known coins. We can add more later if needed.

Hosuke: I will remake the list to shrink the size a bit, and easier for manually editing.

Hosuke:
Hi @tomfutago, there may be some issues with the newly added curvefi.base_trades:
https://dune.com/queries/3357660

The token sold here may be `3CRV` instead of `USDT`/`USDC.e`:
https://optimistic.etherscan.io/token/0x1337bedc9d22ecbe766df105c9623922a27963ec

Please verify.

tomfutago:
> Hi @tomfutago, there may be some issues with the newly added curvefi.base_trades: https://dune.com/queries/3357660
> 
> The token sold here may be `3CRV` instead of `USDT`/`USDC.e`: https://optimistic.etherscan.io/token/0x1337bedc9d22ecbe766df105c9623922a27963ec
> 
> (They have different decimals.)
> 
> Please verify.


Hmm, it looks some curve pools "assume" that all tokens in the pool have 18 decimals:

image

I missed this logic in the old lineage where decimals are hardcoded to 18 decimals:

`curvefi_ethereum_trades` : 
```
dexs.token_sold_amount_raw / power(10, case when swap_type = 'underlying_exchange_base' then 18 else erc20b.decimals end) AS token_sold_amount
```

`curvefi_optimism_trades` :
```
--On Sell: Metapools seem to always use the added coin's decimals if it's the one that's bought - even if the other token has less decimals (i.e. USDC)
--On Buy: Metapools seem to always use the curve pool token's decimals (18) if bought_id = CAST(0 as INT256)
dexs.token_bought_amount_raw / POWER(10 , (CASE WHEN pool_type = 'meta' AND bought_id = CAST(0 as INT256) THEN 18 ELSE COALESCE(erc20a.decimals,p_bought.decimals) END) ) AS token_bought_amount,
dexs.token_sold_amount_raw / POWER(10 , (CASE WHEN pool_type = 'meta' AND bought_id = CAST(0 as INT256) THEN COALESCE(erc20a.decimals,p_bought.decimals) ELSE COALESCE(erc20b.decimals,p_sold.decimals) END) )  AS token_sold_amount,
```

so that means we need to customise `enrich_dex_trades` macro for curve. unless you guys can think of a better solution? @Hosuke @jeff-dude

Hosuke:
> > Hi @tomfutago, there may be some issues with the newly added curvefi.base_trades: https://dune.com/queries/3357660
> > The token sold here may be `3CRV` instead of `USDT`/`USDC.e`: https://optimistic.etherscan.io/token/0x1337bedc9d22ecbe766df105c9623922a27963ec
> > (They have different decimals.)
> > Please verify.
> 
> Hmm, it looks some curve pools "assume" that all tokens in the pool have 18 decimals:
> 
> image
> 
> I missed this logic in the old lineage where decimals are hardcoded to 18 decimals:
> 
> `curvefi_ethereum_trades` :
> 
> ```
> dexs.token_sold_amount_raw / power(10, case when swap_type = 'underlying_exchange_base' then 18 else erc20b.decimals end) AS token_sold_amount
> ```
> 
> `curvefi_optimism_trades` :
> 
> ```
> --On Sell: Metapools seem to always use the added coin's decimals if it's the one that's bought - even if the other token has less decimals (i.e. USDC)
> --On Buy: Metapools seem to always use the curve pool token's decimals (18) if bought_id = CAST(0 as INT256)
> dexs.token_bought_amount_raw / POWER(10 , (CASE WHEN pool_type = 'meta' AND bought_id = CAST(0 as INT256) THEN 18 ELSE COALESCE(erc20a.decimals,p_bought.decimals) END) ) AS token_bought_amount,
> dexs.token_sold_amount_raw / POWER(10 , (CASE WHEN pool_type = 'meta' AND bought_id = CAST(0 as INT256) THEN COALESCE(erc20a.decimals,p_bought.decimals) ELSE COALESCE(erc20b.decimals,p_sold.decimals) END) )  AS token_sold_amount,
> ```
> 
> so that means we need to customise `enrich_dex_trades` macro for curve. unless you guys can think of a better solution? @Hosuke @jeff-dude

Curve may be a special/edge case for `dex.base_trades` and `dex.trades_beta`. We can discuss this issue in a new hotfix PR or an issue.

models/prices/prices_stability_rank.sql
```diff
@@ -0,0 +1,82 @@
+{{ config(
+        schema='prices',
+        alias = 'stablibity_rank',
+        materialized='table',
+        file_format = 'delta',
+        tags = ['static'],
+        post_hook = '{{ expose_spells(\'["ethereum", "arbitrum", "gnosis", "optimism", "bnb", "avalanche_c", "polygon", "fantom", "celo", "base", "zksync", "zora"]\',
+                                    "sector",
+                                    "prices",
+                                    \'["hosuke"]\') }}'
```
jeff-dude, line 10: lets remove the post-hook, i don't think it needs displayed on data explorer. this is more helpful for backend engineering.

models/prices/prices_stability_rank.sql
```diff
@@ -0,0 +1,82 @@
+{{ config(
+        schema='prices',
+        alias = 'stablibity_rank',
```
jeff-dude, line 3: instead of `stability_rank`, i think we simplify to just `trusted_tokens` or `core_tokens`

models/prices/prices_stability_rank.sql
```diff
@@ -0,0 +1,82 @@
...
+                                    \'["hosuke"]\') }}'
+        )
+}}
+
+
+-- Generated from https://dune.com/queries/3355223
+SELECT blockchain AS blockchain
+     , symbol AS symbol
+     , contract_address AS contract_address
+     , rank AS rank
```
jeff-dude, line 19: let's remove `rank` for now

macros/models/_sector/dex/enrich_dex_trades.sql
```diff
@@ -47,10 +47,17 @@ WITH base_trades as (
...
+                WHEN stability_bought.rank IS NOT NULL AND (stability_sold.rank IS NULL OR stability_bought.rank < stability_sold.rank) THEN
+                    base_trades.token_bought_amount_raw / power(10, erc20_bought.decimals) * p_bought.price
+                WHEN stability_sold.rank IS NOT NULL AND (stability_bought.rank IS NULL OR stability_sold.rank < stability_bought.rank) THEN
+                    base_trades.token_sold_amount_raw / power(10, erc20_sold.decimals) * p_sold.price
+                ELSE
+                    coalesce(
+                        base_trades.token_bought_amount_raw / power(10, erc20_bought.decimals) * p_bought.price,
+                        base_trades.token_sold_amount_raw / power(10, erc20_sold.decimals) * p_sold.price
+                    )
+            END AS amount_usd
```
jeff-dude, line 60: let's make this a macro for sure. we should initiate for dex, then reuse across all sectors for consistency as needed

jeff-dude: i think what we can do is similar to the `add_tx_from_to` macro, where we group all logic in a CTE, then pass CTE into macro to join to new prices spell below and write this type of logic

if we can't nest macros like that, then i'd suggest we remove `amount_usd` from this macro altogether, and then try to do directly in final dex spell maybe? hard to say exactly without being hands-on with it here but just giving some ideas

jeff-dude (changes requested):
first off -- thanks for taking on this initiative!

looks like you made the initial list in new spell smaller?
- usdc/usdt on each chain, where applicable
- weth/wavax/whatever other wrapped per chain
- dai?
- other stables?
you could remove any not in that general theme

macros/models/_sector/dex/enrich_dex_trades.sql
Hosuke: Since amount_usd calculation involves `erc20.decimals`, `token.price`, the `add_amount_usd` will be very similar to `enrich_dex_trades`

Hosuke: I just drafted the `add_amount_usd` macro.

I am thinking if I should hard-coded the column names or introduce them as parameters in the signature.

macros/models/_sector/add_amount_usd.sql
Hosuke, line 1: Since `add_tx_columns` is used after CTE blocks in sql files,
it does not use `WITH` in its implementation.

I try to not to use `WITH` and CTE in `add_amount_usd` macro, and use nested query instead, for grammatically correct.

But it may sacrifice performance in this way.

Hosuke: Probably not to separate this macro with enrich_dex_trades.

Hosuke, line 1: I implement another approach in #5281,
we may just choose and merge one of them, and close another.

macros/models/_sector/dex/enrich_dex_trades.sql
```diff
@@ -70,20 +66,16 @@ WITH base_trades as (
...
 
-select
-    *
-from
-    enrichments
+
+{{
+    add_amount_usd(
+        trades_cte = 'enrichments'
+        , prices_model = prices_model
```
jeff-dude, line 75: let's either:
- pass in `prices` CTE above here, similar to `enrichments`
- OR remove `prices` CTE, remove `prices_model` param in `enrich_dex_trades` macro & `dex_trades_beta` model, then just pass in `source(prices,usd)` directly here

models/prices/prices_schema.yml
```diff
@@ -108,3 +108,23 @@ models:
...
+          combination_of_columns:
+            - blockchain
+            - contract_address
+    columns:
+      - name: blockchain
+        description: "Native blockchain of the token"
+      - name: contract_address
+        description: "Contract address of the token"
+      - name: symbol
+        description: "Token symbol"
```
jeff-dude, line 130: not sure it matters, but these columns are flipped based on model

macros/models/_sector/add_amount_usd.sql
```diff
@@ -0,0 +1,58 @@
...
+            , minute
+            , price
+        FROM {{ prices_model }}
+        {% if is_incremental() %}
+        WHERE {{ incremental_predicate('minute') }}
+        {% endif %}
+    ) ps
+    ON bt.token_sold_address = ps.contract_address
+    AND bt.blockchain = ps.blockchain
+    AND ps.minute = date_trunc('minute', bt.block_time)
```
jeff-dude, line 56: to keep consistent with prior design in `enrich_dex_trades` macro, should we put these prices subqueries into a CTE at top, then join to CTE? that looked clean previously

macros/models/_sector/add_amount_usd.sql
```diff
@@ -0,0 +1,58 @@
...
+                FROM {{ ref('prices_trusted_tokens') }}
+                WHERE blockchain = bt.blockchain
+            )
+            THEN bt."{{ bought_amount_column }}" * pb.price
+        WHEN bt.token_sold_address IN (
+                SELECT contract_address
+                FROM {{ ref('prices_trusted_tokens') }}
+                WHERE blockchain = bt.blockchain
+            )
+            THEN bt."{{ sold_amount_column }}" * ps.price
```
jeff-dude, line 22: i'm thinking these `WHEN token IN (...)` statements are what is causing the slowdown in run times in CI. notice this last run took ~2000s, while it used to take ~200s (10x increase in runtime).
reference from curve CI:
https://github.com/duneanalytics/spellbook/actions/runs/7758815487/job/21161436499?pr=5271#step:11:68

if we change this logic to be `left join` to `prices_trusted_tokens`, similar to regular prices joins below, that may be more efficient. of course, we'd have to left join twice, once for bought, once for sold

jeff-dude (changes requested):
left some more feedback, i think we're on the right path

macros/models/_sector/add_amount_usd.sql
Hosuke: https://github.com/duneanalytics/spellbook/pull/5130#discussion_r1469946590

I will try to add them as CTE, but the final usage may be different from `add_tx_columns` macro. It may be a macro quoted in the CTE block.

macros/models/_sector/dex/enrich_dex_trades.sql
Hosuke: I pick option 2.

If option 2 failed then I will go for option 1.

macros/models/_sector/add_amount_usd.sql
Hosuke: 256s now:
https://github.com/duneanalytics/spellbook/actions/runs/7856500247/job/21439334222?pr=5130#step:11:20

jeff-dude (approved):
looking good to me now 🙌 

asking @0xRobin for a quick review, to see if the structure makes sense to use in `nft.trades` too?

also @tomfutago for other sectors being built, such as lending/borrowing

tomfutago:
> looking good to me now 🙌
> 
> asking @0xRobin for a quick review, to see if the structure makes sense to use in `nft.trades` too?
> 
> also @tomfutago for other sectors being built, such as lending/borrowing

Looking slick to me @Hosuke! Just to make sure I understand the idea tho: new macro checks if token is on the trusted list and if either (1) `bought` or (2) `sold` token (in that particular order) is on the list - `usd_amount` will be calculated based on this token. if not, logic reverts to what it was before, correct?

If so, I don't think that applies to lending spells, as all of them (borrow, supply, flashloans) have only 1 amount to convert to `usd_amount `- so no coalesce to find the preference. Unless I'm missing something?

jeff-dude:
> Looking slick to me @Hosuke! Just to make sure I understand the idea tho: new macro checks if token is on the trusted list and if either (1) `bought` or (2) `sold` token (in that particular order) is on the list - `usd_amount` will be calculated based on this token. if not, logic reverts to what it was before, correct?

yep, that is correct.
 
> If so, I don't think that applies to lending spells, as all of them (borrow, supply, flashloans) have only 1 amount to convert to `usd_amount `- so no coalesce to find the preference. Unless I'm missing something?

good point. `nft.trades` is likely one-sided too, as it just gets `amount_usd` translated from the native token used to buy/sell. this macro may only be useful for small subset of sectors

jeff-dude (approved):

PR #5131, opened 2024-01-09
`mummy_finance_optimism` fix to handle denormalized decoded tables

jeff-dude:
- when using a `group by`, ensure no `select *` but explicitly pull columns
- remove join to base table, since decoded has columns now
- update incremental logic
- force changes downstream to see lineage run


jeff-dude:
fyi @kaiblade 

plz see PR description. just this week, we've denormalized decoded tables on optimism blockchain to include `evt_tx_from` and `evt_tx_to`, to avoid joining back to base tables for them. this ended up breaking this model. i've gone ahead and fixed/enhanced.

keep in mind when moving forward, to simplify code submissions. also note, when using a group by, let's avoid `select *` and list columns out that are needed explicitly

aalan3 (approved):

PR #5137, opened 2024-01-10
Added chainhop optimism dexes 

ARDev097:
Added trades data for Chainhop and Saddle_Finance Dexes on Optimism.


models/_sector/dex/trades/optimism/platforms/chainhop_optimism_base_trades.sql
```diff
@@ -0,0 +1,60 @@
...
+        {% if not loop.last %}
+        UNION ALL
+        {% endif %}
+
+    {% endfor %}
+)
+
+SELECT
+    'optimism' AS blockchain,
+    'chainhop' AS project,
```
Hosuke, line 45: Hi @ARDev097,
if `chainhop` and `saddle_finance` are newly added dexes, we may need to add their information in `dex.info` table:

https://github.com/duneanalytics/spellbook/blob/main/models/dex/dex_info.sql

models/_sector/dex/trades/optimism/platforms/saddle_finance_optimism_base_trades.sql
```diff
@@ -0,0 +1,60 @@
...
+    {% for evt_trade_table in saddle_finance_optimism_evt_trade_tables %}
+        SELECT
+            evt_block_time          AS block_time,
+            evt_tx_to               AS taker,
+            evt_tx_from             AS maker,
+            tokensBought            AS token_bought_amount_raw,
+            tokensSold               AS token_sold_amount_raw,
+            CAST(NULL AS double)    AS amount_usd,
+            CAST(NULL AS varbinary)                 AS token_bought_address,
+            CAST(NULL AS varbinary)                AS token_sold_address,
```
Hosuke, line 26: These two fields should be not null if I am not wrong.
Do you have any specific reason to make them null?

Hosuke (changes requested):

models/_sector/dex/trades/optimism/platforms/saddle_finance_optimism_base_trades.sql
ARDev097: Yes. In the decode table of saddle_finance, we do not have these two columns that's why I have added these two columns as null.

Hosuke: In this case we may need to find other sources to join, maybe log?

ARDev097: Yes, it seems better to retrieve those two columns from the log source table. If possible, could you provide any reference or guidance on how to accomplish this?

Hosuke: That might be a little bit complicated, but this should be an example:
https://github.com/duneanalytics/spellbook/blob/main/models/hashflow/ethereum/hashflow_ethereum_raw_trades.sql

Anyway if we can submit new contracts for decoding and use them as decoded sources, it will be easier.

Hosuke: You may need to fill in these columns first, as they are not-null columns by requirements.

ARDev097: Hello @Hosuke ,

As of now, can I remove "saddle_finance" from this PR? Once it is fully ready, I will submit another PR for it.

Hosuke: Yes, you can remove `saddle_finance` first.

ARDev097: Done.

Hosuke (approved):
LGTM.✅
Thank you @ARDev097 

[Sample chainhop_optimism_base_trades]()

jeff-dude (approved):

PR #5143, opened 2024-01-11
Contracts - Make contract_mapping build on incremental refresh

MSilb7:
In https://github.com/duneanalytics/spellbook/pull/5023 the contract_mapping tables were changed from views to materialized tables, since the views were slow to run at the Dune.com query stage (required ~5 mins).

---

Since then, the contract_mapping tables for each chain only build once, and do not refresh on incremental runs. This is an attempt at making these tables rebuild on incremental runs.

cc @jeff-dude on the ideal solution, but putting this up as a starting point.


MSilb7:
**Contract Mapping only has the first 183 days**
image

**Lower-level iterated creators table has the whole history**
image

jeff-dude:
closing until you reopen a new one

PR #5150, opened 2024-01-11
Update referral.rewards spell to Include paragraph mints without a referrer

datadanne:
Kyle noticed that the dashboard is showing that 100% of paragraph mints have a referrer which is isn't correct. I didn't think about that when I added support for paragraph, the event we are using to populate the data is only emitted for mints where there is a referrer. 

This change updates the paragraph query so that it includes all calls to `FeeManger#distributeEvent` where the referrer is the 0-address as rows in the model with no referrer which should fix that! Unfortunately the collection contract isn't available in the call table so I had to join with traces to get that, is that an ok approach @0xRobin?


macros/models/_sector/referral/platforms/paragraph_referral_rewards.sql
```diff
@@ -35,4 +36,39 @@ where
...
+    ,cast(date_trunc('day',call_block_time) as date) as block_date
+    ,cast(date_trunc('month',call_block_time) as date) as block_month
+    ,call_tx_hash as tx_hash
+    ,'NFT' as category
+    ,0x0000000000000000000000000000000000000000 as referrer_address
+    ,minter as referee_address
+    ,{{native_currency_contract}} as currency_contract
+    ,0 as reward_amount_raw
+    ,t."from" as project_contract_address
+    ,cardinality(trace_address) + cast(coalesce(nullif(array_join(trace_address, ''), ''), '0') as integer) + 9999999 as sub_tx_id
```
datadanne, line 57: This is ugly but I couldn't think of a better way to try to make sure that there are no conflicts with `FeeDistributed` `evt_index` values in case there are multiple mint calls in the same transaction

0xRobin: `+9999` is a fine hack to make sure there are no collisions with evt indexes.
A more straightforward way of getting an index from the trace address would be:
`row_number() over (partition by tx_hash, order by trace_address)`

0xRobin:
@datadanne could we try using `ERC721_call_mintWithReferrer` instead of traces to determine the NFT contract address? It should be the `contract_address` column in that table.

datadanne:
> @datadanne could we try using `ERC721_call_mintWithReferrer` instead of traces to determine the NFT contract address? It should be the `contract_address` column in that table.

@0xRobin Do you mean selecting from `FeeManager_call_distributeFee` and joining with `ERC721_call_mintWithReferrer` instead of traces, or just selecting from `ERC721_call_mintWithReferrer` and using that data instead?
 
I'm not sure if joining will work in case there are multiple `mintWithReferrer` calls in one transaction, not sure how we would match them up in that case

But just selecting from `ERC721_call_mintWithReferrer` should give us all the data we need, the only issue with that is that since ERC721 is a proxy contract there will be two rows in `ERC721_call_mintWithReferrer` per mint, do you know of a good way to figure out which call is the non-proxy call without hardcoding the proxy addresses? In a normal transaction with just one call to mintWithReferrer it is easy to figure out since we could just compare `cardinality(call_trace_address)` or something like that, but handling the case with multiple calls in one transaction seems a bit more complicated

0xRobin:
> Do you mean selecting from FeeManager_call_distributeFee and joining with ERC721_call_mintWithReferrer instead of traces, or just selecting from ERC721_call_mintWithReferrer and using that data instead?
> 
I would source everything from `ERC721_call_mintWithReferrer` if that's possible

> But just selecting from `ERC721_call_mintWithReferrer` should give us all the data we need, the only issue with that is that since ERC721 is a proxy contract there will be two rows in `ERC721_call_mintWithReferrer` per mint, do you know of a good way to figure out which call is the non-proxy call without hardcoding the proxy addresses? In a normal transaction with just one call to mintWithReferrer it is easy to figure out since we could just compare `cardinality(call_trace_address)` or something like that, but handling the case with multiple calls in one transaction seems a bit more complicated

Hmm I do find it really strange that we see the duplicate call entries due to the proxy contract. 
I don't understand why `0x3e3255cbe27f34a981a8afa98192e77eb198901a` (on Base) is being decoded in Dune as it's not created through the same factory as all the ERC721 instances.
I noticed that the ERC721Factory has the implementation contract (``0x3e32`) as an immutable variable and it doesn't have methods to update that. So I think it is fine hardcoding the proxy contract and filtering that out. It should just be 1 hardcoded contract for each chain, which we can pass into the macro as a variable.

datadanne:
> I would source everything from `ERC721_call_mintWithReferrer` if that's possible
> 
> Hmm I do find it really strange that we see the duplicate call entries due to the proxy contract. I don't understand why `0x3e3255cbe27f34a981a8afa98192e77eb198901a` (on Base) is being decoded in Dune as it's not created through the same factory as all the ERC721 instances. I noticed that the ERC721Factory has the implementation contract (`0x3e32`) as an immutable variable and it doesn't have methods to update that. So I think it is fine hardcoding the proxy contract and filtering that out. It should just be 1 hardcoded contract for each chain, which we can pass into the macro as a variable.

Nice, if we are ok with that solution it is easy to source everything from `ERC721_call_mintWithReferrer`. I'll push an update that does that tomorrow and will ping you once it is ready for review! Ty for the feedback

macros/models/_sector/referral/platforms/paragraph_referral_rewards.sql
```diff
@@ -35,4 +37,39 @@ where
...
+    ,call_tx_hash as tx_hash
+    ,'NFT' as category
+    ,0x0000000000000000000000000000000000000000 as referrer_address
+    ,c.to as referee_address
+    ,{{native_currency_contract}} as currency_contract
+    ,0 as reward_amount_raw
+    ,c.contract_address as project_contract_address
+    ,row_number() over (partition by call_tx_hash order by call_trace_address) + 9999999 as sub_tx_id
+    ,tx."from" as tx_from
+    ,tx.to as tx_to
```
datadanne, line 60: I noticed something really odd: `call_tx_from` doesn't exist in any of the polygon `Contract_call_functionName` tables. At first I thought it might have been a bug with this specific table, but I checked a few others and none of them have those two columns. Is that a known limitation or something or a bug?

0xRobin: Those 2 columns have only recently been added to the call and event tables.
So I believe it's not out there for all chains yet, it should be added in the near future.

datadanne: Ah I see. Is that documented anywhere? Works without it, just curious if there was a way I could that wasn't write the query -> open PR -> wait for CI to fail

0xRobin: no docs yet I think, I believe the columns are only 'stealth launched' for a couple chains now.
Only way to know this is if you see it in the data explorer on Dune. It'll be properly announced when all chains are covered.

datadanne:
@0xRobin do you think you'll get a chance to review the PR this week?

0xRobin (approved):
@datadanne I just had a look at it yesterday and tests are passing so this looks good! :ship:

PR #5156, opened 2024-01-12
Added arbitrum for yield_yak (including to dex aggregator)

yy-analytics:
Sorry if anything about this PR is incorrect, have tried to follow the guidance as per docs. In short, I'm trying to add Yield Yak's router on Arbitrum into the dex_aggregator.trades spell so that it flows through to some other dashboards. I've largely copy/pasted/edited from the existing spell for Yield Yak on the Avalanche C-Chain, and then created an overall yield_yak_trades.sql file and related yield_yak_trades_schema.yml with inspiration from what was done for Paraswap across multiple chains. As a result, there's a small change to dex_aggregator_trades.sql and I've also added a couple of lines to the related seed CSV file for that spell including a couple of YakSwap events on Arbitrum.


yy-analytics:

yy-analytics:
recheck

ahop1:
recheck

dbt_project.yml
```diff
@@ -871,6 +871,8 @@ models:
       +schema: yield_yak
       avalanche_c:
         +schema: yield_yak_avalanche_c
+      arbitrum:
+        +schema: yield_yak_arbitrum
```
jeff-dude, line 875: can you remove modifications to this file and instead put schema property directly in model config blocks?

models/yield_yak/arbitrum/yield_yak_arbitrum_trades.sql
```diff
@@ -0,0 +1,108 @@
...
+    alias = 'trades',
+    partition_by = ['block_month'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['block_date', 'blockchain', 'project', 'version', 'tx_hash', 'evt_index', 'trace_address'],
+    post_hook='{{ expose_spells(\'["arbitrum"]\',
+                                "project",
+                                "yield_yak",
+                                \'["angus_1"]\') }}'
```
jeff-dude, line 12: can you remove this post hook?
we're moving away from it's usage on spells which feed into downstream spells

models/yield_yak/arbitrum/yield_yak_arbitrum_trades.sql
```diff
@@ -0,0 +1,108 @@
...
+            _tokenOut as token_bought_address, 
+            contract_address as project_contract_address, 
+            evt_tx_hash as tx_hash,
+            ARRAY[-1] AS trace_address,
+            evt_index
+        FROM 
+        {{ source('yield_yak_arbitrum', 'YakRouter_evt_YakSwap') }}
+        {% if is_incremental() %}
+        WHERE evt_block_time >= date_trunc('day', now() - interval '7' Day)
+        {% endif %}
```
jeff-dude, line 39: can you update incremental logic to new design pattern?
target like [this](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/trades/avalanche_c/platforms/openocean_avalanche_c_base_trades.sql#L9)
source like [this](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/trades/avalanche_c/platforms/openocean_avalanche_c_base_trades.sql#L30-L32)

jeff-dude: plz apply across the whole PR

jeff-dude (changes requested):

dbt_project.yml
yy-analytics: Done

models/yield_yak/arbitrum/yield_yak_arbitrum_trades.sql
yy-analytics: Done :)

models/yield_yak/arbitrum/yield_yak_arbitrum_trades.sql
yy-analytics: Have done this, but was not sure if there was also a way to fit something like the below for the `not is_incremental()` scenario into the new design pattern (?)
```
    {% if not is_incremental() %}
    AND tx.block_time >= TIMESTAMP '{{project_start_date}}'
    {% endif %}
```

jeff-dude: nope, the `not` portion will remain filtering on hardcoded date variable when project started. it's helpful to filter base tables in joins as much as possible, especially when data in decoded table starts much later than data in base table (which most cases do).

ideal setup:
```
    {% if not is_incremental() %}
    AND tx.block_time >= TIMESTAMP '{{project_start_date}}'
    {% else %}
    AND {{incremental_predicate('tx.block_time')}}
    {% endif %}
```

the if/then can also be flipped, doesn't matter 🤝 
it's cleaner to see the `else` in there vs. having an `endif` and `if` again.

it's not a blocker by any means, so we can proceed as-is, but just a general design pattern for future spells.

jeff-dude (approved):
pushed a commit to cleanup crosschain view a bit, otherwise should be good 🤝

PR #5174, opened 2024-01-16
[ready for review] Changes to Balancer BPT supply spell

viniabussafi:
This PR creates a new macro for the BPT supply spell. 
What motivates the changes is that the current spell only supports Composable Stable Pools and we need to extend this to all pool types, so we can more accurately calculate BPT prices, which will, then, impact the `liquidity` and `protocol fees` spell.
I'm also updating the `transfers bpt` spell, as https://github.com/duneanalytics/spellbook/pull/5172 was successful in including the transactions that were missing (any idea on what was causing it to fall behind, btw?). My attempt was to make the spell query the `transfers` event, instead of the `logs` table, which made it run faster.


jeff-dude (approved):
awesome cleanup work here 🔥 
i think you'll find this code easier to maintain

PR #5176, opened 2024-01-16
Add token id to balances

aalan3:
I missed adding token id on the views. @hildobby pointed out that it's missing, so I'm adding it here. 

I'm not 100% sure on the name here so suggestions are welcome.


hildobby:
Even just token_id works fine imo since they are only applicable to NFTs, and that way naming matches the one found in other tables

jeff-dude (approved):
updated to `token_id` 🤝

PR #5180, opened 2024-01-18
Add zkSync to Contracts Mapping

MSilb7:
Adds logic to pull extra zkSync contracts (where creation data is in events).

ref: https://github.com/duneanalytics/spellbook/pull/4788
cc @lgingerich for sanity checking

edit: my local dbt is throwing beethovenx schema errors, so testing compiling here.

### DIFF
--- a/macros/models/_sector/contracts/contracts_base_starting_level.sql
+++ b/macros/models/_sector/contracts/contracts_base_starting_level.sql
@@ -70,7 +70,44 @@ FROM (
 
             AND {{ incremental_days_forward_predicate('ct.block_time', 'cd.base_time', days_forward ) }}
             AND {{ incremental_days_forward_predicate('t.block_time', 'cd.base_time', days_forward ) }}
+        {% if chain == 'zksync' %}
+          UNION ALL
+          select 
+            '{{chain}}' AS blockchain
+            ,ct.deployerAddress as trace_creator_address
+            ,ct.contractAddress as contract_address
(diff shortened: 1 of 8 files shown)


lgingerich:
@MSilb7 looks good to me, thanks for handling this

MSilb7:
> @MSilb7 looks good to me, thanks for handling this

Awesome, also want to validate that pulling `deployerAddress` from the event `ContractDeployer_evt_ContractDeployed` is "correct" for determining what address is responsible for deploying the contract? (i.e. this is likely an address of a developer deploying a contract, not a factory, or person who happened to create a DEX pool).

lgingerich:
Good callout. I thought that would work, but turns out it doesn't. The `deployerAddress` field can/will be a factory contract. You should use the `evt_tx_from` field to get the "developer" address.

MSilb7:
> Good callout. I thought that would work, but turns out it doesn't. The `deployerAddress` field can/will be a factory contract. You should use the `evt_tx_from` field to get the "developer" address.

Actually, it should be fine if it's factories, because we iterate up to the top-level deployer (like for creation_trace created contracts).

MSilb7:
Moreso like, is `deployerAddress` ~equivalent to "from" in traces? Which seems like yes.

lgingerich:
Yes I believe they are equivalent

jeff-dude (approved):
thanks both -- this looks good to me

PR #5184, opened 2024-01-18
Curve `dex.trades_beta` override decimals fix

tomfutago:
Aiming to fix `curvefi` logic on amount sold/bought where for some pools (ethereum/optimism) decimals are either:
- defaulted to 18 (regardless of token's decimals)
- swapped between sold and bought token

Adding here 4 additional flags to each `base_trade` models:
- override_bought_decimals
- override_sold_decimals
- swap_bought_decimals
- swap_sold_decimals

and updating `enrich_dex_trades` macro.

linked to #5130


Hosuke:
We can add this tx in the seed for testing:
https://optimistic.etherscan.io/tx/0x9e32c44a4f5dae97fe72f8ff3aa52707180beaafb85fc0823e4e273450667ca5

tomfutago:
> We can add this tx in the seed for testing: https://optimistic.etherscan.io/tx/0x9e32c44a4f5dae97fe72f8ff3aa52707180beaafb85fc0823e4e273450667ca5

Added. Also, I noticed that all seeds for curve used `curvefi` as project name, but all spells (including old lineage in `dex.trades`) are using `curve` as project name, so I updated all seeds.

models/_sector/dex/trades/arbitrum/dex_arbitrum_base_trades.sql
```diff
@@ -53,6 +53,10 @@ WITH base_union AS (
             , project_contract_address
             , tx_hash
             , evt_index
+            , false AS override_bought_decimals
+            , false AS override_sold_decimals
+            , false AS swap_bought_decimals
+            , false AS swap_sold_decimals
```
Hosuke, line 59: I wonder if we can have better approach for these.

tomfutago: I know it's not the most elegant solution, but as defaulting of the decimals to 18 or swapping between bought and sold token decimals goes all the way down to be used in `enrich` macro - I couldn't think of any better solution without touching `base_trades` models.

I'm open to suggestions.

seeds/_sector/dex/curvefi_optimism_base_trades_seed.csv
```diff
@@ -1,18 +1,21 @@
...
+optimism,curve,1,2022-05-20,0xf0a56683a1f5c98f734da37362976c19ced781ea2ca1168c203d605593689308,3,0x94b008aa00579c1307b0ef2c499ad98a8ce58e58,0x7f5c764cbc14f9669b88837ca1490cca17c31607,8659786,112531495,106894737
+optimism,curve,1,2022-05-16,0x1daf9b0e5b53bbb0cf2cc82fc56ea4b03e4b04642326c4c34ee5034edcc419a0,5,0x2e3d870790dc77a83dd1d18184acc7439a53f475,0x94b008aa00579c1307b0ef2c499ad98a8ce58e58,8429446,2028066389647480735900,2002456704
+optimism,curve,1,2022-05-11,0x9e35d672d4202923c30f0e65eeaae4b7e922600b2715dacde56a98cef3903062,5,0x94b008aa00579c1307b0ef2c499ad98a8ce58e58,0x2e3d870790dc77a83dd1d18184acc7439a53f475,7910924,1002932210,1001247450302501874441
+optimism,curve,1,2022-05-14,0xa3178cf9778f11a12c896e7967a38ab72b2ca043e2f569dc56e7fe58de6d7995,3,0x2e3d870790dc77a83dd1d18184acc7439a53f475,0xfb21b70922b9f6e3c6274bcd6cb1aa8a0fe20b80,8234971,6057514570698228151262,30000000000
+optimism,curve,1,2023-09-05,0x9e32c44a4f5dae97fe72f8ff3aa52707180beaafb85fc0823e4e273450667ca5,490,0x8c6f28f2f1a3c87f0f938b96d27520d9751ec8d9,0x94b008aa00579c1307b0ef2c499ad98a8ce58e58,109170536,4000860630461955786,3916237747392440611
```
Hosuke, line 19: I think I may suggest another way of solving this problem, but I am not very sure.
If you search this number `3916237747392440611` in the transaction:
https://optimistic.etherscan.io/tx/0x9e32c44a4f5dae97fe72f8ff3aa52707180beaafb85fc0823e4e273450667ca5

, you can see it is actually swapping to `3.916237747392440611 Curve.fi DAI/USDC/USDT`, and it is not USDT:
https://optimistic.etherscan.io/token/0x1337bedc9d22ecbe766df105c9623922a27963ec

If we can introduce this token, which is not USDT, we may solve this problem perfectly.

seeds/_sector/dex/curvefi_optimism_base_trades_seed.csv
```diff
@@ -1,18 +1,21 @@
...
+optimism,curve,1,2023-09-05,0x9e32c44a4f5dae97fe72f8ff3aa52707180beaafb85fc0823e4e273450667ca5,478,0x8c6f28f2f1a3c87f0f938b96d27520d9751ec8d9,0x94b008aa00579c1307b0ef2c499ad98a8ce58e58,109170536,4000925723992149027,3916301433007904588
```
Hosuke, line 20: Same goes to `3916301433007904588` here and elsewhere, in this event the token involved is `3.916301433007904588 Curve.fi DAI/USDC/USD`. You can have a check.

Hosuke (changes requested):
If we introduce/identify this token [`0x1337bedc9d22ecbe766df105c9623922a27963ec`](https://optimistic.etherscan.io/token/0x1337bedc9d22ecbe766df105c9623922a27963ec) with the accurate decimal `18` defined, we may solve this problem natively.

seeds/_sector/dex/curvefi_optimism_base_trades_seed.csv
tomfutago: Hmm, I don't think so. 0x1337BedC9D22ecbe766dF105c9623922A27963EC is [Curve 3pool](https://curve.fi/#/optimism/pools?search=0x1337BedC9D22ecbe766dF105c9623922A27963EC) - not a single token.

tomfutago:
More thoughts on this (mostly obvious stuff, just to get it clear in my head too):

New `dex.trades_beta` setup has this simplified lineage: 
1: `<project>_<chain>_base_trades` -> 2: `dex_<chain>_base_trades` -> 3: `dex_base_trades` -> 4: `dex_trades_beta`

Level 1: models pull `amount_raw` and `token_address` and only at level 4: `amount` and `amount_usd` are calculated based on level 1: `amount_raw` and `token_address` decimals.

In order to bring `curve` logic to calculate `amount` (only for specific pools) - I don't see any other way than having some info on level 1: in the lineage which txs are "special", so that's why I added new flags, which again is not the prettiest solution but I don't think it adds a lot of overhead - dummy flags on level 2: for majority of chains and another ugly `if` just for ethereum and optimism.

Another possibility that comes to mind (also not very pretty) - is to have a separate branch in the lineage for specific projects like `curve`. It would be something along the lines:
1: `<project>_<chain>_base_trades`
 -> 2a: `dex_<chain>_base_trades` -> 3a: `dex_base_trades`
 -> 2b: `dex_<chain>_base_trades_special` -> 3b: `dex_base_trades_special`
 -> 4: `dex_trades_beta` union all `dex_trades_special_beta`

`dex_trades_special_beta` then would have its own `enrich` logic to calculate `amount` and `amount_usd`.

In addition to `curve`, it might also apply to `balancer` and `oneinch` too:
- `balancer` because `amount_usd` atm doesn't match between `dex.trades` and `dex.trades_beta` - exposed views are bridging this gap, but users will need to use views for their analysis and potentially union them with `dex.trades` for some cross-project stats
- `oneinch` is an absolute masterpiece with intricate net of macros all the way down and not trivial to fit into the new lineage..

Ok, brain dump over.. Thoughts?

seeds/_sector/dex/curvefi_optimism_base_trades_seed.csv
Hosuke: It's a pool, and a special token at the same time.

Its price feed api:
https://api.coinpaprika.com/v1/coins/3crv-lp-3pool-curve
https://coinpaprika.com/coin/3crv-lp-3pool-curve/

Its etherscan:
https://etherscan.io/token/0x6c3f90f043a72fa612cbac8115ee7e52bde6e490

Moreover, you may not find the exact number matched to amount_raw in these transaction to other tokens.

Hosuke:
> More thoughts on this (mostly obvious stuff, just to get it clear in my head too):
> 
> New `dex.trades_beta` setup has this simplified lineage: 1: `<project>_<chain>_base_trades` -> 2: `dex_<chain>_base_trades` -> 3: `dex_base_trades` -> 4: `dex_trades_beta`
> 
> Level 1: models pull `amount_raw` and `token_address` and only at level 4: `amount` and `amount_usd` are calculated based on level 1: `amount_raw` and `token_address` decimals.
> 
> In order to bring `curve` logic to calculate `amount` (only for specific pools) - I don't see any other way than having some info on level 1: in the lineage which txs are "special", so that's why I added new flags, which again is not the prettiest solution but I don't think it adds a lot of overhead - dummy flags on level 2: for majority of chains and another ugly `if` just for ethereum and optimism.
> 
> Another possibility that comes to mind (also not very pretty) - is to have a separate branch in the lineage for specific projects like `curve`. It would be something along the lines: 1: `<project>_<chain>_base_trades` -> 2a: `dex_<chain>_base_trades` -> 3a: `dex_base_trades` -> 2b: `dex_<chain>_base_trades_special` -> 3b: `dex_base_trades_special` -> 4: `dex_trades_beta` union all `dex_trades_special_beta`
> 
> `dex_trades_special_beta` then would have its own `enrich` logic to calculate `amount` and `amount_usd`.
> 
> In addition to `curve`, it might also apply to `balancer` and `oneinch` too:
> 
> * `balancer` because `amount_usd` atm doesn't match between `dex.trades` and `dex.trades_beta` - exposed views are bridging this gap, but users will need to use views for their analysis and potentially union them with `dex.trades` for some cross-project stats
> * `oneinch` is an absolute masterpiece with intricate net of macros all the way down and not trivial to fit into the new lineage..
> 
> Ok, brain dump over.. Thoughts?

I agree that `curvefi`, `balancer` and `oneinch` are 3 most special cases in dex.trades.
But for the current problem I think may be just a token mismatched problem. 
Also the seed records in my point of view may need some fixes, as I think the `bought/sold_amount_raw` cannot match to the `token_bought/sold_address` correctly.

seeds/_sector/dex/curvefi_optimism_base_trades_seed.csv
tomfutago: Ok cool, but I'm still not entirely clear how to distinguish token bought from token sold address in this case. How would you set it up in this example: https://dune.com/queries/3365551 (this is comparing results for 1 trade between old and new model using tx you mentioned above)

Hosuke: I think this may be the problem in old dex.trades as well, and 3CRV's price is different from USDT now (not 1USD at the current rate).

Hosuke: I will first break down the query and see which source these trades come from.

Hosuke: Here are the queries:
https://dune.com/queries/3367645
https://dune.com/queries/3367648

Based on the query results(I search the amount_raw in the transaction details), I think all trades from `curvefi_[optimism].MetaPoolSwap_evt_TokenExchangeUnderlying`, are related to `3CRV` trades.

Details:
If a trade is from `curvefi_[optimism].MetaPoolSwap_evt_TokenExchangeUnderlying`, and its token_bought/sold_address is one of `DAI/USDC/USDT`, then that we should change the address to 3CRV's address.


We may only need to fix this part of the code cross all chains for curvefi trades:
https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/trades/optimism/platforms/curvefi_optimism_base_trades.sql#L60
```
-- MetaPoolSwap TokenExchangeUnderlying
        SELECT
            'meta' AS pool_type, -- has implications for decimals for curve
            t.evt_block_time AS block_time,
            t.evt_block_number,
            t.buyer AS taker,
            CAST(NULL as varbinary) AS maker,
            -- when amount0 is negative it means taker is buying token0 from the pool
            tokens_bought AS token_bought_amount_raw,
            tokens_sold AS token_sold_amount_raw,
            t.contract_address as project_contract_address,
            t.evt_tx_hash AS tx_hash,
            t.evt_index,
            bought_id,
            sold_id
        FROM {{ source('curvefi_optimism', 'MetaPoolSwap_evt_TokenExchangeUnderlying') }} t
        {% if is_incremental() %}
        WHERE {{ incremental_predicate('t.evt_block_time') }}
        {% endif %}
```

Possible way of fixing: we may move this part outside the `union all`, or use `case` when type is `meta` then change the address.

Hosuke: I try to implement a simple fix in #5195 .

tomfutago: Ok, there is also extra logic on ethereum spell: `swap_type = 'underlying_exchange_base'` for `token_sold_amount`.

Just for the record tho, I'm not entirely convinced that traders are buying LP tokens in these transactions. I still think it's actually stables swap, but let's see what comes up from your testing.

Hosuke: But these trades are underlying exchange events. You can see in one single transaction there are a lot of swap events happened. The final outcomes may be stable swaps, but the middle steps (if we want to include them in the dex.trades), involve this 3CRV stabilizing mechanism.

tomfutago: Don't we want to see final outcome in `dex.trades`? I'd say that's what really matters to users/analysts. And that's what's implemented in old lineage too (obv could have been potentially a mistake that no one noticed at the time). I won't pretend that I understand all inner workings of curve tho, so will let you guys to decide where we're going with this.

Hosuke: Or we may fix the `amount_raw` to the corresponding tokens', either way we can verify the data in the explorer is ok.

tomfutago: No worries. I definitely wouldn't say you're interfering - it's actually good that you're questioning my solutions when you disagree, that's actually helpful! Keeps me in check and discussing is always good to clarify things. Even if we both disagree with each others solutions 😂 And don't worry about your English, it reads perfectly to me and btw English is my second language too.

jeff-dude:
closing to focus in #5214

PR #5195, opened 2024-01-22
Fix Proposal: Correct Token Mismatch in curvefi_optimism_base_trades

Hosuke:
**_NOTE: THIS FIX MAY BE WRONG._**

Based on my research, from this query:
https://dune.com/queries/3368710

We can see that, only when the `sold_token` is in `[USDC.e, USDT, DAI]`, the amount_raw in the transaction is matched to the `3CRV` (verified in the explorer).

Trades from `curvefi_optimism.MetaPoolSwap_evt_TokenExchangeUnderlying` are indicated by `pool_type = 'meta'` in the implementation.

### Data correctness checking:
Ensuring `token_address <-> amount_raw <-> evt_index` matches the info in explorers.

If this approach is acceptable I will try to use the similar approach to fix token mismatch in other chains' curve trades.

### Related discussion:
- https://github.com/duneanalytics/spellbook/pull/5184#discussion_r1459917047

### Other possible approach to fix this problem:
The current mismatched `token_sold_address` is actually indicated by `sold_id` in the `log`.
However, the amount_raw may not represent its actual trading amount, but expected (converting underlying).
Here may be the actual implementation of the function:
https://github.com/curvefi/curve-contract/blob/master/contracts/pools/gusd/StableSwapGUSD.vy#L606

cc. @tomfutago


Hosuke:
The current seed records are wrong in my point of view, we may need to fix them as well.

models/curvefi/optimism/curvefi_optimism_trades.sql
Hosuke, line 1: cc. @MSilb7 for reviewing.

jeff-dude:
closing to focus in #5214

PR #5197, opened 2024-01-22
Add dooar to `dex.trades_beta`

tomfutago:
linked to #4759


jeff-dude (approved):
got some momentum going on merges, so pushing this through
fyi @Hosuke 
thanks as always @tomfutago !

PR #5198, opened 2024-01-22
transfers: fix typo bug for native symbol

aalan3:
Small bug that removes double quotes around the native symbol.

Found by @hildobby.


jeff-dude (approved):
good call
@chuxinh also reported this, but we were waiting to implement with the cross-chain implementation. i'm good to push this through soon.

let's wait a bit, as prod is running a bunch of other modified objects this morning

macros/models/_sector/tokens/transfers_enrich.sql
```diff
@@ -21,7 +21,7 @@ SELECT t.blockchain
 , t."from"
 , t.to
 , t.contract_address
-, {{case_when_token_standard("'{{native_symbol}}'", 'tokens_erc20.symbol', 'NULL')}} AS symbol
```
aalan3, line 24: ah it's not possible to do `{{}}` inside of `{{}}` already, @jeff-dude do you know what to do here?

jeff-dude: s/o to @tomfutago for this idea in some of the latest sector spell designs

jeff-dude:
another note:
the spells which call this macro are materialized as views. i went to run the following, to prove our change:
```
select *
from test_schema.git_dunesql_a7a3523c_tokens_ethereum_transfers
limit 10
```

and the execution is spinning for awhile to return any results. not a promising sign on performance of views here. also could be why we're having trouble building the cross-chain, since it reads from all these views and unions them together.

macros/models/_sector/tokens/transfers_enrich.sql
aalan3: nice thank you!

aalan3:
I'll look into performance in a separate PR. Looks like it's due to prices again so I'll try a right join.

PR #5206, opened 2024-01-23
Use right join for transfers due to prices performance

aalan3:
Follow up on https://github.com/duneanalytics/spellbook/pull/5198 where @jeff-dude discovered that transfers are slow.

This can (hopefully) be fixed by selecting from the prices and right joining the transfers table instead.

### DIFF
--- a/macros/models/_sector/tokens/transfers_enrich.sql
+++ b/macros/models/_sector/tokens/transfers_enrich.sql
@@ -30,11 +30,12 @@ SELECT t.blockchain
     'NULL')}} AS usd_amount
 FROM {{ source('prices', 'usd') }} prices
 -- Right join due to performance reasons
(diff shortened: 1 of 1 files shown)


jeff-dude:
not sure if you tested, but fwiw the same query on previous PR is taking up to 30 minutes to run:
```
select
  symbol,
  count(1)
from
  test_schema.git_dunesql_d009ac8c_tokens_ethereum_transfers
where
  token_standard = 'native'
group by
  symbol

```

i actually just canceled it for now. not sure this rewrite of joins helps 🤔

aalan3:
Yes it didn't really help that query which was disappointing, not sure this PR helps at all tbh.

jeff-dude:
lets close for now & talk through considering materializing this level

PR #5207, opened 2024-01-23
Lido liquidity uniswap base, upstream spells' update

ppclunghe:
Upstream spells' update for Uni on Base wstETH pools


jeff-dude (approved):
sounds good!

Issue #4970, opened 2023-12-09
[BUG] Unexpected native token data in `tokens_zksync.transfers`


lgingerich:
### Description
1. zkSync Era emits token transfer events for ETH, which I believe not all EVM chains do. This means that ETH transfers are included in the `tokens_zksync.transfers` for `token_standard` = `erc20`. There is also data in this table for `token_standard` = `native`, which is where I would expect all ETH data to be.
2. The symbol field has the string literal "{{native_symbol}}" for `token_standard` = `native`. This applies for at least zkSync and Arbitrum, but likely stems from an issue in the macro and therefore carries through to all chains.

### Current behavior
Behavior for both of the above problems can be seen in the below picture.
1. For the ETH contract address (0x000000000000000000000000000000000000800a), there is data for both the native and erc20 token standards.
2. Unexpected `symbol` = `{{native_symbol}}`
![image](https://github.com/duneanalytics/spellbook/assets/82065267/a3dfffe5-1b34-4cd6-8bd6-71f2dad0075a)

This can be reproduced for a single tx_hash like:
```
SELECT
    tx_hash,
    contract_address,
    token_standard,
    symbol
FROM tokens_zksync.transfers
WHERE tx_hash = 0xfddfaefa8449a9a1e12668554e8f0411fe1f1748f718b8974587ffae592261d1
```

### Expected behavior
1. I would expect all ETH transfers to only be in the `token_standard` = `native` results. 
2. Expect `symbol` to show the actual symbol, not `{{native_symbol}}`.

### Impacted model(s)

```
1. `tokens_zksync.transfers`
4. Potentially all `tokens_{chain}.transfers` tables
```

### Possible solution
1. I'm aware that this token transfers table is fairly new, so I'm open to suggestions on how to fix this behavior.
2. Seems likely that the string literal is being passed through the macro rather than the actual variable. Likely a simple fix.

jeff-dude:
thank you for raising 🙏 

we haven't fully communicated out these new spells, but the goal was to have users start querying them and finding any potential bugs/enhancements/etc. -- this is a good head start.

fyi @aalan3 something we can look to fix. also, @lgingerich if you are interested in looking to submit a PR with a fix, we're happy to have ya contribute with us. let us know here

lgingerich:
Looks like issue 2 has been fixed. I'll check out issue 1

jeff-dude:
closing in favor of recent PRs to fix, such as #5212

PR #5212, opened 2024-01-23
Fix token standard in `transfers_base`

lgingerich:
The `erc20_transfers` table sometimes includes transfers for ETH. This is particularly in the case of zkSync which emits ERC20-like events for ETH transfers. To avoid labeling ETH transfers as `token_standard` = `erc20`, it is necessary to check the token `contract_address` before specifying the `token_standard`.

Ref: https://github.com/duneanalytics/spellbook/issues/4970

### DIFF
--- a/macros/models/_sector/tokens/transfers_base.sql
+++ b/macros/models/_sector/tokens/transfers_base.sql
@@ -34,7 +34,10 @@ WITH transfers AS (
     , t.evt_index
     , CAST(NULL AS ARRAY<BIGINT>) AS trace_address
     , t.contract_address
-    , '{{token_standard_20}}' AS token_standard
+    , CASE
+        WHEN t.contract_address = {{native_contract_address}} THEN 'native'
+        ELSE '{{token_standard_20}}'
+      END AS token_standard
     , t."from"
     , t.to
     , t.value AS amount_raw
--- a/macros/models/_sector/tokens/transfers_base.sql
+++ b/macros/models/_sector/tokens/transfers_base.sql
@@ -9,7 +9,7 @@ WITH transfers AS (
     , tx_hash
     , cast(NULL as bigint) AS evt_index
     , trace_address
-    {% if native_contract_address%}
+    {% if native_contract_address %}
     , {{native_contract_address}} AS contract_address
     {% else %}
     , CAST(NULL AS varbinary) AS contract_address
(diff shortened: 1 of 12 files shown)


lgingerich:
recheck

jeff-dude:
since you're modifying the macro, CI test will run all blockchain spells which call it. looks like all of them running at the same time is crashing the cluster due to heavy loads.

we could try to add `prod_exclude` tags to all chain spells which call this macro to exclude from running in CI. this would just be for testing purposes, to see CI complete and then validate the output CI tables to check data is as expected.

if the main focus is on zksync, then you could add the tag to all other chain versions in this PR to run only zksync. hope that makes sense.

lgingerich:
I'm not sure how to check output CI tables, but this does pass all tests now after excluding non-zksync chains. It's rather simple logic so I feel good about the data being correct, but of course best to check if possible

lgingerich:
Slightly unrelated, but I wanted to mention it here so I don't forget. Is there a reason that `native_contract_address` is set to null for all chains except ethereum and zksync?

jeff-dude:
> I'm not sure how to check output CI tables, but this does pass all tests now after excluding non-zksync chains. It's rather simple logic so I feel good about the data being correct, but of course best to check if possible

click here:
image

then dig through logs and find all the steps, with this one containing the table built during CI:
image

you can copy / paste that table on dune & query the data. this is v helpful for data validation during development

jeff-dude:
> Slightly unrelated, but I wanted to mention it here so I don't forget. Is there a reason that `native_contract_address` is set to null for all chains except ethereum and zksync?

not 100% sure actually, maybe @aalan3 has a quick answer?

lgingerich:
Nice, this is definitely very helpful.

I'm getting this error: `Error: The schema test_schema.git_dunesql_5c7a98f3_tokens_zksync_base_transfers has corrupted metadata. Please look for an equivalent schema in the data explorer, or recreate it.`

Does the table get deleted some amount of time after the CI run?

lgingerich:
recheck

aalan3:
>Slightly unrelated, but I wanted to mention it here so I don't forget. Is there a reason that native_contract_address is set to null for all chains except ethereum and zksync?

Some chains have a contract address convention while others don't, there are probably some chains that I missed this for.

jeff-dude:
> Does the table get deleted some amount of time after the CI run?

yes, every 24 hours at same time. i forget exact time, need to confirm that. if it doesn't return results, then simply rerun the CI gh action to rebuild

jeff-dude:
> Data in the CI temp table checks out, so I'll remove prod_exclude flags and then this should be ready to merge

sounds good, will get it queued up. thanks for helping validate this data!

jeff-dude (approved):

PR #5213, opened 2024-01-23
Add radiant to lending

tomfutago:


jeff-dude (approved):
looks good. agree with the schema file reordering

PR #5221, opened 2024-01-24
Remove incremental settings from Balancer Pools Labels

viniabussafi:
Following discussions on Telegram, this PR removes the incremental models from the balancer and beethoven X pools labels spells


jeff-dude (approved):
sounds good!

PR #5256, opened 2024-01-29
Drop another price token

aalan3:
Will use this PR to also test the CI validation which is failing for some reason


aalan3:
CI was apparently not an issue

PR #5258, opened 2024-01-29
Add EAS

tomfutago:
**Aim**: add EAS (Ethereum Attestation Service) spells under `attestation` sector and consolidate available data across all chains into 4 models using macro based approach.

**Models**:

1. `eas_schemas` - combines all EAS registered schemas:
     - materialized as incremental model
     - unique key: blockchain, project, version, schema_uid
     - captures schema registration details - most notable ones:
          - `schema_uid` - schema unique identifier
          - `registerer` - address that created the schema
          - `schema` - ABI encoded schema field types
          - `schema_array` - array of 2-element arrays representing schema pairs: [[data_type,field_name],[...]]

3. `eas_schema_details` - list of fields for each EAS registered schema:
     - materialized as view model (built on top of `eas_schemas`)
     - unique key: blockchain, project, version, schema_uid, ordinality
     - using `schema_array` field unnests each pair of `data_type` + `field_name` into a separate row

4. `eas_attestations` - single row per each issued attestation:
     - materialized as incremental model
     - unique key: blockchain, project, version, schema_uid, attestation_uid
     - key fields:
          - `attester` - address that made the attestation
          - `request` - attestation JSON request - that's where core attestation info is being extracted from
          - `raw_data` - ABI encoded data for the attestation
          - `attestation_state` set to 'valid'
          - `is_revoked` set to false
     - if attestation is revoked - the same attestation record is updated (merged on next incremental load) with updated info:
          - `revocation_time` - timestamp when attestation was revoked
          - `attestation_state` set to 'revoked'
          - `is_revoked` set to true

5. `eas_attestation_details` - my best attempt at decoding `raw_data` into human readable form using all the sql magic I was able to put together:
     - materialized as incremental model (built on top of `eas_schema_details` and `eas_attestations`)
     - unique key: blockchain, project, version, schema_uid, attestation_uid, ordinality
     - main data point: `raw_data` field (reminder: it's ABI encoded data) that contains all data compressed together into a hex string without reference to either `field_name` or `data_type` of each field, so no easy way to map it back to `schema` / `schema_array` field (as present in `eas_schemas` or `eas_schema_details` models)
     - my 1st attempt was to try decoding `raw_data` all in one go using lots of nested `bytearray_substring` type of approach, but with attestations against schemas containing 5+ fields (+ additional markers that come with ABI encoding) it resulted in not-so-elegant massive `case-when` construct that wasn't really working in the end (as in: decoded data was incorrect)
     - my 2nd attempt, and the one I decided to go with, is to unnest each `raw_data` into 64 characters (32 bytes) data chunks (each in individual row), so that each can be looked at separately (set-based approach). The steps that I'm following in this approach are as follows:
          - each 64-chars data chunk is thrown against `case when` that aims to find most basic data types: `bool`, `int` (for all ints, e.g. bigint, int128, uint256 etc are lumped together as `int` for simplicity sake), `address`, `bytes32` and `string`. Any other data type, including more complex ones, like arrays etc are ignored. If needed, users can alway fall back on `eas_attestations` table and do custom decoding of `raw_data` from there.
          - based on identified `data_type` - there's a series of `try_casts` / `regexp` expressions etc to uncover hidden treasures
          - some `data_types`, i.e. `string`, depending on the string length can be spread across several 64-chars data chunks, so that's where offset functions come in handy (`lag`/`lead`) to find groups of rows of similar `data_type` and then combine them together using `array_join` / `array_agg` functions
          - from my [tests](https://dune.com/queries/3404879) (mainly based on Arbitrum attestations) - 90% of records show decoded value and doing lots of spot checking using [easscan](https://easscan.org/) - all of them look to be matching. Remaining 10% that are blank after attempted decoding - fairly high for my liking, but it's hard to say how much of these records are blank because (1) they were not filled by the attester anyway, (2) failed decoding per methodology described above or (3) fall into more complex `data_type` category.

**Example**:
Attestation: [0xa6cd54a4d253cf831940ae2672abea5c2d0987fe9e7aeffd34e7bc8f990fe681](https://arbitrum.easscan.org/attestation/view/0xa6cd54a4d253cf831940ae2672abea5c2d0987fe9e7aeffd34e7bc8f990fe681)

`schema`: `string source,bool result`

`raw_data`:
```
0x000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000006474f4f474c450000000000000000000000000000000000000000000000000000
```

`raw_data` split into 64-chars data chunks + how each chunk is interpreted in the decoding process described above:
```
0x                                                               -- <ignored>
0000000000000000000000000000000000000000000000000000000000000040 -- <bytesize marker - ignored>
0000000000000000000000000000000000000000000000000000000000000001 -- True (decoded field)
0000000000000000000000000000000000000000000000000000000000000006 -- <string length (in next chunk) - ignored>
474f4f474c450000000000000000000000000000000000000000000000000000 -- GOOGLE (decoded field)
```

As this example shows, `raw_data` doesn't include `field_name` / `data_type` and encoded fields are not necessarily in the same order as specified in the `schema`.

References: [docs](https://docs.attest.sh/docs/welcome) | [easscan](https://easscan.org/)


jeff-dude (approved):
clean lineage here 🔥

Issue #5214, opened 2024-01-23
`dex.trades_beta` -- how to handle curve decimals logic


jeff-dude:
issue originally found while building new `amount_usd` design in dex lineage:
https://github.com/duneanalytics/spellbook/pull/5130#issuecomment-1895377596

first attempt at fixing from @tomfutago:
https://github.com/duneanalytics/spellbook/pull/5184

alternative attempt at fixing from @Hosuke:
https://github.com/duneanalytics/spellbook/pull/5195

for now, i will close both PRs so we can focus conversation within this issue. both will still be able to be referenced via the links above.

### issue overview
curve has added complexity relative to majority of dex's which feed into `dex.trades`. the specific piece of logic adding complications in the new design revolves around:
- what token to assign in bought / sold, when the 3CRV pool is involved, example on optimism [here](https://optimistic.etherscan.io/token/0x1337bedc9d22ecbe766df105c9623922a27963ec)
- the decimal values of tokens within the pool may vary (6 or 18), as well as the pool itself (18)
- existing code on ethereum [here](https://github.com/duneanalytics/spellbook/blob/main/models/curvefi/ethereum/curvefi_ethereum_trades.sql#L103-L104), which feeds `dex.trades` today, with a focus on decimal logic
- existing code on optimism [here](https://github.com/duneanalytics/spellbook/blob/main/models/curvefi/optimism/curvefi_optimism_trades.sql#L182-L183), which feeds `dex.trades` today, with a focus on decimal logic

in the new dex lineage, still in beta, this extra logic around decimals is missing since we decoupled the raw data from the enhancement metadata (tokens / prices), which contain the decimal values for dividing the raw amount into final amount. since we aren't overriding the decimal in the division to amount for curve, the `amount_usd` value is inflated to wildly high amounts (dividing by 6 rather than 18).

```
select blockchain, project, version, count(1)
from dex.trades_beta
where amount_usd >= 1000000000
group by blockchain, project, version
```
![image](https://github.com/duneanalytics/spellbook/assets/102681548/35d4983e-73ac-4dcf-872e-19856db048d1)

in the new dex lineage, we are **_not_** currently pulling through the fields from curve base level raw spells needed to write the correct amount logic in the final enhancement stage [here](https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/dex/enrich_dex_trades.sql#L46-L47).

### optimism vs. ethereum
- optimism reads from decoded tables
- ethereum reads from raw base tables
[...]

PR #5271, opened 2024-01-31
handle unique decimal logic for curve in `dex.trades_beta`

jeff-dude:
in reference to #5214 

fyi @Hosuke @tomfutago 

spoke a bit with andrew, who built curve on ethereum. preference is to keep logic as close as-is. the back story is that curve contracts on ethereum got all messy when decoded, so the tables are all over the place, hence the base table usage.
ultimately, approach in #5195 won't work, as we shouldn't default to use `3CRV` token necessarily.

i'm not attached to this approach either, but here is an attempt at getting curve logic implemented. this approach avoids new columns, but requires a new macro and modifying final `dex.trades_beta` to union both macros.


jeff-dude:
i haven't fully tested this yet, just wanted to get it out there.

tomfutago:
> i'm not attached to this approach either, but here is an attempt at getting curve logic implemented. this approach avoids new columns, but requires a new macro and modifying final `dex.trades_beta` to union both macros.

I like it. So branching off with project specific logic on the final stage in the lineage without messing up existing intermediate models. Nice!

I guess something similar could be adopted for balancer too, so that `amount_usd` gets calculated as per legacy and included in `dex.trades` rather than exposed view. And there's also 1inch in this ballpark perhaps too..

Hosuke (approved):
Since the decimal mismatch happened in certain `type`/`version`:
https://dune.com/queries/3389004

I think this fix can handle the edge cases well.

jeff-dude:
wrote a quick test query to see differences between new lineage vs. old
```
with
  new as (
    select
      blockchain,
      project,
      block_month,
      sum(token_bought_amount) as total_bought,
      sum(token_sold_amount) as total_sold,
      sum(amount_usd) as total_usd
    from
      test_schema.git_dunesql_22a881e_dex_trades_beta
    where
      blockchain in ('ethereum', 'optimism')
      and project = 'curve'
    group by
      blockchain,
      project,
      block_month
  ),
  old as (
    select
      blockchain,
      project,
      block_month,
      sum(token_bought_amount) as total_bought,
      sum(token_sold_amount) as total_sold,
      sum(amount_usd) as total_usd
    from
      dex.trades
    where
      blockchain in ('ethereum', 'optimism')
      and project = 'curve'
    group by
      blockchain,
      project,
      block_month
  )
select
  new.blockchain,
  new.project,
  new.block_month,
  new.total_bought - old.total_bought as bought,
  new.total_sold - old.total_sold as sold,
  new.total_usd - old.total_usd as usd
from
  new
  join old on new.blockchain = old.blockchain
  and new.project = old.project
  and new.block_month = old.block_month
order by
  new.blockchain,
  new.block_month desc
```

ethereum is pretty spot on. optimism has some differences. think we're good to merge for now, but we can revisit this query in final testing stages

PR #5281, opened 2024-02-02
Trusted Tokens Integration Proposal #2

Hosuke:
In contrast to implementation in #5130 ,
this approach may have better performance but also more coupling.


jeff-dude:
let's close this one and focus on the other. i like it being a macro so it can be used universally across spellbook, once we get it right.

PR #5290, opened 2024-02-04
Balances daily

aalan3:


macros/models/_sector/tokens/balances_daily.sql
```diff
@@ -0,0 +1,23 @@
+{%- macro balances_daily(balances_base) %}
+select
+    *,
+    {{ dbt_utils.generate_surrogate_key(['day', 'type', 'address', 'contract_address', 'token_id']) }} as unique_key
```
Hosuke, line 4: This line handles the `null` value in unique_key.

macros/models/_sector/tokens/balances_daily.sql
```diff
@@ -0,0 +1,23 @@
...
+from (
+    select
+        date_trunc('day', block_time) as day,
```
jeff-dude, line 7: explicitly cast this to `date`?

macros/models/_sector/tokens/balances_daily.sql
jeff-dude: just validating you are stating the use case for this surrogate key, not asking a question?

macros/models/_sector/tokens/balances_daily.sql
```diff
@@ -0,0 +1,23 @@
...
+        "address",
+        contract_address,
+        token_id,
+        amount,
+        row_number() OVER (partition by date_trunc('day', block_time), type, address, contract_address, token_id order by block_number desc) as row_number
+    from {{ balances_base }} balances
+    {% if is_incremental() %}
+    WHERE {{incremental_predicate('block_time')}}
+    {% endif %}
+) where row_number = 1
```
jeff-dude, line 20: is the row number to be safe, or known duplicate issues?

macros/models/_sector/tokens/balances_enrich.sql
```diff
@@ -1,5 +1,6 @@
-{%- macro balances_enrich(balances_base, blockchain) %}
+{%- macro balances_enrich(balances_base, blockchain, daily=false) %}
 select
+    {% if daily %}day,{% endif %}
```
jeff-dude, line 3: add table alias to be consistent with other fields selected?

jeff-dude (changes requested):

macros/models/_sector/tokens/balances_daily.sql
aalan3: This is to pick the latest value for each day, it's a way to do the aggregation it self instead of using `max/max_by`

PR #5293, opened 2024-02-05
Rollup Econ - Handle for Multiple OP Versions

MSilb7:
OP Mainnet batch submitters had used many combinations of to/from across versions, but the current 'Rollup Economics' spell only pulls the most recent combination, which leads to underreported L1 costs.

The OutputOracleProxy was also missed as the to address for state outputs (although this is super small).

This spell:
1. Adds `codebase` to `addresses_ethereum_l2_batch_submitters` for easier filtering (rather than maintaining a list of rollups).
2. Creates a view to generate all OP batch submitter address combinations (batchinbox and outputoracle combos), also handle for pulling the outpout oracle proxy for to addresses.
3. Adjusts the Rollup Econ spell to generate batchinbox and output oracle separately for Optimism chains

cc @Jam516 for eyes


Jam516:
makes sense to me 👍

jeff-dude:
marking as `wip` until CI tests are resolved

models/addresses/ethereum/addresses_ethereum_optimism_batchinbox_combinations.sql
```diff
@@ -0,0 +1,37 @@
+{{ config(alias = 'optimism_batchinbox_combinations',
```
jeff-dude, line 1: can you add `schema` in all config blocks?

jeff-dude (changes requested):

models/addresses/ethereum/addresses_ethereum_optimism_batchinbox_combinations.sql
MSilb7: on it

jeff-dude (approved):

PR #5262, opened 2024-01-30
Updates to bpt prices

viniabussafi:
## Thank you for contributing to Spellbook!

The [docs](docs/) directory has been implemented to answer as many questions as possible. Please take the time to reference each `.md` file within this directory to understand how to efficiently contribute & why the repo is designed as it is 🪄. Example questions to be answered:
- What does each property in the [model config block](docs/models/model_config_block.md) mean?
- What is the [CI test](docs/ci_test/ci_test_overview.md) attached to PRs and how can I best utilize it?
- What are the Spellbook [best practices](docs/general/best_practices.md)?

Please navigate through the [docs](docs/) directory to find as much info as you can.

**Note**: happy to take PRs to improve the docs, let us know!

PR #5304, opened 2024-02-06
[easy] Balancer schema names fix

viniabussafi:
As discussed in https://github.com/duneanalytics/spellbook/pull/5262#issuecomment-1929243133, this PR changes the config block to balancer spells, explicitly calling the blockchain for each one of them


couralex6:
Looks like that solved it! Thank you @viniabussafi !

PR #5322, opened 2024-02-09
New chain added to Bebop

B1boid:
Added BNB chain to bebop aggregator.
There should be no problems since the code is the same as for the current chains


Hosuke (approved):
LGTM.✅
Thank you @B1boid 

[Sample bebop_jam_bnb_trades]()

jeff-dude (approved):

PR #5360, opened 2024-02-16
finalize chain-level transfers spells

jeff-dude:


jeff-dude:
when excluding bnb/eth/poly, all other chains ran fine. trying just eth/poly next.

PR #5330, opened 2024-02-12
Create `cex_ethereum.flows`

hildobby:
# Thank you for contributing to Spellbook!

Thank you for taking the time to submit code in Spellbook. A few things to consider:
- If you are a first-time contributor, please sign the CLA by copy & pasting **_exactly_** what the bot mentions in PR comment
- Refer to [docs](#spellbook-contribution-docs) section below to answer questions
- Dune team will review submitted PRs as soon as possible

## Spellbook sub-projects

To prepare for Spellbook to scale in size, Dune team has started to pilot sub-projects:
- For a quick overview, see [the GH discussion on the topic](https://github.com/duneanalytics/spellbook/discussions/5238)
    - Please ask questions and continue the conversation there
- The first spell lineage to move: `tokens.erc20` and all upstream spells which help build it
    - **_Note_**: Please navigate to [tokens](https://github.com/duneanalytics/spellbook/tree/main/tokens) root directory to find tokens objects

## Spellbook contribution docs

The [docs](https://github.com/duneanalytics/spellbook/tree/main/docs) directory has been implemented to answer as many questions as possible. Please take the time to reference each `.md` file within this directory to understand how to efficiently contribute & why the repo is designed as it is 🪄

Example questions to be answered:
- What does each property in the [model config block](https://github.com/duneanalytics/spellbook/blob/main/docs/models/model_config_block.md) mean?
- What is the [CI test](https://github.com/duneanalytics/spellbook/blob/main/docs/ci_test/ci_test_overview.md) attached to PRs and how can I best utilize it?
- What are the Spellbook [best practices](https://github.com/duneanalytics/spellbook/blob/main/docs/general/best_practices.md)?

Please navigate through the [docs](https://github.com/duneanalytics/spellbook/tree/main/docs) directory to find as much info as you can.

**Note**: happy to take PRs to improve the docs, let us know 🤝

PR #5364, opened 2024-02-16
finalize chain-level transfers spells

jeff-dude:
`bnb` kept failing in prod for #5360 -- reverted in #5363 

try again when possible


jeff-dude:
interesting that the error wasn't cluster crashing, but rather:
`TrinoExternalError(type=EXTERNAL, name=HIVE_WRITER_CLOSE_ERROR, message="Error committing write parquet to Hive", query_id=20240216_231832_44578_4qyde)`
three times in a row.

macros/models/_sector/tokens/transfers_enrich.sql
```diff
@@ -7,33 +7,78 @@ CASE token_standard
...
+    base_transfers as t
+INNER JOIN
+    {{ evms_info_model }} as evms_info
+    ON evms_info.blockchain = t.blockchain
+LEFT JOIN
+    {{ tokens_erc20_model }} as tokens_erc20
+    ON tokens_erc20.blockchain = t.blockchain
+    AND tokens_erc20.contract_address = t.contract_address
+LEFT JOIN prices
+    ON prices.blockchain = t.blockchain
```
aalan3, line 78: it might help if this is static instead of using the transfer blockchain

jeff-dude: i can't even get the `base_` level to rebuild to test out running the `enrich` step. will keep this open to try once we get there though

hildobby:
@aalan3 @jeff-dude Any plans to add some index for native transfers? This would be very useful for uniqueness tests of downstream spells, here is an example of a spell that I currently can't add tests to: https://github.com/duneanalytics/spellbook/pull/5330

Either I test for a combo of column: `flow_type` (= 'inflow' or 'outlfow'), `block_number`, `tx_index`, `evt_index` or have no tests at all. The former currently doesn't work with so many txs having multiple native transfers (with a NULL `evt_index` value).

It would be cool to have some `sub_tx_index` value that is:
- `evt_index AS sub_tx_index` for all erc20/721/1155 tokens
- `CAST(ROW_NUMBER() OVER (ORDER BY "from", to, value) AS BIGINT) AS sub_tx_index` for native transfers (ordering for maximum reproducibility, maybe more columns to order with here)

I think this would end up being similar to the `sub_tx_index` that @0xRobin created in `nft.trades`

jeff-dude:
> @aalan3 @jeff-dude Any plans to add some index for native transfers? This would be very useful for uniqueness tests of downstream spells, here is an example of a spell that I currently can't add tests to: #5330
> 
> Either I test for a combo of column: `flow_type` (= 'inflow' or 'outlfow'), `block_number`, `tx_index`, `evt_index` or have no tests at all. The former currently doesn't work with so many txs having multiple native transfers (with a NULL `evt_index` value).
> 
> It would be cool to have some `sub_tx_trade_id` value that is:
> 
> * `evt_index AS sub_tx_index` for all erc20/721/1155 tokens
> * `CAST(ROW_NUMBER() OVER (ORDER BY "from", to, value) AS BIGINT) AS sub_tx_index` for native transfers (ordering for maximum reproducibility, maybe more columns to order with here)
> 
> I think this would end up being similar to the `sub_tx_trade_id` that @0xRobin created in `nft.trades`

certainly open to new columns to help this table be more useful. can you pull this branch locally, compile, grab a smaller chain output query (for performance reasons/easier testing) and add the column how you'd like to see it in plain sql? then share the query back here? that should help me get it added quickly

PR #5368, opened 2024-02-17
Linea data fees fix

lorenz234:
Linea recently started compressing their data before posting it to L1, this PR adds the new smart contract function to the query


lorenz234:
Also found that polygon ZKEVM has a new proxy contract

lorenz234:
I added the one additional function:'submitData' you had and I missed into this query @MSilb7, now Linea should be correct and this PR ready to merge @jeff-dude

PR #5374, opened 2024-02-18
EAS - add Scroll + tweaks

tomfutago:
Follow up to #5258:
- add Scroll
- update Base sources
- `eas_attestations` - tweak data types for fields that are currently varchars: `is_revocable`, `ref_uid`, `raw_value`, `expiration_time`
- `eas_attestation_details` - add `is_revoked` column

### DIFF
--- a/macros/models/_sector/attestation/eas.sql
+++ b/macros/models/_sector/attestation/eas.sql
@@ -117,65 +117,91 @@ src_EAS_evt_Revoked as (
   {% if is_incremental() %}
   where {{ incremental_predicate('evt_block_time') }}
   {% endif %}
+),
+
+attestations as (
+  select
+    '{{ blockchain }}' as blockchain,
+    '{{ project }}' as project,
+    '{{ version }}' as version,
+    ea.{{ schema_column_name }} as schema_uid,
+    ea.uid as attestation_uid,
+    ea.attester,
+    ea.recipient,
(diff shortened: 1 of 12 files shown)


tomfutago:
Hi @jeff-dude I saw your comment on #5444 - not really sure how that happened tho. Looks like 4 attestations on Optimism got bloated to thousands of rows. The only common thing about all 4 is that they are revoked, which might have something to do with the logic I'm using on incremental load with `{{ this }}` ref, so if you could cast an eye on `eas_attestations` macro if you see anything wrong with it, that'd be much appreciated 🙏 on the other hand, there are tons of other attestations that got revoked at different times and on different chains (including OP) and they are all looking fine, so again - not sure what happened for these 4 🤷‍♂️ I experimentally removed `prod_exclude` tag on this PR - and it all seems to be working fine.

tomfutago:
Oh and btw, I kept this PR in draft mode as I was hoping to update Base sources - but there is some issue with decoding EAS contracts for Base, so will open a separate PR once that's resolved.

jeff-dude:
> I experimentally removed `prod_exclude` tag on this PR - and it all seems to be working fine.

that's odd. must've been an issue with source data? we can let this go again and see if it happens in the future. if so, we can investigate further what the root cause is.

jeff-dude (approved):

PR #5383, opened 2024-02-20
Adds versioning to Balancer spells

viniabussafi:
This PR adds a missing versioning column to a few Balancer spells, as well as updates some joins where those spells are used and their respective schemas.
The reason of this change is to help us prepare our spells for the launch of Balancer V3 later this year


jeff-dude (approved):
sounds good 🫡

PR #5389, opened 2024-02-21
[URGENT] remove PEAKDEFI token

grkhr:
caused issues in volumes. low mcap token, seems to be hacked/scammed/whatever.

https://etherscan.io/tx/0xaffea464a1e3c351314de1294431bf7e2cabe07b382c1379a37a3f7c1e91aaaf
image


jeff-dude (approved):
we may need to develop a test for any new price added, to run a query joining to dex trades and ensure it doesn't blow up

PR #5392, opened 2024-02-21
Add sub_tx_trade_id to `token_{chain}.transfers` tables

hildobby:
Following up on https://github.com/duneanalytics/spellbook/issues/4759#issuecomment-1954818472

This PR introduces a column called `sub_tx_trade_id ` in order to have distinct id for each transfer. One thing to look into is that this will lead to some overlaps between erc20s' evt_index values and the synthetic evt_index created for native transfers. @jeff-dude @aalan3, is it more efficient to avoid this by:
- making native transfers' all negative
- making all of sub_tx_trade_id values be strings with `erc20_` or 'native_' as the start of each value
?


jeff-dude:
sorry that i went ahead and merged the transfers PR -- we wanted to test the performance in prod. we should be set up to handle changes and refresh them now, outside of bnb chain (3x as many transfers, causing issues, more to come there)

will try to catch up on your requests here tomorrow 🙏

aalan3:
@hildobby does the `sub_tx_transfer_id` need to be global for the block number? otherwise I think using row count for the block number (without tx hash, ideally we would have tx_index here to order) and for lookups using that and `type` as the key.

hildobby:
> @hildobby does the `sub_tx_transfer_id` need to be global for the block number? otherwise I think using row count for the block number (without tx hash, ideally we would have tx_index here to order) and for lookups using that and `type` as the key.

yeah good call, tx_hash isn't worth partitioning on.

Also, I agree with @0xRobin who suggests native transfers be just a bigint row_number (as is) with +1eX to avoid collusion. Just checked and current highest evt_index value of erc20 transfers across all evm chains is 56483 (https://dune.com/queries/3457207). So with that info, I propose +100000000 which should give us plenty of buffer space. We might need to update this once we have a new evm being indexed where erc20s reach very high evt_index values but we're far from that ever being a thing rn.

Just pushed a commit that should reflect all those changes. Should I go ahead and make the whole PR mergeable? current code change was more for demonstration purposes thus far so lmk if you're happy to go ahead with this and I'll push some commits @jeff-dude @aalan3

0xRobin:
After closer inspection it seems that we already have a `unique_key` in the tables. 
https://github.com/duneanalytics/spellbook/blob/4a1831f5a46de1b448ce16b9d00d88d7fc7a1899/macros/models/_sector/tokens/transfers_base.sql#L59-L60
@hildobby while it is not numeric, it should be enough to use as unique transfer id in downstream spells?

jeff-dude:
> After closer inspection it seems that we already have a `unique_key` in the tables.
> 
> https://github.com/duneanalytics/spellbook/blob/4a1831f5a46de1b448ce16b9d00d88d7fc7a1899/macros/models/_sector/tokens/transfers_base.sql#L59-L60
> 
> 
> @hildobby while it is not numeric, it should be enough to use as unique transfer id in downstream spells?

let me know if this field is sufficient, then we can close this PR.

if not, i'd like to combine merges for this potential new column along with `amount_usd` fix in #5406

hildobby:
> > After closer inspection it seems that we already have a `unique_key` in the tables.
> > https://github.com/duneanalytics/spellbook/blob/4a1831f5a46de1b448ce16b9d00d88d7fc7a1899/macros/models/_sector/tokens/transfers_base.sql#L59-L60
> > 
> > @hildobby while it is not numeric, it should be enough to use as unique transfer id in downstream spells?
> 
> let me know if this field is sufficient, then we can close this PR.
> 
> if not, i'd like to combine merges for this potential new column along with `amount_usd` fix in #5406

Thanks @0xRobin @jeff-dude, this should be sufficient! Could we stick to a standard name for this column? it's essentially the same as `sub_tx_transfer_id` so the two could use the same name, it's confusing to jump from one table to the next and have diff names when its not necessary

0xRobin:
> Thanks @0xRobin @jeff-dude, this should be sufficient! Could we stick to a standard name for this column? it's essentially the same as `sub_tx_transfer_id` so the two could use the same name, it's confusing to jump from one table to the next and have diff names when its not necessary

There are some differences here:
- `sub_tx_id` should be numeric and aims to preserve within-tx ordering, `unique_key` is a non-numeric hash computed from all the specified columns
- `sub_tx_id` is not enough for uniqueness constraints on its own, and should always be combined with `tx_hash`, while `unique_key` is a unique key on it's own (as the name suggests :smile:). 

I often prefer the `sub_tx_id` approach because it is better suited to use in seed tests and allows to more easily troubleshoot and reverse engineer any errors. That's just personal preference though.

hildobby:
> > Thanks @0xRobin @jeff-dude, this should be sufficient! Could we stick to a standard name for this column? it's essentially the same as `sub_tx_transfer_id` so the two could use the same name, it's confusing to jump from one table to the next and have diff names when its not necessary
> 
> There are some differences here:
> 
> * `sub_tx_id` should be numeric and aims to preserve within-tx ordering, `unique_key` is a non-numeric hash computed from all the specified columns
> * `sub_tx_id` is not enough for uniqueness constraints on its own, and should always be combined with `tx_hash`, while `unique_key` is a unique key on it's own (as the name suggests 😄).
> 
> I often prefer the `sub_tx_id` approach because it is better suited to use in seed tests and allows to more easily troubleshoot and reverse engineer any errors. That's just personal preference though.

Yeah I definitely agree on the preference, unique_key doesnt allow for as thorough of a testing as sub_tx_id does

jeff-dude:
closing this to finalize conversation in #5456

PR #5394, opened 2024-02-21
exclude pharaoh on avalanche_c due to missing sources

jeff-dude:
fyi @discochuck @Hosuke 
these are failing in prod due to missing source (likely a decoded table renamed)

please open a new PR with a fix and we can re-enable in prod. update the source yaml file and the source itself in the model.


jeff-dude:
failing source:
{{ source('glacier_avalanche_c', 'OdosRouterV2_evt_Swap') }}

PR #5406, opened 2024-02-22
bug fix: `amount_usd` for native transfers

jeff-dude:
in order to avoid joining prices more than once, use case logic in join conditions to differentiate native vs. erc20 transfers


macros/models/_sector/tokens/transfers_enrich.sql
```diff
@@ -75,11 +77,22 @@ LEFT JOIN
...
-            prices.contract_address = t.contract_address
-            OR (t.contract_address IS NULL AND prices.contract_address = evms_info.wrapped_native_token_address)
-        )
-    AND prices.minute = date_trunc('minute', t.block_time)
+LEFT JOIN
+    prices
+    ON date_trunc('minute', t.block_time) = prices.minute
+    AND CASE
+        WHEN t.token_standard = 'native'
+        THEN prices.blockchain IS NULL
```
aalan3, line 85: Can't these be one statement by using `AND` in between, similar to this https://github.com/duneanalytics/spellbook/blob/e89871de0a0e6a719382c32bdce5b01cf70e94d2/macros/models/_sector/tokens/balances_enrich.sql#L27-L31

jeff-dude: pushed a change to reformat the case statement. i don't mind the approach either way

PR #5443, opened 2024-02-26
`opensea_polygon_events` filter out bad tx

jeff-dude:


jeff-dude:
short term fix
fyi @0xRobin @aalan3 -- similar to ethereum over the weekend, but same fix won't work here

PR #5464, opened 2024-02-28
balances upgrade + daily aggregation with forward fill

0xRobin:
- rename some of the columns in the balances schema
- addition of daily aggregation with forward fill


0xRobin:
example query: https://dune.com/queries/3475840/5841282

macros/models/_sector/tokens/balances_daily.sql
```diff
@@ -1,23 +1,56 @@
...
+            ON  d.day >= b.day
+            and (b.next_update_day is null OR d.day < b.next_update_day) -- perform forward fill
+)
+
+select
+    b.*
+    ,b.balance * p.price as balance_usd
+from(
+    select * from forward_fill
+    where balance > 1
```
aalan3, line 49: Should this be`>0`?

macros/models/_sector/tokens/balances_enrich.sql
```diff
@@ -1,47 +1,34 @@
...
+from {{balances_raw}} balances
 left join {{ source('tokens', 'erc20') }} erc20_tokens on
-    erc20_tokens.blockchain = '{{ blockchain }}' AND (
+    erc20_tokens.blockchain = balances.blockchain AND (
     CASE
-        WHEN type = 'erc20' THEN erc20_tokens.contract_address = balances.contract_address
+        WHEN token_standard = 'erc20' THEN erc20_tokens.contract_address = balances.token_address
         -- TODO: should not be hardcoded
-        WHEN type = 'native' THEN erc20_tokens.contract_address = 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
+        WHEN token_standard = 'native' THEN erc20_tokens.contract_address = 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
```
aalan3, line 25: This will not work for Polygon I think, we should pass this in instead.

aalan3 (commented):
Looks great! had a couple of comments. Can you also update the description with a summary of the changes and PR name for posterity

macros/models/_sector/tokens/balances_enrich.sql
0xRobin: I have only changed the column name here.

aalan3: We use this for polygon https://github.com/duneanalytics/spellbook/blob/main/tokens/models/tokens/polygon/tokens_polygon_erc20.sql#L14

Most other L2s use ETH, we could actually also just pass in the the symbol and the decimal (it seems to be 8 for BNB?)

aalan3 (approved):

PR #5468, opened 2024-02-28
Zonic `nft.trades` fix

tomfutago:
This is to fix the following for zkSync:
- change default for ETH address from `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0000` to `0x000000000000000000000000000000000000800a` (passed through as `c_alternative_token_address` param in `zonic_base_trades` macro)
- add new parameter: `royalty_fee_receive_address_to_skip` which allows to pass a list of additional addresses to be skipped in the logic that determines `royalty_fee_receive_address`, so in case of zkSync that's `0x0000000000000000000000000000000000008001` which is a system contract (gas fee burn contract maybe?)

Additional notes:
- downstream relevance of `c_alternative_token_address` is that it links to `prices.usd_forward_fill` in `enrich_nft_trades` macro and `0xdeaddeaddeaddeaddeaddeaddeaddeaddead0000` exists in the table only for Optimism, so spells for Scroll and Base need updating too
- the reason for using `0x000000000000000000000000000000000000800a` is that it's listed in `prices.usd_forward_fill` table; alternative option is to use WETH address: `0x5AEa5775959fBC2557Cc8789bC1bf90A239D9a91`

cc: @lgingerich


lgingerich:
Looks good to me

tomfutago:
Updated `c_alternative_token_address` for Base as well to: `0x4200000000000000000000000000000000000006`

Scroll needs updating too but I don't see ETH/WETH for Scroll in `prices.usd_forward_fill`

tomfutago:
Hey @0xRobin this one is ready for review, can you give it a look when you have a chance plz?

0xRobin (approved):
good fix! :+1: 
We can add the WETH address `0x5300000000000000000000000000000000000004` to the scroll model once that is available in prices

jeff-dude (approved):

PR #5470, opened 2024-02-28
Add more tokens for zkSync prices

lgingerich:


lgingerich:
@jeff-dude do you understand what is failing with this? The CoinPaprika ID that it looks like it's failing on is correct, but maybe there's something else I'm missing about it.

jeff-dude:
> @jeff-dude do you understand what is failing with this? The CoinPaprika ID that it looks like it's failing on is correct, but maybe there's something else I'm missing about it.

the [warnings](https://github.com/duneanalytics/spellbook/actions/runs/8086697663/job/22097015962?pr=5470#step:6:81) can be ignored, as the API likely just doesn't output `zksync` yet on all these tokens. it finds a match for other chains though, so it can find a price from there.

there are some actual errors in the logs for inactive tokens though. you can likely remove those ones.

jeff-dude (approved):
thank ya

PR #5490, opened 2024-03-01
Add prices daily

aalan3:
This PR aggregates prices into a daily table to be used for other daily tables, which hopefully should significantly help performance.


jeff-dude (approved):
good idea for daily spells, will be good to keep in mind each time those are built

will queue up to merge in a bit, as there is a large 1inch refresh happening now

models/prices/prices_usd_daily.sql
```diff
@@ -0,0 +1,24 @@
...
+}}
+
+select * from (
+select
+    cast(date_trunc('day', minute) as date) as day,
+    blockchain,
+    contract_address,
+    decimals,
+    symbol,
+    price,
```
aalan3, line 20: After discussing with @0xRobin we should use avg here instead

0xRobin: update: added all of OHLC (open-high-low-close) price points to the schema

jeff-dude:
what's the plan here?

aalan3:
@jeff-dude the table looks good with the additions from @0xRobin.

jeff-dude (approved):

PR #5473, opened 2024-02-29
Create `dex.pools` Macros & Move to Sector - Ready for review

henrystats:
This pr adds dex.pools to the _sector/dex category.

I've set up macros that can be reusable by uniswap forks so it's easier to add new projects to the dex.pools model.

The code used in the macro was written by @hildobby [here](https://github.com/duneanalytics/spellbook/pull/5007#issuecomment-1971295693) and I've asked for his permission to reuse it in this pr for the macros. 

The code also allows for addition of pools with more than 2 tokens to dex.pools

PR #5497, opened 2024-03-03
Migrate `dex.pools` to new _sector structure

lgingerich:
This PR migrates Uniswap V2 on Ethereum to a new `dex.pools` structure that mirrors the recent `dex.trades` change.

I'm keeping this to only Uniswap V2 for now so as to keep the PR as simple as possible.

A few notes:
- `dex_pools_beta.sql` has no additional logic for now. I'm not too familiar with the history of the original `dex.pools` logic, but it looks like there may be some edge cases we need to add in here.
- Standardized column names across different tables are good. The [original `dex.pools` schema](https://github.com/duneanalytics/spellbook/blob/main/models/dex/dex_pools.sql#L38) has some "specialty" names and I believe we should standardize to the basic `block_time`, `block_number`, etc., even if it causes some downstream naming inconsistencies. 

I'm expecting comments and corrections, but hopefully my initial try at this isn't too far off. Let's get this migrated :)


henrystats:
Hola @lgingerich I somewhat already started working on migrating dex.pools to the new sector structure [here](https://github.com/duneanalytics/spellbook/pull/5473).

The pr is awaiting review. There's also a change to the current dex_pools structure to account for pools that may have more than 2 tokens!, I used @hildobby's code in a previous pr that he had submitted! It's also linked in my pr.

lgingerich:
@henrystats Awesome! Guess I should have checked open PRs before this lol, but I'll just close it now.

PR #5507, opened 2024-03-05
Add Garden Finance's $SEED and Ethereans $OS to prices

asnogordo:
EthereansOS rebranded to nonbeta and then back to EthereansOS, hence why the name on coinpaprika doesn't match, but the token contract does.


models/prices/ethereum/prices_ethereum_tokens.sql
```diff
@@ -322,11 +322,13 @@ FROM
     ('sand-the-sandbox', 'ethereum', 'SAND', 0x3845badade8e6dff049820680d1f14bd3903a5d0, 18),
     ('sdao-singularitydao', 'ethereum', 'SDAO', 0x993864e43caa7f7f12953ad6feb1d1ca635b875f, 18),
     ('sent-sentinel', 'ethereum', 'SENT', 0xa44e5137293e855b1b7bc7e2c6f8cd796ffcb037, 8),
+    ('seed-garden', 'ethereum', '$SEED', 0x5eed99d066a8caf10f3e4327c1b3d8b673485eed, 18),
```
jeff-dude, line 325: should symbol be just `SEED`?
https://etherscan.io/token/0x5eed99d066a8CaF10f3E4327c1b3D8b673485eED

models/prices/ethereum/prices_ethereum_tokens.sql
```diff
@@ -322,11 +322,13 @@ FROM
...
     ('shib-shiba-inu', 'ethereum', 'SHIB', 0x95ad61b0a150d79219dcf64e1e6cc01f0b64c4ce, 18),
     ('skl-skale', 'ethereum', 'SKL', 0x00c83aecc790e8a4453e5dd3b0b4b3680501a7a7, 18),
     ('slp-smooth-love-potion', 'ethereum', 'SLP', 0xcc8fa225d80b9c7d42f96e9570156c65d6caaa25, 0),
     ('sngls-singulardtv', 'ethereum', 'SNGLS', 0xaec2e87e0a235266d9c5adc9deb4b2e29b54d009, 0),
     ('snt-status', 'ethereum', 'SNT', 0x744d70fdbe2ba4cf95131626614a1763df805b9e, 18),
+    ('soon-nonbeta', 'ethereum', 'NON', 0x6100dd79fcaa88420750dcee3f735d168abcb771, 18),
```
jeff-dude, line 331: even if the coinpaprika id doesn't match, do you want to update the symbol?
https://etherscan.io/token/0x6100dd79fcaa88420750dcee3f735d168abcb771

jeff-dude (changes requested):

models/prices/ethereum/prices_ethereum_tokens.sql
asnogordo: Looks like when I submitted it to coinpaprika, I included the '$' I've requested they updated that. I saw it failed a test, but is that just a benign check?

models/prices/ethereum/prices_ethereum_tokens.sql
asnogordo: Same thing - is it okay to proceed with OS instead of NON? The failed check is what made me update it. Thank you for you help!

models/prices/ethereum/prices_ethereum_tokens.sql
jeff-dude: the test assumes the coinpaprika symbol is correct. if you updated there, we should be good to update here after. it's also not a huge deal, it can be updated later in the spell

models/prices/ethereum/prices_ethereum_tokens.sql
jeff-dude: yeah, we can proceed with `OS` even if test fails

models/prices/ethereum/prices_ethereum_tokens.sql
asnogordo: Sounds good. I've reverted to the original commit. Let me know if there are any other adjustments you'd like to see. Once again, I appreciate the input jeff-dude!

models/prices/ethereum/prices_ethereum_tokens.sql
asnogordo: Hey jeff, can you take another look when you get the chance? thank you!

jeff-dude: sorry about that, thanks for the nudge!
should be good, let me double check and mark ready for merge

jeff-dude (approved):

jeff-dude:
a recent merge added this token, i've removed the other one. test should pass now.

PR #5521, opened 2024-03-06
Migrate cex_ethereum.flows to a sector spell with macros to easily add new chains

hildobby:
Goal here is to create this macro so i can easily add cex_{chain}.flows for all other evm chains afterwards @jeff-dude

### DIFF
--- a/macros/models/_sector/cex/cex_flows.sql
+++ b/macros/models/_sector/cex/cex_flows.sql
@@ -1,14 +1,6 @@
-{{config(
-        schema = 'cex_ethereum',
-        alias = 'flows',
-        partition_by = ['block_month'],
-        materialized = 'incremental',
-        file_format = 'delta',
-        incremental_strategy = 'merge',
-        unique_key = ['flow_type', 'unique_key']
-        )}}
+{% macro cex_flows(blockchain, transfers, addresses) %}
 
-SELECT t.blockchain
+SELECT '{{blockchain}}' AS blockchain
 , CAST(date_trunc('month', block_time) AS date) AS block_month
 , block_time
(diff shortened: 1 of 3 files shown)


macros/models/_sector/cex/cex_flows.sql
```diff
@@ -29,8 +21,10 @@ SELECT t.blockchain
 , t.tx_hash
 , t.evt_index
 , t.unique_key
-FROM {{ ref('tokens_ethereum_transfers')}} t
-INNER JOIN {{ ref('cex_ethereum_addresses')}} a ON a.address IN (t."from", t.to)
+FROM {{transfers}} t
+INNER JOIN {{addresses}} a ON a.address IN (t."from", t.to)
```
jeff-dude, line 25: can you run a quick test on performance to see how runtime looks when you remove `IN` and do `(a.address = t."from" or a.address = t.to)`?

jeff-dude (commented):
music to my ears!

one quick comment below, if you don't mind checking performance

macros/models/_sector/cex/cex_flows.sql
jeff-dude: old run for reference:
https://github.com/duneanalytics/spellbook/actions/runs/8178326082/job/22361994788

can compare when new is done

jeff-dude: okay, no real change. feel free to keep as-is or revert to your preference

hildobby: it's good as is, feel free to merge

jeff-dude (approved):

PR #5527, opened 2024-03-07
Add BPT decimals to balancer.liquidity

viniabussafi:
This PR simply adds the decimals field on the CTE that queries the `bpt.prices` spell, as well as uses it to properly calculate BPT balances on pools


macros/models/_project/balancer/balancer_liquidity_macro.sql
```diff
@@ -58,11 +58,12 @@ WITH pool_labels AS (
         SELECT 
             day,
             contract_address AS token,
+            decimals,
             bpt_price
         FROM {{ ref('balancer_bpt_prices') }}
         WHERE blockchain = '{{blockchain}}'
         AND version = '{{version}}'
-        GROUP BY 1, 2, 3
+        GROUP BY 1, 2, 3, 4
```
Hosuke, line 66: If we group by all columns, maybe we can use `DISTINCT` here?

Hosuke (changes requested):

macros/models/_project/balancer/balancer_liquidity_macro.sql
viniabussafi: fair enough. Thanks for the review!

Hosuke (approved):
LGTM.✅
Thank you @viniabussafi

jeff-dude (approved):

PR #5534, opened 2024-03-08
Update EAS sources for Base + dupes fix

tomfutago:
* updated sources for Base (decoding issue on current ones)
* removed incremental_predicate on `eas_base_attestations` (all chains) - that was the cause for dupes. this applies only on this spell and not on the other 3, as `block_time` condition when applied on DBT_INTERNAL_DEST excludes revoked attestations that happened in different time window than original attestation and time span defined for `incremental_predicate`.


jeff-dude:
ah, came in here to merge in main and remove exclude tag, but of course you're on top of it 🙏

jeff-dude (approved):

PR #5545, opened 2024-03-10
Contracts - Add Top-Level Contract Address

MSilb7:
Add "top_level_contract_address" field (i.e. for Uniswap pools, this should show the pool factory.

Also simplifies this distinction by removing instances of "top level" from the base tables before iterating.


jeff-dude (approved):

jeff-dude:
will wait to merge this one alone, as we know it can take a bit and needs a few runs after to catchup

PR #5555, opened 2024-03-11
Adds new tokens to tokens.erc20

viniabussafi:
In an effort to keep tokens.erc20 as complete as possible, which will impact in many of our downstream spells, this PR adds a new batch of tokens to their respective chain-specific tables


jeff-dude (approved):
thanks for staying on top of it. we are trying to prioritize implementing the automated source for erc20 token metadata asap 🙏 

if you want to write some queries to test the token coverage, you can query `dune.definedfi.dataset_tokens` -- join to `evms.info` on `networkid` to find blockchain name

PR #5559, opened 2024-03-11
Element ex macro update | `nft.trades` | `EASY`

thetroyharris:
Change the CASE statement so we can get a USD price of the Element Ex trades on zksync.

My apologies @0xRobin for not adding you in as a contributor to this spell originally. This was merely an oversight and was in no way malicious. Thank you for making the spell and macro!


0xRobin (approved):
good catch! :heavy_check_mark:

jeff-dude (approved):

PR #5570, opened 2024-03-12
Add new balances_latest spell

0xRobin:
Add new balances_latest spell


models/_sector/tokens/avalanche_c/tokens_avalance_c_balances_latest.sql
```diff
@@ -0,0 +1,13 @@
+{{ config(
```
aalan3, line 1: This file can be dropped. `avalance`

models/_sector/tokens/avalanche_c/_schema.yml
```diff
@@ -190,3 +190,26 @@ models:
...
+      sector: tokens
+      contributors: 0xRob
+    config:
+      tags: [ 'tokens','balances', 'avalanche_c' ]
+    description: "Latest token balances"
+    columns:
+      - name: day
+      - name: block_date
+      - name: block_number
+      - name: type
```
aalan3, line 206: `token_standard`

aalan3: I think the schema here is copied from the raw balances but it should be the daily.

aalan3 (commented):
Had a few comments. I think this approach is good as a start. Later we can do a aggregation (`max_by`) which we may even be able to materialize daily

jeff-dude:
is this still planned or should we close?

0xRobin:
Closing as we don't have a good setup for this at the moment.
The main problem is correctly materializing the "next_update" column
1. The cluster is breaking due to memory usage. We could revisit once we have a more incremental way of handling these big big models. 
2. The way I wanted to calculate the `next_update` column breaks if we have gaps of non-updates that are bigger than the current incremental window. We could work around this a bit by using a self reference and retrieving the rows that need updates, but that's a bigger lift (and probably not prio at the moment..)

PR #5585, opened 2024-03-13
Create `dex.liquidity` spell 

henrystats:


0xRobin:
I started a discussion around a higher level tvl/balances spell here: https://github.com/duneanalytics/spellbook/issues/5666
This should be able to provide you with the same data.
@henrystats please take a look and provide any comments/feedback you have.

PR #5593, opened 2024-03-13
Local profile for tokens subproject

couralex6:


jeff-dude (approved):
fyi @aalan3 @Hosuke @0xRobin 
for each future subproject, we will need to have the `profiles.yml` copied into the directory so users can run `dbt compile` locally still. note they will need to `cd` into the subproject directory for those models to compile

PR #5595, opened 2024-03-13
SPE-350 introduce automated source for `tokens.erc20` & revamp the entire lineage

jeff-dude:
## goal of this PR:
- introduce new automated source on dune for erc20 token metadata
- clean up old static files
  - reformat each file to be consistent
  - remove all tokens which exist in automated source from static files
    - reduces manifest file size
    - makes files more readable / manageable
- union automated source with static files
  - use `union distinct` to ensure no dupes
- establish `evms.info` as a source, to be able to use in tokens subproject
  - this is a "shortcut" for now, as we will eventually move all of `evms` into a subproject and will be universally a source in spellbook
- de-duplicate the automated source with `row_number()`
- compare performance to old spell
  - compile `dex_trades_beta`, run with old tokens and see runtime, switch to new source and cross-check run times
 

## query support
[this query](https://dune.com/queries/3123536) is used for a variety of analysis. most recently used to build format for chain-level static files, to only output tokens which are missing in automated source.


jeff-dude:
looks like more digging to do to figure out how to leverage these sources
`hive` is still being appended to the front, causing errors.

tokens/models/tokens/tokens_erc20.sql
```diff
@@ -1,34 +1,80 @@
...
+    (
+        {% for key, value in static_models.items() %}
+        select
+            '{{ value.blockchain }}' as blockchain
+            , contract_address
+            , symbol
+            , decimals
+        from
+            {{ value.model }}
+        {% if value.blockchain == 'optimism' %}
```
aalan3, line 62: Could you explain why this is needed?

jeff-dude: optimism lineage for erc20 tokens is a combo of manually curated and dynamic queries. it differs from all other chains. i don't want to mess with that logic until the new design is in place, then we can come back to optimism to simplify as much as possible
cc @MSilb7

macros/dune/source.sql
jeff-dude, line 1: please note, #5603 is being merged separately with this change for a clean git history
once merged, i will rebase this branch so it drops out

jeff-dude:
marking this as blocked now. here is where we stand:
- when running locally, the code properly compiles and reads from `dune.definedfi.dataset_tokens` as expected
- when pushing to PR and CI runs, something in the CI workflow appends `_hive` to the database name, resulting in `dune_hive.definedfi.dataset_tokens`, which doesn't exist and CI fails to find / read it as expected

what is happening in the CI workflow to cause that `_hive` to be appended? 🤔

0xRobin (commented):
@jeff-dude Is `dune` the database name here? (equivalent to `hive`?) 
Or does the table live in `hive.dune.definedfi.dataset_tokens` when you specify the full identifier? :thinking:

jeff-dude:
> @jeff-dude Is `dune` the database name here? (equivalent to `hive`?) Or does the table live in `hive.dune.definedfi.dataset_tokens` when you specify the full identifier? 🤔

`dune` is the database name, so it should override `hive` completely (which is the default db in the dbt profile for CI

0xRobin:
Thinking we can override the database source the correct dbt way as done here:
https://github.com/duneanalytics/spellbook/blob/5427aaf4ca1b0006fba64b260743dab7146f01f7/sources/_datasets/definedfi/_sources.yml#L3-L7

But, that still leaves us with the `_hive` addition, I've done some digging but can't find where it's coming from, might be the schema change code from the spellbook runner? (I can't access that)

jeff-dude:
> Thinking we can override the database source the "correct dbt way" as done here:
> 
> https://github.com/duneanalytics/spellbook/blob/5427aaf4ca1b0006fba64b260743dab7146f01f7/sources/_datasets/definedfi/_sources.yml#L3-L7
> 
> But, that still leaves us with the `_hive` addition, I've done some digging but can't find where it's coming from, might be the schema change code from the spellbook runner? (I can't access that)
> 
> a possible solution might be to override `generate_database_name`.. 🤔

i was doing similar digging yesterday. turns out it's on the engine side / our catalog service. working with them now to get this figured out. seems like no issue on the dbt side.

jeff-dude:
it's working now 👀 

next i'll need to run some queries to test the output and ensure expected coverage, test performance, etc.

jeff-dude:
### performance testing
compiled `dex.trades`, added filter for 2024/01/01 forward
with old tokens: ~6 minutes to return results
with new tokens: ~5 minutes to return results

no noticeable performance regression with larger token metadata table ✅ 

### coverage testing
```
select blockchain, contract_address
from tokens.erc20
except
select blockchain, contract_address
from test_schema.git_dunesql_a50557b_tokens_erc20
```

no returned rows ✅

jeff-dude:
i'm comfortable for merge here, i'll let this sit overnight for everyone to revisit
@Hosuke @0xRobin @aalan3 @couralex6

0xRobin (approved):

tokens/package-lock.yml
```diff
@@ -0,0 +1,8 @@
+packages:
```
aalan3, line 1: Should this be here, any context?

jeff-dude: this file auto-generates when you run `dbt clean`, `dbt deps`, `dbt compile`
since tokens is a new project, we didn't have on in there
you will notice the main spellbook project has the same file:
https://github.com/duneanalytics/spellbook/blob/main/package-lock.yml

jeff-dude: first few commits, i manually removed the file, but it kept coming back. i ended up including in final.

PR #5600, opened 2024-03-14
Balances: change join order to improve performance

aalan3:
This PR changes the order of the joins so that nft metadata is joined first. This seems to create a pretty significant performance gain.


0xRobin (commented):
interesting! how are we evaluating the performance?

aalan3:
@0xRobin you should try it some of your queries to see if this has an effect. 

I've tested the following: 
- Top nft holders: https://dune.com/queries/3517783/5914981 went from 20m to 5m (with prices daily it goes to 1 in)
- Running a query that joins addresses from contracts (for example uniswap tvl) works better now, still not fast with daily but with the non time series table it might actually work now. I'm still investigating this more

PR #5602, opened 2024-03-14
Balances test with daily prices

aalan3:
Follow up on #5600. 

- Testing with daily prices
- Testing with materializing days


jeff-dude:
closing for now

PR #5603, opened 2024-03-14
prep source macro for dune catalog uploads

jeff-dude:
intention of this PR:
- enable capability for spellbook to read data upload tables which are prefixed with `dune` catalog (i.e. `dune.definedfi.dataset_tokens`)
- introduce a `dune` flag in the macro
  - default to `False`
  - apply `True` in sources which need this catalog
  - example: `{{ source("definedfi", "dataset_tokens", dune=True) }}`
- add new `if` condition in macro to look for this flag to be true
  - if true, then override database/catalog to be `dune`


jeff-dude:
canceling CI as it forces all models to run

Hosuke (approved):

macros/dune/source.sql
```diff
@@ -1,9 +1,12 @@
-{% macro source(source_name, table_name) %}
+{% macro source(source_name, table_name, dune=False) %}
```
aalan3, line 1: Could we change this to be `database="delta_prod"` and then we can override it with `database="dune"` where needed.  that way the code below will only change to `database=database` and we don't need the extra if.

jeff-dude: done, good call

we may eventually be able to remove the `if trino` concept altogether, but i'll save for later. i have a task in triage to go through all references to catalog usage in macros to document/understand/enhance

PR #5605, opened 2024-03-14
[EASY] syncswap update | `dex.trades` 

thetroyharris:
Changes:
- Use a decoded table now instead of raw logs
- Taker is now determined by the `to` column in the decoded table and not the raw log column `tx_from`

Ready for review @Hosuke


jeff-dude (approved):
nice, thank you!

PR #5607, opened 2024-03-15
`check_dex_info_relationship` macro

tomfutago:
**Aim**: add generic `check_dex_info_relationship` test on all `dex_<chain>_base_trades` models.

**Rationale**: `dex.info` often gets forgotten when adding new dex and current test only picks it up on building main `dex_base_trades` model, which rarely gets updated in the same PR so existing test is not triggered. On the other hand `dex_<chain>_base_trades` model is always updated when new dex is added, so test in this PR will pick up missing entries early.

**Note**: this could be also resolved by using the same `relationships` test as in `dex_base_trades` model, but I think macro is a bit cleaner.


jeff-dude:
at first glance, makes sense to me. let's have hosuke take a look and see if you guys agree 🤝

models/_sector/dex/trades/arbitrum/_schema.yml
```diff
@@ -2,6 +2,8 @@ version: 2
 
 models:
   - name: dex_arbitrum_base_trades
+    tests:
+      - check_dex_info_relationship
```
Hosuke, line 6: Do you want to use the `relationships` test here?
It may be one test file less.

But either way works for me.

Hosuke (approved):

models/_sector/dex/trades/arbitrum/_schema.yml
tomfutago: I don't mind either (and here comes 😉) but:

1. we're using kinda similar approach of adding more tests with custom seed check macros
2. we'll have 1 test file less, but more lines in each `_schema.yml` file:

with macro:
```
  - name: dex_arbitrum_base_trades
    tests:
      - check_dex_info_relationship
```

with built-in `relationships` test:
```
  - name: dex_arbitrum_base_trades
    columns:
      - name: project
        tests:
          - relationships:
              to: ref('dex_info')
              field: project
```

Hosuke: Convinced me.

jeff-dude (approved):
glad you guys agree, this is good with me -- thank you!

PR #5520, opened 2024-03-06
feat: add bonkbot trades spellbook

whalehunting:
This PR adds a dedicated spellbook for Bonkbots Solana DEX trades, since the data got too big to execute on the query engine itself.

Let me know if there are any changes/adjustments to be made!

PR #5614, opened 2024-03-17
feat: add banana gun trades spellbook

whalehunting:
This PR adds a dedicated spellbook for Bannana Gun Solana DEX trades. Banana Gun is a DEX trading bot (not a dex), which allows users to execute trades using a Telegram chatbot.

Similar to #5520


models/banana_gun/solana/banana_gun_solana_trades.sql
```diff
@@ -0,0 +1,166 @@
...
+      version,
+      token_pair,
+      project_program_id AS project_contract_address,
+      trader_id AS user,
+      trades.tx_id,
+      tx_index,
+      outer_instruction_index,
+      inner_instruction_index
+    FROM
+      {{ ref('dex_solana_trades') }} AS trades
```
Hosuke, line 77: Is `banana_gun` can be considered as a dex aggregator?

jeff-dude: feedback in this PR should mimic the feedback in linked PR in description
not adding to `dex.trades`, but reading from it. eventually these spells can form a cross-chain sector level spell for bots

models/banana_gun/solana/banana_gun_solana_trades.sql
```diff
@@ -0,0 +1,167 @@
...
+   )
+}}
+
+{% set project_start_date = '2024-01-08' %}
+{% set fee_receiver_1 = '8r2hZoDfk5hDWJ1sDujAi2Qr45ZyZw5EQxAXiMZWLKh2' %}
+{% set fee_receiver_2 = 'Cj297UauzMX64FU9dKJZRUBWszJ7tEWpVheasq4CfATV' %}
+{% set fee_receiver_3 = 'HKMh8nV3ysSofRi23LsfVGLGQKB415QAEfZT96kCcVj4' %}
+{% set fee_receiver_4 = '7tQiiBdKoScWQkB1RmVuML7DBGnR31cuKPEtMM7Vy5SA' %}
+{% set fee_receiver_5 = '4BBNEVRgrxVKv9f7pMNE788XM1tt379X9vNjpDH2KCL7' %}
+{% set wsol_token = 'So11111111111111111111111111111111111111112' %}
```
Hosuke, line 18: Maybe we can design a macro for solana bot trades.

whalehunting: what would this look like? also not all bots handle fees in the same way, so the logic to find their trades can differ from bot to bot

Hosuke: Let me study these cases first...

jeff-dude: let's not force a macro, if it requires too many "if this, then that" type coding within. take an hour or two to dig into this @Hosuke, see if possible, then decide if it's worth the effort or not

it's more important that we follow the design for efficiency in orchestration at the end of the day (base level raw data, metadata joins later downstream, etc etc)

Hosuke (approved):
LGTM.✅
Thank you @whalehunting 

[Sample banana_gun_solana_bot_trades]()

models/banana_gun/solana/banana_gun_solana_bot_trades.sql
```diff
@@ -0,0 +1,167 @@
...
+        {% endif %}
+      )
+      JOIN {{ source('solana','transactions') }} AS transactions ON (
+        trades.tx_id = id
+        {% if is_incremental() %}
+        AND {{ incremental_predicate('transactions.block_time') }}
+        {% else %}
+        AND transactions.block_time >= TIMESTAMP '{{project_start_date}}'
+        {% endif %}
+      )
```
Hosuke, line 97: By the way do you think we can make a CTE before inner join to optimize the query like
```
solana_tx as (
    SELECT [ * | neccessary columns ] 
    FROM
    {{ source('solana','transactions') }}
    {% if is_incremental() %}
    WHERE {{ incremental_predicate('transactions.block_time') }}
    {% else %}
    WHERE block_time >= TIMESTAMP '{{project_start_date}}'
    {% endif %}
```

jeff-dude (approved):
this looks good to me

PR #5620, opened 2024-03-18
Init gyro_tokens

viniabussafi:
This PR creates a proxy for the price of the tokens that were submitted on https://github.com/duneanalytics/spellbook/pull/5619#event-12155682375, that were recently launched and will be featured in gyroscope pools

@mendesfabio


models/gyroscope/gyro_tokens.sql
```diff
@@ -0,0 +1,23 @@
+{{ config(
+    schema = 'gyroscope',
+    alias = 'gyro_tokens'
```
Hosuke, line 3: If the schema is `gyroscope`,
then the model name would be `gyroscope_gyro_tokens`.

Hosuke (changes requested):

models/gyroscope/gyro_tokens.sql
viniabussafi: yup, sorry for the oversight

jeff-dude (approved):
if i understand correctly, when these tokens are traded enough on DEXs, eventually the price derived in `dex.prices` will suffice and you can essentially revert this PR?

viniabussafi:
Exactly, @jeff-dude! This is just a way to get things right as soon as pools are launched

PR #5633, opened 2024-03-20
Introduce dex_solana_bot_trades

Hosuke:
In this PR:
- add new spell `dex_solana.bot_trades`
- rename existing bot trades alias to `bot_trades`

### DIFF
--- a/models/dex/solana/dex_solana_bot_trades.sql
+++ b/models/dex/solana/dex_solana_bot_trades.sql
@@ -1,0 +1,50 @@
+{{
+    config(
+        schema = 'dex_solana',
+        alias = 'bot_trades',
+        materialized = 'view',
+        post_hook='{{ expose_spells(\'["solana"]\',
+                                    "project",
+                                    "dex",
+                                    \'["hosuke"]\') }}'
+    )
+}}
+
+
+
+{% set solana_trading_bot = [
+    ref('bonkbot_solana_bot_trades')
+] %}
+
+{% for bot in solana_trading_bot %}
+SELECT block_time,
+       block_date,
+       block_month,
+       blockchain,
+       amount_usd,
+       type,
+       token_bought_amount,
+       token_bought_symbol,
+       token_bought_address,
+       token_sold_amount,
+       token_sold_symbol,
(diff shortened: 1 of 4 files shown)


models/dex/solana/dex_solana_bot_trades.sql
```diff
@@ -0,0 +1,50 @@
+{{
+    config(
+        schema = 'dex_solana',
+        alias = 'bot_trades',
+        materialized = 'view',
+        post_hook='{{ expose_spells(\'["solana"]\',
+                                    "project",
+                                    "dex",
+                                    \'["hosuke"]\') }}'
```
jeff-dude, line 9: can we make it a habit to use the arg names in this `expose_spells` post hook macro, similar to one i linked to in merged PR today for balances?

we don't need to worry about applying to all existing spells (yet), but rather focus on making it a practice moving forward

models/dex/solana/dex_solana_schema.yml
```diff
@@ -91,4 +91,65 @@ models:
...
+      description: >
+         Trades by trading bots on Solana
+      tests:
+        - dbt_utils.unique_combination_of_columns:
+            combination_of_columns:
+              - blockchain
+              - tx_id
+              - tx_index
+              - outer_instruction_index
+              - inner_instruction_index
```
jeff-dude, line 112: if this remains a view, we can remove this test. no need to rerun on a view.

jeff-dude (approved):
we should consider moving this new `bot_trades` spell lineage and the `dex_solana` spell lineage to the new `_sector/dex/` directory path

that way, when we migrate `dex` sector spells to a spellbook subproject to operate on it's own, we can include solana spells, even though they aren't integrated into evm yet.

with that said, we don't have to force on this PR. if you could track as a task in the related gh issue, we can focus on ensuring it's done prior to calling that issue done

Hosuke:
We can continue finishing the current opening `bot_trades` PRs, and then move all relevant spells into `_sector/dex/` in the finalizing PR.

jeff-dude (approved):

PR #5640, opened 2024-03-21
Add coinpaprika tokens to base and fantom

tiagofilipenunes:
Adding more tokens to base and fantom that were recently added to coinpaprika. Tested their API that the ids existed and price information was available.


jeff-dude (approved):
thank you for the due diligence in testing, looks good!

PR #5313, opened 2024-02-08
Added boost_deployed, boost_claimed and boost_completors model

pntemi:
Boost on [boost.xyz](https://boost.xyz), an open source platform to incentivize any onchain action, with any token, on any chain.

In this PR, there are 3 resulting final spells:
- `boost.boost_deployed` : The Boost deployed on Boost Protocols
- `boost.boost_claimed`: Reward claimed after Boost users performed qualifying actions
- `boost.boost_completors`: Summary and stats for each Boost user

There are other spells, per network, which are the smaller building blocks for the 3 spells mentioned above.

PR #5643, opened 2024-03-22
Remove incremental predicate on the join with boost_deployed

pntemi:
This is a follow up PR to the previous [PR on Boost's spells ](https://github.com/duneanalytics/spellbook/pull/5313).
One of the models, `boost.claimed` has a join with `boost.deployed`. With incremental predicate, it's missing a lot of data (see [this](https://dune.com/queries/3551367)) because it only looks at the most recent date, not the entire dataset.

This PR is to remove that because the join actually needs the entire `boost.deployed` model.  `boost.deployed` is also very light (<20k rows)


jeff-dude (approved):
yep, makes sense. let's get that fix in for ya

PR #5646, opened 2024-03-22
clean up opensea bugs and standardize for short-term

jeff-dude:
fyi @0xRobin @aalan3 @couralex6 @Hosuke 
based on short terms fixes we've done over the last month, i'm attempting to "standardize" until we have long term solution.
- in v3, decoded table changed column name due to new submission with updated ABI that no longer contains column names as it used to -- my understanding is the address is a proxy and re-used for new versions of the contract, where ABI can differ, breaking things for us and all users of this table
- in v4, add the short term dedupe fix to polygon to match ethereum

### DIFF
--- a/macros/models/_sector/nft/platforms/seaport_v3_trades.sql
+++ b/macros/models/_sector/nft/platforms/seaport_v3_trades.sql
@@ -285,7 +285,7 @@ with source_ethereum_transactions as (
                   ,contract_address as platform_contract_address
             from (select *
                     from {{ Seaport_call_matchOrders }}
-                    cross join unnest(output_executions) with ordinality as foo(execution,execution_idx)
+                    cross join unnest(output_0) with ordinality as foo(execution,execution_idx)
                    where call_success
                      and contract_address = 0x00000000006c3852cbef3e08e8df289169ede581  -- Seaport v1.1
                  {% if not is_incremental() %}
--- a/macros/models/_sector/nft/platforms/seaport_v3_trades.sql
+++ b/macros/models/_sector/nft/platforms/seaport_v3_trades.sql
@@ -285,7 +285,11 @@ with source_ethereum_transactions as (
                   ,contract_address as platform_contract_address
             from (select *
                     from {{ Seaport_call_matchOrders }}
+                    {% if blockchain in ('ethereum', 'polygon') %}
                     cross join unnest(output_0) with ordinality as foo(execution,execution_idx)
+                    {% else %}
+                    cross join unnest(output_executions) with ordinality as foo(execution,execution_idx)
+                    {% endif %}
                    where call_success
                      and contract_address = 0x00000000006c3852cbef3e08e8df289169ede581  -- Seaport v1.1
                  {% if not is_incremental() %}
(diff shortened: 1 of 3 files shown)


models/opensea/polygon/opensea_v4_polygon_events.sql
```diff
@@ -33,6 +33,13 @@ from trades
...
+-- temporary fix to exclude duplicates
+and tx_hash not in (
+select tx_hash from
+trades
+group by tx_hash, evt_index, nft_contract_address, token_id, sub_type, sub_idx
+having count(*) > 1
+)
+--and tx_hash != 0x42e8f1d5dca4d45678608c58c1f8d0670513787c2acd92a9dc024e5780664121
+--and tx_hash != 0x25ce8a1559a48da77eb55c1b95ee0d8c7b6b4c4d9eece89ac00640db54dfac63
+--and tx_hash != 0xeaa57f3622702b9058ed7529f437aed755a26f132934a38884012c9a4e78799c
```
jeff-dude, line 45: will remove these if CI passes

macros/models/_sector/nft/platforms/seaport_v3_trades.sql
```diff
@@ -285,7 +285,7 @@ with source_ethereum_transactions as (
                   ,contract_address as platform_contract_address
             from (select *
                     from {{ Seaport_call_matchOrders }}
-                    cross join unnest(output_executions) with ordinality as foo(execution,execution_idx)
+                    cross join unnest(output_0) with ordinality as foo(execution,execution_idx)
```
0xRobin, line 288: @jeff-dude I don't think we should put `output_0` (or `_0`, `_1`) arguments in the macros, we should have the proper ABIs that give us the correct names for input and output arguments. 
So I believe this should be fixed in the decode step, not in spellboook.

0xRobin (changes requested):

macros/models/_sector/nft/platforms/seaport_v3_trades.sql
jeff-dude: my understanding after talking with decode team is that the fix isn't as straightforward as we'd like. this PR is aiming for short-term fixes across the board, so i agree that ideally we follow up on this quickly and fix for the long term. this unblocks us for the time being

jeff-dude: we will follow up asap, @aalan3 and i think we should have all these things prioritized for next week, with you leading the effort / us supporting where needed

jeff-dude: this won't work in macros anyway without if/else logic 🫠

aalan3: @0xRobin this is a temporary fix so that we can get through the weekend without removing these models. We will have to fix this next week. 

Regarding the names, it appears that the latest ABI for [this contract](https://etherscan.io/address/0x0000000000000068f116a894984e2db1123eb395#code) has removed the names? Otherwise if this contract is not relevant we can look into removing it and reverting.

jeff-dude:
just reiterating prior to merge, not an ideal solution but short term to get us into next week

PR #5655, opened 2024-03-23
add prod address and rename staging to barn

harisang:
This is a follow-up to PR #5463, that adds the prod version of SeaSolver_v2, and slightly renames the solver, as well as changes the environment to the more standard `barn` terminology.


harisang:
@jeff-dude Should be ready for review and merging

jeff-dude (approved):
sounds good 🤝

PR #5672, opened 2024-03-25
Add seaport v1.6

0xRobin:
Seaport 1.6 is slated to be used from March 25th on OpenSea, this should be all we need to include the new trades :pray:


0xRobin:
@jeff-dude would be good to look to merge this asap as trades are already coming into the v1.6 contracts.

jeff-dude (approved):
i got ya, thanks for coordinating with the other

PR #5684, opened 2024-03-27
Updates to Balancer spells

viniabussafi:
This PR condenses a few updates to balancer spells:
### balancer.protocol_fee
- Adds pool type column
- Adds incremental filter to bpt_prices, given we are currently missing results for fees collected in BPTs on incremental runs of the spell. I've tried a few different approaches and still couldn't get this right on incremental runs, so would be great to get some review on that.

### balancer_v1_ethereum_liquidity
- Adds pool type column

### balancer.liquidity
- Adds pool type column

### balancer.bpt_prices
- Adds pool type column

### balancer_single_recipient_gauges
- Creates a view with mapping for single recipient gauges and their respective pool

### labels_balancer_v2_gauges_ethereum
- Unions balancer_single_recipient_gauges


macros/models/_project/balancer/balancer_protocol_fee_macro.sql
```diff
@@ -77,6 +77,9 @@ WITH pool_labels AS (
         WHERE l.blockchain = '{{blockchain}}'
         AND l.version = '{{version}}'
         AND s.supply > 0
+        {% if is_incremental() %}
+        AND l.day >= now() - interval '2' day
+        {% endif %}
```
jeff-dude, line 82: when doing a left join, i find it easier and most accurate to keep all filtering of the table on the left side (`balancer_bpt_supply`) to be within the join conditions, rather than join then filter. i'm not 100% sure that'll resolve your issue called out in description, but worth a try.

it also seems backwards that data is missing when doing the join without an incremental filter at all vs. adding one in 🤔  you'd think the new approach would have higher chance of missing data. maybe it comes back to points above.

jeff-dude (approved):
code looks good outside of open question, i'll let it sit for a bit 🤝

macros/models/_project/balancer/balancer_protocol_fee_macro.sql
viniabussafi: yeah, makes sense. On commit [851ce02](https://github.com/duneanalytics/spellbook/pull/5684/commits/851ce0246158c0443cc7103ddcda3d2cd44e8430), I've converted the spell from materialized to table and the performance loss doesn't seem too intense, although it's not ideal. So, I made the changes you suggested and the results are looking good so far. 
I say we keep the spell materialized and, if the same issue happens over the next few days, we can maybe change it to a table?

Thanks once again, @jeff-dude 🫡

models/balancer/arbitrum/balancer_v2_arbitrum_protocol_fee.sql
```diff
@@ -4,7 +4,7 @@
     config(
         schema = 'balancer_v2_arbitrum',
         alias = 'protocol_fee', 
-        materialized = 'incremental',
+        materialized = 'table',
```
jeff-dude, line 7: think a typo to leave only arbitrum as `table`?

jeff-dude (changes requested):

models/balancer/arbitrum/balancer_v2_arbitrum_protocol_fee.sql
viniabussafi: my bad. fixed.

jeff-dude (approved):

PR #5698, opened 2024-03-28
Cleanup seaport decoded tables

0xRobin:
Cleanup:
- change seaport schema from `opensea` -> `seaport` on optimism
- I reconciled all sources over all chains with the produce team so we don't have unnamed columns (`_output`, `_0`)


0xRobin:
Once the decoding side is cleaned up, this should run.

jeff-dude (approved):

PR #5717, opened 2024-04-02
Drop balances and transfers prod_exclude models

aalan3:
These tables are wrong and are not being migrated, they will also be cleaned up manually on our end.


aalan3:
@jeff-dude for awareness

jeff-dude:
> @jeff-dude for awareness

nice!

PR #5725, opened 2024-04-03
add tokens to price list - polygon

RantumBits:
add tokens recently added to CoinPaprika to price list - polygon


models/prices/polygon/prices_polygon_tokens.sql
```diff
@@ -123,11 +123,16 @@ FROM
...
+    ('mim-magic-internet-money', 'polygon', 'MIM', 0x49a0400587a7f65072c87c4910449fdcc5c47242,18),
+    ('cone-bitcone', 'polygon', 'CONE', 0xba777ae3a3c91fcd83ef85bfe65410592bdd0f7c,18),
+    ('pzcx-zen-exchange-token', 'polygon', 'PZCX', 0xdd75542611d57c4b6e68168b14c3591c539022ed,18),
+    ('orbs-orbs', 'polygon', 'ORBS', 0x614389eaae0a6821dc49062d56bda3d9d45fa2ff,18),
+    ('mask-mask-network', 'polygon', 'MASK', 0x2b9e7ccdf0f4e5b24757c1e1a80e311e34cb10c7,18),
+    ('ice-decentral-games-ice', 'polygon', 'ICE', 0xc6c855ad634dcdad23e64da71ba85b8c51e5ad7c,18)
 ) as temp (token_id, blockchain, symbol, contract_address, decimals)
 where contract_address not in (
     0xef938b6da8576a896f6e0321ef80996f4890f9c4 -- DG, bad price feed
-    ,0xc6c855ad634dcdad23e64da71ba85b8c51e5ad7c -- ICE DG, bad price feed
```
Hosuke, line 130: `ICE DG` is recovered here.

Hosuke (approved):
LGTM.✅
Thank you @RantumBits

models/prices/polygon/prices_polygon_tokens.sql
jeff-dude: we should be careful when recovering tokens in prices. did we query the CI table for this address to ensure the prices were fixed? we should look for `max(price)` or outliers over any time period of full history. this has downstream impacts to our sector-level spells, if not validated (which is how these exclusions are found)

jeff-dude (approved):
that works too :)

PR #5735, opened 2024-04-04
Add Uniswap Compatible Optimism Dex Pools - Ready for review

henrystats:
This pr adds some optimism dex projects to the dex_pools_beta sector!

Temporarily named the uniswap v3 pools table in the new sector as pools_array as there's already a pools table on dune for optimism with the same schema name!


henrystats:
hola ser @Hosuke, I wanted to add the uniswap compatible pools for all the chains into dex.pools table, how do I go about it?

it's a lot and I was planning to submit multiple pr's with different batches of projects! Submitting 1 pr per project would take a lot of time and a lot of Pr's! I don't mind doing that though if it's the only approach!

But this current one would save some time!

Hosuke:
> hola ser @Hosuke, I wanted to add the uniswap compatible pools for all the chains into dex.pools table, how do I go about it?
> 
> it's a lot and I was planning to submit multiple pr's with different batches of projects! Submitting 1 pr per project would take a lot of time and a lot of Pr's! I don't mind doing that though if it's the only approach!
> 
> But this current one would save some time!

I think you can go for the batch PR, and try to make 1 PR per chain.

henrystats:
okee, thank you ser @Hosuke

henrystats:
@Hosuke this is now ready for review ser, I added all the uniswap v2/v3 compatible pools in this pr!

Will submit other pr's for other chains and also work on a generic macro for other dexes pool addresses!

Hosuke (approved):
LGTM.✅
Thank you @henrystats 

[Sample dex_optimism.pools]()

jeff-dude:
> Temporarily named the uniswap v3 pools table in the new sector as pools_array as there's already a pools table on dune for optimism with the same schema name!

to this point, any reason we can't reuse/refactor the existing? it'd be good to avoid confusion on two spells with same name doing different things, if not reusable

henrystats:
@jeff-dude the tables have pretty much the same data + some enrichments as it's going to be a global spell!

The only difference between the tables is that the new spell uses an array to store the pool tokens versus using two different columns (token0 & token1) and this is to cater for dex with pools that have more than two tokens (e.g curve)!

The old one can be refactored and what I originally was thinking was to get back to deleting the old tables since the new ones already exist but only after checking if there are a lot of dashboards that rely on the table currently and reaching out to the creators regarding the change incase it breaks their queries!

jeff-dude:
> @jeff-dude the tables have pretty much the same data + some enrichments as it's going to be a global spell!
> 
> The only difference between the tables is that the new spell uses an array to store the pool tokens versus using two different columns (token0 & token1) and this is to cater for dex with pools that have more than two tokens (e.g curve)!
> 
> The old one can be refactored and what I originally was thinking was to get back to deleting the old tables since the new ones already exist but only after checking if there are a lot of dashboards that rely on the table currently and reaching out to the creators regarding the change incase it breaks their queries!

gotcha. [this](https://github.com/duneanalytics/spellbook/blob/main/dex/models/_projects/uniswap/optimism/uniswap_optimism_pools.sql) is the one, yes?

we had to move this one to the new dex subproject, since uniswap v3 trades uses it as a reference rather than reading from decoded table directly like all other uniswap trades spells. this is likely due to the ovm1 legacy data needing to be union'd in. [here](https://github.com/duneanalytics/spellbook/blob/main/dex/models/trades/optimism/platforms/uniswap_v3_optimism_base_trades.sql) is what i'm referring to.

@MSilb7 @chuxinh do you guys see any issue with the uniswap v3 pool on optimism spell getting refactored? or do you solely use it to feed into the uniswap v3 trades spell?

the goal here is to build out a new `dex.pools` universal spell and instead of two columns for each token, a new array column is used.

henrystats:
yes ser, thank you! looking at the uniswap macro, we'd still need the previous pools format for the join with the pools table in order to get the token_bought_address and token_sold address!

PR #5736, opened 2024-04-04
Fix version in _schema.yml

Hosuke:
Fix the wrong version used.

### DIFF
--- a/models/_sector/depin/_schema.yml
+++ b/models/_sector/depin/_schema.yml
@@ -1,4 +1,4 @@
-version: 1
+version: 2
 
 models:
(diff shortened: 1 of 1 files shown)


jeff-dude:
> Once upon a time, the structure of these .yml files was very different (s/o to anyone who was using dbt back then!). Adding version: 2 allowed us to make this structure more extensible.
> 
> Resource yml files do not currently require this config. We only support version: 2 if it's specified. Although we do not expect to update yml files to version: 3 soon, having this config will make it easier for us to introduce new structures in the future

PR #5737, opened 2024-04-04
Added balances for Yield Yak vaults

yy-analytics:

Thank you for taking the time to submit code in Spellbook. A few things to consider:

- If you are a first-time contributor, please sign the CLA by copy & pasting **_exactly_** what the bot mentions in PR comment
- Refer to [docs](#spellbook-contribution-docs) section below to answer questions
- Dune team will review submitted PRs as soon as possible

### To speed up your development process in PRs, keep these tips in mind:

- Each commit to your feature branch will rerun CI tests ([see example](https://github.com/duneanalytics/spellbook/actions/runs/8202519819/job/22433451880?pr=5519))
    - This includes *all* modified models on your branch
    - This includes *all* history of the data
- Two tips for faster development iteration:
    - Ensure dbt is installed locally (refer to main `readme`) and run `dbt compile`
        - This will output raw SQL in `target/` directory to copy/paste and run on Dune directly for initial query testing
    - Hardcode a `WHERE` filter for only ~7 days of history on large source tables, i.e. `ethereum.transactions`
        - This will speed up the CI tests and output results quicker -- whether that's an error or fully successful run
        - Once comfortable with small timeframe, remove filter and let full history run

### Incremental model setup
- Make sure your unique key columns are *exactly* the same in the model config block, schema yml file, and seed match columns (where applicable)
- There cannot be nulls in the unique key columns
    - Be sure to double check key columns are correct or `COALESCE()` as needed on key column(s), otherwise the tests may fail on duplicates

### 🪄 Use the built CI tables for testing 🪄

Once CI completes, you can query the CI tables and errors in dune when it finishes running.
- For example:
    - In the `run initial models` and `test initial models`, there will be a schema that looks like this: `test_schema.git_dunesql_4da8bae_sudoswap_v2_base_pool_creations`
    - This can be temporarily queried in Dune for ~24 hours

Leverage these tables to perform QA testing on Dune query editor -- or even full test dashboards!

## Spellbook contribution docs

The [docs](docs) directory has been implemented to answer as many questions as possible. Please take the time to reference each `.md` file within this directory to understand how to efficiently contribute & why the repo is designed as it is 🪄

Example questions to be answered:

- What does each property in the [model config block](/docs/models/model_config_block.md) mean?
- What is the [CI test](/docs/ci_test/ci_test_overview.md) attached to PRs and how can I best utilize it?
- What are the Spellbook [best practices](/docs/general/best_practices.md)?

Please navigate through the [docs](/docs) directory to find as much info as you can.

**Note**: happy to take PRs to improve the docs, let us know 🤝


yy-analytics:
I'm expecting some issues with this PR but would like to discuss the approach. I opened a thread in Discord [here](https://discord.com/channels/757637422384283659/1225002232597053550/1225002234761318490) but haven't heard anything back yet. I've not used the `FROM {{ source(schema, name) }}` style when querying certain tables because the list of tables to UNION is generated dynamically by using `run_query` combined with the `yield_yak_strategies(blockchain)` macro, and I can't use macros in my sources.yml file so don't have a way to dynamically update the sources if using `FROM {{ source(schema, name) }}` type statements. I've done it this way to avoid having to hardcode strategy contract names (new ones can sometimes be added fairly often and I'm just trying to avoid having to open a PR every time a new strategy is added).

jeff-dude:
> I'm expecting some issues with this PR but would like to discuss the approach. I opened a thread in Discord [here](https://discord.com/channels/757637422384283659/1225002232597053550/1225002234761318490) but haven't heard anything back yet. I've not used the `FROM {{ source(schema, name) }}` style when querying certain tables because the list of tables to UNION is generated dynamically by using `run_query` combined with the `yield_yak_strategies(blockchain)` macro, and I can't use macros in my sources.yml file so don't have a way to dynamically update the sources if using `FROM {{ source(schema, name) }}` type statements. I've done it this way to avoid having to hardcode strategy contract names (new ones can sometimes be added fairly often and I'm just trying to avoid having to open a PR every time a new strategy is added).

i believe i responded yesterday. unfortunately, i don't think your approach here will work with `run_query`. does it make sense to build a separate spell solely that outputs a list of tables you need, then call that spell in the main spell to obtain the list? then dbt will handle the dependencies for you?

we don't allow direct access to db, so `run_query` will break some things.

yy-analytics:
> > I'm expecting some issues with this PR but would like to discuss the approach. I opened a thread in Discord [here](https://discord.com/channels/757637422384283659/1225002232597053550/1225002234761318490) but haven't heard anything back yet. I've not used the `FROM {{ source(schema, name) }}` style when querying certain tables because the list of tables to UNION is generated dynamically by using `run_query` combined with the `yield_yak_strategies(blockchain)` macro, and I can't use macros in my sources.yml file so don't have a way to dynamically update the sources if using `FROM {{ source(schema, name) }}` type statements. I've done it this way to avoid having to hardcode strategy contract names (new ones can sometimes be added fairly often and I'm just trying to avoid having to open a PR every time a new strategy is added).
> 
> i believe i responded yesterday. unfortunately, i don't think your approach here will work with `run_query`. does it make sense to build a separate spell solely that outputs a list of tables you need, then call that spell in the main spell to obtain the list? then dbt will handle the dependencies for you?
> 
> we don't allow direct access to db, so `run_query` will break some things.

Thanks @jeff-dude. I could be wrong on this, but the issue I see with what you're describing is that I need the result of the query as a jinja variable (specifically, a list) that I can then loop over to create the "main" query. If I were to have a separate spell and select from that using `ref` then isn't that just the same as having it in a CTE? The main approach I'm going for in this is in the `yield_yak_balances.sql` macro file

yy-analytics:
> > > I'm expecting some issues with this PR but would like to discuss the approach. I opened a thread in Discord [here](https://discord.com/channels/757637422384283659/1225002232597053550/1225002234761318490) but haven't heard anything back yet. I've not used the `FROM {{ source(schema, name) }}` style when querying certain tables because the list of tables to UNION is generated dynamically by using `run_query` combined with the `yield_yak_strategies(blockchain)` macro, and I can't use macros in my sources.yml file so don't have a way to dynamically update the sources if using `FROM {{ source(schema, name) }}` type statements. I've done it this way to avoid having to hardcode strategy contract names (new ones can sometimes be added fairly often and I'm just trying to avoid having to open a PR every time a new strategy is added).
> > 
> > 
> > i believe i responded yesterday. unfortunately, i don't think your approach here will work with `run_query`. does it make sense to build a separate spell solely that outputs a list of tables you need, then call that spell in the main spell to obtain the list? then dbt will handle the dependencies for you?
> > we don't allow direct access to db, so `run_query` will break some things.
> 
> Thanks @jeff-dude. I could be wrong on this, but the issue I see with what you're describing is that I need the result of the query as a jinja variable (specifically, a list) that I can then loop over to create the "main" query. If I were to have a separate spell and select from that using `ref` then isn't that just the same as having it in a CTE? The main approach I'm going for in this is in the `yield_yak_balances.sql` macro file

@jeff-dude , do you know if the "no direct access to db" would also mean that the `dbt_utils.get_column_values` macro also cannot be run? From the [source code](https://github.com/dbt-labs/dbt-utils/blob/main/macros/sql/get_column_values.sql) it looks like it relies on a statement block (same as `run_query`) so I'm guessing not, but thought I might just confirm (?) as I can make a separate spell as you said if `dbt_utils.get_column_values` is working and can then point to that spell in the arguments for this macro. If not possible then I think the only solution is going to be hard-coding the list of strategies and submitting a new PR each time that list changes.

yy-analytics:
Given the "QUERY HAS TOO MANY STAGES" error that's coming now, I'm going to try to break up some of the balances model/macro first into separate Deposits, Withdraws and Reinvests models/macros which can then hopefully be combined to make the balances model/macro without this error.

yy-analytics:
Hi @jeff-dude / @Hosuke ,
I've got to a point where I might need a bit of guidance (please). The last 2 commits on this PR both should work but I've got a problem with "inline functions not being allowed in a view" (when I use "WITH FUNCTION...") and a problem with the function "not being registered" if I try to use "CREATE OR REPLACE FUNCTION..." instead. The calculation relies on the previous 25 reinvests which I've collected using ARRAY_AGG. The ideal would be for this kind of function definition to work in some way - either inline or with CREATE OR REPLACE like the other custom functions in macros/public (am I allowed to add something in here and update the public_macros_schema.yml?). But in the absolute worst case then I can probably make it work only using Trino SQL, however I think performance would be worse in that case. Other alternative might be to have APY as something that's calculated on the "other side", using an inline function in the query editor instead, but that also doesn't feel as neat/useful as trying to get it in the spell. Any guidance you can give would be much appreicated!
Thanks,
Angus

macros/models/_project/yield_yak/yield_yak_reinvests.sql
```diff
@@ -0,0 +1,133 @@
+{% macro get_apy_from_array() %}
+    CREATE OR REPLACE FUNCTION get_apy_from_array(recent_reinvest_info ARRAY(varchar))
```
Hosuke, line 2: Maybe you cannot create UDF here.

jeff-dude: yes, that is correct, we cannot create functions on dunesql engine directly. dbt macros can only be used to output sql code to run

yy-analytics: Okay thanks both, a couple of follow ups:

1. What about if the outputted SQL code includes an inline function ("WITH FUNCTION get_apy_from_array..")? In this format, the compile code created by "dbt compile" works without a problem but when I tried to add this to the PR it failed saying something like "inline functions cannot be used in views" (can see this in the previous commit). Is there any way around this? I thought I was creating incremental models, which I thought were materialized as tables rather than views (?) so I'm a bit confused by the error message.
2. What about the functions in the macros/public folder? Isn't there something I can do similarly to those to make this function work? i.e. defining it in a separate place/whatever is needed.

yy-analytics: Have realised that the historical information from the checks carried out is not visible so let me know if it would be beneficial for me to reverse the latest commit so that we're back to the inline function approach and you can see the "inline functions not allowed in views" error message when carrying out "dbt slim ci / dbt-test (spellbook)" checks

yy-analytics: I found the exact error message from previously by the way:
`Database Error in model yield_yak_arbitrum_reinvests (models/yield_yak/arbitrum/yield_yak_arbitrum_reinvests.sql)
TrinoUserError(type=USER_ERROR, name=NOT_SUPPORTED, message="line 10:1: Views cannot contain inline functions", query_id=20240415_171629_00071_v5q22)`

yy-analytics: Actually, tbh sorry about this guys. I've realised I'm not really making much use of routine-specific functionality in my function declaration for `get_apy_from_array`, so I think I can just put the logic directly in the SQL script and it should work (albeit with slightly worse readability - will add comments).

yy-analytics:
@jeff-dude @Hosuke , I've managed to get the tests to pass for this now. If there are any questions with the content of the PR or anything I could clarify please let me know.

Hosuke:
This is an intense PR, may take some more time to review...

yy-analytics:
Sorry yeah, I tried to keep it simple at first and just do one thing, but that one thing triggered the "QUERY_HAS_TOO_MANY_STAGES" error so I had to break it up. I'll try to give a high level summary in case any of this information helps:

Yield Yak does a few things on both Avalanche and Arbitrum, among which are:
- Dex aggregator ("Yak Swap") - some of the data for this is covered by existing trades spells which feed into the dex aggregator trades spell. Not really in scope for this PR.
- Auto-compounding vaults/pools/farms. These are the focus of this PR.
- A few other things (not in scope).

Auto-compounding vaults/pools/farms (all are "strategies"). These are distinct contracts, not created by a factory, but all of which have some common elements in their ABI (namely Deposit, Withdraw and Reinvest events). The list of strategy names is retrieved by the `yield_yak_strategies.sql` macro. I had wanted to do this dynamically originally using `run_query` but it ended up not being possible. Users deposit into these strategies using the required deposit token (often an LP-type token) and receive YRT tokens in exchange ("Yak Receipt Token").

The Deposit and Withdraw events are very similar so I put them into a single macro `yield_yak_deposits_withdraws.sql` which basically unions the Deposit/Withdraw events for all of the distinct strategies. The models are incremental tables and the macro is done such that if a new strategy (or a new contract with the same ABI as an existing strategy) is added at a later date (even if it's "old") then the "old" data will still be added to the incremental model. The "amount" from these events is the amount of deposit tokens deposited into/withdrawn from the strategy.

The Reinvest events happen each time an address triggers a Reinvest. There is a small reward for that address which typically is about 1% of the pending rewards from the strategy so it happens in a decentralised way (there's not a regular schedule) and will normally occur when the rewards for doing the Reinvest outweigh the approximate gas cost of the tx. In the Reinvest tx, the rewards are sold in exchange for more of the underlying deposit token and then the Reinvest event emits values for `newTotalDeposits` (new amount of deposit tokens in the strategy) and `newTotalSupply` (total supply of the YRT tokens for that strategy currently in circulation). The complex addition for the `yield_yak_reinvests.sql` macro is the addition of the APY calculation (method in more detail is [here ](https://yieldyak.medium.com/yield-yak-upgrades-farm-apy-calculations-d3fc247fbbf2)) and it's the relative change in the ratio between deposit tokens and YRT tokens which tells us how much users are getting as a return.

These 3 event types are the only things needed to calculate the deposit token balance of the strategies at any point in time, which is what the `yield_yak_balances.sql` macro is doing by combining the models for deposits, withdraws and reinvests. It calculates the balance at a point in time by ordering these events, creating a partition for the Reinvest events (because those are effectively like a "reset") and then calculating the cumulative net deposits since the last Reinvest event, combined with the `new_total_deposits` value. Then there's a bit using LEAD to get the time of the following event so that in the end we have a temporal table (`from_time` / `to_time`) for the deposit token balance for each strategy.

models/yield_yak/yield_yak_withdraws.sql
```diff
@@ -0,0 +1,35 @@
+{{ config(
+	    schema = 'yield_yak',
+        alias = 'withdraws',
+        post_hook='{{ expose_spells(\'["arbitrum", "avalanche_c"]\',
+                                "project",
+                                "yield_yak",
+                                \'["angus_1"]\') }}'
```
Hosuke, line 7: We may use the new format for all the `post_hook`:
```suggestion
	schema = 'yield_yak',
        alias = 'withdraws',
        post_hook='{{ expose_spells(
                      blockchains = \'["arbitrum", "avalanche_c"]\',
                      spell_type = "project",
                      spell_name = "yield_yak",
                      contributors = \'["angus_1"]\') }}'
```

models/yield_yak/yield_yak_reinvests.sql
```diff
@@ -0,0 +1,39 @@
+{{ config(
+	    schema = 'yield_yak',
+        alias = 'reinvests',
+        post_hook='{{ expose_spells(\'["arbitrum", "avalanche_c"]\',
```
Hosuke, line 4: Same as above.

models/yield_yak/yield_yak_deposits.sql
```diff
@@ -0,0 +1,35 @@
+{{ config(
+	    schema = 'yield_yak',
+        alias = 'deposits',
+        post_hook='{{ expose_spells(\'["arbitrum", "avalanche_c"]\',
```
Hosuke, line 4: Same as above.

macros/models/_project/yield_yak/yield_yak_deposits_withdraws.sql
```diff
@@ -0,0 +1,63 @@
...
+    , c.evt_index
+    , t.index AS tx_index
+    , c.block_time
+    , c.block_number
+    , c.user_address
+    , c.{{ event_name.lower() }}_amount
+FROM combined c
+LEFT JOIN
+{{ source(blockchain, 'transactions') }} t
+    ON t.hash = c.tx_hash
```
Hosuke, line 61: Do you think we may add incremental filters on `transactions` too?

Hosuke (changes requested):
Impressive, I left minor comments.

macros/models/_project/yield_yak/yield_yak_deposits_withdraws.sql
yy-analytics: Would be nice but I don't think that would work. There's a scenario where a Yield Yak contract has existed for a while but hasn't been decoded yet. If we then decoded that contract it will be picked up in the "combined" CTE (due to the `OR c.contract_address IS NULL` condition, but if we had incremental filters on `transactions` then the older transaction info for any such contract would not get picked up.

models/yield_yak/arbitrum/yield_yak_arbitrum_schema.yml
```diff
@@ -94,7 +91,149 @@ models:
...
+    description: >
+        yield yak deposits
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - blockchain
+            - contract_address
+            - evt_index
+            - tx_index
+            - block_number
```
jeff-dude, line 114: each of these four spells unique combo of columns tests have different columns compared to the unique keys in the model config blocks. can you update these to reflect the same columns as the model? or the opposite, whichever is needed?

macros/models/_project/yield_yak/yield_yak_deposits_withdraws.sql
jeff-dude: even if we can't apply an incremental filter, we should be able to enhance the join.

can you add `block_number` to the join conditions?

transactions is also partitioned on `block_date` -- we could likely add condition for `block_date = cast(date_trunc('day', block_time) as date)`  (assuming `block_date` doesn't exist in the decoded tables. if it does, then simply join on `block_date`)

macros/models/_project/yield_yak/yield_yak_reinvests.sql
```diff
@@ -0,0 +1,138 @@
...
+
+add_tx_index AS (
+    SELECT
+        c.*
+        , t.index AS tx_index
+        , t."from" AS reinvest_by_address
+    FROM combined c
+    LEFT JOIN
+    {{ source(blockchain, 'transactions') }} t
+        ON t.hash = c.tx_hash
```
jeff-dude, line 64: same idea on joins here from deposits/withdraws macro

macros/models/_project/yield_yak/yield_yak_deposits_withdraws.sql
jeff-dude: also, can it be an inner join?
that's more efficient on the dunesql engine.

if inner join isn't possible, it's usually best to first read from the larger of the two tables, then left/right join as needed to smaller tables

models/yield_yak/arbitrum/yield_yak_arbitrum_schema.yml
jeff-dude: same applies to avax file below

jeff-dude (changes requested):
fantastic code here, thank you for applying all the best practices (and even teaching me some more)!

left a few thoughts below prior to finalizing and merging

jeff-dude:
also, there is some complex incremental logic here. are you aware of the output tables built via CI pipelines attached, so you can query them on dune and test the data and ensure all looks good prior to merge?

models/yield_yak/arbitrum/yield_yak_arbitrum_schema.yml
yy-analytics: Yep, I'll make that change, thanks! I think strictly speaking the `contract_address` does not need to appear in the unique combo as `block_number`, `tx_index` and `evt_index` should be sufficient for uniqueness so I'll remove `contract_address` from the arbitrum/avax schemas and the top-level schema, but I'll keep `blockchain` in the top-level schema that unions arbitrum and avax data.

yy-analytics: Actually for balances it's a little different (still need the `contract_address`) but will get them to match by removing `blockchain` from the arbitrum/avax schemas.

macros/models/_project/yield_yak/yield_yak_reinvests.sql
yy-analytics: Yep, can enhance the joins (hadn't realise that would improve performance - good to know!). And yes I think it can be an inner join (there shouldn't be anything present in the decoded tables which does not have a corresponding row in the `transactions` tables). There's no `block_date` in the decoded tables so will do the `block_date` join condition using `date_trunc` as you've written.

yy-analytics: Am taking the approach of adding `block_date` as a column for the deposits/withdraws/reinvests models also (think it might be handy for daily time series data anyway).

yy-analytics:
> also, there is some complex incremental logic here. are you aware of the output tables built via CI pipelines attached, so you can query them on dune and test the data and ensure all looks good prior to merge?

@jeff-dude , no I'm not sure on this part. I know about the tables generated by "dbt compile" and have been testing those to make sure the output looks correct, but as far as I could tell, that's for the "initial model". Where do I find the attached CI pipelines/tables to check the incremental logic is working correctly?

jeff-dude:
> @jeff-dude , no I'm not sure on this part. I know about the tables generated by "dbt compile" and have been testing those to make sure the output looks correct, but as far as I could tell, that's for the "initial model". Where do I find the attached CI pipelines/tables to check the incremental logic is working correctly?

yep, `dbt compile` is useful for testing query in general for full history or initial runs. however, when you introduce complex incremental logic in spellbook, sometimes that can bring bugs into subsequent runs. in order to test the data quality post-incremental run, we open up the CI test tables (from attached gh actions on PRs) to be queried on dune for ~24 hours before being deleted. please note that it's once a day at set time, not 24 hours after build, so a rerun of gh action may be needed if bad timing happens and tables are dropped prior to finishing your testing.

check out these docs, hopefully they help (if not, let me know, we can modify those docs as needed):
https://github.com/duneanalytics/spellbook/blob/main/docs/ci_test/ci_test_overview.md

jeff-dude:
i'll let this sit for a little bit, if you wanted to query those CI output tables for data quality. let me know when ready to merge 🤝

jeff-dude:
> I think the 24 hr window passed. To trigger the actions again I was just going to "Update branch" but it's been showing "Checking for ability to merge automatically..." for a while now (see below). Is there any other way to rerun the gh action?

i was able to merge in main. i think there is a universal gh issue right now, so it's been slow today. other ways to do it:
- on your local setup, merge main from spellbook into your main branch, push changes via git locally (assuming there are new changes on main)
- if no new changes, you can also go into the action itself and manually rerun
  
image

- final approach: push a dummy change to your branch, so a commit is made. each commit will run them again.

yy-analytics:
> i was able to merge in main. i think there is a universal gh issue right now, so it's been slow today. other ways to do it:
> 
> * on your local setup, merge main from spellbook into your main branch, push changes via git locally (assuming there are new changes on main)
> * if no new changes, you can also go into the action itself and manually rerun
> 
> image
> 
> * final approach: push a dummy change to your branch, so a commit is made. each commit will run them again.

Thanks for the tips! I've been able to review the CI tables and I'm happy with how it all looks so am happy for you to merge when ready 🤝

PR #5763, opened 2024-04-10
Finalize dex_solana_bot_trades

Hosuke:
Finalizing dex_solana_bot_trades

cc. @whalehunting


Hosuke:
Not sure about the test failure:
https://github.com/duneanalytics/spellbook/actions/runs/8707669649/job/23883193024#step:11:47

jeff-dude:
> Not sure about the test failure: https://github.com/duneanalytics/spellbook/actions/runs/8707669649/job/23883193024#step:11:47

spill to disk is a config on the CI cluster setup which helps with in-memory processing. since this PR is running the full lineage and they are heavy solana models, the spill to disk is failing. essentially we're crashing the cluster with the amount of data trying to process at once.

i don't know the exact differences between CI and prod cluster setup, but i'm thinking maybe we just try to get this to run and build in prod. we can try CI one more time, see what happens.

jeff-dude (approved):

jeff-dude:
going to try prod now. will try reducing threads to get less concurrent running if needed. worst case scenario, i'll revert.

jeff-dude:
> going to try prod now. will try reducing threads to get less concurrent running if needed. worst case scenario, i'll revert.

fyi, failed in prod on similar error. i'm manually updating prod job to run less threads and try to get to run that way.

Hosuke:
> > going to try prod now. will try reducing threads to get less concurrent running if needed. worst case scenario, i'll revert.
> 
> fyi, failed in prod on similar error. i'm manually updating prod job to run less threads and try to get to run that way.

If so we may revert this PR first and keep the current status for stability.

jeff-dude:
> If so we may revert this PR first and keep the current status for stability.

no need, we're all good now. this lineage will need manual intervention in prod when rerunning the whole thing, to reduce concurrency of models running at same time. there's too much data to be running that much in parallel. we may end up needing to run solana in it's own job(s) at some point, we'll see

PR #5802, opened 2024-04-21
Add CEX addresses for EVM chains, Bitcoin, Solana, Aptos, Cardano, Algorand, EOS, Cosmos, Elrond, Polkadot, Ripple, Tron, Zilliqa & Litecoin

hildobby:


models/cex/algorand/cex_algorand_addresses.sql
```diff
@@ -0,0 +1,23 @@
+{{config(
+        tags = ['static'],
+        alias = 'addresses',
```
jeff-dude, line 3: can you add `schema` property to all of these new ones?

models/cex/algorand/cex_algorand_addresses.sql
```diff
@@ -0,0 +1,23 @@
...
+        post_hook='{{ expose_spells(\'["algorand"]\',
```
jeff-dude, line 4: do we want all of them exposed on the data explorer?
just double checking. should be fine if so.

jeff-dude (changes requested):

models/cex/algorand/cex_algorand_addresses.sql
hildobby: done ☑️

models/cex/algorand/cex_algorand_addresses.sql
hildobby: Yes, having all of them exposed is perfect imo

jeff-dude (approved):

PR #5811, opened 2024-04-22
Add Balancer zkevm spells 

viniabussafi:
This PR adds all necessary models and sources for Balancer spells on Polygon zkevm, following the same pattern for other chains, as well as includes zkevm on all multichain spells.


models/balancer/zkevm/balancer_v2_zkevm_trades_legacy.sql
jeff-dude, line 1: we're in the process of deprecating the legacy lineage -- do you need this?
i see you already have the new view one above this file

viniabussafi: hey, @jeff-dude. not at all, was just unsure about what was the right approach here. Will remove it, thanks!

jeff-dude (approved):
one thing we noticed after building initial zkevm spell in `dex.trades` lineage is that prices are missing for this new chain, so all `amount_usd` values are null. i think maybe it's best to hold off on merging this, until we get the metadata for this chain in place?

i'm not 100% sure if it's being actively worked on this second, but let me know what you think

viniabussafi:
maybe I can work on getting some pricing data on another PR and then we get back to this one. Would that be OK?

viniabussafi:
started a prices.usd table for polygon zkevm [here](https://github.com/duneanalytics/spellbook/pull/5856)

jeff-dude:
thanks again for moving it all along. are you all set here now?

viniabussafi:
all good on my end, @jeff-dude

PR #5816, opened 2024-04-23
Fix magiceden duplicates

0xRobin:
TX that caused the dupes: https://polygonscan.com/tx/0xf70a24c003e88feebf58da309198cd8b83648734a0700794a4475818cd08c253

The duplicates are produced because of a bug in the fee join conditions resulting in multiple matches for this tx.


0xRobin:
Did a manual check on `price_raw`, `royalty_fee_amount_raw` and `platform_fee_amount_raw` and we still have all the current data with this more restricted join. :heavy_check_mark:  
Should be good to merge.

jeff-dude (approved):
fantastic -- thank you for the quick follow-up fix on this one

PR #5820, opened 2024-04-24
Fix blockchains property for dex.sandwiched

vegarsti:
Missing a quote! We have some validation in our indexing pipeline which caught this. (So currently it's not showing in the Data Explorer UI.)


jeff-dude (approved):
good catch -- hard for us to spot here 🤔 i wonder if we could use the same or similar test in our gh actions

vegarsti:
Yeah, feasible for sure! We could look for string arrays and try to validate them

PR #5846, opened 2024-04-28
feat: add camelot pair fee rates spellbook

whalehunting:
The Camelot DEX on Arbitrum has a dynamic fee rate model, both for v2 and v3 architecture:
- Fee rates can be configured by pair
- Fee rates can be set by swap direction (token0 -> token1 and token1 -> token0 can have individual fee rates)
- Fee rates can be updated over time

The [Camelot ⚔️🤴🏼 Dashboard](https://dune.com/whale_hunter/camelot-dex) has been broken since 2 months, because the query logic that determines the latest fee rate for each swap is super inefficient, which leads to >30min execution runtimes, so the query runs fail.

I discussed potential solutions with @0xBoxer and the Camelot team and came up with this structure. The spellbook should contain fee rates for every pair by minute, since the deployment of each pair.
Once live, a swap can be simply joined like this.

```sql
SELECT
    *
FROM
    swaps
    JOIN camelot_arbitrum.pair_fee_rates ON (
        swaps.minute = camelot_arbitrum.pair_fee_rates.minute
    )
```

## Sample
Filtered version with a single pair
https://dune.com/queries/3677413?sidebar=none

## Thoughts
- I formatted the code using `sqlfmt`, lmk if this matches your coding style!


models/camelot/camelot_pair_fee_rates.sql
```diff
@@ -0,0 +1,169 @@
...
+        from dex.trades
+        where block_time >= timestamp '{{project_start_date}}'
+    ),
+    -- Prepare data structure (1 row for every minute since each pair was created)
+    pairs_by_minute as (
+        select time_series.minute, pair, version, token0, token1
+        from pairs
+        cross join time_series
+        where time_series.minute >= pairs.minute
+    )
```
jeff-dude, line 137: as discussed in discord, you are hitting a timeout error on CI (90 minutes). my best guess is this portion of the code. please take the time to figure out where the query really slows down and becomes inefficient. i wouldn't expect data on arbitrum to take this long.

jeff-dude (changes requested):

models/camelot/camelot_pair_fee_rates.sql
whalehunting: yeah probably this cross join, or the last_value() window function a bit further down the code.
but this was the only possibility to build the query imo, do you have suggestions how it could be optimized?

if something is unclear about the code's intentions, pls lmk!

whalehunting: @jeff-dude as you can see in the #5862 PR, the bottleneck is in the `last_value()` window function in the final query statement, not in the cross join

So we should probably focus on this, any thoughts on how to improve it would be very welcome 💯

models/camelot/camelot_pair_fee_rates.sql
jeff-dude, line 1: lets update this directory and file name to:
`models/camelot/arbitrum/camelot_arbitrum_pair_fee_rate.sql`

models/camelot/camelot_pair_fee_rates.sql
```diff
@@ -0,0 +1,163 @@
+{{
+    config(
+        schema="camelot_arbitrum",
+        alias="pair_fee_rates",
+        partition_by=["blockchain"],
```
jeff-dude, line 5: this partition shouldn't be needed, as blockchain is always `arbitrum`

models/camelot/camelot_pair_fee_rates.sql
```diff
@@ -0,0 +1,163 @@
...
+                union all
+                select *
+                from v3_directional_fee_rate_updates
+            )
+        group by minute, pair, version, token0, token1
+    ),
+    camelot_pair_trades_by_minute as (
+        select distinct
+            date_trunc('minute', block_time) as minute, project_contract_address as pair
+        from dex.trades
```
jeff-dude, line 106: plz use jinja ref syntax on dex trades

models/camelot/camelot_pair_fee_rates.sql
```diff
@@ -0,0 +1,163 @@
...
+        where
+            blockchain = '{{blockchain}}' and project = 'camelot'
+            {% if not is_incremental() %}
+                and block_time >= timestamp '{{project_start_date}}'
+            {% endif %}
+            {% if is_incremental() %}
+                and block_time >= date_trunc('day', now() - interval '7' day)
```
jeff-dude, line 113:
```suggestion
                and {{incremental_predicate('block_time')}}
```

jeff-dude: this allows us to control number of days to look back in one spot

models/camelot/camelot_pair_fee_rates_schema.yml
jeff-dude, line 1: `models/camelot/arbitrum/_schema.yml` is fine for naming here

models/camelot/camelot_pair_fee_rates_schema.yml
```diff
@@ -0,0 +1,38 @@
+version: 2
+
+models:
+  - name: camelot_pair_fee_rates
```
jeff-dude, line 4: be sure to update spell name here, if file change above

jeff-dude: also plz add unique combo of columns test in here, it should match the unique keys in model exactly

jeff-dude (changes requested):

models/camelot/camelot_pair_fee_rates.sql
whalehunting: should we change it to something like `pair`? or just remove it

jeff-dude:
i compiled the query so i could run it on dune and see what is happening to get a better understanding. it makes sense to me up until the CTE which reads from `dex.trades`. my understanding was that you introduced `dex.trades` data in order to only pull pairs which have completed a trade, to cut down on the amount of data. however, the code ultimately does a `union all` on dex trades output and all the decoded tables union'd, which introduces more `pair` rows.

do you want to pull pair contracts which exist in `dex.trades`, but not in the decoded tables? if so, i think it's on the right path (but can still be optimized a bit). you could condense from two CTEs to one CTE, and instead of subquery select distinct on the `union all`, then joining back to prior CTE for more columns, you could switch to `union distinct` usage and ideally cut down the query.

if you were to actually only use `dex.trades` output to be a filter on decoded data, then wouldn't this be best?
```
with
  v2_pairs_with_initial_fee_rates as (
    --this CTE output 3,749 unique pair/version outputs
    --this CTE also has 3,749 total rows, so each pair has a minute assigned at creation and that's it, no more rows
    select
      date_trunc('minute', evt_block_time) as minute,
      pair,
      '2' as version,
      token0,
      300 / 1e5 as token0_fee_rate,
      token1,
      300 / 1e5 as token1_fee_rate
    from
      "delta_prod"."camelot_arbitrum"."CamelotFactory_evt_PairCreated"
  ),
  v2_directional_fee_rate_updates as (
    --this CTE output 98 total rows
    --this CTE has 66 unique pair/version combos, so it looks like a pair/version can be repeated at various minutes (fees changing over time, new row when a fee has changed with new minute value)
    select
      date_trunc('minute', evt_block_time) as minute,
      pair,
      version,
      token0,
      token0feepercent / 1e5 as token0_fee_rate,
      token1,
      token1feepercent / 1e5 as token1_fee_rate
    from
      "delta_prod"."camelot_arbitrum"."CamelotPair_evt_FeePercentUpdated" as fee_updates
      join v2_pairs_with_initial_fee_rates as pairs on fee_updates.contract_address = pairs.pair
  ),
  v3_pairs_with_initial_fee_rates as (
    --this CTE output 504 unique pair/version outputs
    --this CTE also has 504 total rows, so each pair has a minute assigned at creation and that's it, no more rows
    select
      date_trunc('minute', evt_block_time) as minute,
      pool as pair,
      '3' as version,
      token0,
      100 / 1e6 as token0_fee_rate,
      token1,
      100 / 1e6 as token1_fee_rate
    from
      "delta_prod"."camelot_v3_arbitrum"."AlgebraFactory_evt_Pool"
  ),
  v3_directional_fee_rate_updates as (
    --this CTE output 4,786,999 total rows
    --this CTE has 460 unique pair/version combos, so it looks like a pair/version can be repeated at various minutes (fees changing over time, new row when a fee has changed with new minute value)
    select
      date_trunc('minute', evt_block_time) as minute,
      pair,
      version,
      token0,
      feezto / 1e6 as token0_fee_rate,
      token1,
      feeotz / 1e6 as token1_fee_rate
    from
      "delta_prod"."camelot_v3_arbitrum"."AlgebraPool_evt_Fee" as fee_updates
      join v3_pairs_with_initial_fee_rates as pairs on fee_updates.contract_address = pairs.pair
  ),
  pairs as (
    --as expected, this CTE returns 4,253 after adding together two initial fee CTE counts above
    select
      *
    from
      v2_pairs_with_initial_fee_rates
    union all
    select
      *
    from
      v3_pairs_with_initial_fee_rates
  ),
  fee_rate_updates as (
    --this CTE returns 2,501,901 rows
    --based on comments for multiple fee changes per minute, it makes sense to see rows drop out when adding from previous CTEs above
    select
      minute,
      pair,
      version,
      token0,
      avg(token0_fee_rate) as token0_fee_rate, -- Handle edge case where pair fees gets changed multiple times per minute
      token1,
      avg(token1_fee_rate) as token1_fee_rate -- Handle edge case where pair fees gets changed multiple times per minute
    from
      (
        select
          *
        from
          pairs
        union all
        select
          *
        from
          v2_directional_fee_rate_updates
        union all
        select
          *
        from
          v3_directional_fee_rate_updates
      )
    group by
      minute,
      pair,
      version,
      token0,
      token1
  ),
  camelot_pair_trades_by_minute as (
    --this CTE returns 5,888,599 trades in history of camelot
    select distinct
      date_trunc('minute', block_time) as minute,
      project_contract_address as pair
    from
      dex.trades
    where
      blockchain = 'arbitrum'
      and project = 'camelot'
      and block_time >= timestamp '2022-06-14'
  ),
  -- Prepare data structure (1 row for every minute where pair trades happened
  -- and/or fee rates got updated)
  pairs_by_minute as (
    select
      fees.*
    from
      fee_rate_updates as fees
      inner join camelot_pair_trades_by_minute as trades on fees.minute = trades.minute
      and fees.pair = trades.pair
  )
select
  'arbitrum' as blockchain,
  pairs.minute,
  pairs.pair,
  pairs.version,
  pairs.token0,
  coalesce(
    token0_fee_rate,
    last_value(token0_fee_rate) ignore nulls over (
      partition by
        pairs.pair
      order by
        pairs.minute rows between unbounded preceding
        and current row
    )
  ) as token0_fee_rate,
  pairs.token1,
  coalesce(
    token1_fee_rate,
    last_value(token1_fee_rate) ignore nulls over (
      partition by
        pairs.pair
      order by
        pairs.minute rows between unbounded preceding
        and current row
    )
  ) as token1_fee_rate
from
  pairs_by_minute as pairs
order by
  minute desc,
  pair asc
limit
  10
```

models/camelot/camelot_pair_fee_rates.sql
```diff
@@ -0,0 +1,163 @@
...
+        select distinct minute, pair
+        from
+            (
+                select minute, pair
+                from camelot_pair_trades_by_minute
+                union all
+                select minute, pair
+                from fee_rate_updates
+            )
+    ),
```
jeff-dude, line 128:
```suggestion
    pairs_by_minute as (
          select minute, pair
          from camelot_pair_trades_by_minute
          union distinct
          select minute, pair
          from fee_rate_updates
    ),
```

jeff-dude: i think where i got confused was this union all then select distinct -- i wonder if this simplified approach works?

models/camelot/camelot_pair_fee_rates.sql
```diff
@@ -0,0 +1,163 @@
...
+    -- Add version, token0, token1 columns
+    pairs_by_minute_with_metadata as (
+        select pairs_by_minute.minute, pairs_by_minute.pair, version, token0, token1
+        from pairs_by_minute
+        left join pairs on pairs_by_minute.pair = pairs.pair
```
jeff-dude, line 133: i was hoping we could avoid the left join back, but i don't think it's possible. only other option is to do left join in dex CTE where there is less data (pre union step) to grab necessary columns, then union fully in the union distinct above with all columns you need

models/camelot/camelot_pair_fee_rates.sql
whalehunting: i didn't know about union distinct, but from naming alone it sounds like it would produce the desired result!
can we add some kind of tests to this spellbook also? so we can see if these changes would change the resultset?

models/camelot/camelot_pair_fee_rates.sql
jeff-dude: likely just remove it. the best option is typically a date column, such as `block_month` or something with relatively even distribution across partitions and enough data to impact performance. if you foresee querying this table and grouping by / filtering by month timeframes, that could help. otherwise, just remove

models/camelot/camelot_pair_fee_rates.sql
jeff-dude: what you can do:
- run CI in current setup, have CI tables ready to query on the dune app
- push a commit to PR, rerun CI on commit, query new CI tables (should have unique git sha value per commit, therefore unique tables)

you can then compare the results with the `exclude` function or just manually

models/camelot/camelot_pair_fee_rates_schema.yml
whalehunting: i moved it to `camelot_arbitrum_schema.yml`, or did you actually mean `_schema.yml`?

models/camelot/camelot_pair_fee_rates.sql
whalehunting: not possible imo, since we also need the metadata for non-trades (pair deployments + fee updates), which also produce rows in the final result set

models/camelot/arbitrum/camelot_arbitrum_schema.yml
```diff
@@ -176,3 +176,46 @@ models:
...
+      Camelot fee rates over time across all pairs and versions on Arbitrum.
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - minute
+            - blockchain
+            - pair
+            - version
+    columns:
+      - blockchain
```
jeff-dude, line 198: i think you need `- &blockchain` format here. leaving out `&` make dbt angry.
however, you also use this above in this schema file.
you should either use:
- `- *blockchain`
- drop this line and go right into `name: blockchain`

jeff-dude (approved):
good stuff here 🔥 
let's get it merged

PR #5857, opened 2024-04-30
Introduce check_dex_null_amount_usd test

Hosuke:
Try to ensure null percentage in `amount_usd` is below certain threshold.

Goal of this PR:
- since `amount_usd` is built in the final stage of `dex.trades` lineage, apply test on `dex.trades`
- when a new chain is added to spellbook, sometimes dex entries can be added before chain metadata, leading to gaps in data coverage
  - i.e. `amount_usd` null for all rows, due to no values in `prices.usd` for that chain
- to test this, set an arbitrary threshold for number of rows to have not null for `amount_usd`
  - this arbitrary number should be relatively loose, as we want to focus on getting the most commonly used tokens into `prices.usd` rather than full coverage, at least on day one
- if below threshold, fail CI to alert both the user and the dune team to get more metadata built and live on prod first


.github/workflows/dbt_slim_ci.yml
```diff
@@ -68,6 +68,15 @@ jobs:
...
         run: "dbt test $PROFILE --select state:new state:modified --exclude tag:prod_exclude tag:remove --defer --state . --project-dir ${PROJECT_DIR}"
 
+      - name: dbt test dex trades specific model(s)
+        run: |
+          if git diff --name-only ${{ github.event.pull_request.base.sha }}..${{ github.sha }} | grep -q "models/_sector/dex/trades/"; then
+            dbt run --models dex_base_trades
+            dbt run --models dex_trades
+            dbt test --models test_check_null_amount_usd
+          fi
+        working-directory: ${{env.PROJECT_DIR}}
```
Hosuke, line 78: cc. @jeff-dude 
This part may be the tricky part.
I am not 100% sure, but it may represent the rough workflow in my head at the moment.

tests/generic/check_dex_null_amount_usd.sql
```diff
@@ -0,0 +1,27 @@
+{% test check_dex_null_amount_usd(model) %}
+
+    {%- set threshold = 0.05 -%}
```
Hosuke, line 3: This threshold is open to discuss.

.github/workflows/dbt_slim_ci.yml
```diff
@@ -68,6 +68,16 @@ jobs:
...
+      - name: dbt test dex trades specific model(s)
+        shell: bash
+        run: |
+          if git diff --name-only ${{ github.event.pull_request.base.sha }}..${{ github.event.pull_request.head.sha }} models/_sector/dex/trades/ | grep -q "models/_sector/dex/trades/"; then
+            dbt run --models dex_base_trades
+            dbt run --models dex_trades
+            dbt test --models test_check_null_amount_usd
+          fi
+        working-directory: ${{env.PROJECT_DIR}}
+
```
Hosuke, line 80: Hi @jeff-dude, I think I may need some helps for this part.

The test seems working now.

tests/generic/check_dex_null_amount_usd.sql
Hosuke: Maybe `0.16`.

jeff-dude:
documenting my testing.
few changes:
- removed test from schema file, as we only care to run it on demand via the CI workflow. we can explore adding it to schema file later, but we won't be running the test ongoing basis anyway.
- updated CI workflow to run new test after CI builds table and runs normal tests

first run:
- only two files modified in PR -- 1) the new test 2) the CI workflow changes
- therefore, step in CI to check for changes in `dex_base_trades` found no match, variable set to false
  - we can feel confident to run on `dex_base_trades` only, since a new chain being added would add here and stop
  - for now, test only needs to run for new chains (in theory this will run if spell is modified at all, but that is unlikely)
- next two steps are skipped, due to false return
- this is all as expected for this run

jeff-dude:
looks like the latest CI cluster setup is failing on dex rebuild, not great 🤔  we have the tools to tweak the cluster settings, so may need to if this keeps failing.

second run goals:
- force change to `dex_base_trades` to see it build in CI, to mimic as if a new chain were being added
- let the normal test step run after build
- new steps *should* run, trigger the variable to be `true` since `dex_base_trades` is in the PR
- ensure the new test runs
- see the results

i'm hitting pause on this for now. feel free to continue with the above steps if you want (if cluster allows).

we will also be raising the threshold to a higher number, as the intention is to ensure not *all* rows are null, because prices source doesn't have that chain. i'm thinking more towards 50% or so threshold.

tests/generic/check_dex_null_amount_usd.sql
Hosuke: I will change to 0.50 first.
```suggestion
    {%- set threshold = 0.50 -%}
```

Hosuke:
Current CI failure due to out of memory:
https://github.com/duneanalytics/spellbook/actions/runs/8971662945/job/24639905782#step:11:30

.github/workflows/dbt_slim_ci.yml
```diff
@@ -68,6 +68,22 @@ jobs:
       - name: dbt test initial model(s)
         run: "dbt test $PROFILE --select state:new state:modified --exclude tag:prod_exclude tag:remove --defer --state . --project-dir ${PROJECT_DIR}"
 
+      - name: Check for new blockchain additions to dex_trades
```
aalan3, line 71: These steps and below shouldn't be needed since a change to the dex_trades model will retrigger a run and then a test anyway or am I missing something?

jeff-dude: we don't join to prices until `dex_trades` spell via the enhance macro & when we add a new chain, it gets added to `dex_base_trades` prior to the enhance step. due to this, we won't know if we're missing prices since we don't have the data yet. CI steps will only runs what's edited, so it stops at `dex_base_trades`. we'd have to force a change to `dex_trades` to see that run each time, not really scalable.

adding a few steps to generic CI workflow isn't ideal for dex specifically. the other idea was introducing a new workflow and only run on PRs that hit this path indicated below.

i guess we could have a dex-specific CI workflow, where we essentially use this updated workflow fully, then keep the old one as-is?

jeff-dude:
i forgot dex isn't as simple as reading from `base_trades` 🤦‍♂️ 
this will require some more thinking, i won't have time to do this week most likely.

Hosuke:
> i forgot dex isn't as simple as reading from `base_trades` 🤦‍♂️ this will require some more thinking, i won't have time to do this week most likely.

We may do an alternative approach like short-cut the price join using the `*base_trades` without building the full `dex.trades`

jeff-dude:
i'm going to close this for now as we work through other priorities and also are building a list to use to integrate new chains into key spells that'll cover things like prices add.

keep the branch though, we may revisit!

PR #5864, opened 2024-05-01
Quick seaport cleanup

0xRobin:
Pending the full rewrite of the seaport macro's, this PR aims to reduce the computation footprint of the existing seaport models so we can actually run/test them in the CI pipeline.
The large runtime of these models was a blocker for https://github.com/duneanalytics/spellbook/pull/5855

In short:
- move existing opensea/seaport models into the nft/trades/old directory
- strip these models from unnecessary joins with nft/erc/prices data
- remove dangling downstream models that relied on the now removed columns

From running on the CI, this has resulted in a -80% runtime reduction for these models. :partying_face: 
The opensea base issue is not resolved in this PR, so that model remains excluded for now.


0xRobin:
@jeff-dude I've finally gotten the CI tests to pass here, should be good now

jeff-dude (approved):
great stuff -- thanks for getting this moving 🙏

PR #5867, opened 2024-05-02
Add zkevm to tokens.erc20

viniabussafi:
This PR creates a file for manual input of tokens on zkevm to tokens.erc20, as well as adds some missing tokens on other chains


jeff-dude (approved):
thanks for helping us stay up-to-date here 🙏

PR #5869, opened 2024-05-02
Added YAK token to prices on avalanche_c

yy-analytics:
Adding Yield Yak token (YAK) to the prices model on avalanche_c


jeff-dude (approved):
sounds good

PR #5872, opened 2024-05-02
Quick fix for balancer.bpt_supply

viniabussafi:
This PR introduces a quick fix to the `joins_and_exits` CTE on the `balancer.bpt_supply` macro. The previous query, which used a join condition for the `joins` and `exits` CTEs was causing a few gaps on the data and skewing results.


jeff-dude (approved):
sounds good 🫡

PR #5873, opened 2024-05-02
op tokens metadata: fix broken col names

jeff-dude:
fyi @MSilb7 
did you resubmit this decoded table possibly?
curious why column names changed. it started failing in prod.


MSilb7:
> fyi @MSilb7 did you resubmit this decoded table possibly? curious why column names changed. it started failing in prod.

Nothing changed to my knowledge. Weird.

PR #5875, opened 2024-05-03
Adding Yield Strategies key information for Yield Yak

yy-analytics:
cc @jeff-dude 

Opening PR as per [this](https://discord.com/channels/757637422384283659/999683200563564655/1235462876345536583) message in Discord.

For the Yield Yak strategies, key pieces of information which can be called from the contracts using `eth_call` are the `depositToken` and the `rewardToken`. Both are essential to be able to get the TVL and fees/revenue in USD. In this PR I've taken the approach of using the newly available LiveFetch `http_post` function to make a JSON RPC call of `eth_call` for the relevant contracts for the relevant function signatures. It has been done in such a way with an incremental model that each call should only be made once per contract. I have presumed this to be a good thing instead of making multiple redundant calls (these attributes do not change for these contracts). The reason for doing this instead of using tables like `StrategyName_call_depositToken` is that these functions are not called for every contract so there would be a lot of missing information.

A question that came up in Discord is whether this could be made more generic. At the moment, the returned output of the `eth_call` for the function `name` returns a string so is of dynamic length. There are no handy decoding functions in DuneSQL as far as I can tell but because I know the output for these contracts (particularly that this string can't be _that_ long) I can use `varbinary_substring()` to get the relevant part. Obviously this would need changing if it was to be further generalised and the decoding part would need to be considered more carefully.


macros/models/_project/yield_yak/yield_yak_eth_call.sql
```diff
@@ -0,0 +1,25 @@
...
+{%- endif -%}
+
+{#-
+Putting this onto a single line because it looks messy otherwise after running dbt compile
+It runs eth_call on the address_column_name for the given function signature and appropriate target_url.
+This returns a JSON object so we then retrieve the result using json_extra_scalar and then convert it
+to varbinary using from_hex
+-#}
+
+from_hex(json_extract_scalar(http_post('{{ target_url }}', concat('{"id":1,"jsonrpc":"2.0","method":"eth_call","params":[{"to":"', CAST({{ address_column_name }} AS varchar), '","data":"{{ function_signature }}"},"latest"]}'), ARRAY['Content-Type: application/json']), '$.result'))
```
jeff-dude, line 23: @kasiafi @Pluies here is the intro PR for trying out `http_post` in spellbook
looks like it's returning an access error:
https://github.com/duneanalytics/spellbook/actions/runs/8939609493/job/24555975242?pr=5875#step:11:25

kasiafi: [This](https://github.com/duneanalytics/arrakis-jobs/pull/4130) should fix the access control issue,

kasiafi: The access control issue is fixed. Now the query fails with
`TrinoQueryError(type=INTERNAL_ERROR, name=GENERIC_INTERNAL_ERROR, message="Your query issued too many HTTP requests", query_id=20240506_105317_00250_xzeks)`

yy-analytics: @kasiafi Yeah thanks for that fix re access control. Is there anything that can be done to get around the too many HTTP requests issue? Is that something being controlled by Dune or is that in reaction to a 429 response code/similar from the endpoint to which the `http_post()` is sent?

yy-analytics: Latest run now has a couple of error messages
- "http_post() function received unsuccessful response for url: https://rpc.ankr.com/arbitrum: 429: Too Many Requests"
- "Your query issued too many HTTP requests"
So seems like there might be a Dune-imposed limit on HTTP requests, but that I may also need to find a way to manage 429 response/error codes?

yy-analytics: I don't know how big/how much capacity Dune's nodes with quicknode are (?) but I would imagine they're pretty big/powerful. Maybe a way to share functionality like this more widely could be by having blockchain-dependent endpoints for JSON RPC declared as variables in `dbt_project.yml` by referencing some environment variables so that they can't be used outside of spellbook? Is this possible do you think?

kasiafi: > Latest run now has a couple of error messages
So seems like there might be a Dune-imposed limit on HTTP requests, but that I may also need to find a way to manage 429 response/error codes?

Apparently you hit both rate limits: Dune's internal limit and the external service limit. It looks like the internal limit is on par with the external service capacity.

Regarding 429, you cannot manage it directly. The `http_post` function will fail on any error code except 404 -- for that one it will return `null`. 

You can wrap the `http_post` function call in `TRY` SQL syntax so that all errors, including 429 code and Dune's internal limit, will be coalesced to `null`. However, it does not let you distinguish between different errors.

yy-analytics: Hi @kasiafi ,

Wrapping the `http_post` function call in `TRY` SQL syntax... That's pretty interesting and I can see potential for that to help make it work actually!

In my case at least, I could add a `WHERE` clause so that rows where one of the columns has a `NULL` value are not added to the incremental model (or similar approach where `NULL` values are retried). That way, if we exceed the external rate limit it will just try again the next time the incremental model runs. I could also add some additional controls so that we never expect more than X requests to be made in a single run (and this way for a full refresh, it would just build up slowly over multiple runs of the incremental model), which should help with the internal limit too.

I would imagine that the scaling issue could arise though if there are suddenly many spells using the LiveFetch functions. That would depend on things that I don't have much knowledge about - sequence in which models are run, whether any are run in parallel, and if so whether those parallel runs count as from the same origin/IP address so get externally rate limited etc. It doesn't seem impossible that something could work though 🤞 .

How often is it that incremental models are refreshed?

kasiafi: > In my case at least, I could add a WHERE clause so that rows where one of the columns has a NULL value are not added to the incremental model (or similar approach where NULL values are retried).

A nice workaround!

> That way, if we exceed the external rate limit it will just try again the next time the incremental model runs.

Note that this way the incremental model will always be lagging. Also, we don't control which calls will pass or fail, so theoretically you could end up missing some old entries forever. I think you'd need some mechanism to make sure that the data is complete beyond some time window.

> I could also add some additional controls so that we never expect more than X requests to be made in a single run

This is very important as using the `TRY` syntax practically disables  the rate limit.

> That would depend on things that I don't have much knowledge about - sequence in which models are run, whether any are run in parallel, and if so whether those parallel runs count as from the same origin/IP address so get externally rate limited etc.

That's exactly what I meant by "add another dimension of complexity for the guys maintaining Spellbook if they have to juggle the rate limits". I believe we cannot make the LiveFetch functions part of Spellbook until we figure out how to maintain them and how it will scale over time. 

> How often is it that incremental models are refreshed? (Not like a full refresh, just updated/added to I mean).

Some are daily, some are hourly. This is subject to change.

yy-analytics: > That's exactly what I meant by "add another dimension of complexity for the guys maintaining Spellbook if they have to juggle the rate limits". I believe we cannot make the LiveFetch functions part of Spellbook until we figure out how to maintain them and how it will scale over time.

@kasiafi , thanks for your response! In that case is there anything further I can do on this? I would be happy to have a go as using `TRY` and the workaround we discussed to make it more practical in this case. But if we need to wait until the maintenance and scaling is more solid before doing that also then I understand.

kasiafi: @yy-analytics let me discuss with the team and get back to you.

yy-analytics: Hi @kasiafi , did anything come from your discussions? If not then I think I will try a different approach for this with a simple view in spellbook and can then save a materialized view query that works from this and enhances it using the LiveFetch functions, so it will kind of be like an incrementally updating table, but as a materialized view instead. That's what I've been doing for now, but without the spellbook view, which I'll add to clean things up a bit if you think that we can't use LiveFetch in spellbook yet.

jeff-dude: gm @yy-analytics 
thank you for the patience here, as we're still experimenting on exactly what this may look like in spellbook. i noticed you've proceeded with the above in another PR, and i think it's for the best right now to proceed that route. once we finish the efforts to scale spellbook, we can hopefully revisit the usage of this function and how to best manage it here. with that said, i'll close this PR for now

PR #5880, opened 2024-05-04
Added Consortium Key Trades for Solana

clizzard7:
Added Consortium Key Trades for Solana. It is a DEX trading bot.


Hosuke (commented):
You can add `consortium_key_solana_bot_trades ` into https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/bot_trades/solana/dex_solana_bot_trades.sql

clizzard7:
> You can add `consortium_key_solana_bot_trades ` into https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/bot_trades/solana/dex_solana_bot_trades.sql

done, thank you.

clizzard7:
This board does not have fee_usd, fee_token_amount etc. columns, but the schema expects it. What should we do here?
Thank you.

Hosuke:
> This board does not have fee_usd, fee_token_amount etc. columns, but the schema expects it. What should we do here? Thank you.

@whalehunting do you have any idea for this question?

whalehunting:
> > This bot does not have fee_usd, fee_token_amount etc. columns, but the schema expects it. What should we do here? Thank you.
> 
> @whalehunting do you have any idea for this question?

@Hosuke  can we fill these columns with NULL for this spellbook? fees are not paid during the trade in this bot. since you unified the schema for all trading bots, it's not possible to change the schema for this one right?

Hosuke:
I think we can put `NULL` for `fee_usd`, `fee_token_amount` at the moment.

whalehunting:
> I think we can put `NULL` for `fee_usd`, `fee_token_amount` at the moment.

where should this be entered? on dune itself you can't add NULL for every row, since the engine will not be able to determine the correct datatype

models/_sector/dex/bot_trades/solana/platforms/consortium_key_solana_bot_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+    token_sold_address,
+    project,
+    version,
+    token_pair,
+    project_contract_address,
+    user,
+    bottrades.tx_id,
+    tx_index,
+    bottrades.outer_instruction_index,
+    coalesce(inner_instruction_index, 0) as inner_instruction_index,
```
Hosuke, line 94: Maybe we can try this way, and also add them in the schema please:
```suggestion
    tx_index,
    bottrades.outer_instruction_index,
    CAST(NULL AS DOUBLE) as fee_usd,
    CAST(NULL AS DOUBLE) as fee_token_amount,
    coalesce(inner_instruction_index, 0) as inner_instruction_index,
```

Hosuke:
> > I think we can put `NULL` for `fee_usd`, `fee_token_amount` at the moment.
> 
> where should this be entered? on dune itself you can't add NULL for every row, since the engine will not be able to determine the correct datatype
> 
> also it seems that the error happens in dex.bot_solana, not in the consortium key spellbook https://github.com/duneanalytics/spellbook/actions/runs/8957404379/job/24600466010?pr=5880#step:11:20

We can try the above approach.

whalehunting:
@Hosuke CI tests finally passing, can we pls get this merged? thank you!

Hosuke (approved):
LGTM.✅
Thank you @clizzard7 

[Sample consortium_key_solana_bot_trades]()

models/_sector/dex/bot_trades/solana/platforms/consortium_key_solana_bot_trades.sql
```diff
@@ -0,0 +1,118 @@
...
+        select distinct transactions.signer as user
+        from {{ source("solana", "account_activity") }} as activity
+        join {{ source("solana", "transactions") }} as transactions on tx_id = id
+        where
+            activity.block_time >= timestamp '{{project_start_date}}'  -- FirstTradeTimestamp
+            and transactions.block_time >= timestamp '{{project_start_date}}'  -- FirstTradeTimestamp
+            and tx_success
+            and address = '{{fee_receiver_1}}'  -- SolanaFeeWallet
+            and balance_change > 0  -- SOL fee payments
+    ),
```
jeff-dude, line 34: is this model unique and requires this CTE, or can we avoid it? i notice most others in this spell lineage do this:
https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/bot_trades/solana/platforms/bonkbot_solana_bot_trades.sql#L95-L110
which seems to be more efficient runtime. 
notice in CI runs here, the incremental step takes 10+ minutes, which appears to be due to this CTE which takes ~6 minutes to return results on it's own

whalehunting: since we need to fetch all users in order to fetch their trades, i guess we can not get rid of this unfortunately @jeff-dude

jeff-dude: can we at least look to be consistent between the two? one filters incrementally on account activity, while the other does not. if possible, please add incremental filter to transactions too

whalehunting: i guess we should remove the incremental filter from the other spellbook. correct me if i'm wrong but when we do incremental filters here, we would probably miss historical users?

jeff-dude: most likely, yes. is this the use case?
- new trades in dex solana could have users from months ago, so if we filter incrementally on users, it would find no match.
- if that isn't the case, then we could be safe to assume new trades find new users

do we know how often new users are output in these CTEs? would it make sense to make their own spell upstream? we could incrementally build that spell to add new users, then call the spell in these ones for much faster runtime

whalehunting: sorry i don't understand what you're trying to say here. the methodology in the CTE is the only way to find solana users, since CK does not pay fees in the trade txs, but waits for a certain threshold until it sends them to the fee address

if the problem is dex.bot_trades, then we could just not include it there, the important thing here is that the data goes live, so the CK dashboard will have solana metrics again

jeff-dude: the concern is [here](https://github.com/duneanalytics/spellbook/actions/runs/9156694489/job/25171548525?pr=5880#step:14:19). that's a long runtime for incremental runs. you can see in [other PR](https://github.com/duneanalytics/spellbook/actions/runs/9157874643/job/25175092596?pr=5881#step:14:21), the two models which don't use this CTE logic run in ~60s during incremental step

i'd like to see that improved potentially by removing the CTE in the slow ones and making the CTE it's own spell. then replace CTE with that spell in join conditions. can we try that?

whalehunting: sure we can! would it be possible to get this merged before? and then i'll submit a new PR with the additional spellbook + changes in this one?

jeff-dude: that works for me. i'll get this in & then we can work on the enhancements.

jeff-dude (approved):

PR #5883, opened 2024-05-05
Added new tokens to Arbitrum prices model

yy-analytics:
Adding a collection of tokens to the prices model for Arbitrum. Reason for the addition is that these tokens are/have been traded on Yield Yak's Yak Swap but do not have price data currently. This is the subset of such tokens which are already present on Coinpaprika for Arbitrum.


Hosuke (approved):
LGTM.✅
Thank you @yy-analytics

PR #5885, opened 2024-05-06
Added new strategies for Yield Yak on Arbitrum

yy-analytics:
Added a couple of new strategy names to the yield_yak_strategies macro so that they are included in other models.


jeff-dude (approved):
sounds good

PR #5886, opened 2024-05-06
add pool_type and pool_symbol to balancer.trades models

viniabussafi:
This PR adds 2 additional fields to balancer.trades models: pool_type and pool_symbol


jeff-dude (approved):
sounds good

PR #5889, opened 2024-05-06
orca_whirlpool: fix table alias, consolidate schema files

jeff-dude:
fyi @thanhlmm 
- consolidate schema files to one per directory
- fix table alias in latest hourly spell


thanhlmm:
Thank @jeff-dude , I will be careful in future MR

PR #5890, opened 2024-05-06
Added Yield Yak deposit tokens for Arbitrum strategies

yy-analytics:
Adding a few tokens manually with their symbol and decimals. These are used as deposit tokens in Yield Yak strategies, so in order to get the balances/TVL of the strategies we need to know the decimals and symbol of these tokens.


jeff-dude (approved):
sounds good

PR #5901, opened 2024-05-08
polygon zkevm: Add back missing tables

jeff-dude:
fyi @viniabussafi @chef-seaweed 
my understanding is the team has already rebuilt. we will test out in CI here.


jeff-dude:
will give it a go in prod then 🙏

PR #5906, opened 2024-05-09
Added user level balance tracking for Yield Yak YRT tokens

yy-analytics:
This PR adds a temporal model for Yak Receipt Token (YRT) balances for the Yield Yak strategies, using `Transfer` events. It allows us to track wallet-level balances in each of the strategies over time, and in combination with the strategy-level balances model, will enable individual users to monitor their deposit token balances over time. It will also make the tracking of the `totalSupply` of any of these YRT tokens pretty straightforward too as we can just track the negative balance of the zero address.

I've taken a similar approach to previous models (e.g. `yield_yak_balances.sql` macro), but if there are any questions on any of it please let me know.

### DIFF
--- a/macros/models/_project/yield_yak/yield_yak_user_yrt_balances.sql
+++ b/macros/models/_project/yield_yak/yield_yak_user_yrt_balances.sql
@@ -1,0 +1,129 @@
+{%- macro yield_yak_user_yrt_balances(
+        blockchain = null
+    )
+-%}
+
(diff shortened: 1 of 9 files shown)


yy-analytics:
Working on more significant changes to address "TOO MANY STAGES" issues

yy-analytics:
Okay looks like that last commit got it working ☺️

jeff-dude (approved):
this works for me. thanks for continuing to write efficient & clean code!

PR #5910, opened 2024-05-09
1inch: added fusion to ar_trades & ar and lop macro improvements

max-morrow:
- added fusion to ar_trades (by analogy with CoW swap: 1inch Fusion, CoW Swap and UniswapX - similar systems)
- added "second side" of LOP calls to ar_trades (these are the volumes of users who came to dApp and exchanged through 1inch classic swap, but the aggregator in the route used only one liquidity source for this exchange: a counter limit order)
- migrated to flags ar_trades & lop_own_trades
- oneinch_swaps: removed legacy columns
- oneinch_ar_macro improvements
- fixed partial & multiple flags setting (important fix)


jeff-dude (approved):

jeff-dude:
canceling full run in CI, as i saw the easy dates run. will try in prod, we've had success the last few.

PR #5911, opened 2024-05-09
[Easy] - Chains Update #1 - Update Get Chain Explorer lInks of new chains 

henrystats:


Hosuke (approved):
LGTM.✅
Thank you @henrystats

jeff-dude (approved):
i want to be sure we're on the same page here. these macro changes will be in effect for spellbook code, but will *not* automatically be built (or updated) as UDF on dunesql engine. that's still a gap in the process, until we build out the full feature stack necessary to build UDF's on the dunesql engine via spellbook.

note: i'm not 100% sure these macros are the ones we've built on dunesql in the past (manually), but just making note in case :)

PR #5921, opened 2024-05-14
Labels fix: ignore transfers above 1 trillion

0xRobin:
This transaction messed with the total stats:
https://etherscan.io/tx/0x35937ed72826da00bdaa784b4b13e703fecbb89e168f900276860f9c68f87d76

Solution is to ignore any transfers with an insane amount


jeff-dude (approved):
we're good to proceed here

PR #5925, opened 2024-05-15
transform balancer spells to incremental

viniabussafi:
This PR transforms a few balancer spells (protocol fees, liquidity, bpt prices and bpt supply) into incremental, in order to have them updated on a higher frequency.

### DIFF
--- a/macros/models/_project/balancer/balancer_bpt_prices_macro.sql
+++ b/macros/models/_project/balancer/balancer_bpt_prices_macro.sql
@@ -23,6 +23,9 @@ WITH pool_labels AS (
             AVG(price) AS price
         FROM {{ source('prices', 'usd') }}
         WHERE blockchain = '{{blockchain}}'
+        {% if is_incremental() %}
+        AND {{ incremental_predicate('minute') }}
+        {% endif %}
         GROUP BY 1, 2, 3
     ),
 
@@ -49,6 +52,9 @@ WITH pool_labels AS (
                     tokenIn AS token,
                     CAST(amountIn as int256) AS delta
                 FROM {{ source('balancer_v2_' ~ blockchain, 'Vault_evt_Swap') }}
+                {% if is_incremental() %}
+                WHERE {{ incremental_predicate('evt_block_time') }}
+                {% endif %}
 
                 UNION ALL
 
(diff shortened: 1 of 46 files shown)


viniabussafi:
hey, @jeff-dude!
I think test are failing due to them taking too long. Let me know if you think we should split this PR into several others.
Also, when you have the time, could you please review the setup of the incremental models and incremental_predicates, just to make sure that I got them right?
The idea is having these spells updated more regularly so we can use Dune as a source of truth when comparing other data sources.
Thanks for your help, as always!

jeff-dude:
> I think test are failing due to them taking too long. Let me know if you think we should split this PR into several others.

looks like it was just unfortunate timing on the last CI run, where you picked up models from other PRs because they were merged at the time this kicked off. the job to upload new dbt manifest to avoid this was likely in progress, rather than complete. i'm trying the CI again now.



> Also, when you have the time, could you please review the setup of the incremental models and incremental_predicates, just to make sure that I got them right?

from a quick glance, the config all looks good to me. just plz double check all sources in the macros that are time based (i.e. event decoded or prices, etc) contain the filter, both in from statements and join conditions 🙏

jeff-dude:
looks like there is a syntax error on the incremental run, meaning somewhere in the is incremental logic.

also, looks like bpt macro is long running for incremental, is it possible you are missing time  filters somewhere?

jeff-dude:
> Am I correct in my assumption? If so, how are incremental runs of balances spells dealt with at the moment?

good question. we need to compile a good list of examples for this. are these the balances you're referring to?
https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/tokens/balances_daily_agg.sql
https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/tokens/balances_daily.sql

viniabussafi:
yes, I've been trying to create similar on this other [PR](https://github.com/duneanalytics/spellbook/pull/5958), which I hope to have finished by tomorrow.
Maybe we can keep this PR on hold for now and see how that other one goes if that's fine by you

viniabussafi:
Closing this one, based on: https://github.com/duneanalytics/spellbook/discussions/6037#discussion-6753023

> Moving forward, all spells outside of the main sub-projects will run on a daily cadence. No changes need to be made to materialization type, rather the orchestration logic will change to make it work.

PR #5926, opened 2024-05-15
Added models for token transfers and balance for Gnosis

hdser:

Thank you for taking the time to submit code in Spellbook. A few things to consider:

- If you are a first-time contributor, please sign the CLA by copy & pasting **_exactly_** what the bot mentions in PR comment
- Refer to [docs](#spellbook-contribution-docs) section below to answer questions
- Dune team will review submitted PRs as soon as possible

### To speed up your development process in PRs, keep these tips in mind:

- Each commit to your feature branch will rerun CI tests ([see example](https://github.com/duneanalytics/spellbook/actions/runs/8202519819/job/22433451880?pr=5519))
    - This includes *all* modified models on your branch
    - This includes *all* history of the data
- Two tips for faster development iteration:
    - Ensure dbt is installed locally (refer to main `readme`) and run `dbt compile`
        - This will output raw SQL in `target/` directory to copy/paste and run on Dune directly for initial query testing
    - Hardcode a `WHERE` filter for only ~7 days of history on large source tables, i.e. `ethereum.transactions`
        - This will speed up the CI tests and output results quicker -- whether that's an error or fully successful run
        - Once comfortable with small timeframe, remove filter and let full history run

### Incremental model setup
- Make sure your unique key columns are *exactly* the same in the model config block, schema yml file, and seed match columns (where applicable)
- There cannot be nulls in the unique key columns
    - Be sure to double check key columns are correct or `COALESCE()` as needed on key column(s), otherwise the tests may fail on duplicates

### 🪄 Use the built CI tables for testing 🪄

Once CI completes, you can query the CI tables and errors in dune when it finishes running.
- For example:
    - In the `run initial models` and `test initial models`, there will be a schema that looks like this: `test_schema.git_dunesql_4da8bae_sudoswap_v2_base_pool_creations`
    - This can be temporarily queried in Dune for ~24 hours

Leverage these tables to perform QA testing on Dune query editor -- or even full test dashboards!

## Spellbook contribution docs

The [docs](docs) directory has been implemented to answer as many questions as possible. Please take the time to reference each `.md` file within this directory to understand how to efficiently contribute & why the repo is designed as it is 🪄

Example questions to be answered:

- What does each property in the [model config block](/docs/models/model_config_block.md) mean?
- What is the [CI test](/docs/ci_test/ci_test_overview.md) attached to PRs and how can I best utilize it?
- What are the Spellbook [best practices](/docs/general/best_practices.md)?

Please navigate through the [docs](/docs) directory to find as much info as you can.

**Note**: happy to take PRs to improve the docs, let us know 🤝


jeff-dude:
hi @hdser 
thanks for marking this ready for review & submitting the code 🙏 

we are slowly moving away from the spells in these directories in favor of our updated approach to transfers / balances, as seen here:
https://github.com/duneanalytics/spellbook/tree/main/models/_sector/tokens/gnosis

it looks like at the moment, we may not have balances for gnosis, but the raw balances data is built/loaded outside of spellbook. you can see examples of new spells on balances raw data if you navigate back a layer to ethereum chain (or a few others).

do these new ones accomplish what you are looking for? ideally we can move away from adding these.

hdser:
@jeff-dude unfortunately for the naive token there are several cases such approach as the one in https://github.com/duneanalytics/spellbook/tree/main/models/_sector/tokens/gnosis fails:

- contracts can be create and in the creation transaction they receive native token. One needs to keep track of the created contract address since in the transaction that address is NULL
- contracts can be destroyed and the native balance of that contract can be refunded to an address. One need to keep track of the balances of the contracts in order to fully construct the "transfers" of native token. This is why there model _balances_suicide_gnosis_xdai_ was added.
- on gnosis there is a block reward contract that "transfers" native token into the address of validator of the block
- there is also a fee collector into which all EIP-1559 "burned fees" are collected and this is done ar the protocol level no transaction.

I think for ERC20 tokens, it should be fine. In the end, in this PR I am not actually creating balance tables but balance diff tables which are essentially transfers. The difference between the transfers and balance diff, is that transfers include the counterparty address of that transaction.

These tables, transfers (xDAI, tokens) and balance diffs, are being heavily used in our dashboards. I am happy to discuss this is more detail and improve the implementation, so that these are not too heavy tables.

hdser:
OK!
Let me try build on top of the tokens_gnosis.transfers

hdser:
@jeff-dude I started working on the tokens spell. Most changes were done for the gnosis network, however in the macro `transfer_base` there is a change done, i.e. 

` COALESCE(to,address) AS to -- Contract Creation has NULL "to" address, but transaction might have value that goes to contract created`

What this does is to account the contract address that has been created in the transaction as the `to` address. In many creation transactions the value of the transaction is not zero and is added to the created contract.
However, because of this change when I run the models all the chains also run and this gets timed out. Is there a way I can test the gnosis specific additions without having to run all these models?

aalan3:
>However, because of this change when I run the models all the chains also run and this gets timed out. Is there a way I can test the gnosis specific additions without having to run all these models?

I would copy the macro and use the copy while developing, that way you can avoid running all the other models.

hdser:
Good point!

Will do that

hdser:
@jeff-dude PR is ready. Suicide model is not incremental, it computes the contract balance up to the suicide call.

jeff-dude:
thank you for continued patience on this PR. i will try to revisit asap to see how we can consolidate transfers as efficiently as possible across spellbook.

jeff-dude:
sorry for the inconvenience, but we completed multi-project setup in spellbook while this PR was open. to fix merge conflicts, your code should now live in `dbt_subprojects/tokens/` directory path

hdser:
@jeff-dude created a new PR https://github.com/duneanalytics/spellbook/pull/6480 to accommodate the new structure.

hdser:
I am closing this PR and moving to the new one

Issue #5504, opened 2024-03-04
[Spell Design] Missing Context on Aave Macro


lgingerich:
The current [macro ](https://github.com/duneanalytics/spellbook/blob/main/macros/models/_sector/lending/lending_aave_compatible_borrow.sql#L256) for Aave V3 has a single column for the user on borrows. `aave_v3_ethereum.Pool_evt_Borrow` has two columns that relate to the user: `user` and `onBehalfOf`. From a quick look, this also affects Aave V2, but I didn't check beyond that.

By omitting data in `onBehalfOf`, there is missing context in `lending.borrow`. It's not possible to calculate net borrow/lend since we don't know who's account funds were borrowed from.

So we need to add this `onBehalfOf` column back in somehow. I'm imagining we want to have both columns and not just swap these out. I wasn't part of the creation of the Aave model or any borrow/lend stuff, so I wanted to open an issue and get input from those that did work on it.

cc: @Hosuke @tomfutago @jeff-dude

tomfutago:
Hi @lgingerich - I used existing spells at the time of adding this macro (one of my early on Spellbook projects). Adding `onBehalfOf` sounds like a valid point - will have a look at adding it in. Any preference for the column name? I'd go with `on_behalf_of` to keep snake case convention as for other column names, but open to suggestions.

lgingerich:
`on_behalf_of` makes sense to me. The main reason I opened an issue for this is that it will of course effect the overall lending table schema. Before officially going with that name, I want to investigate if Compound and others have anything else we need to handle on this.

jeff-dude:
closing issue due to inactivity. if still relevant, and proposed fix is available, feel free to reopen an issue as needed.

PR #5951, opened 2024-05-20
lending sector additions

tomfutago:
Added columns:
* `project_contract_address`
* `on_behalf_of` (as per #5504)

Also, experimentally removed `prod_exclude` tag (ref #5776), but potentially more changes are needed here (@jeff-dude lmk if removing partitions is the way to go).


jeff-dude (approved):
niiice, thank you!

i'm good with removing exclude tags, now that i have a better idea of the issue we ran into and tracking with our engine team. not sure it's the partitions yet, but we are closing in on possible solutions. we can let this ride in prod and i'll be monitoring it.

PR #5952, opened 2024-05-20
Test Missing Null Uniques -  transfers_base_eth.sql

MSilb7:
Optimism Portal transfers are missing from op chains' eth transfer spells.

Testing if the issue stems from the null trace address in the unique ID field cc @chuxinh .

### DIFF
--- a/models/transfers/base/transfers_base_eth.sql
+++ b/models/transfers/base/transfers_base_eth.sql
@@ -25,7 +25,7 @@ with eth_transfers as (
         ,r.block_time as tx_block_time
         ,r.block_number as tx_block_number
         ,substring(to_hex(t.data), 1, 10) as tx_method_id
-        ,cast(r.tx_hash as varchar) || '-' || array_join(r.trace_address,',') as unique_transfer_id
+        ,cast(r.tx_hash as varchar) || '-' || COALESCE( NULLIF(array_join(r.trace_address,','),''), '_') as unique_transfer_id
         ,t.to AS tx_to
         ,t."from" AS tx_from
     from {{ source('base', 'traces') }} as r


MSilb7:
Seems like this resolved the issue, we can push to other OP Chains as well:
```
SELECT * FROM test_schema.git_dunesql_a975727_transfers_base_eth

where tx_hash = 0xab3d87264417ed51346669885cae69822fab1ffb1b1e9920c56f4133f08241f9
AND tx_block_number = 14726349
```

jeff-dude:
now that this PR is open, i was hoping we could determine if we can deprecate these spells? now that we have chain-specific transfers spells in the `_sector/` directory (optimism for example, base spell for pipeline and final spell):
https://github.com/duneanalytics/spellbook/tree/main/models/_sector/tokens/optimism

they seem to query the data v similarly, am i misunderstanding here?

MSilb7:
> now that this PR is open, i was hoping we could determine if we can deprecate these spells? now that we have chain-specific transfers spells in the `_sector/` directory (optimism for example, base spell for pipeline and final spell): https://github.com/duneanalytics/spellbook/tree/main/models/_sector/tokens/optimism
> 
> they seem to query the data v similarly, am i misunderstanding here?

Seems like that's just ERC20s?

This is for native ETH transfers, but it would be way better if this was an expected table for every chain, versus needing a spell.

MSilb7:
Maybe we maintain these as a view, since they're used pretty heavily. We should also check if any columns created here aren't maintained (i.e. tx method id)

Also, the new transfers https://dune.com/discover/content/relevant?q=code:%22transfers.base_eth%22

jeff-dude:
i may be a bit confused on how you're suggesting we proceed. the new transfers spells account for all you need, except pre-bedrock? could we build a view spell for pre-bedrock, then union it in to new transfers, whether that's within queries on the app or in a spell?

or are you suggesting you just turn your version of transfers as-is into views?

jeff-dude:

PR #5954, opened 2024-05-21
cleanup: move limitbreak macro

0xRobin:
cleanup: move limitbreak macro


jeff-dude (approved):
thank you, actually just noticed the limitbreak macro path yesterday and wanted to change

PR #5972, opened 2024-05-22
flashloans: keep downstream field name the same

jeff-dude:
fyi @tomfutago 
hit a bit of a tangled mess downstream with these 🤦‍♂️  i'm going to keep the old name on downstream objects, as it leads to more trouble.


tomfutago:
Oops.. forgot about all those drops 😬 sorry about that!

jeff-dude:
all good, it was my request!

PR #5980, opened 2024-05-23
Nexus Mutual - capital pool

tomfutago:


models/nexusmutual/ethereum/capital_pool/nexusmutual_ethereum_capital_pool_totals.sql
```diff
@@ -0,0 +1,500 @@
...
+  -- AAVE positions
+  aave_collateral_weth_total,
+  avg_aave_collateral_weth_usd_total,
+  aave_debt_usdc_total,
+  avg_aave_debt_usdc_usd_total,
+  avg_aave_debt_usdc_eth_total
+from daily_running_totals_enriched
+{% if is_incremental() %}
+where {{ incremental_predicate('block_date') }}
+{% endif %}
```
jeff-dude, line 500: are you not able to filter incrementally in CTEs above, since you need to sum() for all history? is that why it happens only at the end here?

tomfutago: Unfortunately not easily with all these running totals - many of them depending on each other. I know it seems like a waste to keep re-running the whole thing over and over again, but if I find a better way later on, will re-visit it 🤞

jeff-dude: all good, just confirming. i also need to take the time to find a good solution for that, as it's come up a few times.

jeff-dude:
just a heads up here. i noticed you use a reference to a chainlink spell in this lineage. we were in the process of moving the `chainlink` lineages into a new subproject, as an example to continue testing it out. it made me realize we need to double check downstream dependencies first. sorry if that delays this a little bit.

an alternative could be to proactively change it to a `source()` usage in this lineage, so it wouldn't break if spells moved around.

tomfutago:
> just a heads up here. i noticed you use a reference to a chainlink spell in this lineage. we were in the process of moving the `chainlink` lineages into a new subproject, as an example to continue testing it out. it made me realize we need to double check downstream dependencies first. sorry if that delays this a little bit.
> 
> an alternative could be to proactively change it to a `source()` usage in this lineage, so it wouldn't break if spells moved around.

hmm, not sure if I'm doing this right - I updated `ref()` to `source()` for chainlink but it doesn't seem to work atm.. is there an additional chainlink related change that needs to be merged first for that to work?

jeff-dude:
> hmm, not sure if I'm doing this right - I updated `ref()` to `source()` for chainlink but it doesn't seem to work atm.. is there an additional chainlink related change that needs to be merged first for that to work?

you're on the right path, but we also need to declare as a source in yml file. eventually these spells will be in same subproject as chainlink, so we could eventually change back to ref, but for now we've been putting here:
https://github.com/duneanalytics/spellbook/tree/main/sources/_subprojects/spellbook

it's a bit scattered in the short term, but we're taking immediate steps towards this:
https://github.com/duneanalytics/spellbook/discussions/6037

once it's all set up, we should be clean again :)

jeff-dude (approved):
looking good. last thing, can you add [this](https://github.com/duneanalytics/spellbook/blob/main/models/_sector/dex/trades/dex_trades.sql#L9) to all incremental models, to match source filter? it's a practice we're trying to push more. there is large effect on larger models, while smaller ones not as much, but still good to get in the habit

PR #6000, opened 2024-05-27
Seaport refactor to base trades schema

0xRobin:
Seaport refactor to base trades schema.

The logic changes are mainly in:
- `macros/models/_sector/nft/platforms/seaport_v3_trades.sql`
- `macros/models/_sector/nft/platforms/seaport_v4_trades.sql`

and the other changes are moving the models into the correct chain lineage and adapting the seed tests.


jeff-dude:
hmm did i introduce merge conflicts with a merge today?
i think i'll let you look through those, if you don't mind. i don't want to break it.

PR #6020, opened 2024-05-28
test out the daily subproject

jeff-dude:
notes:
- keep macros at the root of spellbook to be used universally, similar to sources
- eventually clean up dbt project file and remove schema assignments, include in model config directly instead

Edit from Alan: I updated the CI steps here so that each sub project runs in a separate workflow but using a shared workflow (like calling function).


aalan3:
Will open a new PR for this. CI changes have been moved to this PR instead https://github.com/duneanalytics/spellbook/pull/6025

PR #6024, opened 2024-05-29
Yield Yak swap and adapter swap tables

yy-analytics:
In this PR, I've made a couple of new macros `yield_yak_yak_swap.sql` and `yield_yak_yak_adapter_swap.sql` which feed into models on Avalanche and Arbitrum.

The intention is for these to be used/referenced in downstream queries/dashboards that track volume, fees and other metrics at both an overall level and per adapter for our DEX Aggregator product.

The way I've paired the adapter data with the YakSwap event data is actually through direct use of the raw logs tables. The reason for doing it this way is that we don't have a reliable list of adapters in use that could be hard-coded in spellbook and because the models are incremental, if we did hardcode a current selection and then needed to add new adapters retrospectively in future, it would be difficult/messy/might require a full reload. Having this set up incrementally also should make the downstream queries a bit easier as we're not going to be having to `JOIN` the transactions or logs tables.


models/yield_yak/yield_yak_yak_swaps.sql
```diff
@@ -0,0 +1,46 @@
+{{ config(
+	    schema = 'yield_yak',
+        alias = 'yak_swaps',
+        post_hook='{{ expose_spells(
```
Hosuke, line 4:
```suggestion
	schema = 'yield_yak',
        alias = 'yak_swaps',
        materialized = 'view',
        post_hook='{{ expose_spells(
```

models/yield_yak/yield_yak_yak_adapter_swaps.sql
```diff
@@ -0,0 +1,50 @@
+{{ config(
+	    schema = 'yield_yak',
+        alias = 'yak_adapter_swaps',
+        post_hook='{{ expose_spells(
```
Hosuke, line 4:
```suggestion
	schema = 'yield_yak',
        alias = 'yak_adapter_swaps',
        materialized = 'view',
        post_hook='{{ expose_spells(
```

Hosuke (changes requested):
Nice job,
I put minor comments about materialization.

models/yield_yak/yield_yak_schema.yml
```diff
@@ -248,4 +248,118 @@ models:
         description: "UTC time until which the user held this balance of the contract_address YRT"
       - &yrt_balance
         name: yrt_balance
-        description: "Balance of YRT tokens held by the user_address in the time window specified"
\ No newline at end of file
+        description: "Balance of YRT tokens held by the user_address in the time window specified"
+  - name: yield_yak_yak_swaps
```
Hosuke, line 252: You can add a blank line between model definition:
```suggestion
        description: "Balance of YRT tokens held by the user_address in the time window specified"

  - name: yield_yak_yak_swaps
```

models/yield_yak/yield_yak_schema.yml
```diff
@@ -248,4 +248,118 @@ models:
...
+      - &swap_amount_out
+        name: swap_amount_out
+        description: "Amount of the bought token being swapped to / bought in the YakSwap event"
+      - &number_of_hops
+        name: number_of_hops
+        description: "Number of steps/hops the YakSwap has gone through in order to get the desired token at the best rate"
+      - &yak_adapter_swaps
+        name: yak_adapter_swaps
+        description: "JSON array containing detailed information about the individual YakAdapterSwap events which make up this YakSwap"
+  - name: yield_yak_yak_adapter_swaps
```
Hosuke, line 312:
```suggestion
        description: "JSON array containing detailed information about the individual YakAdapterSwap events which make up this YakSwap"

  - name: yield_yak_yak_adapter_swaps
```

Hosuke (approved):
LGTM.✅
Thank you @yy-analytics 

[Sample yield_yak_yak_swaps]()
[Sample yield_yak_yak_adapter_swaps]()

models/yield_yak/yield_yak_schema.yml
yy-analytics: Okay sure, will make this change, plus will add blank lines between other models and will also add the explicit `materialized = 'view'` line on some of the other relevant models that I've put in spellbook a while ago for this project :)

jeff-dude (approved):
another 🔥 PR -- thank you!

PR #6026, opened 2024-05-29
[easy] move solana old models

0xRobin:
Non-functional change to move some solana models without triggering CI


0xRobin:
seems like `state:modified` is still triggered when moving models..
@jeff-dude we might want to merge this and run without the modified flag in prod if we want to spare it from running the `magiceden_solana_events.sql` model

jeff-dude (approved):
noted, i will save this to merge alone and hopefully asap

PR #6034, opened 2024-05-30
Move nft tokens metadata to tokens subproject

aalan3:
This PR moves the existing tokens.nft lineage to the tokens subproject. Solana tokens are not moved for now. This is in preparation for moving the contract mappings lineage


macros/models/_sector/nft/enrich_nft_trades.sql
```diff
@@ -103,7 +103,7 @@ SELECT
         ELSE coalesce(agg_mark.aggregator_name, agg1.name, agg2.name)
         END as aggregator_name
 FROM {{base_trades}} base
-LEFT JOIN {{ref('tokens_nft')}} nft
+LEFT JOIN {{source('tokens', 'nft')}} nft
```
jeff-dude, line 106: fyi @0xRobin 
we are looking to move nft token metadata to tokens subproject, similar to erc20, flipping to source is part of that

tokens/models/tokens/optimism/tokens_optimism_nft_generated.sql
jeff-dude, line 1: i think this one may be backwards?
all of these edition metadata spells should be moved to tokens subproject as they are dependencies for `tokens_nft` lineage. otherwise we create circular dependencies?

tokens/models/tokens/avalanche_c/tokens__schema.yml
jeff-dude, line 1: can be deleted?

tokens/models/tokens/optimism/tokens_optimism_nft_generated.sql
jeff-dude: `dbt ls --resource-type model --output name --select decent_optimism_edition_metadata+1`
`dbt ls --resource-type model --output name --select +decent_optimism_edition_metadata`

jeff-dude: nothing upstream, only token metadata downstream

aalan3: yeah you are right I can move this to the tokens subproject

PR #6036, opened 2024-05-30
Daily spellbook account abstractions

aalan3:
This moves account abstractions which has a pretty clean lineage


aalan3:
@jeff-dude in the last two commits I've updated these models to use the incremental predicate, figured it's a good time to do it since these will run in a full refresh now

jeff-dude (approved):
this is a nice clean one. will be a good example to link for communications / reference

PR #6041, opened 2024-05-31
Daily spellbook contract mappings

aalan3:
This big PR with the goal of moving contract mappings to daily spellbook. `addresses_events` also needed to be moved as part of this. And in turn I had to modify `nft.wash_trades` so that they reference via `source` instead of `ref`


aalan3:
@0xRobin 

>what if some upstream dependencies of a model run only daily but the model runs hourly? This seems like something we should avoid.
This is something that already happens and even in this lineage. Thing materialized as tables or labels are already on a daily schedule. I agree that this is something that should ideally be avoided but I don't think it's always possible. 

In the case of wash trades for example new wash traders will be discovered once a day (and I think this might partly already be the case btw) but new wash trades will still be inserted every hour. 

>can we start to keep a record somewhere of the exposed models (used in other sub-projects) of all the sub-projects, as this is effectively a break in our lineage and we'll need to find a way to correctly propagate full refreshes across these breaks.

This should be possible by looking at the DBT sources for each project (with a bit of filtering).

PR #6032, opened 2024-05-29
eas optimism: attempt to avoid duplicates

jeff-dude:
### Contribution type
Please check the type of contribution this pull request is for:

- [ ] New spell(s)
- [ ] Adding to existing spell lineage
- [x] Bug fix

### For bug fixes
If you are fixing a bug, please provide the following information:

- **Description:** second time in three weeks where this model fails repeatedly in prod on merge duplicates coming from the source
- **Steps to reproduce:** 
```
with
  src as (
    with
      eas as (
        with
          src_SchemaRegistry_evt_Registered as (
            select
              *
            from
              "delta_prod"."attestationstation_v1_optimism"."SchemaRegistry_evt_Registered"
          ),
          src_SchemaRegistry_call_register as (
            select
              *
            from
              "delta_prod"."attestationstation_v1_optimism"."SchemaRegistry_call_register"
          )
        select
          'optimism' as blockchain,
          'eas' as project,
          '1' as version,
          er.uid as schema_uid,
          er.registerer,
          cr.resolver,
          cr.schema,
          transform(split(cr.schema, ','), x -> split(trim(x), ' ')) as schema_array, -- array of 2-element arrays [[data_type,field_name],[...]]
          cr.revocable as is_revocable,
          er.contract_address,
          er.evt_block_number as block_number,
          er.evt_block_time as block_time,
          er.evt_tx_hash as tx_hash,
          er.evt_index
        from
          src_SchemaRegistry_evt_Registered er
          join src_SchemaRegistry_call_register cr on er.evt_tx_hash = cr.call_tx_hash
        where
          cr.call_success
      )
    select
      *
    from
      eas
    where
      schema_uid not in (
        0x026b0f2bfcb9d0e91cb3c2f6c8e84872615c6d5028edec1c9f360a54ae6595d8,
        0x655a5addcf762e79f17ed35afc9e985a9912d57bd02835d529154954dd07a03d,
        0x5d2ec81b1de9d7919b34b41de87073a6f97c914b6c143df0aa15d84d5c5ba391,
        0xed7c89d83e631fc93115ac5cb92260b859dc70ff3ba8d686cd5e5b54800d0137,
        0x11bd7f94b4ce942bd11d1e02e94a3ab386993845211dc937fd0076e38f0c084e,
        0x45ca2f188d07f561fa0c4c83ffe204a86780c59cd39b1dcefcea6e64540adde4
      )
  )
select
  schema_uid,
  count(1)
from
  src
group by
  schema_uid
having
[...]

PR #6042, opened 2024-05-31
EAS schemas - bug fix

tomfutago:
### Contribution type

- [ ] New spell(s)
- [x] Bug fix

---

- **Description:** current logic assumes only 1 EAS schema can be created per tx, which optimism proves it wrong
- **Steps to reproduce:** [sample tx](https://optimistic.etherscan.io/tx/0xccf2e1b79be8b005fa5f6cd531e4aa0895a715b5a7e65003376a599c366c0bab) - 6 schemas created here
- **Implementation details:** fixed join in the query to ensure `evt` joins to only 1 corresponding `call`
- **Test instructions:** if the logic is still wrong, CI run will fail on uniqueness check
- **Related issue(s):** #6032 

---

### DIFF
--- a/macros/models/_sector/attestation/eas.sql
+++ b/macros/models/_sector/attestation/eas.sql
@@ -44,7 +44,7 @@ select
   er.evt_tx_hash as tx_hash,
   er.evt_index
 from src_SchemaRegistry_evt_Registered er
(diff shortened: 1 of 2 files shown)


jeff-dude (approved):
thanks for digging in and fixing 🙏

PR #6048, opened 2024-06-01
Added a few strategies to Yield Yak models

yy-analytics:
Just adding a few strategies to existing Yield Yak models. PR consists of adding a few constants to the hardcoded strategies model and updating sources appropriately.


Hosuke (approved):
LGTM.✅
Thank you @yy-analytics

jeff-dude (approved):

PR #6050, opened 2024-06-02
add token2022 into token transfers and fungible spells

andrewhong5297:
### Contribution type

- [ ] New spell(s)
- [ ] Bug fix

---

Token2022 is a new main token program on Solana that's starting to gain adoption (most recently from paypal with PyUSD). I'm writing up a longer guide on differences and nuances, but it requires me to update the fungible table and transfers table logic. 

Token2022 is already included in the tokens_solana.nft spell because that one is not token program dependent ([see this query](https://dune.com/queries/3791470)). We're currently missing the libreplex 2022 standard, metaplex core, and mpl-404 nft standards still.


andrewhong5297:
okay there is currently a 90 day window filter that it runs in, we'll need to boost the cluster for this to full refresh.

models/tokens/solana/tokens_solana_transfers.sql
```diff
@@ -14,56 +14,235 @@
...
-    , call_outer_executing_account as outer_executing_account
-FROM (  
-      SELECT account_source, account_destination, amount, call_tx_id, call_block_time, call_block_slot, call_outer_executing_account, call_tx_signer, 'transfer' as action, call_outer_instruction_index, call_inner_instruction_index
+with 
+token2022_fee_state as (
+      --we need the fee basis points and maximum fee for token2022 transfers because the fee amount is not emitted in transferChecked
+      SELECT 
+      call_account_arguments[1] as account_mint
+      , try(bytearray_to_uint256(bytearray_reverse(bytearray_substring(call_data,
+                  1+1+1+1+1+case when bytearray_substring(call_data,1+1+1,1) = 0x01 and bytearray_substring(call_data,1+1+1+32+1,1) = 0x01
```
aalan3, line 23: this looks painful 🙈

aalan3: could this be turned into a UDF?

andrewhong5297: Haha maybe one day when we have a solana udf. Nothing that can be done for now

models/tokens/solana/tokens_solana_transfers.sql
```diff
@@ -14,56 +14,235 @@
...
+    , COALESCE(call_inner_instruction_index,0) as inner_instruction_index
+    , call_outer_executing_account as outer_executing_account
+FROM base tr
 --get token and accounts
 LEFT JOIN {{ ref('solana_utils_token_accounts') }} tk_s ON tk_s.address = tr.account_source 
 LEFT JOIN {{ ref('solana_utils_token_accounts') }} tk_d ON tk_d.address = tr.account_destination
 WHERE 1=1
 {% if is_incremental() %}
 AND {{incremental_predicate('call_block_time')}}
 {% endif %}
```
jeff-dude, line 247: are we able to move this incremental filter up into the CTEs instead, so it filters on each source read directly, rather than altogether at the end? that has helped on other spells

jeff-dude (changes requested):

models/tokens/solana/tokens_solana_transfers.sql
andrewhong5297: yes, I've moved it up now

jeff-dude (approved):

PR #6051, opened 2024-06-02
remake bridges.flows

hildobby:
gm, after 123 commits this PR is now ready :)

here is are the exact specs:

PixelSnap 2025-07-14 at 18 29 40@2x
PixelSnap 2025-07-14 at 18 32 08@2x

some design decisions:
- made it as easy as can be to add new bridges
- only materialised on the main deposits/withdrawals spells
- bridges.flows is currently a view since deposits/withdrawals may be added later, i might materialise eventually since this will be the main queried spell
- bridges.flows is a FULL OUTER JOIN of deposits and withdrawals so that even if one is missing, all info available is shown
- deposits/withdrawals have a few more columns than the flows one to keep the flows one as light as can be
- tried to make all column names super explicit (and detailed them in the yml files) but lmk if there are any questions on what those are


hildobby:
I'm really confused here, how can uniqueness tests fail for `bridge.flows_beta` when the spell simply calls another spell (`bridge_ethereum.flows`) which itself has been tested and has no dupes. any clues @jeff-dude @aalan3 @0xRobin?

0xRobin:
@hildobby if duplicates show up in the incremental run, but not in the first run of the CI. It has often been a problem with null values being present in one of the merge columns.

jeff-dude:

hildobby:
Once my contract name change from `ovm` to `base` is approved, this pr should be close to ready

jeff-dude:
this won't run in CI until it's migrated over to new subproject structure. can this be moved to daily?

0xBoxer:

jeff-dude:
going through another cleanup effort on PRs. if reopen is necessary, plz do one at a time until finalized and merged 🙏

0xBoxer:
@hildobby can you write a paragraph on what you are trying to build here, not really following the initialized, finalised schema.

hildobby:
> @hildobby can you write a paragraph on what you are trying to build here, not really following the initialized, finalised schema. 

Yeah i'll rewrite a detailed PR description once this is ready, which it almost is.

In short:

The initiated is the send tx on original chain and finalised is the receive tx on the destination chain.

Benefits are:
- when both chains are on dune: the final bridge.flows can have bridge time metric + total bridge fees data.
- when only 1 chain is on dune: deeper coverage, tracking sends+receives gets us a higher chance of catching bridge events since only 1/2 chains need data coverage. that's not relevant for base's canonical bridge but will be for global bridges like laterzero and across.

hildobby:
phew 103 commits of iteration so far ☝️

but i'm almost done!

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_deposits.sql
```diff
@@ -0,0 +1,75 @@
...
+        , deposit_amount_raw
+        , sender
+        , recipient
+        , deposit_token_standard
+        , deposit_token_address
+        , tx_from
+        , tx_hash
+        , evt_index
+        , contract_address
+        , bridge_id
```
0xBoxer: bridge_id is very ambigious, `bridge_tx_id` or `bridge_transfer_id` would be more descriptive

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_flows.sql
```diff
@@ -0,0 +1,34 @@
...
+, d.deposit_token_standard
+, w.withdrawal_token_standard
+, d.deposit_token_address
+, w.withdrawal_token_address
+, d.tx_from AS deposit_tx_from
+, d.tx_hash AS deposit_tx_hash
+, w.tx_hash AS withdraw_tx_hash
+, bridge_id
+FROM {{ ref('bridges_deposits') }} d
+FULL OUTER JOIN {{ ref('bridges_withdrawals') }} w USING (bridge_name, bridge_version, deposit_chain, withdrawal_chain, bridge_id)
```
0xBoxer: probably helps to add `block_date` to speed up the join?

on second thought probably did not use that by intention because that breaks the condition in extreme cases?

0xBoxer: do you strictly need all these conditions, wouldn't just id be sufficient?

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_flows.sql
```diff
@@ -0,0 +1,34 @@
+{{ config(
+    schema = 'bridges',
```
0xBoxer: all the schema here should be `bridges_evms.flows` etc., we will probably want to have a truely multichain bridges_crosschain.flows` eventually

0xBoxer (approved):

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_flows.sql
hildobby: no unfortunately joining on block_date would mean it's not catching all events, especially since we have intent based bridges where a user can execute withdrawal whenever he wants

yes, those ids are bridge+version specific. and every chain where there are deposits has its ids. so these need to be all joined on

ids are very often just the event nonce used in deposit events as their bridge_transfer_id. but some use other stuff like leaf ids, hashes or other data types so i converted the value to a string for future compatibility with every type of bridge_transfer_id values

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_deposits.sql
hildobby: i changed all references to `bridge_transfer_id`

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_flows.sql
hildobby: but what would that mean for ethereum>solana bridge transfers? they would be bound to be incomplete in both `bridges_evms.flows` and `bridges_solana.flows`, so i think it makes sense for `bridges_evms.deposits` and `bridges_evms.withdrawals` but probably not for the final `bridges.flows` model. wdyt?

0xBoxer: yeah that makes sense to me, but that would mean that we should set this up as  `bridges_crosschain.flows` and not use any varbinary columns to start with IMO

hildobby: ah if you want output to be strings then `bridges_evms.flows` with varbinary values does make sense. this would really reduce compute costs when joining on other evm tables. and i guess other vms may have their specificities (datatypes and occasional different columns) so you're right that it's better to have a table for each vm type and have it all joined into a crosschain table

also, wouldn't `bridges_crosschain.flows` just be `bridges.flows`?

hildobby: i just moved it all into an `evms` subfolder and renamed `bridges.deposits/withdrawals/flows` to `bridges_evms.deposits/withdrawals/flows` and created a view for each on top, it might make sense to materialise those crosschain tables once bridge events from other vms are added

hopefully this is is what you had in mind, lmk if not

0xBoxer: > also, wouldn't bridges_crosschain.flows just be bridges.flows?

in a perfect world that's how things would work, but now that we have the very confusing `dex.trades `and `tokens.transfers`  which are EVM specific, it kinda would make things even more confusing to have a truly multichain `bridges.flows`

0xBoxer: I suggest `bridges.flows` just stay vacant, for now and ever, the tablename should specify what it contains

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_flows.sql
0xBoxer: fair enough, took a look myself and seems reasonable to match on these

0xBoxer:
let me do another pass tmrw, looks pretty good already

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_flows.sql
hildobby: alright, done

hildobby:
i changed `bridges.` to `bridges_crosschain.` and updated the two slides in the PR description @0xBoxer

0xBoxer:
This is good to go from my POV, fantastic job @hildobby. Thank you!

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/bridges_crosschain_deposits.sql
```diff
@@ -0,0 +1,38 @@
+{{ config(
+    schema = 'bridges_crosschain'
+    , alias = 'initiated'
```
jeff-dude:
```suggestion
    , alias = 'deposits'
```

jeff-dude: assuming this should be deposits, similar to file name and the withdrawals equivalent model

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/evms/bridges_evms_deposits.sql
```diff
@@ -0,0 +1,74 @@
+{{ config(
+    schema = 'bridges_evms'
+    , alias = 'initiated'
```
jeff-dude:
```suggestion
    , alias = 'deposits'
```

jeff-dude: assuming this should be deposits, similar to file name and the withdrawals equivalent model

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/chains/base/bridges_base_deposits.sql
```diff
@@ -0,0 +1,40 @@
...
+
+{{ config(
+    schema = 'bridges_' + blockchain,
+    alias = 'deposits',
+    materialized = 'view'
+    )
+}}
+
+{% set bridges_platforms = [
+    'bridges_base_cctp_v1_deposits'
```
jeff-dude: not a must at all, but while in here, curious if you'd like to add the `blockchain` variable usage in these ref lists to futureproof typos of other chains when adding in the future

jeff-dude: applies to all of these base/ethereum level models, will only comment once

jeff-dude (changes requested):
minor feedback below

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/chains/base/bridges_base_deposits.sql
hildobby: i'm not sure what you mean here?

jeff-dude: `'bridges_' + blockchain + '_cctp_v1_deposits'`
this type of format, like you do on schema names and such

jeff-dude: to be clear, not 100% sure this is correct syntax, but hopefully gives the idea

hildobby: ah alright makes sense, just did that on all 4 files and pushed the PR

PR #6052, opened 2024-06-02
Rework `addresses_events.first_funded_by` + make a crosschain version

hildobby:
I reworked first_funded_by spells so they're a lot more efficient and simpler to read the code. And I also added a crosschain version to know the chain on which an address first appeared (+ details of who funded it)

pls lmk when this is merged so i can make a new pr to move it to the hourly spellbook like we discussed @jeff-dude


hildobby:
It tells me that `tokens_arbitrum_transfers` doesn't exist, but it does, why is that? is `ref('tokens_' + blockchain + '_transfers')` incorrect to use here @jeff-dude ?

hildobby:
> Is something wrong with zksync data? maybe missing some? I've got it causing CI test failures across multiple PRs

@jeff-dude @aalan3 @0xRobin

jeff-dude:
> > Is something wrong with zksync data? maybe missing some? I've got it causing CI test failures across multiple PRs
> 
> @jeff-dude @aalan3 @0xRobin

this looks like a failure due to the seed test. i believe this seed hadn't been created in prod, so when it's not included in the PR to build via CI, it can't find it. can you modify the seed file that these schema files call and pull into this PR to rebuild?

if it still fails, you may have missing and/or incorrect data in the seed file itself to test against -- or the query itself has a bug and is outputting incorrect results vs. what is expected in seed file

hildobby:
@jeff-dude lmk when you can take a look so i can make a follow-up pr when this is merged to move `first_funded_by` spells into the hourly spellbook section (which might be a decently big PR given its heavy downstream usage)

jeff-dude:
thanks for the patience. let's go ahead and move them to hourly project in this PR, should be as simple as moving all of them to proper directory. i ran a list command to see if any other models are in the lineage, but looks like no:
```
addresses_events_arbitrum_first_funded_by
addresses_events_avalanche_c_first_funded_by
addresses_events_base_first_funded_by
addresses_events_blast_first_funded_by
addresses_events_bnb_first_funded_by
addresses_events_celo_first_funded_by
addresses_events_ethereum_first_funded_by
addresses_events_fantom_first_funded_by
addresses_events_first_funded_by
addresses_events_gnosis_first_funded_by
addresses_events_linea_first_funded_by
addresses_events_mantle_first_funded_by
addresses_events_optimism_first_funded_by
addresses_events_polygon_first_funded_by
addresses_events_scroll_first_funded_by
addresses_events_zkevm_first_funded_by
addresses_events_zksync_first_funded_by
addresses_events_zora_first_funded_by
```

hildobby:
1. @jeff-dude any clues what's causing the errors? Is it just the amount of data that's too high to compute here? It seems to only happen on the biggest models (solana, and ethereum)

2. Also, how can I see which spells are affected downstream? Are there any spells where i need to change the referenced table (changing it from ref to source of vice versa because of the daily > hourly migration)

jeff-dude:
> 1. @jeff-dude any clues what's causing the errors? They've been persistant on the last few reruns and I can't seem to have a run without them, is it just the amount of data that's too high to compute here? It seems to only happen on the biggest models (solana, and ethereum)
> 2. Also, how can I see which spells are affected downstream? Are there any spells where i need to change the referenced table (changing it from ref to source of vice versa because of the daily > hourly migration)

looks like you are past the cluster crash errors with time filters.

as for affected spells downstream, you can use `dbt ls --resource-type model --output name --select +addresses_events_first_funded_by+` to find upstream and downstream dependencies, if a `ref()` is used. if it's a `source()` used, you may need to use vs code project search functionality to find via source name 🤔

hildobby:
> > 1. @jeff-dude any clues what's causing the errors? They've been persistant on the last few reruns and I can't seem to have a run without them, is it just the amount of data that's too high to compute here? It seems to only happen on the biggest models (solana, and ethereum)
> > 2. Also, how can I see which spells are affected downstream? Are there any spells where i need to change the referenced table (changing it from ref to source of vice versa because of the daily > hourly migration)
> 
> looks like you are past the cluster crash errors with time filters.
> 
> as for affected spells downstream, you can use `dbt ls --resource-type model --output name --select +addresses_events_first_funded_by+` to find upstream and downstream dependencies, if a `ref()` is used. if it's a `source()` used, you may need to use vs code project search functionality to find via source name 🤔

@jeff-dude wash trades tables seem to be the only downstream spells, should I change those from sources to refs in this same PR?

jeff-dude:
> @jeff-dude wash trades tables seem to be the only downstream spells, should I change those from sources to refs in this same PR?

if wash trades is in the same subproject as this, then yes, that is fine to be a ref

hildobby:
> > @jeff-dude wash trades tables seem to be the only downstream spells, should I change those from sources to refs in this same PR?
> 
> if wash trades is in the same subproject as this, then yes, that is fine to be a ref

nvm, wash trade spells are actually part of the the `nft` subproject so I left them as sources in the end.


Otherwise, the spell seems to work fine so I removed the block_time filters. CI tests failing is normal given the block_time filter which means many addresses showed up before but the block_time filter isn't letting it see that, hence the failure. It looks ready to be merged to me.

jeff-dude:
pushed a commit to clean up the files a bit more & remove duplicate tests. since table are unique on one field, you already had a unique test on the column directly and don't need the multi-column test.

my last questions is the setup of final two spells:
- union all chains
- aggregate union of all chains

does the first union spell have a use case outside of feeding the final aggregate spell? if not, can we consolidate into one spell? we could wrap the union into a CTE, then write the aggregate on top of the CTE.

if they do need to remain separate for various use cases, can we maybe update table naming standards? using `crosschain` in schema name hasn't been used before, typically without the specific chain called out in schema name it's known to be crosschain

hildobby:
> pushed a commit to clean up the files a bit more & remove duplicate tests. since table are unique on one field, you already had a unique test on the column directly and don't need the multi-column test.
> 
> my last questions is the setup of final two spells:
> 
> * union all chains
> * aggregate union of all chains
> 
> does the first union spell have a use case outside of feeding the final aggregate spell? if not, can we consolidate into one spell? we could wrap the union into a CTE, then write the aggregate on top of the CTE.
> 
> if they do need to remain separate for various use cases, can we maybe update table naming standards? using `crosschain` in schema name hasn't been used before, typically without the specific chain called out in schema name it's known to be crosschain
> 
> edit: just deprecated goerli too, as it was left behind. feel free to add as needed.

The only reason I kept the union of all chains is because it was already there and removing it might break some queries, happy to remove it if found unused but unsure I can check myself if that's the case. We can definitely remove the union if unused

dbt_subprojects/hourly_spellbook/models/_sector/addresses_events/addresses_events_first_funded_by.sql
jeff-dude, line 1: rather than remove this union spell, can we just move the logic from `dbt_subprojects/hourly_spellbook/models/_sector/addresses_events/addresses_events_crosschain_first_funded_by.sql` into this spell directly? it's the same columns, just one new one introduced, then aggregate it all

jeff-dude: then can remove the need for new spell there

hildobby: alright, just did that @jeff-dude 👌

jeff-dude (approved):

jeff-dude:
this is going to be a heavy one in prod, with combo of the large EVMs + solana + bitcoin. i'll have to plan this carefully. i will likely look to do monday morning 🙏

jeff-dude:
going to give this a shot now 🙏

PR #6054, opened 2024-06-03
Incremental predicate chainlink

aalan3:
This changes the chainlink models to use the `incremental_predicate` macro and adds a incremental predicate strategy (`incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_month')]`. I didn't add the strategy everywhere because there are so many models but these should be the most heavy models. 

I used the following regex for search and replacing for the macro `(\w+\.\w+|\w+) >= date_trunc\(.*?\).*?\)`. This can be used for other models as well.


jeff-dude (approved):

jeff-dude:
this one is in daily spellbook project, so i will disable modified there too & test the new trigger cluster spinup by running the job manually until it's up

PR #6059, opened 2024-06-03
sources: one file per subproject

jeff-dude:
fyi @aalan3 
a little cleanup on sources, to match how tokens subproject is built. also finding it easier when modifying some of this.


aalan3:
lgtm this is actually what I do while developing

PR #6005, opened 2024-05-28
Organize Balancer Spells into folders

viniabussafi:
With the upcoming V3  basically doubling the number of spells for Balancer, this PR aims to organize our folders into sectors, making blockchain-specific folders easier to navigate. 
There are no changes to the spells themselves, except for removing legacy trade spells. 
All schema files have been renamed with "_schema" and improved for syntax, formatting, and column descriptions. 
Outputs have been validated and remain unchanged, as expected.

PR #6075, opened 2024-06-04
Revert "Revert "Organize Balancer Spells into folders (#6005)" (#6074)"

jeff-dude:
fyi @viniabussafi 
part 2 (and we pray 🙏 )


jeff-dude:
this did not work 🤦 

i tried running the whole `models/_project/balancer/` directory, but it fails:
```
19:23:32    Database Error in model balancer_v2_ethereum_lbps (models/_project/balancer/pools/ethereum/balancer_v2_ethereum_lbps.sql)
  TrinoUserError(type=USER_ERROR, name=FUNCTION_NOT_FOUND, message="line 35:13: Unexpected parameters (varbinary, integer, integer) for function substring. Expected: substring(char(x), bigint), substring(char(x), bigint, bigint), substring(varchar(x), bigint), substring(varchar(x), bigint, bigint)", query_id=20240604_192317_75643_268bq)
```

but it looks like this model wasn't changed, just moved directories. i'm really confused.

i'm not sure what i can do tbh. i might have to fully revert, let the modified models in prod run the old code, then see if we can add back again. this assumes the revert will help successfully rerun the old code.

PR #6082, opened 2024-06-05
Materialize dex raw pools

aalan3:
This spell is used as input for a lot of the oneinch spells. We should materialize it to improve performance of the downstream spells.

### DIFF
--- a/models/dex/dex_raw_pools.sql
+++ b/models/dex/dex_raw_pools.sql
@@ -1,8 +1,10 @@
 {{ config(
         schema='dex',
         alias = 'raw_pools',
-        materialized = 'view',
+        materialized = 'incremental',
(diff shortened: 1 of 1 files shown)


models/dex/dex_raw_pools.sql
```diff
@@ -1,8 +1,10 @@
 {{ config(
         schema='dex',
         alias = 'raw_pools',
-        materialized = 'view',
+        materialized = 'incremental',
         file_format = 'delta',
+        incremental_strategy = 'merge',
+        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.creation_block_time')],
```
aalan3, line 7: @grkhr just want to double check that this column works.

PR #6084, opened 2024-06-05
Add incremental predicates on cow 

aalan3:
More performance fixes

### DIFF
--- a/models/cow_protocol/ethereum/cow_protocol_ethereum_batches.sql
+++ b/models/cow_protocol/ethereum/cow_protocol_ethereum_batches.sql
@@ -57,6 +57,9 @@ batch_values as (
     from {{ ref('cow_protocol_ethereum_trades') }}
         left outer join {{ source('prices', 'usd') }} as p
             on p.contract_address = 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
+            {% if is_incremental() %}
+            and {{ incremental_predicate('minute') }}
+            {% endif %}
             and p.minute = date_trunc('minute', block_time)
             and blockchain = 'ethereum'
     {% if is_incremental() %}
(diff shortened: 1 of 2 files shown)


models/_sector/dex/bot_trades/solana/platforms/consortium_key_solana_bot_trades.sql
```diff
@@ -27,8 +27,14 @@ with
...
-            activity.block_time >= timestamp '{{project_start_date}}'  -- FirstTradeTimestamp
-            and transactions.block_time >= timestamp '{{project_start_date}}'  -- FirstTradeTimestamp
+            {% if is_incremental() %}
+                {{ incremental_predicate("activity.block_time") }}
+                and {{ incremental_predicate("transactions.block_time") }}
+            {% else %}
+                activity.block_time >= timestamp '{{project_start_date}}'  -- FirstTradeTimestamp
+                and transactions.block_time >= timestamp '{{project_start_date}}'  -- FirstTradeTimestamp
+            {% endif %}
+
```
jeff-dude, line 37: this is one that i tried to add in PR phase, but the use case needs to see full history. it is slow, but we couldn't find a better way around it. ultimately the output is a distinct list of users, but filtering outputs missing data.

i wonder if another option is to query the spell itself with `{{ this }}` and union in the source data with the incremental flag, so that at least getting history of users is faster on the spell

jeff-dude: here is thread with thoughts:
https://github.com/duneanalytics/spellbook/pull/5880#discussion_r1610273106

jeff-dude (changes requested):

models/_sector/dex/bot_trades/solana/platforms/consortium_key_solana_bot_trades.sql
aalan3: Good catch, I think I'll just remove it for now but we need some sort of filter here otherwise this will stop working at some point. We can come back to this when we look closer at solana models.

PR #6086, opened 2024-06-05
Zeroex native_fills performance improvements

aalan3:
Added filters on prices as this seems to be taking a lot of time. 

Edit: I dropped the api_fills related changes. https://github.com/duneanalytics/spellbook/actions/runs/9382039480/job/25832612176?pr=6086 shows that native_fills are working


jeff-dude (approved):
nice catch, this is def something we try to catch up front lately

PR #6089, opened 2024-06-05
Zeroex api fills performance

aalan3:
Didn't find much to change here but left a comment as well on the bnb_fills where it's doing a full scan on logs which is most of runtime for that query. I'm not sure if there's a decoded table we can use for that instead

### DIFF
--- a/models/zeroex/bnb/zeroex_bnb_api_fills.sql
+++ b/models/zeroex/bnb/zeroex_bnb_api_fills.sql
@@ -347,6 +347,7 @@ uni_v2_swap as (
 
 
 )
+-- TODO: this is really slow, need to optimize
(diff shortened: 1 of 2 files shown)


models/zeroex/bnb/zeroex_bnb_api_fills.sql
```diff
@@ -347,6 +347,7 @@ uni_v2_swap as (
 
 
 )
+-- TODO: this is really slow, need to optimize
```
jeff-dude, line 350: this will trigger full refresh, but the plan for this PR will be to turn off modified anyway, so we can let it be

jeff-dude (approved):

jeff-dude:
merging now with modified off

PR #6094, opened 2024-06-05
fix cex flows column

jeff-dude:
fyi @hildobby 
failing without quotes


hildobby:
woops, thx for the quick fix @jeff-dude !

PR #6097, opened 2024-06-06
Don't materialize rolling celo transfers

aalan3:
Materializing these takes a very long time and has almost no usage. Will keep these around until there's balances for Celo but after that these can be dropped.


aalan3:
@jeff-dude do we need to manually drop these tables first before deploying?

jeff-dude:
> @jeff-dude do we need to manually drop these tables first before deploying?

the bug may have only been from full table --> view, but not 100% sure. may be worth merging solo and monitoring

PR #6108, opened 2024-06-06
keep as base58 (opensea and magiceden solana)

andrewhong5297:
### Contribution type

- [ ] New spell(s)
- [x ] Bug fix

Not sure when we unioned in the old solana models for nft trades, but the old models were in varbinary for addresses/hashes. I've removed that so they are in base58, consistent with other solana models. 

cc @0xRobin 

I'm guessing that when this runs we are likely to see duplicate trades, since before it would not have caught it since the hashes were in different data types.


andrewhong5297:
Need a bot on github that tags all contributors of a model being edited in spellbook lol

0xRobin (approved):
changes look good. 
The models were moved in a recent NFT cleanup, preparing for the NFT subproject.
Looking at the volume, these did not overlap with existing solana models so we should not see duplicates I think. 
(the 2 old models mainly contain historical data)

marking as blocked for now as we'll need to adapt this PR when https://github.com/duneanalytics/spellbook/pull/6090 is merged

andrewhong5297:
> changes look good. The models were moved in a recent NFT cleanup, preparing for the NFT subproject. Looking at the volume, these did not overlap with existing solana models so we should not see duplicates I think. (the 2 old models mainly contain historical data)
> 
> marking as blocked for now as we'll need to adapt this PR when #6090 is merged
> 
> cc: @jeff-dude, not sure how we proceed when the CI is unable to run these solana models (changes here are small and fine to manually approve, but not sure for larger changes)

maybe we split it into two PRs to make the merge easier

jeff-dude:
a bit confused on this one. are we removing the solana spells from main `nft.trades`? is that how we get away with changing the data types in these?

andrewhong5297:
Yeah they got moved which is how i ran into the error on types in one of my queries

jeff-dude:
> Yeah they got moved which is how i ran into the error on types in one of my queries

okay, i see. maybe we wait until we get the prod env prepared for the other solana spells before we try to push this one and get it to run, so we know the correct sizing to use

andrewhong5297:
Makes sense. Not in a rush to fix this one.

jeff-dude:
@0xRobin let me know if there are next steps here, otherwise plz feel free to close 🙏

0xRobin:
closing since now solana trades stand on their own and a proper refactor of these 2 old models is needed to get them on the solana schema.

andrewhong5297:
reopening, let's just remove the legacy models since it's causing duplicate data

0xRobin:
@andrewhong5297 I don't think we can safely remove those models at the moment as we'd be losing all historical volume for 2022.
Can you expand on the duplicates, how big is the impact? It's not directly visible clear in the aggregated data as both magiceden models diverge quite a bit.
Could we filter the duplicates out or expand the new models to also cover 2022 data?
looks like it's mainly magiceden v2 that's covered in both models? 


image
(old models in black)

0xRobin:
closing in favor of: https://github.com/duneanalytics/spellbook/pull/6519

andrewhong5297:
The impact is we are overstating volumes by 5x because of the v2 model. Its likely because they started supporting trades in more tokens than just SOL, and I made that adjustment on the new models but never touched the old ones.

PR #6110, opened 2024-06-06
add account_type column to token_accounts solana

andrewhong5297:
### Contribution type

- [ ] New spell(s)
- [ ] Bug fix

@aalan3 mentioned it would be good if added an account type column to token_accounts, so we can filter on the heavy JOINs. I'm adding it here.

### DIFF
--- a/solana/models/solana_utils/solana_utils_schema.yml
+++ b/solana/models/solana_utils/solana_utils_schema.yml
@@ -14,6 +14,8 @@ models:
         name: token_mint_address
       - &address
         name: address
+      - &account_type
+        name: account_type
       - &updated_at
         name: updated_at
   
--- a/solana/models/solana_utils/solana_utils_token_accounts.sql
+++ b/solana/models/solana_utils/solana_utils_token_accounts.sql
@@ -25,7 +25,6 @@ WITH
             {% endif %}
             group by 1
       )
-
 SELECT
 *
 , case when nft.account_mint is not null then 'nft'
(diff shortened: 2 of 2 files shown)


models/solana_utils/solana_utils_token_accounts.sql
```diff
@@ -26,4 +26,11 @@ WITH
             group by 1
       )
       
-SELECT * FROM distinct_accounts
\ No newline at end of file
+SELECT 
+* 
+, case when nft.account_mint is not null then 'nft' 
+      when token_mint_address is not null then 'fungible'
+      else 'basic'
```
aalan3, line 33: this would never happen right? Because above on line 22 we do `WHERE aa.token_mint_address is not null`. so it will always either be `fungible` or `nft`?

andrewhong5297: Mmm good point, forgot I did that

andrewhong5297:
@aalan3 @jeff-dude this should be merged next, so we can add a nft filter out for all the dex trades spells. this will shrink the join by a ton.

jeff-dude:
> @aalan3 @jeff-dude this should be merged next, so we can add a nft filter out for all the dex trades spells. this will shrink the join by a ton.

noted. we can likely work on this soon as we wrap up some of the other subprojects early this week. we'll also be having a discussion tomorrow on where solana spells live.

aalan3 (approved):
Should be merged after https://github.com/duneanalytics/spellbook/pull/6197. This might trigger a bunch of full refreshes btw but we can skip those since we are just adding a column here.

PR #6118, opened 2024-06-09
Yield strategies models and Mantle expansion (including tokens.erc20 for Mantle)

yy-analytics:
### Contribution type

- [x] New spell(s)
- [ ] Bug fix

---

### For new spell(s)

- **Spell name(s):** *yield_yak.yield_strategies* (across avalanche_c, arbitrum and mantle)
- **Description:** These spells act as a stop gap and should be replacing [this PR](https://github.com/duneanalytics/spellbook/pull/5875) until such a time when LiveFetch functions might be able to be used in spellbook. They are views which retrieve all Yield Yak strategy contracts, along with key information, and will be used with LiveFetch functions in a query to update a materialized view in Dune SQL.
- **Who are the new spell(s) for?** Internal team
- **How will the new spell(s) be used downstream?** These will feed into a materialized view and the main intention is cleaning up code in these views so that we don't have overly long queries on the Dune SQL side and can re-use parts. This will then be used in queries to show performance of various yield strategies over time, with the basic labelling info and necessary static info (deposit token address etc.) to get prices coming from the materialized view that depends on this spell (hope that makes sense).
- **Implementation details:** It's a view which takes information from the decoded `contracts` tables, and adds some info from the `creation_traces` table to give a list of yield strategies.
- **Test instructions:** Uniqueness of `contract_address` column (tests in schema.yml files)
- **Related issue(s):** None

---

- **Description:**
  - Addition of Mantle network means we can now include all Yield Yak strategies on the Mantle network (so lots of copy/paste/replace for those).
  - Refactoring the `yield_yak_trades` models so that they make use of a common macro instead of repeating code.
  - Creating a manual addition of erc20 tokens for Mantle and adding a few tokens which were missing.

---

### Additional information

- I wasn't completely sure what counts as "new spell" versus adding to existing. The only new spell which is "exposed" in the UI is listed in the "new spells" section but there are lots of new models created here where it's repeating an existing spell for a new blockchain.
- The various pieces of work are split into the different commits in the PR, with the final few commits being a few corrections after testing locally with `dbt compile` and some corrections after initially raising the PR.
- Any questions or anything I can explain further please let me know.

---


tokens/models/tokens/mantle/tokens_mantle_erc20.sql
Hosuke, line 1: Thank you for initiating `tokens_mantle.erc20`.

PR #6140, opened 2024-06-11
Try adding doc blocks

0xBoxer:
showcasing dbt docs blocks 

Dbt documentation on this: https://docs.getdbt.com/docs/collaborate/documentation#using-docs-blocks

to edit content in the docs blocks and get it to compile on local docs: 
- change md or .yml
- ``$dbt docs generate`` 
- ``$dbt docs serve --port 1234``


models/_sector/dex/trades/_schema.yml
```diff
@@ -8,6 +8,7 @@ models:
       contributors: 0xRob, hosuke, jeff-dude, tomfutago
```
philippWassibauer, line 8:
```suggestion
      contributors: 0xRob, hosuke, jeff-dude, tomfutago
      long_description: '{{ doc("dex_trades_doc") }}'
```

0xBoxer: this doesn't work, tried to compile this in these configs:
- `'{{ doc("dex_trades_doc") }}'`
- `{{ doc("dex_trades_doc") }}`
- "{{ doc("dex_trades_doc") }}"

ChatGPT also doesn't know of any workaround, we don't have access to the `doc` function within the `meta` context.

0xBoxer:
okay tried out the things we talked about:
- we can add arbitrary values to meta fields
- we cannot use jinja template within the meta fields
- this branch is adapted for having a short_description in the `meta` field
- for the data explorer you will need to adapt from reading `description:value` to read `meta:short_description:value`

philippWassibauer (approved):
This looks good to me. 

Remind me again what our upside is of having the long table description in spellbook? 
As we will only show them in docs.dune.com we might be overengineering this. 

I know this comes as a bit late feedback and I might have forgotten the reason for this.

0xBoxer:
@philippWassibauer your initial reasoning was to keep code and docs closer together, I actually thought that makes sense so happy to continue with this approach. 

makes it easier for us to get spellbook contributors to do docs for their creations if it's all in one place.

For sources it is indeed not that useful, but o well, better to have a standardized pattern everywhere imo

jeff-dude (approved):

0xRobin:
image
thankyouuuuu :raised_hands:

PR #6158, opened 2024-06-12
Prices drop normie

aalan3:
This has some price deviations that we would like to get rid of/


Hosuke (approved):

jeff-dude (approved):

jeff-dude:
confirmed with 1inch team, they are good to drop too

PR #6168, opened 2024-06-13
simplify `dbt compile` step in CI pipeline

jeff-dude:
thought process:
- initial commit, let the normal CI run to see runtimes
- remove the profile flag from compile step to force the dummy profile at root of spellbook to be used (similar to wizard flow)
- ideally no queries run on compile due to this setup
- goal: much faster compile in CI

### DIFF
--- a/.github/workflows/dbt_run.yml
+++ b/.github/workflows/dbt_run.yml
@@ -42,7 +42,7 @@ jobs:
         run: "./scripts/ensure_cluster.sh"
 
       - name: dbt compile to create manifest to compare to
-        run: "dbt --warn-error compile $PROFILE --project-dir ${PROJECT_DIR}"
+        run: "dbt --warn-error compile --project-dir ${PROJECT_DIR}"
 
       - name: check schemas
         run: |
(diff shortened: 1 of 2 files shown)


jeff-dude:
image

jeff-dude:
removing the dummy model and prepping for merge

jeff-dude:
tagging the team here for awareness. this seems to speed up `dbt compile` time in CI in new subproject setup. we were seeing 15+ minute compile times on all PRs. let's monitor this change post deploy to ensure no side effects.

summarizing thoughts on this change:
- the profile flag passed in our custom profile file on the GH runner (that has access credentials to db)
- when compile ran, it was throwing out a bunch of small queries to check schemas/tables exist
- without this profile flag, the set up mimics the wizard flow with no db connection, i.e. no queries
- compile step is solely to build a manifest file that downstream steps in CI can use for state comparison (`state:modified`). it appears to work fine this way.

aalan3 (approved):
very nice!

Hosuke (approved):

PR #6169, opened 2024-06-13
Fix Linea and Zksync for Rollup Economics mapping

lorenz234:
### Contribution type

- [ ] New spell(s)
- [x] Bug fix

---

### For new spell(s)

---

- **Description:** [Description of the changes made]

---

### For bug fixes

---

### Additional information

---


lorenz234:
Hi @jeff-dude,

Just updated some of the mapping to new addresses, though now getting:

TrinoQueryError(type=INSUFFICIENT_RESOURCES, name=CLUSTER_OUT_OF_MEMORY, message="Query killed because the cluster is out of memory. Please try again in a few minutes.", query_id=20240614_074019_00004_fii5e)

is it my query or something on dunes end?

Thanks

jeff-dude:
i see @0xRobin self-assigned this -- i'm wondering if CI failing is a sign we should split this spell up like we normally do per blockchain, then union together after each chain spell is built?

0xRobin:
yes @jeff-dude I agree, these queries could use some optimizations. 
Currently the `ethereum.transactions` and `prices.usd` table is joined in for every chain subquery, which is not optimal. 
I think we can reduce the queries to:
1. for each chain, get the `block_number` and `tx_hash` of all the relevant txs (this is different for each chain, some use events, some use a filter on transactions..)
2. union these all together
3. join the result with transactions and prices to extract the relevant fee information.

We could do (1) in chain specific models and then (2) and (3) in the final model, both should then be materialized I think.
thoughts on this design @lorenz234 @Jam516?

lorenz234:
> 1. for each chain, get the `block_number` and `tx_hash` of all the relevant txs (this is different for each chain, some use events, some use a filter on transactions..)

I totally agree, this spell is slowly growing into being unsustainable. I would even say we should fully switch to filter on transaction level rather than events. While events might not change as frequently we could then just create one consistent mapping, a long table of from_address, entity and function method. This would help to more quickly add new methods into the spell. 

e.g.
`
FROM (
    values
      (0xAddress, 'zkSync Era', 0xMethod)
      , (...) 
    AS x(address, entity, method)`

Jam516:
Yeah, I agree with those changes @0xRobin @lorenz234 . Makes sense at this point.

jeff-dude:
refactor is happening in #6426 -- plz contribute there if interested in the design / output.

PR #6194, opened 2024-06-17
Stablecoin transfers spell

thetroyharris:
### Contribution type

- [x] New spell(s)
- [ ] Bug fix

---

### For new spell(s)

New spell macro that isolates stablecoin transfers.
Community
Create dashboards that run quickly
Via macro

---

New spell macro that isolates stablecoin transfers.


thetroyharris:
@jeff-dude 

Once the stablecoins view is created I will be moving on to this PR, however I figured it would be good to chat with you and whomever else about this PR ahead of time.

Essentially, I am looking to create an aggregate table of all stablecoin transfers across all supported blockchains. I planned on creating a macro that I would use get this information for each chain. However, the information I would like to support in this table would be more then just a filtered transfers table. It would also include:
- transaction cost in gas tokens & dollars
- number of stablecoins present in the transaction
- the total amount transfered in the transaction by each token

That is where it stands at the moment, however I would also like to add some trace inforamtion at some point but I am not ready to do that atm.

What are your thoughts on how to best design this set of spells such that they are quick to query and easy to maintain?

macros/models/_sector/transfers/erc20_stablecoins/transfers_erc20_stablecoins.sql
```diff
@@ -0,0 +1,150 @@
...
+WITH
+    stables_in_tx AS (
+        SELECT 
+            evt_tx_hash
+            , COUNT(*) AS stables_in_tx 
+        FROM (
+            SELECT distinct 
+                t.contract_address
+                , t.evt_tx_hash 
+            FROM {{ source('erc20_' + blockchain, 'evt_transfer') }} t
```
jeff-dude, line 11: rather than read from source transfers again, can we take advantage of the new transfers spells in the `_sector/tokens/` directory?

macros/models/_sector/transfers/erc20_stablecoins/transfers_erc20_stablecoins.sql
```diff
@@ -0,0 +1,150 @@
...
+                CASE 
+                    WHEN symbol = 'ETH' THEN 'ethereum'
+                    WHEN symbol = 'MATIC' THEN 'polygon'
+                    WHEN symbol = 'XDAI' THEN 'gnosis'
+                    WHEN symbol = 'BNB' THEN 'bnb'
+                    WHEN symbol = 'AVAX' THEN 'avalanche_c'
+                    WHEN symbol = 'TRX' THEN 'tron'
+                    WHEN symbol = 'FTM' THEN 'fantom'
+                    WHEN symbol = 'CELO' THEN 'celo'
+                END AS chain
```
jeff-dude, line 134: if we're passing in a blockchain variable, we should be able to code this in a way to simplify this column logic rather than look at all chains?

jeff-dude (changes requested):

macros/models/_sector/transfers/erc20_stablecoins/transfers_erc20_stablecoins.sql
thetroyharris: Is it better for me to use the transfers_enrich macro for ethereum or just use the `tokens_ethereum.transfers` as the source?

macros/models/_sector/transfers/erc20_stablecoins/transfers_erc20_stablecoins.sql
thetroyharris: To my knowledge, the best price data for native tokens is where blockchain is `NULL` in the `prices.usd` table. So if I want to use that data I have to map them to a blockchain, which I thought was easiest by making the above case statement. 

What is the better solution to this problem?

macros/models/_sector/transfers/erc20_stablecoins/transfers_erc20_stablecoins.sql
thetroyharris: Tired using them and now get: "Error: Query exceeds cluster capacity (memory/pages), please optimize your query"

jeff-dude: it's better to read from them directly, either as ref or source, not reuse macro. they are heavy tables, especially on larger chains, so not too surprising to see cluster crash. it would be advised to start with one chain in the PR to keep it simple and give cluster more dedicated space to run

macros/models/_sector/transfers/erc20_stablecoins/transfers_erc20_stablecoins.sql
jeff-dude: you are passing in `blockchain` value through the model to macro, so the transfers data above is already focused on one chain. we should keep the prices data focused to one chain as well to match, rather than all. you can filter on this variable already used, then also pass in another new variable for symbol as needed

jeff-dude:

PR #6197, opened 2024-06-18
Solana subproject

aalan3:
Moving solana to its own sub project.

We will not be able to run this in CI fully.


0xRobin (approved):
I'll move the NFT solana models in here once it's set up

PR #6198, opened 2024-06-18
interim integration of paraswap v6.2: replace old v6.1 for now

alexshchur:
The intent of the PR is to update Paraswap v6 models to refer to the new v6.2 contract.

_It is planned for later to adjust the configuration to make the models include both v6.1 and v6.2._


jeff-dude:
looking at CI results prior to looking at code -- looks like a source table doesn't exist? are we still waiting on it to be created?

jeff-dude:
> looking at CI results prior to looking at code -- looks like a source table doesn't exist? are we still waiting on it to be created?

code looks good otherwise. just gotta look at these:
https://github.com/duneanalytics/spellbook/actions/runs/9561513986/job/26355896575?pr=6198#step:11:42

alexshchur:
@jeff-dude indeed contracts were added _after_ creating PR, which I now realize was an oversight.
After synching the branch CI tests look good.

Does that mean we're good to merge?

aalan3 (approved):

PR #6199, opened 2024-06-18
Make NFT models incremental

0xRobin:
These are currently the less performant NFT models in dbt cloud, making them incremental will save a couple more minutes runtime.

### DIFF
--- a/nft/models/_sector/aggregators/chains/nft_ethereum_aggregators_gem.sql
+++ b/nft/models/_sector/aggregators/chains/nft_ethereum_aggregators_gem.sql
@@ -1,14 +1,20 @@
 {{config(
-    
     schema = 'nft_ethereum',
-    alias = 'aggregators_gem'
-)}}
+    alias = 'aggregators_gem',
+    materialized = 'incremental',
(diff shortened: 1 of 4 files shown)


nft/models/_sector/aggregators/chains/nft_ethereum_aggregators_gem.sql
```diff
@@ -1,14 +1,21 @@
 {{config(
-    
     schema = 'nft_ethereum',
-    alias = 'aggregators_gem'
-)}}
+    alias = 'aggregators_gem',
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
```
aalan3, line 6: Can we also add `incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')],` here and below

aalan3: ah yeah you are right ofc. Saw that you added it where it's possible.

aalan3 (approved):

PR #6202, opened 2024-06-19
move NFT solana models

0xRobin:
this moves the NFT Solana models to the new subproject.
Should be able to run this in dbt cloud without a full refresh.


0xRobin:
cancelled the Solana CI as it won't be able to finish anyway, compile and schema checks passed though :heavy_check_mark:

PR #6192, opened 2024-06-17
zeroex - model updates on gasless txns - ethereum

RantumBits:
zeroex - model updates on gasless txns - ethereum 
continuation of #6180

PR #6203, opened 2024-06-19
Zeroex performance

aalan3:
This PR hopes to improve performance on some zeroex models.

For ethereum I added an incremental filter on prices where it was missing.

For BNB I broke out the part of the query that's doing a full scan on bnb logs into its own model. We should then be able to run that model incrementally and joining on it should be much lighter as well.


aalan3:
@RantumBits could you please review this to make sure I'm not messing up the logic on BNB here?

dex/models/_projects/zeroex/bnb/zeroex_bnb_api_fills_uni_v2_pair_creation.sql
```diff
@@ -0,0 +1,35 @@
+{{  config(
+        schema = 'zeroex_bnb',
+        alias = 'fills_uni_v2_pair_creation',
+        materialized='incremental',
+        unique_key = ['pair', 'makerToken', 'takerToken'],
+        file_format ='delta',
+        incremental_strategy='merge'
```
jeff-dude, line 7: target incremental predicate?

dex/models/_projects/zeroex/bnb/zeroex_bnb_api_fills_uni_v2_pair_creation.sql
```diff
@@ -0,0 +1,35 @@
+{{  config(
+        schema = 'zeroex_bnb',
+        alias = 'fills_uni_v2_pair_creation',
+        materialized='incremental',
+        unique_key = ['pair', 'makerToken', 'takerToken'],
```
jeff-dude, line 5: not sure all three of these should be unique keys, otherwise only `block_time` is left to be updated. likely just `pair`?

jeff-dude: or potentially this could be append-only or something rather than merge? if so, we'd have to left join `{{ this }}` to make sure not already processed. merge may still be easier with correct keys

jeff-dude (changes requested):

jeff-dude:
fyi on this one @Hosuke

dex/models/_projects/zeroex/bnb/zeroex_bnb_api_fills_uni_v2_pair_creation.sql
aalan3: My understanding of how this works is that you get the latest events for each pair, the old query implied there could be duplicates so we'd want to replace those values with the latest. This could be wrong however.

aalan3: block time might actually not even be needed here

jeff-dude: i think `pair` should be the only unique key, if we partition by that field in the `rn` and then the maker/taker tokens can change?

Pluies:
(Rebasing this to trigger a new CI run and hopefully reproduce issues that happened before for investigation 👍 )

aalan3:
Overlaps with #6192

RantumBits:
> @RantumBits could you please review this to make sure I'm not messing up the logic on BNB here?

looks good ser, thanks for your help with this 🫡

aalan3:
@RantumBits thanks for reviewing this!

PR #6226, opened 2024-06-21
solana bot_trades improvement

0xRobin:
Pulled out the users into it's own incremental model for soul_sniper and consortium_key


aalan3:
great to see solana spells actually run in CI!

aalan3 (approved):

PR #6244, opened 2024-06-24
Lifinity migrate

andrewhong5297:
migrating solana lifinity

PR #6245, opened 2024-06-24
Solana dex trades: enrichment + base trades setup

0xRobin:
Solana dex trades: enrichment + base trades setup.

This PR moves the following columns from the base (project) models to the enriched model.

```
token_bought_symbol
token_sold_symbol
token_pair
token_bought_amount
token_sold_amount
amount_usd
fee_usd
```


andrewhong5297:
pls keep me in the loop as you work on these! I have a bunch of PRs I need to make to migrate and add a few columns to dex solana first. might be worth you waiting until I'm done there.

0xRobin:
@andrewhong5297 thanks for notifying, currently only this one right: https://github.com/duneanalytics/spellbook/pull/6244?

Seems like it shouldn't impact this PR, but will wait with the followup PR (that'll remove these columns from the marketplace models itself)

jeff-dude (approved):
i'll get this in on the current larger cluster once we see the projects succeed (other than phoenix)

jeff-dude:
reiterating before merge, this won't impact project level spells yet. rob will be following up with more PRs to clean up project level that feed into this. there will then be final project level views which replicate the current state today, so no queries will break.

PR #6258, opened 2024-06-25
referral rewards: remove model with dupes

jeff-dude:
fyi @0xRobin 
this started to fail with duplicates. i didn't see any in the actual spell, so the source query must start producing dupes. i would imagine in the enrichment macro?
i need to exclude for now to get some merges to run


0xRobin:
thanks for notifying, I'll investigate

PR #6261, opened 2024-06-26
fix referral duplicates

0xRobin:
fix referral duplicates


0xRobin:
The core of this issue was a bas price token. 

This is fixed in this PR: https://github.com/duneanalytics/spellbook/pull/6262

We should be able to merge this once that is merged and `prices.usd` is updated

jeff-dude:
to be safe, i will let prices run in prod first and then have CI run here

jeff-dude (approved):

PR #6262, opened 2024-06-26
Fix price tokens and set up tests

0xRobin:
There's a a bad price token currently that is producing duplicates as it's reported as the 0x000.. address. (which is used to represent native tokens in many spells) 
This will remove it from tokens and set up tests to discover it.
image

cc: @MSilb7 the core of this is originating from `tokens_optimism.erc20_bridged_mapping`, can you handle this edge case there? It might be a bad emitted event?


0xRobin:
here are 2 transactions that have caused this:
```
select * from optimism_ethereum.L1StandardBridge_evt_ERC20DepositInitiated
where l2Token = 0x0000000000000000000000000000000000000000
```

`0xd251d574ec579f51f3d4761fa69b6f704ad302a392da430e2b2059693af10490`
`0x04215b5421cf276e1345afdfaad0aa564320085d6001447419389b67c9bb068f`

jeff-dude (approved):
discussed on our call, good to go here. next steps:
- lets this run and refresh `prices.usd`
- run `nft_trades_beta+`
- run `dex_base_trades+` (jeff to confirm that's sufficient)

this will fix rows that incorrectly joined to prices and grabbed wrong price, therefore assigning incorrect `amount_usd`

jeff-dude:
> discussed on our call, good to go here. next steps:
> 
> * lets this run and refresh `prices.usd`
> * run `nft_trades_beta+`
> * run `dex_base_trades+` (jeff to confirm that's sufficient)
> 
> this will fix rows that incorrectly joined to prices and grabbed wrong price, therefore assigning incorrect `amount_usd`

this is all done

PR #6269, opened 2024-06-26
Fix Bebop duplicates bug

B1boid:
Looks like some kind of weird inconsistent behavior of UNION ALL, because tests were passed successfully, but then in prod a couple of old duplicates appeared.

I think I fixed it, so returning eth model back to prod (https://github.com/duneanalytics/spellbook/pull/6257#issue-2373672651)


Hosuke (approved):
LGTM.✅
Thank you @B1boid 

(Maybe we can have the testing sql as a test for bebop.trades in future.)

PR #6270, opened 2024-06-26
Fix prices.usd

jeff-dude:
```
SELECT
  token_id,
  blockchain,
  symbol,
  contract_address,
  decimals
FROM
  prices_native.tokens
WHERE
  contract_address != 0x0000000000000000000000000000000000000000
```

returns no results, therefore drops all native tokens out of `prices.usd`

### DIFF
--- a/models/prices/optimism/prices_optimism_tokens.sql
+++ b/models/prices/optimism/prices_optimism_tokens.sql
@@ -6,28 +6,19 @@
         
         )
 }}
-
-WITH prices_op AS (
-    SELECT
-        token_id
-        , blockchain
-        , symbol
-        , contract_address
-        , decimals
-    FROM {{ ref('prices_optimism_tokens_curated') }}
-    UNION ALL
-    SELECT
-        token_id
-        , blockchain
-        , symbol
-        , contract_address
-        , decimals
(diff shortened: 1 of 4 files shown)


aalan3 (approved):
Ah yeah contract_address can be null. Need to do [`DISTINCT FROM`](https://trino.io/docs/current/functions/comparison.html#is-distinct-from-and-is-not-distinct-from) instead for these cases.

0xRobin:
this should do the trick.
Let's verify the CI output

jeff-dude:
adding context, as users have reached out for support.
- [this](https://github.com/duneanalytics/spellbook/pull/6262) PR was merged, due to an issue with an optimism token address (more info in description there)
- the `!=` approach caused all native tokens to drop out of `prices.tokens`, which feeds into `prices.usd`
- rather than the above approach, the `IS DISTINCT FROM` function will help us get what we need

jeff-dude:
we have an ongoing bug in CI. for our sake here:
```
SELECT
  token_id,
  blockchain,
  symbol,
  contract_address,
  decimals
FROM
  prices_native.tokens
WHERE
  contract_address IS DISTINCT
FROM
  0x0000000000000000000000000000000000000000
```
this returns results as expected. i'm good to merge.

PR #6273, opened 2024-06-26
Update claim fee logic on boost.claimed

pntemi:
2 updates:
1. Boost recently removed the claim fee from our protocol. This PR is to update the logic to reflect the current fee.
2. We had a bug in our logging where all Drakula's boost with action on Base are logged as Optimism instead.


Hosuke (approved):
LGTM.✅
Thank you @pntemi

models/boost/boost_claimed.sql
```diff
@@ -44,7 +44,10 @@ quest_claims_enriched as (
         coalesce(b.reward_token_address, c.reward_token_address) reward_token_address,
         claim_tx_hash,
         c.block_time,
-        coalesce(claim_fee_eth, 0.000075) claim_fee_eth,
+        case 
+            when block_time >= timestamp '2024-06-06 16:00:00' then 0.0
```
Hosuke, line 48: We may introduce these timestamps as jinja variables.

jeff-dude: @Hosuke do you mind pushing a commit to show how it's done? then @pntemi can leverage these things moving forward 🤝

Hosuke: Just added.

jeff-dude: i think the variables need single quotes around them, went ahead and pushed. let's see CI run now.

jeff-dude: looks to work now

jeff-dude (approved):

models/boost/boost_claimed.sql
Hosuke: Forgot to push the local commits.

PR #6280, opened 2024-06-27
Update Orca

andrewhong5297:
updating orca for new transfers spell and removing some columns


0xRobin (approved):
looks good from my side.
@andrewhong5297 I appended the changes for a `base_trades` model and the compatibility view here in your PR.

PR #6281, opened 2024-06-27
Phoenix dex fix

andrewhong5297:
redo phoenix fix without seed


0xRobin (approved):
Looks good to merge when CI passes :heavy_check_mark: 
I've made a couple small adjustments that should also be replicated once we remove columns from other base spells.
1. rename the model to `_base_trades`
2. add the `check_columns_solana_dex_trades` schema test 
3. add a compatibility view at the old model name, @aalan3 notified me these solana dex models do see some usage on the base level so we try and not break those user queries.

andrewhong5297:
> Looks good to merge when CI passes ✔️ I've made a couple small adjustments that should also be replicated once we remove columns from other base spells.
> 
> 1. rename the model to `_base_trades`
> 2. add the `check_columns_solana_dex_trades` schema test
> 3. add a compatibility view at the old model name, @aalan3 notified me these solana dex models do see some usage on the base level so we try and not break those user queries.

good callouts and foresight! I'll copy the same pattern for meteora/goose when I get to it later today/tomorrow

PR #6287, opened 2024-06-28
feat: add ape.store spellbook

whalehunting:
This spellbooks records trades from tokens that get deployed on https://ape.store -  a pump.fun style token launchpad
It includes both trades on the bonding curve, as well as trades that happen on Uniswap, after the token leaves the bonding curve

Related dashboard:
https://dune.com/whale_hunter/ape-store


models/ape_store/base/ape_store_base_schema.yml
```diff
@@ -0,0 +1,67 @@
+version: 2
+
+models:
+  - name: ape_store_base_trades
```
Hosuke, line 4: Do you want to add this `ape_store_base_trades` into any roll-up spell?

whalehunting: no, it should be a standalone spell. once they add more chains i'll add a rollup @Hosuke

Hosuke (approved):
LGTM.✅
Thank you @whalehunting 

(The view may be too large to build so it ran out of time).

[Sample ape_store_base_trades](https://dune.com/queries/3898168)

models/ape_store/base/ape_store_base_schema.yml
Hosuke: In this case do you want to add a `post_hook` to the spell?

whalehunting: yes! will do that

whalehunting: done! @Hosuke

Hosuke (approved):

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
+{{ config(
+    alias = 'trades',
+    schema = 'ape_store_base',
+    partition_by = ['block_month'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
+    unique_key = ['blockchain', 'tx_hash', 'tx_index']
```
Hosuke, line 8:
```suggestion
    unique_key = ['blockchain', 'tx_hash', 'tx_index'],
```

models/ape_store/base/ape_store_base_schema.yml
jeff-dude: can we actually remove the post hook to expose on data explorer? if we know this isn't the end spell, even if final not built yet, i'd rather hold off. we need to clean out the data explorer rather than adding more at the moment. when in doubt, we should be looking to add if it's a final level spell only and is built for the overall community (i.e. a `dex.trades` or `dex.bot_trades` type)

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
+{{ config(
+    alias = 'trades',
+    schema = 'ape_store_base',
+    partition_by = ['block_month'],
+    materialized = 'incremental',
+    file_format = 'delta',
+    incremental_strategy = 'merge',
```
jeff-dude, line 7: please add incremental predicate property in the model config block. this should filter the target on the same column which the source is being filtered incrementally.

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+    unique_key = ['blockchain', 'tx_hash', 'tx_index'],
```
jeff-dude, line 8:
```suggestion
    unique_key = ['block_month', 'blockchain', 'tx_hash', 'tx_index'],
```

models/ape_store/base/ape_store_base_schema.yml
```diff
@@ -0,0 +1,67 @@
...
+      project: ape_store
+      contributors: whale_hunter
+    config:
+      tags: ["base", "ape_store", "dex", "whale_hunter"]
+    description: >
+      Trades for tokens that get deployed through ApeStore
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - blockchain
```
jeff-dude, line 16:
```suggestion
            - block_month
            - blockchain
```

models/ape_store/base/ape_store_base_trades.sql
jeff-dude: lets add the partition column to make joins more efficient

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+    post_hook = '{{ expose_spells(
+                        blockchains = \'["base"]\',
+                        spell_type = "project",
+                        spell_name = "ape_store",
+                        contributors = \'["whale_hunter"]\') }}'
```
jeff-dude, line 13: as noted above, let's remove for now. i'd rather play it safe, as we don't have a process to remove from data explorer at the moment.

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+   )
+}}
+
+{% set project_start_date = '2024-04-04' %}
+{% set blockchain = 'base' %}
+{% set weth_contract_address = '0x4200000000000000000000000000000000000006' %}
+
+with
+    deployments as (
```
jeff-dude, line 22: since this CTE can't be filtered incrementally (assumption), i wonder if we take a similar approach to previous spells and make this it's own spell that is incremental, then read from it here. this only applies if we expect the result set here to continue to grow (especially if exponentially). do you foresee that happening?

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+            ) as token_bought_address,
+            (amount0in + amount1in) / 1e18 as token_sold_amount,
+            if(amount0out > 0, 'WETH', token.symbol) as token_sold_symbol,
+            if(
+                amount0out > 0, {{ weth_contract_address }}, token
+            ) as token_sold_address,
+            sender as user,
+            evt_tx_hash as tx_hash,
+            evt_index as tx_index
+        from {{ source('ape_store_base', 'Router_evt_swap') }} as swaps
```
jeff-dude, line 54: add where clause to filter incrementally

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+        left join
+            {{ source('tokens', 'erc20') }} as token
+            on (
+                token.blockchain = '{{blockchain}}'
+                and token.contract_address = swaps.token
+            )
+        left join
+            {{ source('prices', 'usd') }} as ethPrice
+            on (
```
jeff-dude, line 63: add join condition to filter incrementally

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+            token_bought_amount,
+            token_bought_symbol,
+            token_bought_address,
+            token_sold_amount,
+            token_sold_symbol,
+            token_sold_address,
+            tx_from as user,
+            trades.tx_hash,
+            evt_index as tx_index
+        from deployments
```
jeff-dude, line 87: to be safe, lets read from `dex.trades` and join in `deployments`. trino usually can plan the query efficiently enough on it's own, but it's best to put the larger table in the from statement.

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+        join
+            {{ source('dex', 'trades') }} as trades
```
jeff-dude, line 89: we need to filter dex incrementally

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+            on (
+                trades.blockchain = deployments.blockchain
+                and trades.block_time >= deployments.evt_block_time
```
jeff-dude, line 92: wonder if we can use `block_month` here instead, as it's the partition column on dex and should help as this spell grows? we'd need to introduce `block_month` in above CTE to match

models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,104 @@
...
+        where
+            trades.blockchain = '{{blockchain}}'
+            and trades.block_time >= timestamp '{{project_start_date}}'
+    )
+select *
+from bondingcurvetrades
+union all
+select *
+from dextrades
+order by block_time desc
```
jeff-dude, line 104: can remove order by, won't do anything on the pipeline side

jeff-dude (changes requested):
in addition to below feedback, we should consider where this spell lives. we're still in a bit of a fragile state with the main `models/` directory as we continue to move some around into their new subprojects. typically if new spells don't fit into `dex`, `nft` or `solana`, we will put into `daily_spellbook` subproject

models/ape_store/base/ape_store_base_trades.sql
whalehunting: wdym with this? i don't understand it 😆

models/ape_store/base/ape_store_base_trades.sql
whalehunting: this should not be necessary, since in the join condition we already filter time right?

```sql
left join
            {{ source('prices', 'usd') }} as ethPrice
            on (
                ethPrice.blockchain = '{{blockchain}}'
                and ethPrice.contract_address = {{ weth_contract_address }}
                and ethPrice.minute = date_trunc('minute', evt_block_time)
            )
```

models/ape_store/base/ape_store_base_trades.sql
jeff-dude: it looks like you already have it applied in line 8
`incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')],`
maybe you just added since?

to hopefully be more clear:
this property filters the target to the same timeframe as the source, i.e. where you use incremental filter on line 67 for example. based on your model configs, this model will build incrementally via merge statements. this will ensure the merge statement filters the target to a smaller timeframe to scan less rows, therefore becoming more performant

models/ape_store/base/ape_store_base_trades.sql
jeff-dude: it's still necessary, as trino will scan the entire prices dataset on left joins in order to match on `minute`. prices is quite large. there's no reason to try and read more data than our base source. plz do apply to ensure proper planning by the trino engine

jeff-dude:
CI timing out on a long incremental run is a red flag. we expect it to happen on full history step for large datasets, but if that succeeds and incremental takes 30+ minutes, we know we're missing some filters somewhere (likely the prices filter in this instance)

jeff-dude:
please also acknowledge the comment about where these spells live. we're taking the final steps to reorg spellbook subprojects now, but the "default" spot for new spells which don't feed to existing lineages will live in the `daily_spellbook` directory. by default, new spells like these would run on a daily cadence rather than hourly

models/ape_store/base/ape_store_base_trades.sql
jeff-dude: to be clear:
this question is related to the solana bot trades spells where we had to move initial CTEs to their own spells upstream rather than the CTE within due to performance constraints. i'm wondering if this spell fits a similar pattern and to get ahead of it. base is much more performant than solana, but just curious

PR #6297, opened 2024-07-01
move jupiter out of dex directory

0xRobin:
This moves Jupiter out of the /dex directory as it's not a part of the `solana_dex_trades` lineage.


jeff-dude (approved):
will mark to merge without modified to avoid unnecessary refresh

PR #6301, opened 2024-07-01
[CoW Protocol] Add and update solver names

harisang:
This PR adds the new "Horadrim" solver addresses and also updates "Otex" to "GlueX Protocol" (and also does a small change in "Quasimodo").


Hosuke (approved):
LGTM.✅
Thank you @harisang

PR #6167, opened 2024-06-13
incremental predicate update for lido liquidity

ppclunghe:
It's ready for review. Thanks for help @jeff-dude 

To be sure I got your comment in this [PR](https://github.com/duneanalytics/spellbook/pull/6033) @jeff-dude
Is the incremental_predicate correct for this sub-spell (time is the target spell column name)? If it is, I proceed with other sub-spells in the model

PR #6306, opened 2024-07-02
Fix blockchains list for lido_liquidity_arbitrum spells

vegarsti:
SyncSense caught these, regression from #6167.


aalan3 (approved):
Nice thank you. We have (had?) a check for this, will take a look into why this wasn't caught.

PR #5995, opened 2024-05-25
Create `attacks.address_poisoning`

hildobby:
Impossible to make more efficient now since I need to check every address for some heuristics. I'm now waiting on https://github.com/duneanalytics/spellbook/pull/6308 to be approved which will be a huge help in lowering the runtime

PR #6308, opened 2024-07-02
Create `addresses.info`

hildobby:
This PR creates `addresses.info` with aggregated high level information for all EVM addresses with chain-specific spells culminating into a crosschain one. I have thought of usecases for all addresses in there, it can be used:

- to easily find a subset of addresses based on some heuristics
- to join on as a filter to only query addresses in the time ranges they appeared in, potentially making queries and downstream spells more efficient (for example I think I can finally optimise https://github.com/duneanalytics/spellbook/pull/5995 into a more optimised runtime)
- to do some address segmentation by chain it shows up on, if its a contract on any chain, where/when it was first funded, etc
- I'll also be creating `addresses.index` using it, a spell that solely contains `address` and `index` where index is a `BIGINT` (or `INT256` if there's more addresses than I anticipate) created incrementally based on when an address first appeared anywhere. This can then be used:

  -  to efficiently store all kinds of spells with network graph data or spells containing a lot of addresses, ie: address `1039` sent to those distinct addresses `[10, 212, 1334]`
  - or even in spells with CTEs querying a lot of addresses, this can make that data more storage efficient allowing for more complex queries while staying under cluster capacity

For the incremental updates, the chain-specific macros constantly fetches the last block_number for every address (across txs and token transfers) so it can easily incrementally update based on those with no missing data. For the croschain version, a map is created for every address which has this data for all chains whre the address appeared:

```
map_from_entries(array[
            ('last_seen', CAST(last_seen AS varchar))
            , ('last_seen_block', CAST(last_seen_block AS varchar))
            , ('executed_tx_count', CAST(executed_tx_count AS varchar))
            , ('is_smart_contract', CAST(is_smart_contract AS varchar))
            , ('sent_count', CAST(sent_count AS varchar))
            , ('received_count', CAST(received_count AS varchar))
            ]) AS chain_stats
```

those maps are then updated on incremental runs with this line which ensures that only the chains with new data get overwritten:

```
map_concat(map_filter(t.chain_stats, (k, v) -> NOT contains(map_keys(nd.chain_stats), k)), nd.chain_stats) AS chain_stats
```

I think the spell is now ready, this PR will have most chains (and the crosschain spell) deactivated in prod so it can first build for ethereum and a couple of others, then I'll do some follow up PRs for the other chains and end with the crosschain spell.

@0xRobin lmk if you spot anything missing here, I've looked through extensively but a second pair of eyes might surface stuff I missed!


hildobby:
hey @jeff-dude, i'm trying to build this spell that will have 1 line per address with high level info (`executed_tx_count`, `max_nonce`, `is_smart_contract`, `namespace`, `name`, `first_funded_by`, `first_tx_block_time`, `last_tx_block_time`, `first_tx_block_number`, `last_tx_block_number`, `first_received_block_time`, `first_received_block_number`, `last_transfer_block_time` & `last_transfer_block_number`) 

the goal is to easily get aggregated data for any evm address but this would also be super useful for filtering other spells and making them more efficient (ie filtering each address by first tx). this is just for ethereum for now but i want to make it for all evm chains and have a crosschain version of the spell.

in order to build it in spellbook, I use the `last_updated` column as an incremental predicate which for now is the last block_time that address had any in/outflow/tx but i might switch it to MAX(block_time) in the end. on each incremental run, all new data is fetched and aggregated by address, and if that address is already there, the new data is added to the new one but i'm failing to the row replacement to work properly, jeff you mentioned that this would work do you know what i did wrong here? i just want to take new data, index by address and replace the row with that address if it already exists

jeff-dude:
at first glance, looks like it could be how you use timestamps in the incremental phase. for source filter, you consistently filter on `block_time`, but for target filter (in model config), it's based on a new field you build for timestamp, where greatest is obtained across various sources? it could be that the address + timestamp combo across tables differ, so when you filter both source/target on incremental, it doesn't properly find the row to match to. that leads to rewriting it instead, hence the failure on dupes post-incremental run.

i also notice you left join to `{{ this }}` in incremental code, but don't actually select any fields or filter on it at all. this won't actually do anything then? with the block number filter in the join condition, it would need to be inner join to work properly i believe.

hildobby:
> at first glance, looks like it could be how you use timestamps in the incremental phase. for source filter, you consistently filter on `block_time`, but for target filter (in model config), it's based on a new field you build for timestamp, where greatest is obtained across various sources? it could be that the address + timestamp combo across tables differ, so when you filter both source/target on incremental, it doesn't properly find the row to match to. that leads to rewriting it instead, hence the failure on dupes post-incremental run.
> 
> i also notice you left join to `{{ this }}` in incremental code, but don't actually select any fields or filter on it at all. this won't actually do anything then? with the block number filter in the join condition, it would need to be inner join to work properly i believe.

So I made some fixed but the issue remains. I changed the filter to a value called `last_seen`. this value will change on every update, how can i make it so that it replaces the row with the same `address` regardless of all other values including `last_seen`?

jeff-dude:
> So I made some fixed but the issue remains. I changed the filter to a value called `last_seen`. this value will change on every update, how can i make it so that it replaces the row with the same `address` regardless of all other values including `last_seen`?

i'm not fully sure without digging in deep and running myself. can we try to simplify first and remove the target filter from model config blocks altogether? see how that performs?

hildobby:
> i'm not fully sure without digging in deep and running myself. can we try to simplify first and remove the target filter from model config blocks altogether? see how that performs?

I simplified it heavily so that we can look deeper into the logic and see what breaks @jeff-dude

hildobby:
Ok here we are, back to the same issue caused by incremental filters. here the incremental filter appends new lines (thus causing tests to fail because it leads to dupes) when it should be replacing lines with the same `address` value, how can i fix that?

jeff-dude:
@Hosuke can you take some time to look into the logic attempted in this PR, and help us make it run as expected?

dbt_subprojects/daily_spellbook/models/addresses/chains/addresses_ethereum_info.sql
```diff
@@ -0,0 +1,24 @@
+{% set blockchain = 'ethereum' %}
+
+{{
+    config(
+        schema = 'addresses_' + blockchain,
+        alias = 'info',
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'merge',
```
Hosuke, line 9: Maybe we can try `delete+insert` here.

```suggestion
        incremental_strategy = 'delete+insert',
```

dbt_subprojects/daily_spellbook/models/addresses/chains/_schema.yml
```diff
@@ -0,0 +1,68 @@
...
+      - &name
+        name: name
+        description: "If the address is a smart contract that is decoded on Dune, here is the associated smart contract name on Dune"
+      - &first_funded_by
+        name: first_funded_by
+        description: "Which other address first ever funded this address"
+      - &first_tx_block_time
+        name: first_tx_block_time
+        description: "The block time of the first transaction executed by this address"
+      - &last_tx_block_time
```
Hosuke, line 45: Not sure if this is equivalent to `last_seen`, but this column seems not appeared in the final selection.

dbt_subprojects/daily_spellbook/models/addresses/chains/addresses_ethereum_info.sql
```diff
@@ -0,0 +1,24 @@
+{% set blockchain = 'ethereum' %}
+
+{{
+    config(
+        schema = 'addresses_' + blockchain,
+        alias = 'info',
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'delete+insert',
+        unique_key = ['address'],
```
Hosuke, line 10: Update strategy again:
```suggestion
        incremental_strategy = 'merge',
        unique_key = ['address'],
        merge_update_columns = ['executed_tx_count', 'max_nonce', 'is_smart_contract', 'namespace', 'name', 'last_seen', 'last_tx_block_number'],
```

dbt_subprojects/daily_spellbook/models/addresses/chains/addresses_ethereum_info.sql
```diff
@@ -0,0 +1,25 @@
...
+{{
+    config(
+        schema = 'addresses_' + blockchain,
+        alias = 'info',
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'merge',
+        unique_key = ['address'],
+        merge_update_columns = ['executed_tx_count', 'max_nonce', 'is_smart_contract', 'namespace', 'name', 'last_seen', 'last_tx_block_number'],
+        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.last_seen')]
```
Hosuke, line 12:
```suggestion
```

Hosuke:
Fixed now. The problem is due to `incremental_predicates`:

https://docs.getdbt.com/docs/build/incremental-strategy#about-incremental_predicates

Although I think I may simplify the spell further, through removing unused params.

dbt_subprojects/daily_spellbook/models/addresses/chains/addresses_ethereum_info.sql
```diff
@@ -0,0 +1,24 @@
...
+    )
+}}
+
+{{
+    addresses_info(
+        blockchain = blockchain
+        , transactions = source(blockchain, 'transactions')
+        , token_transfers = source('tokens_' + blockchain, 'transfers')
+        , creation_traces = source(blockchain, 'creation_traces')
+        , first_funded_by = ref('addresses_events_' + blockchain + '_first_funded_by')
```
Hosuke, line 21: `token_transfers` and `first_funded_by` seems unused in the macro.

Hosuke (approved):
[addresses_ethereum_info preview]()

dbt_subprojects/daily_spellbook/models/addresses/chains/_schema.yml
hildobby: no this is not the same column, last_tx_block_time is for the last executed tx. Then there's last_seen which is for transfers (executed or not by the address). The two are purposely separate because it's nice to be able to filter for just executed txs by time if thats what you're after

hildobby:
> Fixed now. The problem is due to `incremental_predicates`:
> 
> https://docs.getdbt.com/docs/build/incremental-strategy#about-incremental_predicates
> 
> Also the spell may be simplified further, through removing unused params.

All the unused columns are a remanent of the spell from a few weeks back. @jeff-dude asked me to simplify it so that we can focus on it's iteration logic, but I'm going to add all those columns back before it gets merged.

dbt_subprojects/daily_spellbook/models/addresses/chains/addresses_ethereum_info.sql
hildobby: they were removed as asked by jeff, i'll add them back.

dbt_subprojects/daily_spellbook/macros/models/sector/addresses/addresses_info.sql
```diff
@@ -0,0 +1,94 @@
...
+, nd.address
+, nd.executed_tx_count + t.executed_tx_count AS executed_tx_count
+, COALESCE(nd.max_nonce, t.max_nonce) AS max_nonce
+, COALESCE(COALESCE(nd.is_smart_contract, t.is_smart_contract), false) AS is_smart_contract
+, COALESCE(nd.namespace, t.namespace) AS namespace
+, COALESCE(nd.name, t.name) AS name
+, GREATEST(nd.last_tx_block_time, t.last_seen) AS last_seen
+, GREATEST(nd.last_tx_block_number, t.last_tx_block_number) AS last_tx_block_number
+FROM new_data nd
+LEFT JOIN {{this}} t ON t.address=nd.address
```
jeff-dude, line 90: (commenting on this line because i believe this join to `{{ this }}` is likely all that's needed)

i think we're close, but not quite accomplishing what you're looking for. i still think the left joins in CTEs above to the target table aren't doing anything, even with the block number filter. the `count(*) as executed_tx_count` is still just performing on the base transactions table filter to the last one day. once you add that value to existing count in line 82, you will likely be double counting tx's each day.

maybe the move is to introduce a CTE which reads from `{{ this }}` and grabs max block number as the only output. you could then potentially inner join to this new CTE in `executed_txs` and filter where source block number is greater than the target in the conditions. does that make sense?

we can probably remove the left join in `is_contract` altogether. the merge statement behind the scenes should handle these updates for us. i believe the only field requiring lookup to target table is the count

hildobby: I'm not sure I fully understand, the `new_data` cte is fetched from the `executed_txs` cte which has:

```
LEFT JOIN {{this}} t ON txs."from"=t.address
    AND txs.block_number>t.last_tx_block_number
```

afaik, that should mean that it only adds transactions which are past the `last_tx_block_number` value that gets updated every time there's new data. where would you see the double counting?

jeff-dude: let me share a query that hopefully visualizes what i'm saying. if you run the below, you will notice that the current incremental setup will return full tx_count for the entire one last day (our incremental predicate outputs column passed in minus one day). the left join to target table with a filter on block number helps to null out values from the target, however, the base table which performs the left join will retain all of it's data. that means all 300k+ transactions in last one day. if you run this model multiple times in a row, it's going to add that value over and over (or something close to 300k, as timestamps will slightly shift each run).

```
WITH
  executed_txs AS (
    SELECT
      "from" AS address,
      COUNT(*) AS executed_tx_count,
      COALESCE(MAX(nonce), 0) AS max_nonce,
      MIN(block_time) AS first_tx_block_time,
      MAX(block_time) AS last_tx_block_time,
      MIN(block_number) AS first_tx_block_number,
      MAX(block_number) AS last_tx_block_number
    FROM
      "delta_prod"."arbitrum"."transactions"
    where
      "from" = 0x00000000000000000000000000000000000a4b05 --example address
      and block_time <= timestamp '2024-09-16 17:14' --block time at runtime of historical data
      and block_time >= timestamp '2024-09-15 17:14' --incremental filter reads from last 1 day
    GROUP BY
      1
  )
select
  txs.*,
  t.address as target_lookup_value --this is null due to left join with block number filter
from
  executed_txs as txs
  LEFT JOIN test_schema.git_dunesql_99fe47c_addresses_arbitrum_info t ON txs.address = t.address
  AND txs.last_tx_block_number > t.last_tx_block_number
```

0xRobin (commented):
hey @hildobby, really good idea for a spell!

What you're trying to do here is very much non trivial.
Basically we need a way of doing cumulative aggregate calculations that we can update incrementally without fully materializing the daily stats. 

Here's what I would do:
- Split the macro into 1 macro for each source that is used. This will help both with the incremental complexity and runtimes (they can run in parallel). Bringing them back together will be easy using `left join some_stats using (address)`, the only downside is that we will materialize everything twice, but that's fine as optimizing compute is more important than optimizing storage. (and we're optimizing storage anyway by not materializing the daily stats)
- Don't do cross chain incrementally, make it a view (or materialized table). The extra logic needed to do this correctly incrementally is not worth it, and since we're not dealing with raw sources in this model, it should be pretty lightweight. Let's monitor the performance though when we do this. 

Splitting the macro has a couple other benefits:
- we can do the incremental logic separate for each source, which takes care of the fact that not all sources will be in sync.
- easier to define the correct incremental aggregation for each metric.

---
To help with the total work of replicating this for each chain, we have recently built a simple 'evm replication script'. The prerequisite for this are that we have 1 example chain in it's separate directory (including the schema file). Basically this means, just develop everything for `ethereum` under `../chains/ethereum/` and then run the script, and it'll replicate it for all EVMs. I'll give a more proper walk through when we can run it.

hildobby:
> What you're trying to do here is very much non trivial. 

haha I'm aware but I actually think I can nail this with incremental spells. I think materialising the crosschain spell is needed here for proper usage so I'll stick to this new non-time-based incremental spell type. This will also be a huge help for any similar spells going forward, serving as a "skeleton" of sorts.
 
> To help with the total work of replicating this for each chain, we have recently built a simple 'evm replication script'. The prerequisite for this are that we have 1 example chain in it's separate directory (including the schema file). Basically this means, just develop everything for `ethereum` under `../chains/ethereum/` and then run the script, and it'll replicate it for all EVMs. I'll give a more proper walk through when we can run it.
> 
> Thoughts?

This is very cool, how can I use it? Although copy/pasting models isn't my biggest time killer tbh. Also, ethereum might not always be the best chain to test on (ie spells for L2-only stuff), any chance we could make ethereum the default and maybe add a flag to the cli command where we can input source blockchain.

A few months back I started working on a script to create/add spells to schema files but then a spellbook re-architecture I wasn't aware of happened which rendered my wip work useless so I scrapped it. Coms between internal spellbook team and external contributors aren't optimal so these efforts are likely better handled internally unfortunately, have you guys considered making a schema file maker/adder script?

hildobby:
Asked on telegram but posting here too in case anyone knows what the error is:

Is the function `map_union_agg()` available in DuneSQL?

It's in the Dune docs here: https://docs.dune.com/query-engine/Functions-and-operators/aggregate#map-union-agg

But throws line 610:3: Function 'map_union_agg' not registered on spellbook on the CI tests here

I also tested it straight onto Dune which results in the same error: https://dune.com/queries/4096813

What am I doing wrong? It would be super useful to have for this spell, otherwise not sure how to make it without super complex logic
@jeff-dude @0xRobin @Hosuke

dbt_subprojects/daily_spellbook/macros/models/sector/addresses/addresses_info.sql
```diff
@@ -0,0 +1,265 @@
...
+, first_tx_block_time
+, last_tx_block_time
+, first_tx_block_number
+, last_tx_block_number
+, GREATEST(last_tx_block_time, last_transfer_block_time) AS last_seen
+, GREATEST(last_tx_block_number, last_received_block_number, last_sent_block_number) AS last_seen_block
+FROM executed_txs
+LEFT JOIN is_contract USING (address)
+LEFT JOIN transfers USING (address)
+LEFT JOIN {{ source('addresses_events_'~blockchain, 'first_funded_by')}} ffb USING (address)
```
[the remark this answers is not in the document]
hildobby: Ah thanks I swapped it so sources is being joined onto and changed it to a full outer join

dbt_subprojects/daily_spellbook/macros/models/sector/addresses/addresses_info.sql
```diff
@@ -0,0 +1,265 @@
...
+    , received_volume_usd
+    , sent_volume_usd
+    , first_tx_block_time
+    , last_tx_block_time
+    , first_tx_block_number
+    , last_tx_block_number
+    FROM executed_txs
+    LEFT JOIN is_contract USING (address)
+    LEFT JOIN {{ source('addresses_events_'~blockchain, 'first_funded_by')}} ffb USING (address)
+    LEFT JOIN transfers USING (address)
```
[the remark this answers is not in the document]
hildobby: Ah thanks I swapped it so first_funded_by is the `FROM` table with the rest being joined onto and executed_txs + transfers as `FULL OUTER JOINs`

dbt_subprojects/daily_spellbook/models/addresses/chains/addresses_arbitrum_info.sql
```diff
@@ -0,0 +1,25 @@
...
+{{
+    config(
+        tags=['prod_exclude'],
+        schema = 'addresses_' + blockchain,
+        alias = 'info',
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'merge',
+        unique_key = ['address'],
+        merge_update_columns = ['executed_tx_count', 'max_nonce', 'is_smart_contract', 'namespace', 'name', 'first_funded_by', 'first_funded_by_block_time', 'tokens_received_count', 'tokens_received_tx_count', 'tokens_sent_count', 'tokens_sent_tx_count', 'first_transfer_block_time', 'last_transfer_block_time', 'first_received_block_number', 'last_received_block_number', 'first_sent_block_number', 'last_sent_block_number', 'received_volume_usd', 'sent_volume_usd', 'first_tx_block_time', 'last_tx_block_time', 'first_tx_block_number', 'last_tx_block_number', 'last_seen', 'last_seen_block'],
```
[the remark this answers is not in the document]
hildobby: It didn't work until those columns were added. Those are specified here as columns to be replaced, this is the first time I make (and see) a spell of this type, afaik all other incremental spells only work as append with no replace component unlike here where this appears to be necessary

dbt_subprojects/daily_spellbook/models/addresses/addresses_info.sql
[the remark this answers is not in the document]
hildobby: I agree on the complexity but if it were a view it would be unusable. Crosschain views are only useful when heavily filtered but can't be used much for macro analysis. Views are also going to get increasingly bad given the speed at which new EVM chains pop up.

I finally got first_funded_by's crosschain materialised making it useful for things that always timed out before, here I would really like to do the same.

I also really think that this spell, if done properly, would be in high demand through the API and across queries overall, reinforcing the need for it to be efficient to query.

dbt_subprojects/daily_spellbook/models/addresses/chains/addresses_arbitrum_info.sql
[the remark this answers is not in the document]
0xRobin: That's very strange behavior, normal incremental spells definitely work by replacing the incremental results.
I'll have a look at this through the CI.

dbt_subprojects/daily_spellbook/models/addresses/addresses_info.sql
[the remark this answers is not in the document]
0xRobin: ok that's fine on reviewing this next time.

dbt_subprojects/daily_spellbook/models/addresses/chains/addresses_arbitrum_info.sql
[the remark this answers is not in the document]
0xRobin: I have removed these statements and everything runs fine from what I can tell.

0xRobin (approved):
@hildobby
I have pushed some changes.
- removed the `merge_update_columns`, they don't seem needed
- replaced `greatest` with a combo of `array_max` and `filter` to allow it to handle null values
- moved everything into the new `addresses` sector, this deserves it's own spot.

PR #6325, opened 2024-07-05
Dex trades uint256

aalan3:
This should be a uint256 and there should not be negative values in dex.trades amounts


jeff-dude:
should we be doing a "catch all" here at the end, or digging into the project/versions which are causing this to happen in the first place and fix at the base trades level?

jeff-dude (approved):

PR #6327, opened 2024-07-05
nft.wash_trades small migration

hildobby:
MERGE WITH https://github.com/duneanalytics/spellbook/pull/6331

### DIFF
--- a/nft/macros/nft_wash_trades.sql
+++ b/nft/macros/nft_wash_trades.sql
@@ -112,14 +112,17 @@ WITH filter_1 AS (
     SELECT unique_trade_id
     , true AS flashloan
     FROM {{ ref('nft_trades') }} nftt
-    INNER JOIN {{ source('dex','flashloans') }} df ON df.blockchain='{{blockchain}}'
-        AND df.block_time=nftt.block_time
+    INNER JOIN {{ source('lending','flashloans') }} df ON df.blockchain='{{blockchain}}'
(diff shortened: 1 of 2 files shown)


nft/macros/nft_wash_trades.sql
```diff
@@ -112,14 +112,17 @@ WITH filter_1 AS (
     SELECT unique_trade_id
     , true AS flashloan
     FROM {{ ref('nft_trades') }} nftt
-    INNER JOIN {{ source('dex','flashloans') }} df ON df.blockchain='{{blockchain}}'
-        AND df.block_time=nftt.block_time
+    INNER JOIN {{ source('lending','flashloans') }} df ON df.blockchain='{{blockchain}}'
+        AND df.block_number=nftt.block_number
```
jeff-dude, line 116: another good rule of thumb:
use the partition column in join conditions. both of these tables use `block_month`. if you add that as an extra condition, the query engine will have less files to scan through. then can further filter on that small subset by `block_number`, `tx_hash`

jeff-dude: i think it's okay to proceed for now as there isn't a performance concern in current table sizes. something for us to track when finalizing a refactor on `nft.wash_trades` (once we complete `nft.trades`)

jeff-dude (approved):

0xRobin (approved):

PR #6343, opened 2024-07-08
Meteora and GooseFX changes

andrewhong5297:
changes to make these two tables faster, and updating base_trades to include block_slot and project_main_id


jeff-dude:
i assume this one is also just blocked by CI. we could look to test both open PRs on an increased cluster as needed.

solana/models/_sector/dex/goosefx/goosefx_ssl_v2_solana_trades.sql
```diff
@@ -60,36 +61,33 @@
...
+            , trs_1.to_token_account as token_sold_vault
         FROM {{ source('goosefx_solana', 'gfx_ssl_v2_call_swap') }} sp
-        INNER JOIN {{ source('spl_token_solana', 'spl_token_call_transfer') }} trs_1 
-            ON trs_1.call_tx_id = sp.call_tx_id 
-            AND trs_1.call_block_time = sp.call_block_time
-            AND trs_1.call_outer_instruction_index = sp.call_outer_instruction_index 
-            AND trs_1.call_inner_instruction_index = COALESCE(sp.call_inner_instruction_index,0) + 1
+        INNER JOIN {{ ref('tokens_solana_transfers') }} trs_1 
+            ON trs_1.tx_id = sp.call_tx_id 
+            AND trs_1.block_time = sp.call_block_time
```
jeff-dude, line 71:
```suggestion
            AND trs_1.block_date = CAST(date_trunc('day', sp.call_block_time)
```

jeff-dude: we should consider always joining to solana transfers on `block_date`, as this is the partition column on that spell and we've seen performance enhancements when doing this

jeff-dude: this would require introducing `block_date` to all solana dexs -- which we likely should be doing anyway, to match `dex.trades`. 
`block_time`
`block_date`
`block_month`
should be standard across sector spells

jeff-dude: in order to keep this PR clean, we could take the first step to just use the `CAST` directly in join condition, then follow up with introducing `block_date` and cleaning up downstream joins.
cc @0xRobin

jeff-dude (changes requested):

0xRobin (approved):

jeff-dude:
> i've updated these to match all other DEXs in solana lineage -- plz let me know if you approve @andrewhong5297

for regression testing, i ran this:
```
with
  ci_table as (
    select
      project,
      version,
      count(1) as num_rows,
      sum(token_sold_amount_raw) as total_sold,
      sum(token_bought_amount_raw) as total_bought
    from
      test_schema.git_dunesql_cc0b1c5_dex_solana_trades
    where
      block_time >= DATE '2024-08-01'
      and block_time < DATE '2024-08-15'
    group by
      project,
      version
  ),
  prod_table as (
    select
      project,
      version,
      count(1) as num_rows,
      sum(token_sold_amount_raw) as total_sold,
      sum(token_bought_amount_raw) as total_bought
    from
      dex_solana.trades
    where
      block_time >= DATE '2024-08-01'
      and block_time < DATE '2024-08-15'
    group by
      project,
      version
  )
select
  *
from
  prod_table
except
select
  *
from
  ci_table
```
and it returns no rows, as expected, meaning the data matches between the two env's. i left out `amount_usd` intentionally because the updated pipeline modifies the logic slightly to use trusted tokens price first, so the values differ slightly (but are more accurate with the new logic 🙏)

jeff-dude (approved):

PR #6353, opened 2024-07-11
Change lending sector's `usd_amount` to `amount_usd`

hildobby:
@jeff-dude @aalan3 @0xRobin could we get some spellbook standardisation going? seems to me like such a low hanging fruit to only accept columns that have the same name when present across spells. that includes columns such as `amount_raw`, `amount`, `amount_usd`, `block_time`, `block_date`, `block_month`, `block_number`, `tx_hash`, `evt_index`, `tx_index`, `blockchain`, `project`, `version`, `project_contract_address`, `tx_from`, `tx_to`.

afaik none of those have any reason to have varrying names across spells, that includes `evt_*`, `call_*` variations as well as any other custom variations (like usd_amount). There should also be some checks for columns to always be there, some examples:
- if `block_time` is there, `block_number` should also be there
- `block_date`/`block_month` is there, `block_time` should also be there
- `amount_usd` is there, `amount` should also be there
- `project_contract_address` is there, `project` & `version` should also be there
- `evt_index` is there, `tx_hash` & `tx_index` should also be there
- `tx_from` is there, `tx_hash` should also be there
- `tx_to` is there, `tx_hash` should also be there
- `tx_hash` is there, `tx_index` should also be there

A lot of spellbook's confusion is spell discoverability & credibility but lack of standardisation also really doesn't help. This is so much better to enforce upon spell creation rather than once it's been used in many queries, please do something about it thx


jeff-dude:
> A lot of spellbook's confusion is spell discoverability & credibility but lack of standardisation also really doesn't help. This is so much better to enforce upon spell creation rather than once it's been used in many queries, please do something about it thx

we are aware and putting a larger emphasis on curated data quality, credibility & standardization. we've even had recent internal team shifts to put more focus here. it won't happen overnight, but we're starting to list out the major spells and then how to get them to higher quality & featured in data explorer for discoverability.

right now, many of these sector spells outside of dex/nft get v low usage query execution wise, so these changes won't happen overnight, but we will be working towards it.

as to new spells, fully agree on being more strict moving forward. we are also planning to do that. @Hosuke and myself will be trying to spot these pre-merge.

jeff-dude (approved):
added supply as well

hildobby:
> right now, many of these sector spells outside of dex/nft get v low usage query execution wise, so these changes won't happen overnight, but we will be working towards it.

yeah i would never use a dune spell i didn't write/contribute to/review because of the unclear (unquantified) trust assumptions, and that's not even factoring in that dune discoverability is horrible. so i can't imagine others knowing about, let alone using many spells.

but assuming those are things you're currently fixing (afaik), i bet those will eventually be surfaced better and thus used more. it's probably worth spending a 🤏 amount of time to review col names before they get surfaced and thus potentially used more, no? obviously not all, but there's some (imo) no-brainer sectors ie: staking, lending, cex, bridge [WIP], mev [WIP], addresses [WIP], attacks [WIP], etc (i'm biased i work(ed) on all those, but i would be surprised if those don't get much usage later)

it's also still unclear to me what curated means for you guys, do you just maintain them? add new protocols yourself? just a stamp of approval?

jeff-dude:
> yeah i would never use a dune spell i didn't write/contribute to/review because of the unclear (unquantified) trust assumptions, and that's not even factoring in that dune discoverability is horrible. so i can't imagine others knowing about, let alone using many spells.
> 
> but assuming those are things you're currently fixing (afaik), i bet those will eventually be surfaced better and thus used more. it's probably worth spending a 🤏 amount of time to review col names before they get surfaced and thus potentially used more, no? obviously not all, but there's some (imo) no-brainer sectors ie: staking, lending, cex, airdrop, bridge [WIP], mev [WIP], addresses [WIP], attacks [WIP], etc (i'm biased i work(ed) on all those, but i would be surprised if those don't get much usage later)
> 
> Also, contributing to spellbook is already riddled with points of friction, so i definitely won't make a gh issue for every new sector i'm making, that's just asking too much from contributors (who's incentives to contribute are already negative). i would rather the standards be defined first so i can follow those. pls just lower friction to contribute, it's only been increasing over the past 2 years.
> 
> it's also still unclear to me what curated means for you guys, do you just maintain them? add new protocols yourself? just a stamp of approval?

long story short, the objective is more query executions against curated data. using the term curated vs. spells as we now include labels, balances and such which live outside of spellbook.

to achieve that, all the above is part of the effort to get there. our product team is actively working on answering many of these questions. once we are in a good spot, we will surface that out publicly. i'll be sure to keep hannah (our PM) closely aligned here.

PR #6361, opened 2024-07-12
Feature/app data sources

olgafetisova:
### Contribution type

- [X] New spell(s)
- [X] Adding to existing spell lineage
- [X] Bug fix
---

The PR solves two separate issues:

1) Adding new logic to the trades table 
- cow_protocol_arbitrum_trades.sql
- cow_protocol_gnosis_trades.sql
- cow_protocol_ethereum_trades.sql

### For bug fixes

- **Description:** Usd_volume and fee_usd fields were not calculated correctly due to the null values
- **Implementation details:** The new fix solves the problem by treating null value as 0, which will not affect the final value of the calculation but will prevent the break of the logic.
- **Test instructions:** PoC query can be found here https://dune.com/queries/2360196

2) Adding new data sources + creating new tables

**New sources** - files uploaded to Dune
- dune_upload.dataset_app_data_gnosis
- dune_upload.dataset_app_data_arbitrum
- dune_upload.dataset_app_data_mainnet

**Modifying existing table** with new source 
- cow_protocol_ethereum_app_data
- cow_protocol_ethereum_schema.yml

**Creating new tables and adjusting yml files**
- cow_protocol_gnosis_app_data.sql
- cow_protocol_arbitrum_app_data.sql
- cow_protocol_arbitrum_schema.yml
- cow_protocol_gnosis_schema.yml


olgafetisova:
DBT fails potentially due to the fact that in the same PR I am adding both sources and query logic that depends on these sources. Do you recommend splitting it into two different PRs or ignoring the error? 
@dune-eng 
@jeff-dude 

Source mentioned here https://github.com/duneanalytics/spellbook/blob/b87ade472180c4cf5198ac638765d3863273937d/sources/cow_protocol/gnosis/cow_protocol_gnosis_sources.yml#L125

And this is the query it is referenced in 
https://github.com/duneanalytics/spellbook/blob/b87ade472180c4cf5198ac638765d3863273937d/models/cow_protocol/gnosis/cow_protocol_gnosis_app_data.sql#L31 .

There are 3 new sources added to different files with the same issue.

jeff-dude:
> DBT fails potentially due to the fact that in the same PR I am adding both sources and query logic that depends on these sources. Do you recommend splitting it into two different PRs or ignoring the error? @dune-eng @jeff-dude
> 
> Source mentioned here
> 
> https://github.com/duneanalytics/spellbook/blob/b87ade472180c4cf5198ac638765d3863273937d/sources/cow_protocol/gnosis/cow_protocol_gnosis_sources.yml#L125
> 
> And this is the query it is referenced in
> 
> https://github.com/duneanalytics/spellbook/blob/b87ade472180c4cf5198ac638765d3863273937d/models/cow_protocol/gnosis/cow_protocol_gnosis_app_data.sql#L31
> 
> .
> There are 3 new sources added to different files with the same issue.

hi @olgafetisova!

can you explain a bit more what these sources are? i'm unable to query them. are they limited to your cow team?
```
select *
from dune_upload.dataset_app_data_gnosis
limit 10
```
this returns table does not exist.

is there any reason you can't use materialized views on the app to build on top of these instead?

olgafetisova:
> > DBT fails potentially due to the fact that in the same PR I am adding both sources and query logic that depends on these sources. Do you recommend splitting it into two different PRs or ignoring the error? @dune-eng @jeff-dude
> > Source mentioned here
> > https://github.com/duneanalytics/spellbook/blob/b87ade472180c4cf5198ac638765d3863273937d/sources/cow_protocol/gnosis/cow_protocol_gnosis_sources.yml#L125
> > 
> > And this is the query it is referenced in
> > https://github.com/duneanalytics/spellbook/blob/b87ade472180c4cf5198ac638765d3863273937d/models/cow_protocol/gnosis/cow_protocol_gnosis_app_data.sql#L31
> > 
> > .
> > There are 3 new sources added to different files with the same issue.
> 
> hi @olgafetisova!
> 
> can you explain a bit more what these sources are? i'm unable to query them. are they limited to your cow team?
> 
> ```
> select *
> from dune_upload.dataset_app_data_gnosis
> limit 10
> ```
> 
> this returns table does not exist.
> 
> is there any reason you can't use materialized views on the app to build on top of these instead?

I found this example on how to reference to the files uploaded to dune 
https://github.com/duneanalytics/spellbook/blob/40c2ae5888e82bd0d51bad78446b252775e010fd/sources/_base_sources/dune_upload_sources.yml (which is this table dune.oplabspbc.dataset_op_airdrop1_addresses_detailed_list, but referenced as in the example in the link.

So the tables I am actually referring to are the following:

- dune.cowprotocol.dataset_app_data_arbitrum
- dune.cowprotocol.dataset_app_data_gnosis
- dune.cowprotocol.dataset_app_data_mainnet

Is there a better way to reference such uploads?

jeff-dude:
> Is there a better way to reference such uploads?

yeah, you use:
`{{ source("cowprotocol", "dataset_app_data_arbitrum", database="dune") }}`

then the source yml file should be the standard schema / table name, without worrying about `dune` database prefix.
```
sources:
  - name: cowprotocol
    tables:
      - name: dataset_app_data_arbitrum
```

we should be careful about adding these types of sources though. we don't use often in spellbook due to concerns that any changes on the app to these tables can break pipelines in spellbook.
- are these static uploads?
- what happens if you need to change them?

i'd like to ask again, any reason mat views on the app won't work for your use case instead so you have full control?

olgafetisova:
> i'd like to ask again, any reason mat views on the app won't work for your use case instead so you have full control?

There is no issue in creating materialized views. 
However, if I rely solely on the materialized view in my dashboards, the old table `cow_protocol_ethereum.app_data` won't be updated and will have outdated data. I am trying to prevent this so that the community continues using the old table with the newly updated data since there are already many queries created using the previous structure. 
Additionally, I wanted to add the same structure for arbitrum and gnosis.

I 'd like to confirm: should I proceed with updating the script to use materialized views? or should I use my materialized view (ex. dune.cowprotocol.result_cow_protocol_ethereum_app_data) directly without updating the Spellbook?

jeff-dude:
> There is no issue in creating materialized views. However, if I rely solely on the materialized view in my dashboards, the old table `cow_protocol_ethereum.app_data` won't be updated and will have outdated data. I am trying to prevent this so that the community continues using the old table with the newly updated data since there are already many queries created using the previous structure. Additionally, I wanted to add the same structure for arbitrum and gnosis.
> 
> I 'd like to confirm: should I proceed with updating the script to use materialized views? or should I use my materialized view (ex. dune.cowprotocol.result_cow_protocol_ethereum_app_data) directly without updating the Spellbook?

i apologize but i'm not fully following. the existing ethereum spell reads from an old community upload table, right? now you are attempting to override that with a new API upload table specific to cow namespace?

all i'm mentioning is to essentially use the same SQL query you use here to build spellbook views to instead just run the query on the app and build mat view on top. it should accomplish the exact same thing, but you'll have full control.

PR #6386, opened 2024-07-18
add token granularity to label spells

0xRobin:
add token granularity to label spells


0xRobin:
I've come across some token contracts that have multiple symbols in the `tokens.transfers` dataset. This is due to a change that was made to the symbol in `erc20.tokens` but a full refresh was never done. 

@jeff-dude we should do a full refresh of:
- `tokens_polygon_transfers`
- `tokens_bnb_transfers`
- `tokens_ethereum_transfers`
to resolve this. 

https://dune.com/queries/3928601
image

jeff-dude:
> I've come across some token contracts that have multiple symbols in the `tokens.transfers` dataset. This is due to a change that was made to the symbol in `erc20.tokens` but a full refresh was never done.
> 
> @jeff-dude we should do a full refresh of:
> 
> * `tokens_polygon_transfers`
> * `tokens_bnb_transfers`
> * `tokens_ethereum_transfers`
> 
> to resolve this. (these are the final enrichment models so shouldn't be to heavy I believe) https://dune.com/queries/3928601 image

oof, this is when i wish we used views for metadata. seems feasible for erc20 metadata, just not prices (at least in current state). something to continue to consider. while we still materialize, we also need to consider consistent full refreshes. something we plan to do when we fully migrate to new orchestration tool (started steps on that).

this isn't blocking this PR, correct? seems CI still passes fine?

jeff-dude (approved):

0xRobin:
Correct, although we should do the full refresh before merging this one. Otherwise this table will as well contain duplicate symbols for those contract addresses.

jeff-dude:
> Correct, although we should do the full refresh before merging this one. Otherwise this table will as well contain duplicate symbols for those contract addresses.

running a full refresh on those three now. trying all in one step, if too much on current cluster size, i'll do per step or increase size

PR #6388, opened 2024-07-18
DUX-551 move AA models to main project

jeff-dude:


jeff-dude:
fyi @0xRobin 
this lineage was requested to move back to hourly. for now, moving to main project. it's in `_sector/` directory so should be easy to track (and can mimic this PR if needed). we will move this to new hourly subproject. i will update the spreadsheet to reflect this too

PR #6401, opened 2024-07-22
Create Hourly Spellbook subproject

0xRobin:
1. Split out hourly models into its' own new 'hourly_spellbook' dbt subproject
```
├── _project
│   ├── balancer
│   ├── cow_protocol
│   ├── ens
│   ├── lido
│   └── safe
└── _sector
    ├── abi
    ├── account_abstraction
    ├── blobs
    ├── cex
    ├── lending
    ├── perpetual
    ├── prices
    ├── rollup_economics
    └── staking

```
2. This PR also moves the following lineages to the `tokens` subproject:
```
- prices.tokens
- transfers
- balances
```


hourly_spellbook/models/_sector/staking/solana/staking_solana_schema.yml
jeff-dude, line 1: should likely move these solana spells to solana project?

hourly_spellbook/models/_sector/staking/solana/staking_solana_stake_actions.sql
jeff-dude, line 1: should likely move these solana spells to solana project?

sources/_subprojects_outputs/tokens/transfers.yml
jeff-dude, line 1: any reason transfers has its own source file here rather than add to _sources?

jeff-dude (commented):
in addition to the minor comments below:
- any reason we don't move project specific macros to their new subproject, i.e. all the balancer macros to `hourly_spellbook/macros/`? i think we're gonna have some follow up cleaning to do in terms of getting all macros in the correct spot, across various projects
- can we use this PR to continue moving more around and finalize the projects? my thought was to reuse the existing dbt cloud project and cluster for `hourly_spellbook`, then all others move to their final spots too, which would require no extra setup since already done. that would mean pushing all remaining spells to `daily_spellbook` project within this PR too. otherwise, we'll have to set up new dbt cloud project, new cluster, to support `hourly_spellbook` while we continue to run the other jobs on legacy

sources/_subprojects_outputs/tokens/transfers.yml
0xRobin: no particular reason, I was trying some new formats out to see how we can clean up the shared sources in the future.
I think I do prefer this setup with some more structure and split into more different files.
From a dbt standpoint this doesn't matter at all so it's just a repository organisation question.

0xRobin:
> * any reason we don't move project specific macros to their new subproject, i.e. all the balancer macros to `hourly_spellbook/macros/`? i think we're gonna have some follow up cleaning to do in terms of getting all macros in the correct spot, across various projects

good point, will clean them up while we're at it. :+1: 



> * can we use this PR to continue moving more around and finalize the projects? my thought was to reuse the existing dbt cloud project and cluster for `hourly_spellbook`, then all others move to their final spots too, which would require no extra setup since already done. that would mean pushing all remaining spells to `daily_spellbook` project within this PR too. otherwise, we'll have to set up new dbt cloud project, new cluster, to support `hourly_spellbook` while we continue to run the other jobs on legacy

Initially I wanted to limit the PR to just splitting out the hourly models (as it is already huge). 
But happy to take this further in one go if we're fine with the CI not running. :ok_hand:

0xRobin:
yeah github is **not** liking these big PRs, will revert and split back into 2.

0xRobin:
@jeff-dude my advise would be to merge this, setup a new dbt cloud job for hourly spellbook but re-use the current legacy cluster.
After that we can open this PR against main and move everything to it's final location.

0xRobin:
I'm closing this and reopening the PR here (https://github.com/duneanalytics/spellbook/pull/6421) from a branch within the spellbook repo (instead of my fork). 
Ci is behaving weird and I think it's due to the branch being on the fork.

PR #6408, opened 2024-07-22
Refactor l2_revenue from `rollup_economics`

lgingerich:
### Contribution type

- [ ] New spell(s)
- [X] Adding to existing spell lineage
- [ ] Bug fix

---

- **Description:** Refactor `rollup_economics.l2_revenue` to reference the `gas.fees` table rather than duplicating the gas logic again in this query. These rollup_economics spells are already very heavy and this is an improvement in reducing query load. This also helps keep gas calculation logic in a single place for updates (i.e `gas.fees`) rather than duplicating over multiple spells.


Hosuke (approved):
Thank you @lgingerich 

Overall looks good, though there are some discrepancies when comparing with the original result:
https://dune.com/queries/3939250

lgingerich:
Thanks for catching that. Six of the differences are on 2024-07-23, three on 2024-03-29, and one on 2023-06-06. Is there anything you can think of which would cause this clustering?

If it was an issue with the logic, I would expect discrepancies across multiple chains to be randomly distributed in time.

lgingerich:
I checked `gas.fees` for zksync on 2024-07-23 and it is correct with 2.19 ETH revenue: [https://dune.com/queries/3939289](https://dune.com/queries/3939289)

Hosuke:
I think the new implementation is good to me.

0xRobin (approved):
this looks good. :heavy_check_mark: 
Any discrepancy we find in the future should be solved by making the `gas.fees` models more accurate.

Hosuke:
There is a CI issue:
https://github.com/duneanalytics/spellbook/actions/runs/10158841309/job/28091739966?pr=6408#step:8:15

lgingerich:
@Hosuke I'm fixing now

lgingerich:
Working now

jeff-dude (approved):

jeff-dude:
i like the simplicity here for sure 🙏 

one note:
i think this is a bit confusing, so maybe something we can clean up to make more clear. for spells which are listed as sources in dbt, they should live in the file of which project they belong. built in that project, output from that project. ultimately, having one source file for all spells might make this a bit easier to follow after seeing it in a few spots'
fyi @0xRobin @Hosuke

PR #6415, opened 2024-07-23
Add more tokens for ZKsync

lgingerich:


Hosuke (approved):

jeff-dude:
oof, even with the new automated source to feed token metadata, we still have this many tokens missing?
how did you generate this long list @lgingerich ?

lgingerich:
[ZettaBlock](https://zettablock.com/) indexes tokens so I went there to get this data.

Honestly I'm just not fully aware of the new token situation in Dune. Can you describe the current process and when/how I should be manually uploading token metadata? Is it possible to just index all tokens like block explorers do?

jeff-dude:
> Honestly I'm just not fully aware of the new token situation in Dune. Can you describe the current process and when/how I should be manually uploading token metadata? Is it possible to just index all tokens like block explorers do?

we have a new automated source (not in spellbook), which gets called here:
https://github.com/duneanalytics/spellbook/blob/main/tokens/models/tokens/tokens_erc20.sql#L41-L59

the coverage wasn't as extensive as expected, so we left the process for union'ing in manual token additions for now. you can see that in the same model above.

the goal was to lessen the need for PRs like this, but we seem to be falling short there. you can query the source directly on dune too, to see tokens covered:
`dune.definedfi.dataset_tokens`
we're basically just reading from the definedfi API.

lgingerich:
Thanks, this does look pretty good. Pricing is obviously great but another part of me wants these things indexed just for address labeling. There's many times where I see a trending contract and it's some random ERC20/721/1155 token. It's not that difficult to figure it out manually, but when address labeling is one of the biggest missing pieces of understanding crypto data, it would be extremely nice to have natively in Dune.

Defined seems great for pricing but it seems lacking for this other use case (and NFTs in general)

jeff-dude:
thanks for sharing. helpful to know for us to revisit internally on how to index this metadata efficiently.

jeff-dude (approved):

PR #6426, opened 2024-07-24
Refactor `rollup_economics` spell

lgingerich:
**IN WORK**


dbt_subprojects/hourly_spellbook/models/_sector/rollup_economics/ethereum/_schema.yml
Hosuke, line 1: I just refactored these spells into `dbt_subprojects/hourly_spellbook/models/..`.
Please check if I miss anything.

0xRobin (commented):
Overall this looks very good! :heavy_check_mark: 

Can we try to have these as views:
- `rollup_economics_l1_data_fees.sql`
- `rollup_economics_l1_verification_fees.sql`

I don't expect these to be queried directly, and the aggregated data is present in `l1_fees`. Trying to reduce the amount we're storing the same data multiple times across models.

> RE: blockchain naming. l1_blob_fees pulls from an upstream source with human readable chain names which differ from the standard Dune naming (i.e. should be "zksync", not "zkSync Era"). I have not made the change yet but I would prefer to move to the standard Dune naming model for each of these chains, and potentially change the name column to be blockchain. Let's discuss this further.

I agree on this but would make the changes in a follow up PR to keep the refactor clean.

lgingerich:
@0xRobin 
- `rollup_economics_l1_data_fees.sql`
- `rollup_economics_l1_verification_fees.sql`

In the [existing model](https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/hourly_spellbook/models/_sector/rollup_economics/ethereum/l1_data_fees.sql), these are both set as incremental runs. I'm fine changing to views but just want to make sure you're good changing this behavior.

RE: naming — sounds good will leave for a later PR

lgingerich:
@jeff-dude I'm still trying to understand how to handle sources in this new format. How should I fix this?
```
Compilation Error
  dbt found two sources with the name "gas_fees".
  
  Since these resources have the same name, dbt will be unable to find the correct resource
  when looking for source("gas", "fees").
  
  To fix this, change the name of one of these resources:
  - source.hourly_spellbook.gas.fees (../../sources/_subprojects_outputs/hourly_spellbook/_sources.yml)
  - source.hourly_spellbook.gas.fees (../../sources/_subprojects_outputs/daily_spellbook/_sources.yml)
  ```

0xRobin:
> @jeff-dude I'm still trying to understand how to handle sources in this new format. How should I fix this?
> 

@lgingerich to prevent having this issue of double source definitions we've come to the following standard:
Any dependencies between different dbt subprojects can be listed as am "output" with the project that it's build in.
For this case, gas.fees resides in daily_spellbook so you should keep the source definition in `_subprojects_outputs/daily_spellbook` and remove the one in `_subprojects_outputs/hourly_spellbook`

0xRobin (approved):

jeff-dude:
there seems to be some inconsistencies with usage of `rollup_economics` and `rollup_economics_ethereum` in terms of schema names / corresponding file names. all spells fall under `../rollup_economics/ethereum/` directory path, so i'd imagine having `_ethereum` in names is expected?

changing schema names can cause some impact downstream, so likely fine to leave as-is, but wanted to call this out

jeff-dude (approved):

lgingerich:
@jeff-dude good callout, fixed now

PR #6441, opened 2024-07-26
Update Ethereum `gas.fees` for EIP4844

lgingerich:


lgingerich:
```
Compilation Error
  Model 'model.daily_spellbook.gas_ethereum_fees' (models/_sector/gas/fees/ethereum/gas_ethereum_fees.sql) depends on a node named 'ethereum_blob_submissions' which was not found
  ```
  `gas.fees` is a daily model while `ethereum.blob_submissions` is run hourly. Could that by why it doesn't find the node?
  cc: @jeff-dude

jeff-dude:
> ```
> Compilation Error
>   Model 'model.daily_spellbook.gas_ethereum_fees' (models/_sector/gas/fees/ethereum/gas_ethereum_fees.sql) depends on a node named 'ethereum_blob_submissions' which was not found
> ```
> 
> `gas.fees` is a daily model while `ethereum.blob_submissions` is run hourly. Could that by why it doesn't find the node? cc: @jeff-dude

we just merged a PR to finalize movement of spells across dbt subprojects. if spells live across projects, use `source` instead of `ref` in jinja syntax.

sorry to do during development, but there'll always be open PRs :) it's tough to communicate fully, so we will help each PR that was open.

jeff-dude:
the `readme` has more info on subprojects, along with a link to gh discussion around it.

lgingerich:
@jeff-dude thanks for the insight! Changing from `ref` to `source` didn't seem to work. Any ideas?

jeff-dude:
> @jeff-dude thanks for the insight! Changing from `ref` to `source` didn't seem to work. Any ideas?

i think this is still the right approach, we just need to establish the output spell from the other project as a source. please see [this](https://github.com/duneanalytics/spellbook/blob/main/sources/_subprojects_outputs/hourly_spellbook/_sources.yml) source file, which lists all spells within the hourly project which are used as sources in other projects. if you add there, dbt should find it

lgingerich:
Finally good to go on this

jeff-dude (approved):
incremental was already pretty efficient here, but pushed a commit to move prices incremental filter to join conditions since it's a left join. we've seen improved performance on that in general. i think this one was quick previously since it filters to one chain/token only to cut down on the data more

jeff-dude:
should this PR be merged and built in prod prior to merging #6408, since that will read from `gas.fees` now?

lgingerich:
I don't think it should matter either way. https://github.com/duneanalytics/spellbook/pull/6408 does use `gas.fees` but only touches L2's, while this is for Ethereum

jeff-dude:
> I don't think it should matter either way. #6408 does use `gas.fees` but only touches L2's, while this is for Ethereum

ah good point, you filter on chains there

PR #6480, opened 2024-08-01
Gnosis transfers v2

hdser:
### Update!

### Contribution type

- [ ] New spell(s)
- [ ] Bug fix

---

### For new spell(s)

---

- **Description:** [Description of the changes made]

---

### For bug fixes

---

### Additional information

---


dbt_subprojects/tokens/models/transfers_and_balances/gnosis/_schema.yml
```diff
@@ -2,105 +2,114 @@ version: 2
...
-        description: "The transaction index"
-      - name: from
-        description: "The sender of the transfer"
-      - name: to
-        description: "The receiver of the transfer"
-      - name: contract_address
-        description: "The contract address of the transfer"
+      tags: ['tokens', 'transfers', 'gnosis']
+    columns: 
+      - <<: *base_columns
```
jeff-dude, line 55: TIL of this usage -- so efficient!

jeff-dude: fyi @Hosuke 
pair this with line 12 above for clean column reusability

dbt_subprojects/tokens/models/transfers_and_balances/gnosis/tokens_gnosis_suicide_transfers.sql
```diff
@@ -0,0 +1,134 @@
...
+        , tx_from
+        , tx_to
+        , address 
+        , refund_address 
+    FROM 
+        {{ source('gnosis', 'traces') }}
+    WHERE
+        type = 'suicide'
+        AND
+        success
```
jeff-dude, line 32: are you intentionally leaving out incremental filters on this one?

dbt_subprojects/tokens/models/transfers_and_balances/gnosis/tokens_gnosis_suicide_transfers.sql
```diff
@@ -0,0 +1,134 @@
...
+        address
+        ,event_sequence
+        ,SUM(amount_raw) AS amount_raw
+    FROM (
+        SELECT
+            t2.address 
+            ,-t1.amount_raw AS amount_raw
+            ,t2.event_sequence
+        FROM    
+            {{ ref('tokens_gnosis_base_without_suicide_transfers') }} t1
```
jeff-dude, line 55: same question on this one, if the above adds incremental, we should here too to match

dbt_subprojects/tokens/models/transfers_and_balances/gnosis/tokens_gnosis_suicide_transfers.sql
```diff
@@ -0,0 +1,134 @@
...
+            t1.block_time < t2.block_time
+
+        UNION ALL
+
+        SELECT
+            t2.address 
+            ,t1.amount_raw AS amount_raw
+            ,t2.event_sequence
+        FROM    
+            {{ ref('tokens_gnosis_base_without_suicide_transfers') }} t1
```
jeff-dude, line 74: same question on this one, if the above adds incremental, we should here too to match

jeff-dude (commented):
really well written and easy to follow, thank you 🙏 

a few questions below, mostly around incremental on one model

jeff-dude:
in addition to above, hoping to leverage your expertise here. would logic throughout this PR apply to all EVM chains? should i be making an effort to apply to all chains?

jeff-dude:
another question:
a teammate internally noticed and called out transfers where `from = to` -- should we be filtering those out? essentially saying `where "from" != "to"`?'

this is example tx (on another chain):
https://etherscan.io/tx/0xedf6692b9ed3284766dc3ca9b9815b7eeb421d43909814af04f4ed10d0f6d400#eventlog

hdser:
Agree that one has to be careful in filtering `"from" != "to"` 
There are even token implementations where you can emit a transfer event where the value transfered is bigger than the supply of the token. Take MANA token implementation
`  function transferFrom(address _from, address _to, uint256 _value) returns (bool) {
    var _allowance = allowed[_from][msg.sender];

    // Check is not needed because sub(_allowance, _value) will already throw if this condition is not met
    // require (_value <= _allowance);

    balances[_to] = balances[_to].add(_value);
    balances[_from] = balances[_from].sub(_value);
    allowed[_from][msg.sender] = _allowance.sub(_value);
    Transfer(_from, _to, _value);
    return true;
  }`
  
  because they do not apply `require (_value <= _allowance);` the event is still emitted when to = from for example https://etherscan.io/tx/0xc9b19be499695b68f4da5c1c9bbbc8ccabb72f76624b63f1b6dc392c1da1f718#eventlog
  
 This does not affect balances but affect volume transfer metrics. I am not sure that the transfer model should remove those. They are still an event produced on-chain even though it is not "real". One might want to create a metric to track which tokens allow this and who are the users doing this. But any model that tracks volume should have this filter.

dbt_subprojects/tokens/models/transfers_and_balances/gnosis/tokens_gnosis_suicide_transfers.sql
hdser: This was intentional, mostly because it feels the implementation of the incremental is non-trivial and it would be hard for me to test it. Some more details:

The `tokens_gnosis_suicide_transfers.sql` is a particular model. The whole point of the `transfers` model is to have a table that will allow to build the balances. However, we need to know the balances of the contracts that are destroyed in order to build this table.

So the idea is that every time a contract is destroyed one can refund the native currency that the contract has into a refund address. So one needs to know the balance, which means we need the cumulative sum of all transfers (in/out) of this contract up to the point of destruction. 

I see two pain points on the incremental implementation:
1. This would mean that the incremental should keep track of the real balance block by block, and not loose this info when moving to an new partition. Not that difficult, but needs to be tested that we do not "leak" balances when changing partitions during incremental updates; 
2. We do not know which contract will be destroyed, so we would need to keep the balance of all contracts even though we are just trying to build the transfers table.

So, I am happy is there is an incremental implementation for this model. But I think it can not be the standard incremental filters used for the other models.

hdser:
Concerning the application to other EVMs. The change in the macro to account for the contract creation events that add fund is directly applicable for all, I believe.
Same for the suicide events.
For the `non_standard_transfers` this is no longer the case and might be different for each chain. On Gnosis, there is a block reward contract for when it was PoA. It follows the same fee mechanism as Ethereum but the fees go into an address and not burned. So this model is chain specific one.

dbt_subprojects/tokens/models/transfers_and_balances/gnosis/tokens_gnosis_suicide_transfers.sql
jeff-dude: all good, the explanation as to why is all i needed 🤝 

the performance on this spell looks pretty good as-is, taking only 33s to build full history

jeff-dude:
thank you for all of the clarifications. i think i'm good with these changes. let me know when you're all set with your commits

hdser:
Thanks!
Just learns that there are some type of transactions that pay no fee, even though block explorers show this a fee paid and burned. Just doing some small tweaks

jeff-dude:
> Thanks! Just learned that there are some type of transactions that pay no fee, even though block explorers show this a fee paid and burned. Just doing some small tweaks

all good, tag me when ready!

hdser:
@jeff-dude done some last tests. With the last small changes on the `tokens_gnosis_base_non_standard_transfers` part one is now able to get the correct value of the balance for the fee collector contract
Screenshot 2024-08-05 at 17 08 11
see comparison with https://gnosisscan.io/address/0x6BBe78ee9e474842Dbd4AB4987b3CeFE88426A92

Run some test for checking [negative balances](https://dune.com/queries/3966006), found a small typo on the suicide model, now looks that no address has negative balances over their entire history, block by block. 

PR seems ready to me.

jeff-dude:
> PR seems ready to me. Should the macro be left as _new, or this be merged with the standard one?

i'm curious to see how the macro looks overwritten, rather than `_new`. can you try a commit to see it?
we also merged a PR to move around macros, so you may need to move that around a bit.

hdser:
> > PR seems ready to me. Should the macro be left as _new, or this be merged with the standard one?
> 
> i'm curious to see how the macro looks overwritten, rather than `_new`. can you try a commit to see it? we also merged a PR to move around macros, so you may need to move that around a bit.

Updated brach with new macro layout and merged the _new to be the `transfer_base`.
pipeline will probably fail because it will run all blockchains and timeout.

jeff-dude:
canceling CI for now. i will finalize review here shortly

jeff-dude (approved):
looks good to me. thanks again for helping contribute to the transfers for gnosis. hopefully we can use these learnings to consider for other chains in the future. i'll get this queued up for early next week

PR #6496, opened 2024-08-04
Solana staking improvements

andrewhong5297:
Adding a few solana improvement tables.

"solana_utils.epochs": this is a simple table that I'm creating for easier query UX around epochs on Solana

"staking_solana.stake_account_delegations": vote delegations of accounts on solana are non-trivial to pull because of merge/split/withdraw accounts. This is an intermediary table to pull all historical validators delegated by a stake account

"staking_solana.validator_stake_account_epochs": this is a table that combines the prior two to get all stake accounts delegated to a validator BEFORE an epoch began. This then allows us to more accurately calculate active stake and stake rewards in in-app queries, which has been heavily requested by solflare, steakhouse, helius, and other teams.


andrewhong5297:
hmm this used to run in 37 minutes but idk why it is suddenly now timing out. maybe we can bump the cluster for this one, it really shouldn't be that compute expensive

andrewhong5297:
oh it ran @jeff-dude , data looks good to me

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_stake_account_delegations.sql
```diff
@@ -0,0 +1,78 @@
...
+                        FROM {{ ref('staking_solana_stake_actions') }} a
+                        LEFT JOIN (SELECT *
+                                        , COALESCE(call_block_time, timestamp '2000-01-01 00:00:00') as filled_block_time
+                                        , call_block_slot as filled_block_slot
+                                    FROM {{ source ('stake_program_solana','stake_call_DelegateStake') }}
+                                    ) del
+                            ON del.account_stakeAccount = a.source
+                            AND del.filled_block_time <= a.block_time
+                        WHERE action = 'split'
+                    ) as old_vote
```
jeff-dude, line 59: since the main query is filtering incrementally, should this subquery that is left joined also include incremental filters? or does it need history?

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs.sql
```diff
@@ -0,0 +1,30 @@
...
+    , b.epoch_time
+    , b.epoch_start_slot
+    , b.epoch_next_start_slot
+    , b.stake_account
+    , b.vote_account
+    , max_by(bal.sol_balance, bal.day) as sol_balance 
+FROM {{ ref('staking_solana_validator_stake_account_epochs_raw')}} b
+LEFT JOIN {{ ref('solana_utils_daily_balances')}} bal ON bal.address = b.stake_account 
+    AND bal.token_mint_address is null
+    AND bal.day <= date_trunc('day', b.epoch_time)
```
jeff-dude, line 26: since main table is filtered incrementally, should this one left joined be too?

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs_raw.sql
```diff
@@ -0,0 +1,40 @@
...
+with 
+    base as (
+        SELECT 
+            epoch.epoch
+            , epoch.block_time as epoch_time
+            , epoch.epoch_start_slot
+            , epoch.epoch_next_start_slot
+            , vote.stake_account
+            , max_by(vote.vote_account, vote.block_time) as vote_account --latest delegated vote before epoch started. then get the rewards from accounts per epoch.
+        FROM {{ ref('staking_solana_stake_account_delegations') }} vote
```
jeff-dude, line 22: the table left joined is filtered incrementally, should this one be too?

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_stake_account_delegations.sql
```diff
@@ -0,0 +1,78 @@
+{{ config(
+    schema = 'staking_solana'
+    , alias = 'stake_account_delegations'
+    , materialized = 'incremental'
+    , file_format = 'delta'
+    , incremental_strategy = 'merge'
+    , unique_key = ['block_time', 'block_slot', 'stake_account', 'vote_account']
```
jeff-dude, line 7: looks like this model is failing the unique test, but only after incremental run. that usually means incorrect keys. i noticed this model has 4 total columns, 4 keys, meaning nothing to update when match found. can we simplify the keys to less columns?

then the other thing to consider is null values in a key, which would fail the lookup. can any be null?

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_stake_account_delegations.sql
andrewhong5297: this needs history unfortunately. we could move it into a sub incremental model if we wanted to

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs.sql
andrewhong5297: can't, because we don't know the day the balance was last updated (need the max_by() to get it)

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs_raw.sql
andrewhong5297: no, the base table needs full history since we don't know the last delegation changed date

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_stake_account_delegations.sql
andrewhong5297: none of the keys can be full. and yes this would only ever append, and not update. I put this unique key here so that we just don't end up with duplicates for overlapping incremental runs. we can't remove any of them because stake accounts could move delegations between vote accounts (and vice versa) any time.

jeff-dude: okay, only other option should be the target filter. i've removed to try again.

jeff-dude (approved):
okay, looks better now

PR #6502, opened 2024-08-05
EVM blockchain metrics

0xRobin:
EVM blockchain metrics

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/blockchain_metrics/arbitrum/_schema.yml
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/blockchain_metrics/arbitrum/_schema.yml
@@ -1,0 +1,25 @@
+version: 2
+
+models:
+  - name: arbitrum_address_metrics
+    columns:
+      - name: blockchain
+      - name: address
+        tests:
+          - unique
+      - name: min_block_time
+        description: "block_time of the first transaction initiated by the address"
+
+  - name: arbitrum_metrics
+    meta:
+      blockchain: arbitrum
+      sector: metrics
+      contributors: 0xRob
+    config:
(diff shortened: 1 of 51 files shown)


0xRobin:
On issue I see here is that you can't properly aggregate the address and contract metrics (if you want to show daily or weekly stats in the dashboard, there's no way to correctly calculate these from the hourly stats as they are dependent on the time window)

hannahblue:
> On issue I see here is that you can't properly aggregate the address and contract metrics (if you want to show daily or weekly stats in the dashboard, there's no way to correctly calculate these from the hourly stats as they are dependent on the time window) Hourly might not be the best timeframe for this dataset.

@0xRobin  That's a good point. And even if the timeframe was daily, it would push the problem to broader timeframes like weekly or monthly. The new_addresses and new_contracts are ok at least, because they only ever get counted once. But we should remove active_addresses, returning addresses, and active_contracts.

jeff-dude:
expected schema, per requirements doc:
- `blockchain`: name of blockchain
- `chain_id`: unique id associated to blockchain
- `block_hour`: truncate block time to hour
- `tx_count`: count distinct tx_hash values
- `tx_success_count`: count of successful txs
- `tx_success_rate`: count of successful txs / total txs
- `avg_block_time_seconds`: current block time - previous block time, recorded in seconds granularity
- `tx_per_second`: count distinct tx's / block_time
- `new_addresses`: distinct "from" where not in previous block(s)
- `new_contracts`: distinct "to" where not in previous block(s), in contracts list

columns removed, per feedback:
- `active_addresses`: removed per hannah comment above
- `returning_addresses`: removed per hannah comment above
- `active_contracts`: removed per hannah comment above

hannahblue:
This looks good. Thank you @jeff-dude and @0xRobin

PR #6512, opened 2024-08-06
Fix bug in gas_ethereum.fees

0xRobin:
The bug was still present due to the inclusion of the incremental predicate for blobs in the last where statement.
This meant only blob transactions were included when the model was run incrementally. 
cc: @lgingerich


lgingerich:
Makes sense, thanks for fixing

jeff-dude (approved):
thanks for quick fix!

PR #6518, opened 2024-08-07
make epochs table instead of incremental

andrewhong5297:
### Update!

### Contribution type

- [ ] New spell(s)
- [x] Bug fix

I'm making solana_utils.epoch a table to refresh fully each time versus incremental, because the incremental is messing with the row_number() to get `first_block_epoch`. because each time it runs, it will add a new first_block_epoch window function ordering which then messes up the query. Open to creative solutions here too if anyone thinks of one, too late in the night for me to find one haha

### DIFF
--- a/dbt_subprojects/solana/models/solana_utils/solana_utils_epochs.sql
+++ b/dbt_subprojects/solana/models/solana_utils/solana_utils_epochs.sql
@@ -2,11 +2,9 @@
   config(
         schema = 'solana_utils',
         alias = 'epochs',
-        materialized='incremental',
+        materialized='table',
         file_format = 'delta',
-        incremental_strategy = 'merge',
         unique_key = ['block_slot'],
-        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')],
         post_hook='{{ expose_spells(\'["solana"]\',
                                     "sector",
                                     "solana_utils",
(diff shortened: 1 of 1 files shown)


andrewhong5297:
seems to be some solution to using incremental tables with window function columns but would make this sql maybe 100-200 lines longer lol. it's a cheap query so maybe we just do the full refresh each time. I'm sure we've run into this in the past with other models in here.

https://butterflymx.com/blog/dbt-incremental-models/

0xRobin:
@andrewhong5297 a creative solution here is to manually look back more then the standard incremental window. So that the first result in the incremental window still has all the necessary rows to completely fill it's window.

So if an epoch here is ~2 days, we need to look back 1d (incr window) + 2d (epoch duration) on blocks, let's do 4d just to be safe.
So on `solana.blocks` add:
```
{% if is_incremental() %}
        WHERE time > now() - interval '4' day
{% endif %}
```
And on the final select we still add:
```
{% if is_incremental() %}
WHERE {{incremental_predicate('block_time')}}
{% endif %}
```
So we only merge the results in the incremental window.

andrewhong5297:
But then wouldn't it overwrite the past epoch it runs into? i.e it might get the current epoch right but then mess up the past epoch.

andrewhong5297:
Like basically using a day filter will always end up in the middle of one epoch or another

jeff-dude (approved):

jeff-dude:
lets merge for sake of time, but i need to revisit existing spells to find if we had a good solution to window functions / incremental. building an example one could be useful too.

andrewhong5297:
agree, worth revisiting in the future as this is a simple table we can use as an example

PR #6541, opened 2024-08-12
Remove price for old GALA token

fhenneke:
### Contribution type

- [ ] New spell(s)
- [x] Bug fix

---

- **Description:** The price of the [old GALA](https://etherscan.io/address/0x15d4c048f83bd7e37d49ea4c83a07267ec4203da) token is wrong. Instead of using its actual price, the price of the [new GALA](https://etherscan.io/address/0xd1d2Eb1B1e90B638588728b4130137D262C87cae) token is used.
- **Steps to reproduce:** The actual price of the old token [according to some liquidity pools](https://app.uniswap.org/explore/tokens/ethereum/0x15d4c048f83bd7e37d49ea4c83a07267ec4203da) is around 2-3 orders of magnitude smaller than the price given on Dune by this query:
```sql
select * from prices.usd
where contract_address in (0x15d4c048f83bd7e37d49ea4c83a07267ec4203da, 0xd1d2Eb1B1e90B638588728b4130137D262C87cae)
order by minute desc
limit 2
```
The prices of the new and old token are the same on Dune, which is not correct.
- **Implementation details:** An association of the old GALA token with the token id of the new GALA token is removed. This PR does not advocate for using a better price feed. It only advocates for removing a wrong price feed.
- **Test instructions:**
- **Related issue(s):**


fhenneke:

dbt_subprojects/tokens/models/prices/ethereum/prices_ethereum_tokens.sql
```diff
@@ -157,7 +157,6 @@ FROM
     ('fx-function-x', 'ethereum', 'FX', 0x8c15ef5b4b21951d50e53e4fbda8298ffad25057, 18),
     ('fxc-flexacoin', 'ethereum', 'FXC', 0x4a57e687b9126435a9b19e4a802113e266adebde, 18),
     ('fxs-frax-share', 'ethereum', 'FXS', 0x3432b6a60d23ca0dfca7761b7ab56459d9c964d0, 18),
-    ('gala-gala', 'ethereum', 'GALA', 0x15d4c048f83bd7e37d49ea4c83a07267ec4203da, 8),
```
Hosuke, line 160: Do you think we can fix the token address to reuse this price feed?

jeff-dude: good question. can we update address to the below?
`0xd1d2Eb1B1e90B638588728b4130137D262C87cae`

coinpaprika has a price that matches what i see on etherscan for the new address:
https://coinpaprika.com/coin/gala-gala/

would this work?

jeff-dude: ah, nevermind, i see in your PR description now a query that shows this token is in prices twice already with both addresses. we are simply removing the old one.

jeff-dude (approved):

PR #6550, opened 2024-08-13
Dex_trades: remove partition by project

couralex6:


jeff-dude (approved):
pushed a commit to remove from base trades too

0xRobin:
note that this will decrease the performance of any project-specific queries (like `uniswap.trades`)

jeff-dude:
> note that this will decrease the performance of any project-specific queries (like `uniswap.trades`)

yeah, testing performance now.

background for this PR:
- partitions with spaces (`0x API`, `1inch LOP`) are causing issues in pricing pipeline downstream

try two things:
- remove `project` altogether, assuming no major performance impact
- rename those projects, not fully sure the downstream impact on queries

jeff-dude:
okay, yes, performance on the downstream project views is degraded enough to not make this change right now. i will close this PR, then ask those above teams if we can change project name.

PR #6557, opened 2024-08-14
fix curve duplicate pools

0xRobin:
fix curve duplicate pools
```
select * from curvefi_ethereum.view_pools where pool_address = 0x4d8842511cadcc65125cb9353b9520cc7f424688
```
The model did not handle tx'es where multiple pools where created, I added an extra condition on the event <> call join to resolve this.


jeff-dude:
heads up @andrewhong5297 -- for your awareness on a model you helped build awhile back. looks like a new use case came through, but rob was on it to fix 🫡

jeff-dude (approved):

PR #6562, opened 2024-08-15
DUX-612 Gas fees revamp

0xRobin:
Gas fees revamp [WIP]
todo's:
- apply and test macro's
- unify schema's
- finalize final `gas.fees` schema and spell

PR #6570, opened 2024-08-16
Alphabetize and Include all Gas Fee Tables

### LINKED ISSUE #6569: Missing Chains in Table Aggregation
https://github.com/duneanalytics/spellbook/blob/ea76a11fcc7800da0b430c10ac73a5b1579427ab/dbt_subprojects/daily_spellbook/models/_sector/gas/fees/gas_fees.sql#L11-L21

I can see that several other gas tables are available here, have they been intentionally excluded because the native asset differs from ETH?

bh2smith:
There are missing tables that already exist and are otherwise inaccessible in the Dune UI (e.g. polygon, gnosis, fantom, mantle). I suspect this has to do with the native asset being different from ETH, but One can expect analysts to be able to distinguish this fact (at least I think its better than not having access to them). We could add a USD value column to the table so people making bar charts with prices have something to rely on. 

closes #6569 

I also alphabetized the list here.


jeff-dude:
gm @bh2smith 
i'm going to work on finalizing #6562 today, which aims to fix `gas.fees` for all evm chains. i can reference work here as needed, or you can hop over and give a quick review if you'd like 🙏

bh2smith:
That looks like a major overhaul. I don't see that these other chains are going to be included, but I will wait and leave it to you for now. Then come back after the structural changes if the info remains unavailable.

PR #6579, opened 2024-08-19
Add Event Parsing for GMX

AIDataMaster:
Created a new **gmx** folder under the **daily_spellbook** subproject, including an "event" directory for parsing events from the main contract address **0xc8ee91a54287db53897056e12d9819156d3822fb**. 

The goal is:
- to add models for **PositionIncrease** and **PositionDecrease** events for **Arbitrum** and **Avalanche_C** blockchains. 
- to add model for **MarketCreated** event
- to add model for **OrderCreated** event
- to add model for **OrderExecuted** event
- to add model for **OrderCancelled** event
- to add model for **PositionFeesCollected** event
- to add model for **PositionFeesInfo** event
- to add all token contracts using public API (real + synthetic assets)
- to add macros for efficient data processing

**Ways to contribute to Spellbook**:
- [x] Build spells

Progress:
- [x] Arbitrum PositionIncrease
- [x] Arbitrum PositionDecrease
- [x] Avalanche_c PositionIncrease
- [x] Avalanche_c PositionDecrease
- [x] Macro
- [x] Configuration
- [x] Add Synthetic Assets
- [x] Add MarketCreated Model
- [x] Join PositionIncrease/Decrease with MarketCreated
- [x] Tests
- [x] Add OrderCreated Event Model 
- [x] Add OrderExecuted Event Model
- [x] Add OrderCancelled Event Model
- [x] Add PositionFeesCollected Event Model
- [x] Add PositionFeesInfo EventModel


AIDataMaster:
> I have read the CLA Document and I hereby sign the CLA

AIDataMaster:
@jeff-dude This is my answer to your question on Discord

I decided to change the logic of working with assets on GMX protocol
This protocol has event "MarketCreated". It generates all assets that are traded on dex.
There are some real assets like SOL and some synthetic assets like DOGE.
So, it means that tokens.erc20 does not have synthetic tokens - yes, I did investigation on Dune first! 

I did 2 iterations:
1st iteration: I added synthetic tokens to tokens_<chain>.erc20 but I did it manually.
It means that if I don't insert a new token manually then the subsequent models can crash on the next build. 
Why? 
We have erc20 -> MarketCreated -> EventModel
If erc20 is not updated then MarketCreated will have NULL in some of rows because I use LEFT JOIN for 2 tables: MarketCreated and ERC20Tokens
If MarketCreated is crashed or has NULL values in decimals it means that EventModel will not apply POWER function correctly.
POWER function return NULL for POWER(10,NULL). It means that all rows for new market will not be updated correctly.

My decision: I reverted all updated for synthetic tokens in subproject tokens

2nd iteration:
Inside of the GMX team we decided to use API endpoint to fetch all markets available on exchange. My colleague told me that it's crucial endpoint that has impact on trading. So if clients trade on GMX, backend somehow incorporate logic of that endpoint (endpoint is listed in model in PR). 
So, we decided to create a separate model of all assets on GMX based on API call.

You can see that model in path: ".../daily_spellbook/models/_projects/gmx/tokens/<chain>/<model>.sql

AIDataMaster:
@jeff-dude, could you please review the macro **process_variable.sql** that was added?

I need help converting **CamelCase** function names into **snake_case** within the macro. The SQL code that performs this operation looks like this:

`SELECT lower(regexp_replace('positionFeeUsd', '([A-Z])', '_$1')) AS snake_case;`

However, I’m finding it challenging to create a new macro named **camel_to_snake_case** to perform this operation.

If you could assist me with this macro, I would be very grateful.

AIDataMaster:
@dune-eng, I am working on five additional events: **PositionFeesCollected**, **PositionFeesInfo**, **OrderCreated**, **OrderExecuted**, and **OrderCancelled**.

Could you suggest the best approach for adding these events? Is it better to include everything in one pull request (PR), or would it be more effective to break down the work into separate PRs? What is the most convenient way for you to review new models?

jeff-dude:
i want to take a step back and start simple here.

this may be a silly question (apologies ahead of time, if so), but i notice you are reading straight from raw logs tables frequently and decoding fields on the fly within the models -- are you aware of the decoding process on dune which will decode these logs for you and output clean source tables, which you can then read in spellbook to simplify the logic?

AIDataMaster:
Yes, I know about this table: gmx_v2_arbitrum.EventEmitter_evt_EventLog1 and other ones.
But it contains info in EventData (json data type). It's not convenient to work with it. 
What do you suggest? replace <chain>.logs with decoded tables ? and build a table on top of decoded tables?

jeff-dude:
the most common process in spellbook is to build curated datasets on top of decoded source tables for sure.
raw --> decoded --> spellbook curated data

do you mean there's one column within that is json, even in decoded? wouldn't it still be less processing, like being done on logs here?

AIDataMaster:
I agree with you, however, that json has multiple nested jsons that makes it very complicated to process.
I have parsed all data until the end but still facing some errors in final parsing of json. I can share the code. Maybe, you can see light improvements that I can do to finish that work.

Anyway, it's quite hard to parse the whole json data to extract all info. However, it could ensure quality of data and speed up fetching data from decoded data and speed up other models creation.

AIDataMaster:
@jeff-dude please, get familiar with this partial solution too.

I’ve extracted data for one tx_hash, focusing on the addressesItems array within the JSON EventData. The result is a pivoted table where columns represent keys and records contain the corresponding values. This approach can be extended to other arrays in EventData and all tx_hashes, by joining the initial and EventData tables.

However, extracting data from EventData may take longer than directly from arbitrum.logs.

Please review the [query](https://dune.com/queries/4010350) and suggest any improvements for parsing all JSON data. If no better approach is identified, we can stick with this method or continue using raw data for these spellbooks, or improve the way how we store data in JSON EventData to speed up its parsing.

PR #6590, opened 2024-08-21
DUX-620 cleanup and fix data explorer entries

jeff-dude:


dbt_subprojects/nft/models/nft_metrics/ethereum/metadata/nft_ethereum_metadata_xcopy_collections.sql
```diff
@@ -4,8 +4,8 @@
         ,alias = 'xcopy_collections'
         ,materialized = 'table'
         ,post_hook='{{ expose_spells(blockchains = \'["ethereum"]\',
-                                    spell_type = "nft",
-                                    spell_name = "metadata",
+                                    spell_type = "sector",
+                                    spell_name = "nft_ethereum_metadata",
```
jeff-dude, line 8: fyi @catherine-2 
the param names are confusing, but `spell_type` needs to be either: `sector` or `project` and then we can name `spell_name` what we'd like, but i'll choose schema name here to make it easy to find in data explorer

jeff-dude: we want to clean up usage of this post hook to simplify but need to find a way to do so without breaking anything, hopefully soon!

jeff-dude:
canceling CI and running in prod without modified enabled to avoid unnecessary refreshes

PR #6595, opened 2024-08-22
Add more docs slugs

0xBoxer:
Add more docs slug to spells elevated in data explorer


jeff-dude (approved):
all good here?
hate that schema files kick off CI. feel free to manually cancel those with each commit, to avoid running so much for no reason

0xBoxer:
yep this is finished, any label I can use to not trigger the CI runs?

I feel like I saw that pattern somewhere

jeff-dude:
> yep this is finished, any label I can use to not trigger the CI runs?
> 
> I feel like I saw that pattern somewhere

nah, not really, only option is `prod_exclude` dbt tag in model configs. wouldn't help this scenario.

0xBoxer:
ok, will cancel manually going forward 🫡

PR #6598, opened 2024-08-23
fix col description find replace accident

0xBoxer:
accidentally used find and replace on base_fee_per_gas for col descriptions, fixing that in this PR


jeff-dude (approved):
ah yes, love base chain but hate the name for data engineering purposes 💀

PR #6599, opened 2024-08-23
fix source files: cleanup source definitions

jeff-dude:
fyi @0xBoxer 
last merged PR had fix for source table name on optimism, which then dbt found another entry for in different directly. apparently source names are case sensitive, so once you fixed, the other had already been established and clashed. the reason we didn't see in CI is source file updates don't trigger CI. we'll have to run `dbt compile` locally to see

while i'm here, moving all three sources from that random labels directory into base sources


0xBoxer:
ya this was strange stuff, somehow they are also not consistent upstream. thanks for fixing

PR #6603, opened 2024-08-23
DUX-621 initiate new address level metrics

jeff-dude:
todo:

- [x] test data output
- [x] validate schema
- [x] add all EVM chains
- [x] remove hardcoded date filter for quicker run times


dbt_subprojects/hourly_spellbook/macros/blockchain_transaction_address_metrics.sql
```diff
@@ -0,0 +1,109 @@
...
+    select
+        distinct address
+    from
+        {{ source(blockchain, 'creation_traces') }}
+    where
+        1 = 1
+        and block_time >= timestamp '2024-08-01'
+        {% if is_incremental() %}
+        and {{ incremental_predicate('block_time') }}
+        {% endif %}
```
0xRobin, line 13: we should not filter this incrementally, as a contract could have been deployed at any moment in the past.

dbt_subprojects/hourly_spellbook/macros/blockchain_transaction_address_metrics.sql
```diff
@@ -0,0 +1,109 @@
...
+    , tx.tx_count
+    , cast(tx.tx_success_count as double)/cast(tx.tx_count as double) as tx_success_rate
+    , case
+        when tx.block_hour = from_new_address.min_block_hour then True
+        else False
+        end as from_is_new_address
+    , case
+        when from_contract.address is not null then True
+        else False
+        end as from_is_contract
```
0xRobin, line 82: no harm in leaving this in, but it's currently not possible to initiate a transaction from a contract address so these will all be False.

0xRobin: nvm, I stand corrected this is possible on zksync

dbt_subprojects/hourly_spellbook/models/_sector/blockchain_metrics/_schema.yml
```diff
@@ -13,3 +13,19 @@ models:
...
+    config:
+      tags: [ 'evms','metrics', 'address' ]
+    description: "fundamental blockchain transaction address metrics"
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - blockchain
+            - block_hour
+            - from_address
+            - to_address
```
0xRobin, line 31: Probably more of a comment on the requirements than on the PR itself but I would have expected these address metrics to have just 1 address as the subject of the data instead of a (from, to) address pair as it currently is defined.
Currently the `tx_count` column is '1' for ~90% of the rows meaning we mainly just have duplicate data from evms.transactions in here. 
image
https://dune.com/queries/4020812/6768658

0xRobin (commented):
did you happen to try out the generation script for this?

dbt_subprojects/hourly_spellbook/macros/blockchain_transaction_address_metrics.sql
jeff-dude: good call, likely wrote properly on historical then updated to incorrect value on incremental with filter. i have removed.

dbt_subprojects/hourly_spellbook/macros/blockchain_transaction_address_metrics.sql
jeff-dude: always one chain to have an outlier 😅

dbt_subprojects/hourly_spellbook/models/_sector/blockchain_metrics/_schema.yml
jeff-dude: i have confirmed with hannah the from/to both existing is the intended output. there are use cases it enables for their usage downstream.

jeff-dude:
> did you happen to try out the generation script for this?

i got halfway through and thought "shit, i should have used the script". next time, now that it's on the mind!

PR #6611, opened 2024-08-26
`tokens.transfers` update model schema to models, not sources

jeff-dude:
fyi @0xBoxer 
we had wrong format in yml file here, should be model, not source


0xBoxer:
damn sorry about that, created this via my script and didn't think about it.

PR #6627, opened 2024-08-28
Make aave_flashloans and dex_flashloans views.

0xRobin:
Make aave_flashloans and dex_flashloans views.

Hopefully this should resolve the recurring ghost partition issue we're seeing.


jeff-dude:
i may look to continue this branch/PR to move all flashloan spells into one project, they are fragmented atm making it more difficult to fix issues

0xRobin:
yes, that would be great. From what I could see there's only minor column name differences between the "dex" flashloans and the "lending" ones.

jeff-dude (approved):
nvm, let's proceed as-is. i started to document what is happening, i just can't explain why. we can catch up on it in 1:1 if you want. i will likely ask a teammate to help me debug, as it looks to be an outside process in s3 affecting us and dbt not knowing it happened.

jeff-dude:
i lied, one small commit to refresh lending until investigation is complete

PR #6631, opened 2024-08-29
Duplicates dodo dex

0xRobin:
- `union distinct` instead of `union all` to resolve duplicate decodings
- changing all dodo versions from `2_[DSP|DVM|..]` to just `2`


0xRobin:
cc @tomfutago @Hosuke on the version changes.

jeff-dude:
adding slight context here, in case we come back to this PR to update these again if fixed at decoding level.

> this appears to be a decoding issue:
> - contract that were submitted for decoding are actually DSP contracts
> - our decoding pipeline added decoding for DSP on existing DVM contracts (no clue why, might be because deployed through the same factory?)
> - the spell does union all on these decoded tables, which then breaks uniqueness, due to each decoded table returning the same tx's

jeff-dude:
oops, accidental close

jeff-dude (approved):

PR #6647, opened 2024-09-02
Setup token transfer metrics

0xRobin:
Setup token transfer metrics


0xBoxer:
just raising that I think these should just be `token_metrics` instead of `token_transfer_metrics. The former will allow us to extend these with balances, distribution etc. metrics while the former boxes us in quite heavily.

jeff-dude:
closing this for now, but keep the branch for when we revisit 🙏

Issue #6577, opened 2024-08-19
Representing native tokens in prices models


0xRobin:
# Representing native tokens in prices models
 Native tokens are often annoying to deal with when pulling in price info.
I'll describe the current state with it's problems and solutions, and what I think should be the desired state. 

## 1. Current state

Currently the native tokens are defined [here](https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/tokens/models/prices/prices_native_tokens.sql):
https://github.com/duneanalytics/spellbook/blob/d948aee66f5f8b7a89882b45a7d70117d56a449b/dbt_subprojects/tokens/models/prices/prices_native_tokens.sql#L34-L37

They have `blockchain`, `contract_address` and `decimals` as `null` values.

This is also how they show up in `prices.usd` ([dune query](https://dune.com/queries/3998695))
![image](https://github.com/user-attachments/assets/f0029f3f-a22d-4767-a7b0-54e079189d8c)

### Problems and current solutions

When you have a model that includes trades from both erc20 as native tokens, you have to adapt your query to deal with this.
Most solutions follow any of these setups
1. **replace the native rows with a wrapped alternative**
```sql
select * 
from (
    select
       blockchain
      ,tx_hash
      ,case when currency_contract = 0x0000000000000000000000000000000000000000 
         then 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2 
         else currency_contract 
       end as currency_contract
      ,amount
      from trades
)
left join prices.usd
    on blockchain = blockchain
    and contract_address = currency_contract
```
here we end up using the wrong price feed (WETH instead of ETH), this logic will break whenever there's a depeg event and we also end up with the wrong symbol in our end table. 

2. **Add extra logic in the join condition**
```sql
select
   blockchain
  ,tx_hash
  ,currency_contract
  ,amount
  ,price
from trades
left join prices.usd
on (blockchain = blockchain 
    and currency_contract = contract_address)
or (currency_contract = 0x0000000000000000000000000000000000000000
    and blockchain is null
    and symbol = 'ETH')
```
This works fine, but requiring blockchain `null` is confusing, and we're relying solely on the symbol column here to determine the price feed, which is a very unsafe column that can hold arbitrary data.
This is the current default that I've seen the most.
a note from the prices beta announcement reiterates the danger of relying on the symbol column:

[...]

PR #6651, opened 2024-09-03
Update balancer_cowswap_amm_ethereum_liquidity.sql

viniabussafi:
### Update!

### Contribution type

- [ ] New spell(s)
- [ ] Bug fix

---
Quick update to fix the protocol_liquidity_eth and pool_liquidity_eth fields on the balancer liquidity models as the former syntax (WHERE symbol = 'ETH') was causing undesired results, likely on either one of these results: 
2024-09-03 08:47 | solana | 0x0b931265156395f7491826c94b706e34717e5a9c60e6f668ff9301bbc347b8c1 | 8 | ETH | 0.00000884
-- | -- | -- | -- | -- | --
2024-09-03 08:47 | ethereum | 0x0d505c03d30e65f6e9b4ef88855a47a89e4b7676 | 18 | ETH | 0.00000884

Side note: I've seen that some nexusmutual spells also use this syntax, so tagging @tomfutago here for visibility

### DIFF
--- a/dbt_macros/shared/balancer/balancer_liquidity_macro.sql
+++ b/dbt_macros/shared/balancer/balancer_liquidity_macro.sql
@@ -74,7 +74,8 @@ WITH pool_labels AS (
             DATE_TRUNC('day', minute) as day,
             AVG(price) as eth_price
         FROM {{ source('prices', 'usd') }}
-        WHERE symbol = 'ETH'
+        WHERE blockchain = 'ethereum'
+        AND contract_address = 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2
         GROUP BY 1
     ),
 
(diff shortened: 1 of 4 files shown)


tomfutago:
> Side note: I've seen that some nexusmutual spells also use this syntax, so tagging @tomfutago here for visibility

hey @viniabussafi thanks for bringing this up! nexusmutual spells are using this type of syntax (hence I think they're fine for the time being):

```
  where symbol = 'ETH'
    and blockchain is null
    and contract_address is null
```

there's an open issue tho: #6577 with suggestions how price feed for native tokens might change, so worth following and I agree with point raised there that using WETH for ETH prices is not the best approach (risk of depeg events).

jeff-dude (approved):
yes, plz stay tuned for enhanced approach to pricing ideally released this month 🫡

PR #6678, opened 2024-09-05
`dex.prices_beta` initial build

jeff-dude:
few callouts:
- added `distinct` on pull from `dex.trades` to reduce row count written
- switched join to erc20 metadata to `inner join` to ensure we have decimal value to divide by (hardcoding 18 decimals may be inaccurate, as tokens can be various decimals)
- `union` rather than `union all` on bought / sold, in order to reduce rows written (adds some runtime on dbt side)

### DIFF
--- a/dbt_subprojects/dex/models/prices/_schema.yml
+++ b/dbt_subprojects/dex/models/prices/_schema.yml
@@ -1,6 +1,17 @@
 version: 2
 
 models:
+  - name: dex_prices_beta
+    description: >
+      obtain pricing info from dex.trades to feed into prices pipeline
+    tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - block_month
+            - blockchain
+            - contract_address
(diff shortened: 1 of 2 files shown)


jeff-dude:
note: at this time, code is filter to august 1st forward, to reduce run time. revert prior to merge.

jeff-dude:
moved away from merge with unique keys, as it is expected that tokens can be bought / sold on the same `block_time` and the prices differ v slightly across various scenarios. this is where we pick up the calculations to smooth out the price per minute (and then hour, day) in the next steps (@couralex6 testing now)

couralex6 (approved):

couralex6:
* Removed partition by `blockchain` to better suit clickhouse delta imports. 
* Filter out null values from symbol and amount which were also causing issues with clickhouse Delta imports

jeff-dude:
moved the filters earlier in the query, otherwise i think it's good to go

PR #6698, opened 2024-09-10
sushiswap arbitrum trades fix

### LINKED ISSUE #6687: [BUG] no data in dex.trades for sushiswap 2 on arbitrum
### Description

in the dex.trades, swaps that happened on sushiswap version 2 are not showing up .

this is due to the sushi_v2_arbitrum.Pool_evt_Swap table that is being pointed to in the spell is empty

https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/dex/models/trades/arbitrum/platforms/sushiswap_v2_arbitrum_base_trades.sql#L18

### Current behavior
no swap data because source table is null

### Expected behavior
data should be populated if source table has the decoded transactions

### Impacted model(s)

https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/dex/models/trades/arbitrum/platforms/sushiswap_v2_arbitrum_base_trades.sql#L18

### Possible solution

there is another table that decodes the swap transactions for sushiswap

```
select * FROM sushiswap_v3_pool_arbitrum.UniswapV3Pool_evt_Swap
limit 100
```

changing the source tables to this table above in the base trades model should work

Hosuke:
Attempt to fix #6687

### DIFF
--- a/dbt_subprojects/dex/models/trades/arbitrum/platforms/sushiswap_v2_arbitrum_base_trades.sql
+++ b/dbt_subprojects/dex/models/trades/arbitrum/platforms/sushiswap_v2_arbitrum_base_trades.sql
@@ -15,7 +15,7 @@
         blockchain = 'arbitrum',
         project = 'sushiswap',
         version = '2',
-        Pair_evt_Swap = source('sushi_v2_arbitrum', 'Pool_evt_Swap'),
+        Pair_evt_Swap = source('sushiswap_v3_pool_arbitrum', 'UniswapV3Pool_evt_Swap'),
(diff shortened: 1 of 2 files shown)


jeff-dude (approved):
interesting scenario here. appears that the swap decoded table was recreated and the old table name persisted -- but data was truncated -- so job continued to run on empty data. now the swap and factory table names are a bit out of sync too. might be worth asking the team exactly what happened here, i can raise it internally.

jeff-dude (approved):
the team is going to look closer at what happened, but i think we can merge with the tests passing

PR #6701, opened 2024-09-10
Polygon migration: from `MATIC` to `POL`

jeff-dude:
start simple:
- ensure metadata spells contain the update to symbol, as native address remains the same
  - prices, erc20, etc.
- ensure any downstream joins to metadata changed with hardcoded filters on symbol is updated


jeff-dude:
as mentioned in linked issue / PR description, starting simple here. there are a lot more hardcoded `MATIC` instances in spellbook, however, they are used in various scenarios throughout different lineages. 
- many tokens on other chains still use `MATIC` tied to the contract address
- looks like wrapped MATIC is still `WMATIC` on polygonscan, not `WPOL`
- the `0x0000000000000000000000000000000000001010` address stills [shows as MATIC](https://polygonscan.com/token/0x0000000000000000000000000000000000001010)

dbt_subprojects/tokens/models/prices/prices_native_tokens.sql
```diff
@@ -41,7 +41,7 @@ FROM
     ('kava-kava', null, 'KAVA', null, null),
     ('kuji-kujira', null, 'KUJI', null, null),
     ('ltc-litecoin', null, 'LTC', null, null),
-    ('matic-polygon', null, 'MATIC', null, null),
```
0xRobin, line 44: considering prices and the big impact on user queries, think we should add POL here and keep MATIC around for now as well

0xRobin:
>  we keep the old API id value and do one row per symbol. essentially just introduce a new way to join, but keep backwards compatability.

yes that should work for now, let's hope they add the full history to the POL feed in the future.

PR #6707, opened 2024-09-11
`ethereum_blobs` exclude from prod

jeff-dude:
fyi @lorenz234 @0xRobin @hildobby @MSilb7 @Jam516 
tagging as history shows all involved
this has started to fail in prod on the following error:
```
19:54:30    Database Error in model ethereum_blobs (models/_sector/blobs/ethereum/ethereum_blobs.sql)
  TrinoUserError(type=USER_ERROR, name=GENERIC_USER_ERROR, message="Task 20240911_193845_06449_mhikh.5.1.0 is failed, due to containing long running stuck splits.", query_id=20240911_193845_06449_mhikh)
```

i raised to the platform team to see if it's query engine related, but looks like it's new data coming through which is breaking the query planning.
first thought revolves around:
> The blobs table contains lots of fun UTF8ish data, right? And that file in particular has a very fun blob that appears to contain a single blob with 300,000 null characters

seems like lots of nulls coming through unexpectedly somewhere?

if we want this spell re-enabled, feel free to open PR to remove prod exclude tag and apply fixes as necessary (or continue the conversation on how to fix).

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/blobs/ethereum/ethereum_blobs.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/blobs/ethereum/ethereum_blobs.sql
@@ -1,4 +1,5 @@
 {{ config(
+    tags = ['prod_exclude'],
     schema = 'ethereum',
     alias = 'blobs',
     materialized = 'incremental',


Pluies:
I've managed to reproduce this on dune.com, see [query 4063379](https://dune.com/queries/4063379).

The:
```
,ceil(cast(length(regexp_replace(cast(blob as varchar), '0*$', '')) - 2 as double) /2 ) AS used_blob_byte_count
```

Is an absolute killer on one of these blocks (not sure which!). Getting the blocks themselves, without computing `used_blob_byte_count`, is instant.

belen-pruvost:
[Here](https://dune.com/queries/4063389) is an alternative for trimming empty characters in a performant, varbinary native way. You can find more data about these functions [in our docs](https://docs.dune.com/query-engine/Functions-and-operators/varbinary#varbinary-ltrim)

PR #6709, opened 2024-09-12
better formula for used blob bytes

0xRobin:
Follow up from https://github.com/duneanalytics/spellbook/pull/6707


0xRobin:
Cummulative distributions of `used_blob_byte_count` are identical, so calculation is consistent enough (it sometimes if off by 1 or 2 compared to the old calculation, which is negligible)
image
https://dune.com/queries/4064303/6843516
image
https://dune.com/queries/4064311/6843539

---
FYI @MSilb7 I think you originally wrote the formula for this.

jeff-dude (approved):
thanks for quick follow up 🙏

PR #6710, opened 2024-09-12
[CoW Protocol] Update mainnet Barter staging address

harisang:
This PR updates the mainnet Barter staging address in the CoW Protocol solver competition. PR #6699 introduced the same address for Barter both for the staging and the production environment of CoW Protocol, but since the intention was only to test that address in staging, that address is exclusively used now in production while a new one has been created by the Barter team for staging. This PR addresses this


harisang:
cc @bram-vdberg

harisang:
@jeff-dude Would it be possible to get this done today? Sorry for putting pressure on this.

jeff-dude (approved):
no problem, this is usually smooth to run in prod so i'm comfortable pushing quicker.

PR #6725, opened 2024-09-17
move RWA balances to daily subproject

0xRobin:
This PR moves RWA balances to the daily subproject. 
The data is on a daily timeframe anyway and the models are currently quite heavy to run hourly. 

FYI @darvinrio


darvinrio:
works @0xRobin

jeff-dude (approved):
thanks both

PR #6773, opened 2024-09-19
`prices_trusted_tokens` add native tokens as-is for now

jeff-dude:
for the new prices pipeline, we need to include native tokens. i wanted to do this in a way that didn't simply replicate approach in old pipeline, as we agreed to begin new design here.

@0xRobin i tried to follow your recommendation in #6577 -- hopefully i did it properly, otherwise we can fix later 🙏

### DIFF
--- a/dbt_subprojects/tokens/models/prices/prices_native_tokens.sql
+++ b/dbt_subprojects/tokens/models/prices/prices_native_tokens.sql
@@ -15,7 +15,7 @@ SELECT
 FROM
 (
     VALUES
-
+    --maintain legacy native token approach for historical queries
     ('ada-cardano', null, 'ADA', null, null),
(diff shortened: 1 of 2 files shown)


jeff-dude:
due to safeguards in place from OP bug in #6262 and then #6270 to avoid zero addresses, i have moved the logic around to handle for native tokens

jeff-dude:
i'm going to close this for now, as i know for sure i would have to modify referral rewards to remove the `OR` in join to prices to avoid duplicates. i also searched the repo and nft trades, gas fees may also end up with a conflict (didn't fully test). then we also know we have `prices.usd_native` temporary placeholder spell.

it's likely best we plan this together to finalize.

for now, i will pull through the old approach into trusted tokens in a separate PR.

PR #6781, opened 2024-09-20
prices ethereum: add new sky governance token

jeff-dude:


jeff-dude:
we are aware that the token is not active yet, will manually override the failed test on that.

PR #6801, opened 2024-09-22
fix epochs 

andrewhong5297:
incremental causing all sorts of issues, changing this to a full table refresh instead.


dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs_raw.sql
```diff
@@ -22,7 +22,7 @@ with
         LEFT JOIN {{ ref('solana_utils_epochs') }} epoch
             ON first_block_epoch = true --cross join
             {% if is_incremental() %}
-            AND {{incremental_predicate('epoch.block_time')}}
+            AND epoch.block_time >= date_trunc('day', now() - interval '7' day)
```
jeff-dude, line 25: the default `incremental_predicate` only looks back one day. rather than update the macro for this default, we can revert to a hardcoded approach to look back 7 days. since an epoch is ~2 days on average, this should be plenty of time to look up values needed

jeff-dude (approved):

PR #6811, opened 2024-09-23
adding token2022 support for orca swaps

andrewhong5297:
adding token2022, this is more difficult because the offset of transfer indices is not strictly known like with spl_token

### DIFF
--- a/dbt_subprojects/solana/models/_sector/dex/dex_solana_base_trades.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/dex_solana_base_trades.sql
@@ -13,6 +13,7 @@
 
 {% set solana_dexes = [
     ref('orca_whirlpool_base_trades')
+    , ref('orca_whirlpool_v2_base_trades')
     , ref('raydium_v3_base_trades')
     , ref('raydium_v4_base_trades')
     , ref('phoenix_v1_base_trades')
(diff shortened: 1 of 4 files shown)


andrewhong5297:
I've checked it works for token extensions like PyUSD. Note that this spell could potentially break in the future if some new kind of token extension comes out and changes the instruction invocation pattern.

https://solscan.io/tx/2P2hcK79CV82MqjXa6S4xxKoTHkgqmYkszWkxY1KmASvCek8YfSekSrB8rtdRJSTUYYyuCVxMXFrQd11bXDZqn6K

jeff-dude:
> Note that this spell could potentially break in the future if some new kind of token extension comes out and changes the instruction invocation pattern.

this scares me a bit. is there any way we can set up the spell to avoid failing in this scenario, but rather ignore swaps for new token extensions that change the pattern? then it'd be a new spell and/or fix here if it happens?

andrewhong5297:
> > Note that this spell could potentially break in the future if some new kind of token extension comes out and changes the instruction invocation pattern.
> 
> this scares me a bit. is there any way we can set up the spell to avoid failing in this scenario, but rather ignore swaps for new token extensions that change the pattern? then it'd be a new spell and/or fix here if it happens?

it will fail silently. it won't stop the spell from running

andrewhong5297:
lmk if you have other questions! orca team and blockworks team have been requesting this for a while now haha

jeff-dude (approved):
sorry for the delay. this looks good, follows the design from v1 as expected. thank you!

PR #6838, opened 2024-09-25
`dex.prices_beta` validate trade contains a trusted token

jeff-dude:
trying logic again, this time separate the joins to avoid `OR` within join conditions. try to check in the where clause.


jeff-dude:
hitting the memory limit per node still. i will need to keep trying other approaches.

jeff-dude:
closing this one now, to try and implement the latest logic agreed upon. we will revisit this if/when we get the other piece done.

PR #6846, opened 2024-09-26
`gas.fees`: enable celo for gas fees

jeff-dude:


dbt_subprojects/hourly_spellbook/macros/sector/gas/fees/gas_fees.sql
```diff
@@ -92,7 +83,7 @@ WITH base_model as (
         ,txns.gas_used as gas_used
         ,{{tx_fee_raw(blockchain) }} as tx_fee_raw
         ,{{tx_fee_breakdown_raw(blockchain)}} as tx_fee_breakdown_raw
-        ,{{fee_currency(blockchain)}} as tx_fee_currency
+        ,{{var('ETH_ERC20_ADDRESS')}} as tx_fee_currency
```
jeff-dude, line 86: during the fix, looks like blockchain team has cleaned up the raw table to match other chains and the `fee_currency` field no longer exists
```
select fee_currency, *
from celo.transactions
limit 10
```

jeff-dude: i don't think we can simply remove this here. our non-raw values will break without the decimal assignment, which requires the correct address to join to metadata. i have raised to the team internally again.

jeff-dude:
currently the test is failing for `tx_fee_raw` to equal the sum of parts within `tx_fee_breakdown_raw`

- `tx_fee_raw`: 
```
cast(gas_price as uint256) * cast(txns.gas_used as uint256)
````
- `tx_fee_breakdown_raw`:
```
map_concat(
    map()
          ,case when txns.priority_fee_per_gas is null or txns.priority_fee_per_gas < 0
                then map(array['base_fee'], array[(cast(gas_price as uint256) * cast(txns.gas_used as uint256))])
                else map(array['base_fee','priority_fee'],
                         array[(cast(base_fee_per_gas as uint256) * cast(txns.gas_used as uint256))
                                ,(cast(priority_fee_per_gas as uint256) * cast(txns.gas_used as uint256))]
                         )
                end)
 as tx_fee_breakdown_raw
```


`base_fee_per_gas` comes from `blocks` table. maybe that doesn't align with values in `transactions.gas_price`?

need to investigate further. celo docs:
https://docs.celo.org/protocol/transaction/gas-pricing

PR #6850, opened 2024-09-28
feat: add ape store spellbook

whalehunting:
This PR adds new spellbook for the ApeStore token deployer on Base + Ethereum networks.
It's a continuation of PR #6287 , which was closed due to inactivity (went on vacation)

### Contribution type
- [x] New spell(s)

---

### For new spell(s)

- **Spell name(s):** *ape_store.trades*
- **Description:** Deployments and trades for tokens deployed by [ape.store](https://ape.store/)
- **Who are the new spell(s) for?** General community
- **How will the new spell(s) be used downstream?** [Dashboard](https://dune.com/whale_hunter/ape-store)
- **Test instructions:** Seed tests
- **Related issue(s):** None


whalehunting:
@jeff-dude i tried to implement [your request](https://github.com/duneanalytics/spellbook/pull/6287#discussion_r1681565340) in [this commit](https://github.com/duneanalytics/spellbook/pull/6850/commits/ee9c6c44c5ccbc779dfb9f360581daf1b027974c), hope that is what you meant :)

dbt_subprojects/daily_spellbook/models/ape_store/base/ape_store_base_trades.sql
```diff
@@ -0,0 +1,110 @@
...
+            tx_from as user,
+            trades.tx_hash,
+            evt_index as tx_index
+        from {{ source('dex', 'trades') }} as trades
+        join
+            deployments
+            on (
+                trades.blockchain = deployments.blockchain
+                and trades.block_month >= deployments.block_month
+                and (token_bought_address = token or token_sold_address = token)
```
jeff-dude, line 95: this `or` condition really messes with the engines query planning and doesn't run efficiently. i actually ran into the same problem this week, but finally found logic that works more efficiently. you can see here:
https://github.com/duneanalytics/spellbook/pull/6847/files#diff-5dceb884deab38e05245be5fa15b51577bf1a3a220ec6b7c07d27f3e4ce2e52cR52

jeff-dude: same idea below on ethereum

jeff-dude (changes requested):

dbt_subprojects/daily_spellbook/models/ape_store/base/ape_store_base_trades.sql
whalehunting: so should i change it to:
```sql
-- ...
and contains(ARRAY[token_bought_address, token_sold_address], token)
```
?

jeff-dude:
i'm still a bit confused on how base can take so much more time than ethereum. i can see the decoded event swap for base is much larger, ~2.2mil vs. ~1.7k rows, but in general that 2.2mil number shouldn't be much of an issue to process anyway. the incremental run is also taking just as long as the historical run, which is usually a red flag.

can we break down the base chain query to find out what is causing the slowdown? is it the event swap decoded table left joining a large prices dataset? is it the dex trades read and join to deployments?

whalehunting:
@jeff-dude I'm like 90% sure it's the dex.trades joined on deployments. The normal swap part is running fine on the dune dashboard itself, without any spellbooks:
https://dune.com/whale_hunter/ape-store

Base was live a few months before Ethereum, so makes sense that it has more data. Ethereum has little activity rn, also pretty clear when looking at the dashboard

jeff-dude:
> @jeff-dude I'm like 90% sure it's the dex.trades joined on deployments. The normal swap part is running fine on the dune dashboard itself, without any spellbooks: https://dune.com/whale_hunter/ape-store
> 
> Base was live a few months before Ethereum, so makes sense that it has more data. Ethereum has little activity rn, also pretty clear when looking at the dashboard

gotcha, that is what i figured. since the deployments token list is relatively large (at least compared to token list used in my other spell example), that may end up making the array / contains approach slow again.

is it possible to read from `dex.trades` twice, both inner join to deployments on token (once bought, once sold) and then union back together?

whalehunting:
> 
> gotcha, that is what i figured. since the deployments token list is relatively large (at least compared to token list used in my other spell example), that may end up making the array / contains approach slow again.
> 
> is it possible to read from `dex.trades` twice, both inner join to deployments on token (once bought, once sold) and then union back together?

yes good idea, will try this out and report back

whalehunting:
@jeff-dude that run was a lot quicker than the previous ones 🔥 

hope you're happy with the result - so now we can revert the date filters (marked with a 2x TODO comments) and deploy the spellbook to production 👀

jeff-dude:
> @jeff-dude that run was a lot quicker than the previous ones 🔥
> 
> hope you're happy with the result - so now we can revert the date filters (marked with a 2x TODO comments) and deploy the spellbook to production 👀

blazing fast ⚡ 
just reverted to original dates, let's see it happen

jeff-dude:
good news: still v fast

bad news: some dupes in base chain model. maybe due to `union all` rather than `union`?

whalehunting:
@jeff-dude i think it figured it out, this happens when arbitrage/mev bots are swapping directly between 2 ape.store tokens, so the trade get's included in both "dex buys" and "dex" sells. any idea how this can be prevented?

Duplicates:
https://dune.com/queries/4170904?sidebar=none

jeff-dude (approved):
nice!

all the proof we need that the query engine really hates `OR` on address matching, whether join conditions or filters.

do you want to revert the additional unique keys added? or leave as-is? i am marking ready for merge, and can do either way

jeff-dude:
merging with a few other PRs, i think it's fine either way

PR #6899, opened 2024-10-07
Update staking_solana_validator_stake_account_epochs_raw.sql

smyyguy:
I noticed the lookback was bumped from 2 to 7 days, but it seems the same issue is still occurring where it does not recognize a new epoch has began and thus does not run the incremental model. I think this needs a full refresh every time in that case

### Update!

### Contribution type

- [ ] New spell(s)
- [x ] Bug fix

I noticed the lookback was bumped from 2 to 7 days, but it seems the same issue is still occurring where it does not recognize a new epoch has began and thus does not run the incremental model. I think this needs a full refresh every time in that case

### DIFF
--- a/dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs_raw.sql
+++ b/dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs_raw.sql
@@ -1,9 +1,8 @@
 {{ config(
     schema = 'staking_solana'
     , alias = 'validator_stake_account_epochs_raw'
-    , materialized = 'incremental'
+    , materialized = 'table'
     , file_format = 'delta'
-    , incremental_strategy = 'merge'
     , unique_key = ['epoch', 'epoch_time', 'stake_account', 'vote_account']
 )
 }}
@@ -21,9 +20,6 @@ with
         FROM {{ ref('staking_solana_stake_account_delegations') }} vote
         LEFT JOIN {{ ref('solana_utils_epochs') }} epoch
             ON first_block_epoch = true --cross join
-            {% if is_incremental() %}
-            AND epoch.block_time >= date_trunc('day', now() - interval '7' day)
-            {% endif %}
         WHERE vote.block_slot < epoch.block_slot --only get changes to accounts before start of epoch
         GROUP BY 1,2,3,4,5
     )


smyyguy:
lmk if this makes sense to you @jeff-dude

jeff-dude:
> lmk if this makes sense to you @jeff-dude

ideally we can find a way to remain incremental. can you remind me with what exactly the issue is, whether it's this spell directly or spell(s) downstream of this? can you share a query to show the issue?

smyyguy:
Only the current epoch should be NULL, but it is populating NULL values for epoch_next_start_slot after the last full run which was andrew's patch about two weeks ago. Prior to that, the same issue occurred after epoch 652 which was when the model first launched. 

solana_utils_epochs seems to be properly populating the values, so im not sure why the cross join returns null for these periods. I believe its that solana_utils.epochs is a full table refresh each run, so when new epochs start, epoch_next_start_slot gets populated for previous epochs. However, this other spell reads from solana_utils.epochs but filters incrementally, so it doesn’t find when an epoch ends and get the value to pull forward. 

Also recreating the table in dune seems to work fine, the only part i removed was the incremental_predicate -- is this causing the problem?

Here is an example for the most recent epoch, 680
with 
    base as (
        SELECT 
            epoch.epoch
            , epoch.block_time as epoch_time
            , epoch.epoch_start_slot
            , epoch.epoch_next_start_slot
            , vote.stake_account
            , max_by(vote.vote_account, vote.block_time) as vote_account --latest delegated vote before epoch started. then get the rewards from accounts per epoch.
        FROM staking_solana.stake_account_delegations vote
        LEFT JOIN solana_utils.epochs epoch
            ON first_block_epoch = true --cross join
        WHERE vote.block_slot < epoch.block_slot --only get changes to accounts before start of epoch
            AND epoch.epoch = 680
        GROUP BY 1,2,3,4,5
    )

SELECT
    epoch
    , epoch_time
    , epoch_start_slot
    , epoch_next_start_slot
    , stake_account
    , vote_account
FROM base
WHERE vote_account is not null

smyyguy:
this table impacts dune.dune.result_solana_validator_epoch_revenues (which i believe is a mat view created by your team)

jeff-dude:
gotcha. i can see ~150 null values for next slot column on non-current epochs. i don't fully understand how that's possible to be honest, as the incremental window to search is 7 days back, while the min/max combo of block_time for each epoch is two days. that's why we tried to raise it to 7 days.

is it possible this other spell it reads from (before left joining epochs spell) has late arriving data or something? or the filter for `vote.block_slot < epoch.block_slot` make the join miss results?

we can probably end up making this a full table, but it would still be good to know how or why this is happening, as it could impact data quality in general

smyyguy:
Honestly I cannot figure it out either. I would have thought Andrew's PR to change from 2 to 7 days would have solved this

PR #6907, opened 2024-10-08
DUX-699 remove old spark code from nft.transfers macro

0xRobin:
remove old spark code from nft.transfers macro


0xRobin:
verified that data is here now :heavy_check_mark: 
```
select count(*) from test_schema.git_dunesql_af871d0_nft_polygon_transfers
where amount > pow(2,32) and token_standard = 'erc1155' and transfer_type = 'batch'
```

PR #6916, opened 2024-10-09
DUX-712: add support for xDAI native price feed (as alternative for XDAI)

0xRobin:
add support for xDAI native price feed (as alternative for XDAI, all caps).
I'm keeping the original in there as well to not break any queries.

This should resolve the null gas fees on gnosis.
evms.info has `xDAI` and prices.native_tokens had `XDAI`, which resulted in the null values for gnosis.


0xRobin:
after merge and full refresh, we should refresh `gas_gnosis_fees` to propagate the changes.

jeff-dude (approved):
works for me -- feel free to merge when you see fit 🤝

PR #6923, opened 2024-10-10
Query and doc fixes for first_funded_by + add a few columns

hildobby:
This PR:
- fixes some documentation issues given that not all native tokens are ETH
- fixes some query issues not factoring in `trace_address`, an essential inclusion for txs funding multiple addresses
- added `token_symbol`, `trace_address`, `amount` and `amount_usd` to the crosschain spell:
  - `token_symbol` is necessary given not all EVM chains use ETH as their native currency
  - `trace_address` is must to easily and efficiently join this spell onto other spells and remain accurate
  - `amount` & `amount_usd` are very useful for sleuthing stuff like CEX deposit addresses (they're usually funded in tiny amounts of native token)


jeff-dude:
can you help validate if these seed errors are real?

hildobby:
> can you help validate if these seed errors are real?

no those are due to the date range, those addresses here first appear after the date filter although they appeared before then, removing the filter will fix it!

jeff-dude (approved):

PR #6929, opened 2024-10-11
Refactor evms.contracts

Hosuke:
Make [`blockchain`, `address`] unique in final `evms.contracts`.


Hosuke:
[evms.contracts preview](https://dune.com/queries/4150578)

dbt_subprojects/daily_spellbook/models/evms/evms_contracts.sql
```diff
@@ -32,25 +32,29 @@
...
+            , "from"
+            , code
+            , name
+            , namespace
+            , dynamic
+            , base
+            , factory
+            , detection_source
+            , created_at
+            , row_number() over (partition by address order by created_at desc) as duplicates_rank
```
jeff-dude, line 52:
```suggestion
            , row_number() over (partition by blockchain, address order by created_at desc) as duplicates_rank
```

jeff-dude: if we're unique on these two fields, don't they both need added to partition in row number?

jeff-dude (changes requested):

dbt_subprojects/daily_spellbook/models/evms/evms_contracts.sql
Hosuke: You are right, I did the partition by address before union by chain.
But would be safer to partition by blockchain and address together.

dbt_subprojects/daily_spellbook/models/evms/evms_contracts.sql
```diff
@@ -32,25 +32,29 @@
...
+            , "from"
+            , code
+            , name
+            , namespace
+            , dynamic
+            , base
+            , factory
+            , detection_source
+            , created_at
+            , row_number() over (partition by blockchain, address order by created_at desc) as duplicates_rank
```
Hosuke, line 52: Sorry, at the partition moment, the `blockchain` feild is not existed yet.

jeff-dude:
understood on address being unique per chain level contracts table, thank you for clarification.

however, now that we do this approach, we should remove `created_at` from unique key in model. then we should be all good to go.

jeff-dude (approved):

jeff-dude:
will run this in prod in a few minutes, should refresh in next 20-30 minutes hopefully

PR #6938, opened 2024-10-12
Init worldchain

Hosuke:
Init worldchain base tables.


jeff-dude (approved):
thank you!

once we get all of these worldchain spells live, we should query `dex.trades` for `worldchain` and find most frequently traded tokens (bought or sold), then revisit adding more trusted tokens if needed

PR #6939, opened 2024-10-13
Create `get_explorers()` udf

hildobby:
Here's an example of the output idea: https://dune.com/queries/4156218/6995640


0xRobin:
@hildobby unfortunately, since migrating to DuneSQL, we aren't able to define UDFs in SQL from within the spellbook repo. :disappointed:

0xRobin:
@hildobby 
These functions have been re-implemented in the backend, and are not pulled from this repo anymore. The team is looking into ways to allow user contributions to UDFs, but it's not high priority at the moment I'm afraid. 
I think your best bet is to try and get this working with a parameterized view on Dune itself.

hildobby:
> @hildobby These functions have been re-implemented in the backend, and are not pulled from this repo anymore. The team is looking into ways to allow user contributions to UDFs, but it's not high priority at the moment I'm afraid. I think your best bet is to try and get this working with a parameterized view on Dune itself.

Sadly a parameterised view wouldn't work here since those only accept 1 parameter per call, I can't run it on every row :/

PR #6962, opened 2024-10-16
metrics spells: move from hourly to daily

jeff-dude:
towards DUX-722
- move from hourly to daily, since all are aggregated to day anyway
- establish cross-project sources
- update refs to sources
- move remaining `evms` metric spells (older iteration) to one parent directory above, removing `/evms/`, but remain hourly since at `block_hour` level

a bit riskier change, but curious on including here (fine without too):
- increase the time period to 3 days on daily project for incremental reads, to handle any issues that may persist over a weekend
  - curious if this has large impact on run times, i have a feeling not really, but is much safer around data quality and not missing any rows


jeff-dude:
> * increase the time period to 3 days on daily project for incremental reads, to handle any issues that may persist over a weekend

reason i thought about this:
our hourly jobs look back a full day. that's 24x the amount needed, but helps with lagging source data and generally is fast enough for our needs.
our daily job also looks one full day back, that's 1x the amount needed, introducing some more risk

jeff-dude:
manually canceling CI, as i know transfers metrics spells will timeout / crash and there are no changes to code here other than incremental look back period

0xRobin (approved):
+1 on the incremental time increase for the daily job.

PR #6970, opened 2024-10-18
create `tokens_{chain}.info`

hildobby:
This can be considered as an improved version to `tokens_{chain}.erc20` which imo is poorly named, lacks native tokens and a lot of other columns.

A crosschain spell doesn't make sense for now I will eventually work on a spell that maps tokens across different chains (ie all USDC addresses across chains). Once that's in place, we can combine it with the results from this spell to get a great crosschain spell.

For downstream usage, this will come in handy to rank tokens based on heuristics to filter out spam tokens from the rest for example. That will in turn be useful for spells like: `attacks.address_poisonning`


0xBoxer:

jeff-dude:
going through another cleanup effort on PRs. if reopen is necessary, plz do one at a time until finalized and merged 🙏

PR #6975, opened 2024-10-18
Nexus Mutual - deprecate old spells & convert v1 views to static tables

tomfutago:
### Contribution type

- [ ] New spell(s)
- [ ] Bug fix

---

- **Description:**
  - deprecate old unused spells: `capital_pool_eth_daily_transaction_summary` & `nexusmutual_ethereum_capital_pool_eth_daily_transaction_summary`
  - convert `claims_v1` & `covers_v1` views to static tables - no new records expected (assuming materialized as 'table' and 'static' tag will do the trick here?)

---


dbt_subprojects/daily_spellbook/models/nexusmutual/ethereum/claims/nexusmutual_ethereum_claims_v1.sql
```diff
@@ -2,12 +2,13 @@
   config(
     schema = 'nexusmutual_ethereum',
     alias = 'claims_v1',
-    materialized = 'view',
+    materialized = 'table',
+    tags = ['static'],
```
jeff-dude, line 6:
```suggestion
```

jeff-dude: you should remove static, as it reads from tables. we use `static` when the spell is just hardcoded strings and no table reads

dbt_subprojects/daily_spellbook/models/nexusmutual/ethereum/covers/nexusmutual_ethereum_covers_v1.sql
```diff
@@ -2,11 +2,9 @@
   config(
     schema = 'nexusmutual_ethereum',
     alias = 'covers_v1',
-    materialized = 'incremental',
-    file_format = 'delta',
-    incremental_strategy = 'merge',
+    materialized = 'table',
+    tags = ['static'],
```
jeff-dude, line 6:
```suggestion
```

jeff-dude (changes requested):

dbt_subprojects/daily_spellbook/models/nexusmutual/ethereum/claims/nexusmutual_ethereum_claims_v1.sql
tomfutago: ah cool, so if all I want is to have this data as a table which is loaded once and just sits there - `materialized = 'table'` is all I need?

jeff-dude: we have removed usage of `static` tag in our prod orchestration atm. you could actually leave in then if you want, so if ever in the future we want to revisit adding, we would know this model can be static

tomfutago: good to know, let's leave it it then

jeff-dude (approved):

PR #6977, opened 2024-10-18
Update Solana token transfer models and add prices

0xBoxer:
Updated Solana token transfer models to improve consistency and performance across SPL, Token2022, and SOL transfers. Added price information to enhance token transfer data.


dbt_subprojects/solana/models/tokens/solana/tokens_solana_transfers.sql
```diff
@@ -76,7 +126,55 @@ SELECT
...
 LEFT JOIN {{ ref('solana_utils_token_accounts') }} tk_d ON tk_d.address = tr.to_token_account
 WHERE 1=1
--- AND call_block_time > now() - interval '90' day --for faster CI testing
\ No newline at end of file
+{% if is_incremental() %}
+AND {{incremental_predicate('block_time')}}
+{% endif %}
+{% if not is_incremental() %}
+AND block_time > now() - interval '7' day
+{% endif %}
```
Hosuke, line 137: Do we need to remove those `if not is_incremental()` clauses?

0xBoxer: ye definitely, just still testing this is to make sure the trial runs don't time out

0xBoxer:
@0xRobin @jeff-dude this is good to go from a logic perspective, but the job has memory limit problems for 30 days.
Any hints from past experiences how we could solve this/move this forward?

0xRobin:
I would set the partition to `block_date` instead of `block_month`, that should already limit the data a bit if read by partition.
And we probably want to adapt the `delete+insert` merge strategy here as well? 
That should be interchangeable with the `merge` strategy without modifying the logic

dbt_subprojects/solana/models/tokens/solana/tokens_solana_transfers.sql
```diff
@@ -4,25 +4,28 @@
         alias = 'transfers',
         materialized = 'incremental',
         file_format = 'delta',
-        incremental_strategy = 'merge',
+        incremental_strategy = 'delete+insert',
         partition_by = ['block_date'],
-        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')],
+        incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_date')],
```
0xRobin, line 9: @0xBoxer apparently the `DBT_INTERNAL_DEST` prefix is specific to the `merge` strategy. Just using `block_date` should work :pray:

0xBoxer: aha, thanks 🙏

0xBoxer:

0xBoxer:
CI runs for 30 days of history now.

This will modify the existing `solana_transfers` spell, we should have a quick think about that, but IMO this is how the model should have been like from the start and is more akin to our other `blockchain_transfers` spell, we should just bite the bullet and do the switch here.

The final output is currently defined as a view that would respond well to filtering on `token_version` or any `block_date` filter. 

querying the data via test schemas in dune seems fine to me, trino is really good at doing it's predicate pushdown thing.

 here's a matview of net transfers based on this: https://dune.com/queries/4198707

0xBoxer:
can probably be closed in favour of https://github.com/duneanalytics/spellbook/pull/7074

PR #6979, opened 2024-10-18
exclude solana gas fees

0xRobin:
exclude solana gas fees


0xRobin:
@0xBoxer incremental run is taking ~50min on this and delaying the hourly job :disappointed: 
We'll have to revisit next week to see if we can optimize it further.

PR #6982, opened 2024-10-21
re-enable solana fees

0xRobin:
solana fees performance testing

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/solana/gas_solana_fees.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/solana/gas_solana_fees.sql
@@ -5,6 +5,7 @@
     materialized = 'incremental',
     file_format = 'delta',
     incremental_strategy = 'delete+insert',
+    incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_date')],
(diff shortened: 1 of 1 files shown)


0xRobin:
Result of some testing with different incremental strategies:
Here's a markdown table of the results with the run column including the links:

| Method | Initial Run | Incremental Run | Run Link |
|--------|-------------|-----------------|----------|
| Append | 2m10s | 36s | [Run](https://github.com/duneanalytics/spellbook/actions/runs/11439343213/job/31822804183) |
| Delete+Insert | 1m48s | 3m51s | [Run](https://github.com/duneanalytics/spellbook/actions/runs/11439917071/job/31824636814) |
| Merge | 1m42s | 33m11s | [Run](https://github.com/duneanalytics/spellbook/actions/runs/11440094755/job/31825182092) |

Note that all the initial runs are the same exact SQL so any time difference is just random or due to cluster traffic.

I'm currently leaning into `delete+insert` as it still has a low enough runtime without requiring custom logic. `append` is the best performant, but requires custom logic and is very susceptible to duplicates or incomplete data making it into the prod table.

0xBoxer (approved):
I was wondering if something like that was possible, thanks for making this work 🙏

jeff-dude:
i can try and give this a merge today

PR #6983, opened 2024-10-21
Fix worlcoin safes

Jam516:
The current safe worldcoin spell assumes that activity started in October. In reality, activity started in July, and millions of Safes were deployed in July and August. Without this change, any query using the worldcoin safes table will be inaccurate


jeff-dude (approved):

0xBoxer:

PR #7001, opened 2024-10-23
Upgrade dbt version to 1.8

jeff-dude:
towards DUX-635
this PR will be a bit noisy due to tests change in latest version. summary of changes below:

- upgrade to dbt v1.8 ([here](https://docs.getdbt.com/docs/dbt-versions/core-upgrade/upgrading-to-v1.8#resources))
- loosely follow the guide [here](https://docs.getdbt.com/blog/upgrade-dbt-without-fear#step-2-add-require-dbt-version-to-your-dbt_projectyml-file)
- [this](https://docs.getdbt.com/docs/dbt-versions/core-upgrade/upgrading-to-v1.8#new-data_tests-syntax) change was outputting warnings on all projects, so i universally changed `tests:` to `data_tests:`
- changed dbt version in pipfile
- added minimum version requirement in each dbt project file to ensure local users aren't on outdated version


jeff-dude:
not 100% sure what kicked off the daily CI job here, as clicking into files changed crashes my browser 🫠 

we shouldn't expect any impact to models and the current CI test failure is fine, due to model not building as it would expect.

jeff-dude:
to be safe, we may want to merge this with modified turned off. maybe not necessary, as other subprojects didn't kick off. either way, we'll want to monitor closely post merge and likely isolate it.

jeff-dude:
merging now, as it's blocking all other PRs from running (due to forcing warnings to error out in compile step)

PR #7014, opened 2024-10-24
Update ethereum_blobs.sql

shlee-lab:
Removed 'ltrim' while retaining 'rtrim'.

In blob raw data, padding with 0 bytes occurs only on the right side, while left-side 0 bytes are part of the actual data and included in the compression algorithm. Therefore, only 'rtrim' should be applied, without using 'ltrim'.

### Update!

### Contribution type

- [ ] New spell(s)
- [v ] Bug fix

### For bug fixes

The issue was that both ltrim and rtrim were used to remove 0 bytes from blob raw data, but only right-side 0s are padding. Left-side 0 bytes are part of the actual data, so applying ltrim was incorrect.

.

I removed the use of ltrim and kept rtrim to ensure that only right-side padding is removed, while preserving left-side 0 bytes as part of the actual data.

1. Apply the fix and process blob raw data.
2. Check that only right-side padding is removed and left-side 0 bytes are retained.
3. Verify that the compression algorithm includes left-side 0 bytes.

---

### Additional information

None
---

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/blobs/ethereum/ethereum_blobs.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/blobs/ethereum/ethereum_blobs.sql
@@ -21,7 +21,7 @@ SELECT
     , b.index AS blob_index
     , b.proposer_index AS beacon_proposer_index
     , b.kzg_commitment AS blob_kzg_commitment
-    ,bytearray_length(varbinary_ltrim(varbinary_rtrim(blob))) as used_blob_byte_count
+    ,bytearray_length(varbinary_rtrim(blob)) as used_blob_byte_count
     ,bytearray_length(blob) AS blob_byte_count
     ,bytearray_length(blob) AS blob_gas_used
     -- GPT to the rescue


shlee-lab:

0xRobin:
@shlee-lab from my understanding this is completely up to the blob poster, as you pay for a full blob of bytes either way.
Do you have an example of where this is incorrect?

shlee-lab:
@0xRobin Basically it's not incorrect. Actually, it was ambiguous to check the bug category but I didn't find the proper category. I'm suggesting for improvement. 

Now, let's talk about data. The blob_used_data indicates the actual data used within the blob. Therefore, we must ensure accuracy when trimming padded zeros. For instance, consider the data '00012345'. While the right side may be padded with 000000...000000 and those zeros can be safely removed, if we also remove the left-side 0s, the data would incorrectly appear as 12345, which is not the correct representation of the original data. Although left-side 0 padding may exist in certain cases, to maintain consistency and minimize distortions in actual data usage, it's crucial to only remove right-side padding. This ensures we report the true amount of data used without altering its integrity.

0xRobin:
@shlee-lab I will close this as we are more comfortable with the current approach.

We would rather understate the usage by a small number of leading zero-but-useful bytes then risk overstating the data usage by a lot when different conventions are used. 
Consider the bytes used column as a lower limit estimate to the actual bytes used.

PR #7023, opened 2024-10-24
need to make this table also refresh fully not incremental

andrewhong5297:
related to some of the other issues with validator_stake_account_epochs_raw, this downstream table needs to also be changed from incremental to full run.

### DIFF
--- a/dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs.sql
+++ b/dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs.sql
@@ -2,10 +2,8 @@
     schema = 'staking_solana'
     , alias = 'validator_stake_account_epochs'
     , materialized = 'incremental'
-    , file_format = 'delta'
-    , incremental_strategy = 'merge'
+    , file_format = 'table'
     , unique_key = ['epoch', 'epoch_time', 'stake_account', 'vote_account']
-    , incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.epoch_time')]
     , post_hook='{{ expose_spells(\'["solana"]\',
                                 "sector",
                                 "staking",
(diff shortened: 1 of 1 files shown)


dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs.sql
```diff
@@ -2,10 +2,8 @@
     schema = 'staking_solana'
     , alias = 'validator_stake_account_epochs'
     , materialized = 'incremental'
-    , file_format = 'delta'
-    , incremental_strategy = 'merge'
+    , file_format = 'table'
```
jeff-dude, line 5:
```suggestion
    , materialized = 'table'
```

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_validator_stake_account_epochs.sql
```diff
@@ -2,10 +2,8 @@
...
     , unique_key = ['epoch', 'epoch_time', 'stake_account', 'vote_account']
```
jeff-dude, line 6:
```suggestion
```

jeff-dude (changes requested):

jeff-dude:
> related to some of the other issues with validator_stake_account_epochs_raw, this downstream table needs to also be changed from incremental to full run.

can you better help me understand why these don't work incrementally? i struggled to see when looking a bit myself

andrewhong5297:
my best guess is it is because [in this model](https://github.com/smyyguy/spellbook/blob/18d41eb7d24ed63177c316f282e07b1c899ee7a4/dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_stake_account_delegations.sql) we are getting all delegated stake accounts, but that can happen before/during an epoch. So then we are getting a bunch of stake accounts that are delegated and already added into the row, that aren't getting captured again in an incremental? 

honestly stumped, I tried looking at the full dag but didn't find anything obvious and it's slow to test because it works when you run full so I don't have a way of testing to see where the incremental went wrong (at least in the dune ui).

PR #7026, opened 2024-10-24
Bump polygon balances version to `0002`

couralex6:


jeff-dude (approved):
looking at polygon balance spells [here](https://github.com/duneanalytics/spellbook/tree/main/dbt_subprojects/tokens/models/transfers_and_balances/polygon).

looks like we would need to refresh [this](https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/tokens/models/transfers_and_balances/polygon/tokens_polygon_balances_daily_agg_base.sql) one, as it's materialized, and others are views.

i don't remember off the top of the head if a source change merged into prod will trigger state modified for models using it.

jeff-dude:
CI won't run here due to only source change

PR #7032, opened 2024-10-25
fix: correct base fee calculation logic for solana transactions

0xBoxer:
This PR updates the logic for calculating base fees on Solana transactions to ensure accurate fee computation.


jeff-dude:
marking WIP until we determine our path forward

jeff-dude:
closing this, as it should be fixed in the new macro

PR #7034, opened 2024-10-25
feat: creating macros to be able to split out backfill runs

0xBoxer:
This PR introduces new macros that enable splitting out backfill runs for more efficient data processing.


0xBoxer:
kind of expected this to not fully run through, but none of the parts finishing does seem a bit sketch, do we have to go down to monthly granularity to chop this up into right sized pieces or can can the prod cluster potentially handle this on merge?

jeff-dude:
> kind of expected this to not fully run through, but none of the parts finishing does seem a bit sketch, do we have to go down to monthly granularity to chop this up into right sized pieces or can can the prod cluster potentially handle this on merge?

it's a tough balance to find. factors we know:
- CI cluster is smaller (5 nodes, smaller size machines each)
- we now run 8 models in parallel, rather than cluster focusing on one model, sharing resources across all 8

that means we have 8 quarters running at same time, i.e. 2 years of data. it's still a lot when in parallel. we can lower the thread count and increase cluster size to see how it goes.

i was considering scaling up CI to test, but there's no reason to run so heavy both CI and then prod after merge. we should just try in prod again

0xBoxer:
wait a second why do I not trigger any incremental runs on earlier attempts 🤔

looking into that now, think I have to pass parameters even though they won't be applied since logic lives in incremental

0xBoxer:
nvm I just had `prod_exclude` tags that carried over from trying to merge this

0xBoxer:
I am unsure about the materialization of tables here. 

- I materialize all quarterly models and give them the `[static]` tag
- union of all quarterly models is a view
- combination of incremental + backfill is materialized
- that means that we'd materialize the quarterly models again, I am not sure how bad that is. 

 maybe better to build a materialized incremental model with just the past few days and create a view between the materialized submodels and the backfill instead?

jeff-dude:
okay let me try to summarize what i pushed and why:
- simplify macros to use `else` for non-incremental
- add partitions to all tables, even if full table and static
- rename current quarter to `_current` that will need to be manually maintained each quarter until we revisit and finalize design on these
- make current incremental type, so it triggers incremental clause in macros
- remove backfill union all views
- make final spells views on top of static tables + current incremental

i believe this matches what our original convo was for the design. when 2025 Q1 comes around, we will need to push a PR to add a static spell for 2024 Q1 and update date params for the `_current` model (start date at least).

i probably messed something up logic wise and have to step away for the night. if it fails, should be able to walk through the bug fixes and continue on.

let me know what ya think of this approach

jeff-dude:
> wouldn't we only need to do this once we have to re-backfill?

technically yes, but it would be nice to have the models in place and ready to go so we aren't confused where we stand. i guess it's either:
- each quarter, be proactive and update current model with new date params, move old current model to a static full table for prior quarter
- don't do anything, if a backfill is required for whatever reason, catch up on missing quarters in the models and build all at that time

jeff-dude:
latest commit forces each quarterly model to run only start date + 1 day, to see end-to-end in CI pipeline.

the good news: it ran successfully

the bad news: data volume is insane, as expected

some numbers:
- `solana.transactions` has 60,294,113,546 total rows for all time
- seems like ~75mil rows/day is about the max here
- at high end, that is 75mil rows/day * 90 days in a quarter = 6.7bil rows/quarter model
- `solana.vote_transactions` has 298,032,640,890 total rows for all time
- seems like ~300mil rows/day is about the max here
- at high end, that is 300mil rows/day * 90 days in a quarter = 27bil rows/quarter

jeff-dude (approved):

PR #7035, opened 2024-10-25
upstream solana gas fees spells: move to solana subproject

jeff-dude:


0xBoxer:
ready whenever you are

0xBoxer (approved):

jeff-dude:
> ready whenever you are

if we do this in multiple steps, i will need to update your leftover spells in hourly to use source rather than ref here. then when we move those too, we will have to revert back to refs

0xBoxer:
aren't those all prod excluded?

I can change my refs to sources or move my project to solana folder in my branch

0xBoxer:
we can also do it all at once, however you prefer

jeff-dude:
okay, let's move full lineage here. then you current working branch will need to adapt that we have moved them to solana.

0xBoxer:
yep I'll merge main after this is done and move to solana in my branch

jeff-dude:
both projects are compiling = this is good for me. i will turn off modified on each and let it rip. we are slightly blocked on hourly job from another model failing, so i will need to resolve that first

jeff-dude:
> yep I'll merge main after this is done and move to solana in my branch

if you don't want to wait, you could also merge this feature branch into yours to expedite the main branch merge and get a headstart. then main should be clean.

PR #7065, opened 2024-10-29
fix: temporarily exclude polymarket daily prices model from data explorer

0xBoxer:
excluding daily prices model because of unknown bug, no time to investigate, hourly and latest models work fine


jeff-dude (approved):

jeff-dude:
not sure this will actually do anything in prod fyi.
since it's a view, it doesn't run anyway. once the model ran once to build view, post hook already ran and applied. not sure if it'll remove from DE or not. we have (or had) an external script which removed spells in this scenario from DE, but it might be paused. i can't recall exactly tbh.

PR #7074, opened 2024-10-30
feat: restructure solana token transfers with backfills

0xBoxer:
Macro Creation: Defined new macros for each transfer type to streamline code and standardize transfer patterns.

- solana_sol_transfers_macro
- solana_spl_transfers_macro
- solana_spl_transfers_call_transfer_macro
- solana_token22_spl_transfers_macro

Quarterly Backfill Models: Implemented quarterly backfill models spanning 2020 Q4 to 2024 Q3, covering:
- SOL transfers
- SPL transfers
- SPL call transfers
- Token2022 transfers

Model Updates: Revised current models to utilize the newly created macros.

Unified Views: Established comprehensive views to combine backfills and current models for a streamlined data view.

PR #7076, opened 2024-10-31
metrics: initiate solana transfers

jeff-dude:
~~dependent on #7074 merged & fully live in prod.~~

initial thought is to keep solana transfers separate for two reasons:
- data volume -- the net transfers for EVM is already compute intensive, adding solana will be difficult
- lack of labels relative to EVMs, leading us to remove join to labels on solana side here and aggregating to address only instead


dbt_subprojects/daily_spellbook/models/_metrics/transfers/metrics_net_solana_transfers.sql
```diff
@@ -0,0 +1,113 @@
...
+
+with raw_transfers as (
+    /*
+    - get the transfer amount sent per address in a transaction hash, set the amount sent as a negative value
+    - union the transfer amount received per address in a transaction hash, keep the amount received as a positve value
+    */
+    select
+        blockchain
+        , block_date
+        , tx_id as tx_hash
```
jeff-dude, line 20: is it confusing to rename solana column to match EVM, if model is separate?

dbt_subprojects/daily_spellbook/models/_metrics/transfers/metrics_net_solana_transfers.sql
```diff
@@ -0,0 +1,113 @@
...
+        , from_base58(from_owner) as address
```
jeff-dude, line 21: is `from_base58()` needed, since solana is separate model here?

0xBoxer: nope, would leave it as base58 instead

dbt_subprojects/daily_spellbook/models/_metrics/transfers/metrics_net_solana_transfers.sql
0xBoxer: kind of, let's stick with solana terms in this model

0xBoxer:
we'll probably have to run this in quarters as well I suppose, but let's see what happens

PR #7085, opened 2024-10-31
metrics: add solana to tx metrics

jeff-dude:
broke out into separate CTEs in case we add more in the future

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/_metrics/transactions/metrics_transactions_daily.sql
+++ b/dbt_subprojects/daily_spellbook/models/_metrics/transactions/metrics_transactions_daily.sql
@@ -22,7 +22,7 @@ with evm as (
                 {{ source('evms', 'transactions') }}
             where
                 1 = 1
-                {% if is_incremental() or true %}
+                {% if is_incremental() %}
                 and {{ incremental_predicate('block_time') }}
                 {% endif %}
         )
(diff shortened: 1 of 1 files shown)


dbt_subprojects/daily_spellbook/models/_metrics/transactions/metrics_transactions_daily.sql
```diff
@@ -9,39 +9,58 @@
...
+), solana as (
+    select
+        'solana' as blockchain
+        , block_date
+        , id as tx_hash
+    from
+        {{ source('solana', 'transactions') }}
+    where
+        1 = 1
+        and block_date is not null --200m+ rows of nulls
```
jeff-dude, line 45: this is an odd finding here
either of you know why there are so many rows with null `block_date`?

jeff-dude:
(need to remove short timeframe prior to merge)

dbt_subprojects/daily_spellbook/models/_metrics/transactions/metrics_transactions_daily.sql
0xRobin: no clue, sadly :confused:

0xBoxer: solscan seems equally confused: https://solscan.io/block/31748087 

probably some client issue, looks like that block slot range that doesn't have dates attached to it was 4 years ago, a few of them have timestamps and are dated around that time: https://solscan.io/block/38877915 

Can safely ignore from my POV

jeff-dude: fwiw, `solana.blocks` also has the nulls (as expected on your findings here). i will leave the filter then, thanks both

PR #7090, opened 2024-11-01
feat: update evm.info

0xBoxer:
This PR updates the evm.info table.


dbt_subprojects/daily_spellbook/models/evms/evms_schema.yml
```diff
@@ -514,6 +514,8 @@ models:
         description:  "Chain where proofs are submitted to"
       - name: settlement
         description:  "Chain that is being settled on"
+      - name: is_on_dune
+        description: "Is the chain on Dune"
```
Hosuke, line 518: Not sure my description is correct enough.

Hosuke (approved):

dbt_subprojects/daily_spellbook/models/evms/evms_schema.yml
0xBoxer: lgtm

0xBoxer:
I was using this for something, honestly forget what, but seems useful even if not always 100% up to date. 

we could just make it part of the onboarding list I guess, but that lives on your side

jeff-dude:
> I was using this for something, honestly forget what, but seems useful even if not always 100% up to date.
> 
> we could just make it part of the onboarding list I guess, but that lives on your side

are you aware of this table `dune.dune.dataset_chainlist`? that could serve your use case for chain coverage on dune.

the reason i'm pushing back is because we have a backlog task DUX-741 to improve how we build, maintain and use `evms.info` with it's growing importance for other spells and such. we should look to prioritize that and leverage this other `dune.dune.dataset_chainlist` table as part of that (need to confirm how that is built and maintained too)

0xBoxer:
> are you aware of this table dune.dune.dataset_chainlist

was not aware, but it's missing a bunch of data that this table has. I guess we could use that to fetch a bunch of core info from there, but realistically that doesn't save us a lot of time 🤷 

looks like the blockchain team maintains this, we could ask them if we can add to their stuff somehow, but that seems harder than just keeping it here and also allow for outside contributions

jeff-dude (approved):
okay okay, we can push this column, just know that we will be enhancing this table more in near future :)
also pushed fix for two chains that are on dune to `true`

Issue #7093, opened 2024-11-02
[BUG] tokens_solana.transfers returns no data when directed at Token2022 programs


rsellers:
### Description
I maintain a few dune tables which query over transfers pertaining to a certain project with token_mint_address = 'xNETbUB7cRb3AAu2pNG2pUwQcJ2BHcktfvSB8x1Pq6L'. Starting 10/31 or 11/01, the tables broke with my queries retuning no data. I had no issues before, and I've isolated the issue to stem from the tokens_solana.transfers query when filtering by certain contract addresses. 

My contract address program is a little unique in that its a token2022 version. That seems to be the only thing which is special about it. So either the root cause is related to token2022 or something totally different. 

### Current behavior
A query from this table returns blank when token_mint_address points towards a token 2022 mint address. 

Example query retuning blank with token 2022 program: 
SELECT *
    FROM
        tokens_solana.transfers
    WHERE
        token_mint_address = 'xNETbUB7cRb3AAu2pNG2pUwQcJ2BHcktfvSB8x1Pq6L' -- XNET Token 2022 program
    LIMIT 2

### Expected behavior
I need the tokens_solana.transfers to function correctly and list all the transfer data so I can continue to operate my tables. 

Example *successful* query retuning expected information for a regular token program (dogwifhat for example): 
SELECT *
    FROM
        tokens_solana.transfers
    WHERE
        token_mint_address = 'EKpQGSJtjMFqKZ9KQanSqYXRcF8fBopzLHYxdM65zcjm' -- dogwifhat regular token program
    LIMIT 2

### Impacted model(s)
I'm not sure what to put for models. 

### Possible solution
I don't have any suggestions right now, unfortunately.

jeff-dude:
thank you for raising. as mentioned by boxer above, i will link PR with fix

PR #7094, opened 2024-11-02
fix: token22 mint address

0xBoxer:
The token_mint_address field in our token22 models requires correction. This necessitates a rerun of the affected models. Given the relatively small data volume involved, the reprocessing should be completed swiftly. Note that this will require triggering backfill operations to ensure data consistency.


0xBoxer:
this will fix https://github.com/duneanalytics/spellbook/issues/7093

0xBoxer:
This won't require a rerun of the downstream metrics as far as I can tell, the prices join worked as intended, I just pulled this one column from the wrong source

Hosuke (approved):
LGTM.✅
Thank you @0xBoxer

jeff-dude (approved):
i will look to merge a bit later today

jeff-dude:
okay, yes, i think you are right that only modified needs run and not anything downstream. you can see in the CI logs, there are lots of models downstream of this change due to reading from `tokens_solana.transfers` in various spells. i will turn off downstream in prod for this merge, to avoid that large refresh.

PR #7095, opened 2024-11-03
Update test table schema example

0xdatawolf001:
Smol update to the readme example


Hosuke (approved):

jeff-dude (approved):
good catch. we should likely update the CI to get rid of `_dunesql_` as it's leftover from the migration to new engine, but we can leave for now and get this updated

PR #7114, opened 2024-11-07
Optimize zeroex_ethereum_native_fills

Hosuke:
Trying:
1. extract common prices and tokens CTE


Hosuke:
[Mtaching test](https://dune.com/queries/4257983)

jeff-dude:
> Benchmark original building time in #7118
> 
> Original incremental building: https://github.com/duneanalytics/spellbook/actions/runs/11723854595/job/32656431197?pr=7118#step:15:19
> 
> New implementation incremental building: https://github.com/duneanalytics/spellbook/actions/runs/11717726253/job/32637851623?pr=7114#step:15:19
> 
> The new version completes in just 8.52 seconds compared to the previous 48.69 seconds, despite handling a similar number of rows. This represents an improvement of about 82.5% in execution time.

thank you. definitely worth merging and see where it takes us in prod. the incremental run in prod is like 25-30x longer than that 48s incremental, but prod also runs on a smaller cluster + runs in parallel with other models at the same time so it's hard to compare.

let's plan to merge this and see what it looks like, then evaluate from there if further optimization is needed.

thanks again!

jeff-dude (approved):
fyi @RantumBits 
we are trying to optimize run time on this model. it is an outlier in prod in terms of run time and holds up the overall job

jeff-dude:
could also likely merge this without `state:modified` enabled in prod, as there is no change to the logic. that is likely the best bet.

PR #7119, opened 2024-11-07
Fix solana dex.trades issue

0xBoxer:
This PR is to trigger a run of dex.trades to investigate the issue with token22 null values.


jeff-dude:
i will move the changes to native sol transfers here to other PR, so i can merge that one alone first

jeff-dude:
as noted in slack, should be good to remove here without modified enabled and increased incremental read back timeframe. that will need handled in dbt cloud manually.

to confirm, changes here:
- no logic change in dex pipeline, just need to refresh data with how it joins to fixed transfers data (in progress now)
- unique key change on sol transfer current model, no impact on history, just logic moving forward

jeff-dude (approved):

PR #7123, opened 2024-11-08
metrics: filter out wrap actions in solana transfers

jeff-dude:


0xBoxer (approved):
that should be all that's needed

0xRobin (approved):

PR #7130, opened 2024-11-09
Unmaterialise `blobs.based_submitters`

hildobby:
With `blobs.based_submitters` not always up to date prior to `ethereum.blobs` runs, it leads to some accidentally untagged blobs when they should be tagged, like here you can see all those untagged from nov 7th onwards showing up when they should be tagged as taiko:

PixelSnap 2024-11-09 at 17 57 36@2x

It's also misleading people who think those are new blob submitters: https://x.com/trent_vanepps/status/1855170775915065474

And regardless, blobs.based_submitters currently runs in ~2 seconds so this should be no big deal in terms of compute/runtime


jeff-dude (approved):

jeff-dude:
noting for self:
dbt-trino seems to have bug when switching between table/view in materialization, possible this fails in prod and requires manual step to remove previous table first.

PR #7147, opened 2024-11-18
add bob tokens

mastilver:
### Description:

- Add bob all bob tokens from https://github.com/bob-collective/ethereum-optimism.github.io
- Add bob stablecoin tokens from https://github.com/bob-collective/ethereum-optimism.github.io

PR #7158, opened 2024-11-19
add bob tokens to tokens_erc20

mastilver:
### Description:

I opened this PR earlier: https://github.com/duneanalytics/spellbook/pull/7147 but I forgot to add it to `tokens_erc20`, this fix that


jeff-dude (approved):
good catch, we can add.

if you peek at code below, you'll see we also have an automated source for erc20 metadata. however, if i'm looking correctly, BOB is not covered at this time:
```
select *
from dune.definedfi.dataset_tokens
where networkid = 60808
limit 10 
```
i pulled networkid from here:
https://chainlist.org/chain/60808

PR #6290, opened 2024-06-30
feat: add banana gun evm bot trades spellbook

whalehunting:
This PR will implement the first spellbook for EVM bot trades, starting with BananaGun. The corresponding solana spellbook(s) are already in use and have proven to significantly reduce execution times for dashboards,

Related dashboards:
- https://dune.com/whale_hunter/banana-gun-bot
- https://dune.com/whale_hunter/dex-trading-bot-wars

## Tasks
- [x] Add banana gun ethereum query + schema
- [x] Add banana gun base query + schema
- [x] Add banana gun blast query + schema
- [x] Add banana gun bsc query + schema
- [x] Add banana gun avalanche query + schema
- [x] Add banana gun evm query + schema
- [x] Add evm query
- [x] Add seed files (?)

PR #7180, opened 2024-11-24
feat: add banana gun evm bot trades spellbook

whalehunting:
This PR reimplements #6290 and aims to add the first EVM bot trades spellbook.

### Description:
EVM bot trades data is growing, so it seems like a reasonable time to move them to Spellbook. I'm planning to do this for most EVM bots, so this spellbook should act as a blueprint for other bots.

Adds 3 new spellbooks:
- banana_gun_ethereum_bot_trades
- banana_gun_base_bot_trades
- dex_evm_bot_trades


0xRobin:
@whalehunting Fixed the seed for you. 
I did `describe test_schema.git_dunesql_37fe81f_banana_gun_ethereum_trades_seed` on Dune to see the actual schema and then added all the varbinary columns to the seed schema file. (The engine does not infer the varbinary type automatically, so they need to be specified)

dbt_subprojects/dex/models/bot_trades/_schema.yml
```diff
@@ -0,0 +1,134 @@
...
+      Banana Gun trades on Ethereum
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - blockchain
+            - tx_hash
+            - evt_index
+      - check_bot_trades_seed:
+          seed_file: ref('banana_gun_ethereum_trades_seed')
+          blockchain: ethereum
```
0xRobin, line 114: no need to add the blockchain filter here, since the seed is only for this chain.

dbt_subprojects/dex/models/bot_trades/_schema.yml
```diff
@@ -0,0 +1,134 @@
...
+      Banana Gun trades on Base
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - blockchain
+            - tx_hash
+            - evt_index
+      - check_bot_trades_seed:
+          seed_file: ref('banana_gun_base_trades_seed')
+          blockchain: base
```
0xRobin, line 134: no need to add the blockchain filter here, since the seed is only for this chain.

dbt_subprojects/dex/models/bot_trades/banana_gun/ethereum/banana_gun_ethereum_bot_trades.sql
```diff
@@ -0,0 +1,195 @@
...
+  evt_index,
+  IF(evt_index = highestEventIndex, true, false) AS is_last_trade_in_transaction
+FROM
+  botTrades
+  JOIN highestEventIndexForEachTrade ON botTrades.tx_hash = highestEventIndexForEachTrade.tx_hash
+  LEFT JOIN botETHDeltas ON botTrades.tx_hash = botETHDeltas.tx_hash
+  LEFT JOIN {{ source('prices', 'usd') }} ON (
+    blockchain = '{{blockchain}}'
+    AND contract_address = {{weth}}
+    AND minute = DATE_TRUNC('minute', block_time)
```
0xRobin, line 191: should add an incremental block here to filter prices on the incremental run

dbt_subprojects/dex/models/bot_trades/banana_gun/base/banana_gun_base_bot_trades.sql
```diff
@@ -0,0 +1,193 @@
...
+  evt_index,
+  IF(evt_index = highestEventIndex, true, false) AS is_last_trade_in_transaction
+FROM
+  botTrades
+  JOIN highestEventIndexForEachTrade ON botTrades.tx_hash = highestEventIndexForEachTrade.tx_hash
+  LEFT JOIN botETHDeltas ON botTrades.tx_hash = botETHDeltas.tx_hash
+  LEFT JOIN {{ source('prices', 'usd') }} ON (
+    blockchain = '{{blockchain}}'
+    AND contract_address = {{weth}}
+    AND minute = DATE_TRUNC('minute', block_time)
```
0xRobin, line 189: should add an incremental block here to filter prices on the incremental run

0xRobin (changes requested):

dbt_subprojects/dex/models/bot_trades/banana_gun/base/banana_gun_base_bot_trades.sql
whalehunting: how would that look like in a join condition? it already has minute = minute, is that not sufficient?

dbt_subprojects/dex/models/bot_trades/banana_gun/ethereum/banana_gun_ethereum_bot_trades.sql
whalehunting: see below

dbt_subprojects/dex/models/bot_trades/banana_gun/base/banana_gun_base_bot_trades.sql
0xRobin:
```suggestion
  LEFT JOIN {{ source('prices', 'usd') }} ON (
    blockchain = '{{blockchain}}'
    AND contract_address = {{weth}}
    AND minute = DATE_TRUNC('minute', block_time)
    {% if is_incremental() %}
    AND {{ incremental_predicate('minute') }}
    {% endif %}  
```

0xRobin: while indeed we are already filtering on minute, adding the incremental predicate adds a static filter which is more efficient and reduces the total amount of rows that the engine need to scan to look for the exact minute match

0xRobin: For these models, the performance difference is negligible but we try to be consistent in adding incremental filters to all sources/refs when possible.

0xRobin (approved):

PR #7188, opened 2024-11-25
add sanctum router to solana dex trades

senyor-kodi:

### Description:

Add Sanctum Router to dex_solana.trades

### DIFF
--- a/dbt_subprojects/solana/models/_sector/dex/dex_solana_base_trades.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/dex_solana_base_trades.sql
@@ -23,6 +23,7 @@
    , ref('meteora_v2_solana_base_trades')
    , ref('goosefx_ssl_v2_solana_base_trades')
    , ref('pumpdotfun_solana_base_trades')
+   , ref('sanctum_router_base_trades')
 ] %}
 
 {% for dex in solana_dexes %}
--- a/dbt_subprojects/solana/models/_sector/dex/sanctum_router/sanctum_router_base_trades.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/sanctum_router/sanctum_router_base_trades.sql
@@ -1,0 +1,202 @@
+{{
+  config(
+        schema = 'sanctum_router_solana',
+        alias = 'base_trades',
+        partition_by = ['block_month'],
+        materialized = 'incremental',
+        file_format = 'delta',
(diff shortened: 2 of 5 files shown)


senyor-kodi:

jeff-dude:
> Hey jeff, adding a time filter for last week did it!
> 
> Had to fix a couple typos before all checks passed but seems to be good now. Do I need to do anything else on my end? Should I keep the time filter?

if you haven't already, i would recommend querying the CI tables directly on the app and test the data output to ensure it meets your expectations (even though its subset of history). if you click into the gh action logs and expand the dbt model run step, you can find table names and query them directly 🙏 

once you are comfortable with the output, we can revert the hardcoded date commit and prep for merge

senyor-kodi:
> if you haven't already, i would recommend querying the CI tables directly on the app and test the data output to ensure it meets your expectations (even though its subset of history). if you click into the gh action logs and expand the dbt model run step, you can find table names and query them directly 🙏
> 
> once you are comfortable with the output, we can revert the hardcoded date commit and prep for merge

the tables for the two incremental models look fine!

however the view model test_schema.git_dunesql_e513efa_sanctum_router_trades comes up empty. is this normal or should I be getting a result here as well?

jeff-dude:
> however the view model test_schema.git_dunesql_e513efa_sanctum_router_trades comes up empty. is this normal or should I be getting a result here as well?

good catch, another reason to test these output tables!

in this scenario, i believe we are actually all good. `dex_solana_trades` model isn't modified in this PR, since you only need to add to `*_base_trades`. CI will redirect to read from prod for any models that aren't in the PR itself. this means the view is reading from prod table, where we know this data isn't integrated yet.

we can confirm that the `_base_trades` table has rows as expected:
```
select *
from test_schema.git_dunesql_e513efa_dex_solana_base_trades --test_schema.git_dunesql_e513efa_sanctum_router_trades
where project = 'sanctum_router' and version = 1
limit 10
```

this will be all good in prod.

let me give the PR a more general review prior to you reverting the hardcoded date filter for historical data.

jeff-dude (approved):
this is looking clean to me 🙏 
feel free to revert the short timeframe and we can prep for merge

senyor-kodi:
Done! Thanks a lot for your help @jeff-dude and thank you for all the work y'all do on Spellbook 🙏🙏

PR #7189, opened 2024-11-25
Revert 0x changes: downstream is broken

jeff-dude:
@RantumBits unfortunately everything blew up in prod with all these changes merged :(
it's too difficult to pinpoint exactly what at the moment. we will need to revert and get prod running again, then revisit how we roll these changes out. we should always start as upstream in the pipeline as possible, ensure all is good, and continue from there


jeff-dude:
@RantumBits even after the reverts to original state prior to today, when i tried to run a full refresh of the 0x spells, they were failing. example here:
```
23:17:16    Database Error in model zeroex_polygon_nft_fills (models/_projects/zeroex/polygon/zeroex_polygon_nft_fills.sql)
  TrinoUserError(type=USER_ERROR, name=TYPE_MISMATCH, message="line 42:28: Cannot apply operator: timestamp(3) with time zone <= varchar(10)", query_id=20241125_231241_02189_cv74k)
```

i'm not sure why a model that wasn't touched would do this. i assume it hasn't had a full refresh in a long time, so this error had been hidden or something?

to get prod running again, i had to exclude all spells in `models/_projects/zeroex/` directory. we need to figure out how to get these running again. feel free to dm me on TG if you want to work through this and re-enable the spells

PR #7196, opened 2024-11-26
zeroex - exclude nft spells 

RantumBits:
### Description:

zeroex - add tag prod_exclude to nft spells


jeff-dude (approved):

jeff-dude:
short term:
exclude all `nft_fill` spells until follow up PR with macro approach and bug fixes to re-enable. there are only ethereum and polygon at this time.

PR #7199, opened 2024-11-27
Update autosnipe_solana_bot_trades.sql

clizzard7:
Project start was false - no we have the right start date


0xRobin (approved):
approving even with failing CI because it's a trivial change

PR #7209, opened 2024-11-28
Init boba chain

Hosuke:
Init boba chain


Hosuke:
prices automation CI ❌ can be ignored in this PR.

sources/_base_sources/evm/boba_base_sources.yml
jeff-dude, line 1: i assume this source list is shorter due to decoded not available yet, just raw data?
example, returns no rows:
```
select * from boba.logs_decoded limit 10
```

Hosuke: Yes, I can put them as placeholders first, as they existed in the namespace.

Hosuke, line 1: I have tried to match all raw tables and schema with https://dune.com/data?category=canonical&namespace=boba

Hosuke: Updated now, this PR should be ready to go.

jeff-dude (approved):
thank you. we can iterate and add more tokens as needed when adding projects to spellbook and can dig into the data a bit further

PR #7213, opened 2024-11-29
New Yield Yak strategies

yy-analytics:

### Description:

Just adding a couple of newly deployed strategies to the strategies macro/model in Spellbook for Yield Yak.

---


yy-analytics:
@jeff-dude this one is failing with a "QUERY_HAS_TOO_MANY_STAGES" error (502 versus a 500 limit it seems), but when I run `dbt compile` locally and then paste the SQL for the related model (`yield_yak_avalanche_c_user_yrt_balances`) into Dune, it runs without a problem. Any advice on what to do here?

Hosuke:
Nested macro in the jinja syntax may be a cause as well:
https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/daily_spellbook/macros/project/yield_yak/yield_yak_user_yrt_balances.sql#L34

jeff-dude:
> @jeff-dude the `dbt-test` gh action for this one seems to be failing with a "QUERY_HAS_TOO_MANY_STAGES" error (502 versus a 500 limit it seems), but when I run `dbt compile` locally and then paste the SQL for the related model (`yield_yak_avalanche_c_user_yrt_balances`) into Dune, it runs without a problem. Any advice on what to do here?

good question, let me raise internally to see what the latest is around the stages error and why the app would differ from CI here.

yy-analytics:
Sorry for the slow response @jeff-dude , I was only using the free engine to run it. I've just done it again to confirm that I wasn't misremembering and again the query runs fine on the free engine. You can see the query I'm talking about here: https://dune.com/queries/4383108

I'm getting this query by running `dbt compile` locally within the `daily_spellbook` directory, then taking the model from the following path: `dbt_subprojects\daily_spellbook\target\compiled\daily_spellbook\models\yield_yak\avalanche_c\yield_yak_avalanche_c_user_yrt_balances.sql` and pasting it into Dune to run.

I've also double checked that it contains the new additions from this PR and it does, so I'm at a bit of a loss as to what's going on now.

jeff-dude:
> Sorry for the slow response @jeff-dude , I was only using the free engine to run it. I've just done it again to confirm that I wasn't misremembering and again the query runs fine on the free engine. You can see the query I'm talking about here: https://dune.com/queries/4383108
> 
> I'm getting this query by running `dbt compile` locally within the `daily_spellbook` directory, then taking the model from the following path: `dbt_subprojects\daily_spellbook\target\compiled\daily_spellbook\models\yield_yak\avalanche_c\yield_yak_avalanche_c_user_yrt_balances.sql` and pasting it into Dune to run.
> 
> I've also double checked that it contains the new additions from this PR and it does, so I'm at a bit of a loss as to what's going on now.

you're taking the right steps. each cluster has a different config for number of stages, so it's possible the free engine is also above 500. we can explore updating CI cluster to have more stages, but i'm worried that could become an issue again in the future if you continue to add source tables.

we may need to find the right balance of design in dbt vs. complexity of output query compiled and ran on clusters. it's worth considering a redesign here to possibly not include so many sources in one query? maybe we have more models that break it up a bit? would there be a logical way to split that up within a chain?

yy-analytics:
Hi @jeff-dude , there's not really an obvious/natural way to split the strategy types because each type (i.e. source table) has many actual strategies within it that just share the same ABI, and some of those might still be active whilst some might be deprecated. So we can't filter at the "strategy type" level for deprecated ones to e.g. create an archived group, because some may still be active. What I would say is that the number of sources should only be increasing quite slowly from now on - for instance if the stages limit was increased to 1000 then I don't think we'd hit that for many years (or maybe ever).

The only alternative I can think of would be de-coding contracts again from scratch and manually inputting an ABI that relates to the shared functions/events amongst all of the strategies - and then using the same name (e.g. YakStrategy) for all of them so that they're all coming under a single (or a handful) contract name. But would rather not do this approach as then we lose extra information in the decoded contract tables that might be useful elsewhere.

jeff-dude:
> Hi @jeff-dude , there's not really an obvious/natural way to split the strategy types because each type (i.e. source table) has many actual strategies (contract addresses) within it that just share the same ABI, and some of those might still be active whilst some might be deprecated. So we can't filter at the "strategy type" level for deprecated ones to e.g. create an archived group, because some may still be active. What I would say is that the number of sources should only be increasing quite slowly from now on - for instance if the stages limit was increased to 1000 then I don't think we'd hit that for many years (or maybe ever).
> 
> The only alternative I can think of would be de-coding contracts again from scratch and manually inputting an ABI that relates to the shared functions/events amongst all of the strategies - and then using the same name (e.g. YakStrategy) for all of them so that they're all coming under a single (or a handful of) contract name(s). But would rather not do this approach as then we lose extra information in the decoded contract tables that might be useful elsewhere.

okay, thank you for the context. the platform team sounds willing to try 1k stages, so i can look into that. just want to reiterate, if you go over that 1k in future (even if looks unrealistic), we will have to exclude and find other solutions :)

let me pass this along and see how we can test here.

yy-analytics:
Thanks for that @jeff-dude , let me know if there's anything needed from my side :)

jeff-dude:
> Thanks for that @jeff-dude , let me know if there's anything needed from my side :)

an interesting finding: the reason it works on free engine, which is 500 stages, is because the raw sql is only ~200 stages. however, the merge statement built via dbt ends up being 510 stages, pushing us over that limit here in CI. not much we can do to change the merge logic, without digging into the dbt-core code which isn't realistic at the moment. might be worth looking into why the merge would have so many stages relative to raw sql, not sure if that is something you have full access / transparency into?

for reference, this is generated merge statement from dbt logs:
```
merge into hive.test_schema.git_dunesql_5439151_yield_yak_avalanche_c_user_yrt_balances as DBT_INTERNAL_DEST
            using hive.test_schema.git_dunesql_5439151_yield_yak_avalanche_c_user_yrt_balances__dbt_tmp as DBT_INTERNAL_SOURCE
            on (DBT_INTERNAL_DEST.to_time >= date_trunc('day', now() - interval '3' day)) and (
                    DBT_INTERNAL_SOURCE.user_address = DBT_INTERNAL_DEST.user_address
                ) and (
                    DBT_INTERNAL_SOURCE.contract_address = DBT_INTERNAL_DEST.contract_address
                ) and (
                    DBT_INTERNAL_SOURCE.from_time = DBT_INTERNAL_DEST.from_time
                )

        
        when matched then update set
            "blockchain" = DBT_INTERNAL_SOURCE."blockchain","user_address" = DBT_INTERNAL_SOURCE."user_address","contract_address" = DBT_INTERNAL_SOURCE."contract_address","from_time" = DBT_INTERNAL_SOURCE."from_time","to_time" = DBT_INTERNAL_SOURCE."to_time","yrt_balance" = DBT_INTERNAL_SOURCE."yrt_balance"
        

        when not matched then insert
            ("blockchain", "user_address", "contract_address", "from_time", "to_time", "yrt_balance")
        values
            (DBT_INTERNAL_SOURCE."blockchain", DBT_INTERNAL_SOURCE."user_address", DBT_INTERNAL_SOURCE."contract_address", DBT_INTERNAL_SOURCE."from_time", DBT_INTERNAL_SOURCE."to_time", DBT_INTERNAL_SOURCE."yrt_balance")
```

jeff-dude:
i'll be out for rest of the year, so i'm tagging @0xRobin to support as you need 🙏 

plz do let us know level of urgency here and can help determine how we can proceed to help here. thank you for the patience.

yy-analytics:
Thanks @jeff-dude and happy holidays! It's not extremely urgent (no critical dependencies) but would like to get it resolved as soon as I can ideally. @0xRobin if you're able to shed any light on my questions above about `is_incremental` and changes to the macro and/or how to gauge the number of stages that would be much appreciated, thanks!

0xRobin:
@yy-analytics I believe you're right on your analysis, we will need to bring that left join out of the union.
Since we're reading straight from decoded events, the performance impact shouldn't be to large.

0xRobin (approved):

PR #7234, opened 2024-12-03
Only keep validators with an assigned index in `staking_ethereum.flows`

hildobby:
…_ethereum.flows

### Description:

[...]

---


0xRobin:
@hildobby can you give more details on the external dataset you're using here?

hildobby:
> @hildobby can you give more details on the external dataset you're using here?

This dataset only contains validator_index and their corresponding pubkey values. This is because the BLS key check happens offchain due to it being quite compute heavy. So this dataset was fetched from beaconcha.in and using it as an inner join ensures that validator_index is correct while excluding non-valid deposits (there are a few, currently throwing off the validator_index values the current spell calculates by counting incrementally).

TLDR; this fixes a bug I've had for a while and makes the data accurate

0xRobin:
Ok, thanks for explaining, how is the dataset maintained?

hildobby:
I defined the source (+ included it in a sources.yml) like on here: https://github.com/duneanalytics/spellbook/blob/588ca2762f9436ecf79a57b147f01c8d49ad91c3/dbt_subprojects/tokens/models/prices/prices_solana_day.sql#L25


But it's still failing, is there something I'm missing?

dbt_subprojects/hourly_spellbook/models/_sector/staking/ethereum/staking_ethereum_flows.sql
```diff
@@ -113,3 +99,102 @@ SELECT w.block_time
...
+
+{{ config(
+    
+    schema = 'staking_ethereum',
+    alias = 'flows',
+    post_hook='{{ expose_spells(\'["ethereum"]\',
+                                "sector",
+                                "staking",
+                                \'["hildobby", "0xBoxer"]\') }}')
+}}
```
0xRobin, line 122: @hildobby you duplicated the full model below, which uses incorrect syntax for the source (hence the compile error)

jeff-dude:
minor note: we are also in sync in TG group for design patterns related to the source in this PR

0xRobin:
@hildobby we have new data available from Lido which should be able to resolve this bug without us depending on external scripts.
Can you look at the `beacon_lido.validators` table to resolve the index <> pubkey matching?

hildobby:
how should I reference an external dataset in spellbook? where should i place the schema file? and should the jinja part be a ref or a source? @0xRobin @jeff-dude

0xRobin:
@hildobby from spellbooks view, this is just a normal source like a raw blockchain table.
So you can add a sources yaml file with it and use `{{source(..)}}`

I would add it under `/sources/lido/beacon`

hildobby:
> @hildobby from spellbooks view, this is just a normal source like a raw blockchain table. So you can add a sources yaml file with it and use `{{source(..)}}`
> 
> I would add it under `/sources/lido/beacon`
Screenshot 2025-01-08 at 14 17 10


seems like there's something weird here, what am I missing that's causing the CI error @0xRobin ?

sources/_subprojects_outputs/staking/_sources.yml
jeff-dude, line 1: you should be good to remove this file now

jeff-dude (changes requested):
looks good, just one minor request to remove old source file

sources/_subprojects_outputs/staking/_sources.yml
hildobby: done

jeff-dude (approved):

PR #7280, opened 2024-12-09
DUX-798 ronin: initiate katana dex

jeff-dude:


dbt_subprojects/dex/models/trades/ronin/platforms/katana_ronin_base_trades.sql
jeff-dude, line 1: fyi @Hosuke 
i made a decision to not use the uni v2 macro here, as many column names have `_` prefix and i didn't want to modify the macro for lots of `if blockchain = 'ronin' then `_` +` type of logic. it would also then run all models which use the macro both in CI and prod.

jeff-dude:
CI is likely to fail on node crash. successful run can be found in commit logs.

dbt_subprojects/dex/models/trades/ronin/platforms/katana_ronin_base_trades.sql
0xRobin: If we'd still want to use the macro we can perform any needed column renames in a CTE and then pass the cte references in the macro. 
```
with evt_swap as (
  select
  _amount1Out as amount1Out
  ,_amount1In as amount1In
  ,...
  )
, evt_paircreated as (
  select
  ...
  )

{{
    uniswap_compatible_v2_trades(
        blockchain = 'ronin',
        project = 'katana',
        version = '1',
        Pair_evt_Swap = 'evt_swap`,
        Factory_evt_PairCreated = 'evt_paircreated'
    )
}}
```

Hosuke: Both approaches are ok for me.

Hosuke (approved):
LGTM.✅

dbt_subprojects/dex/models/trades/ronin/platforms/katana_ronin_base_trades.sql
jeff-dude: i actually started down this approach via CTE renames, but then also realized the source tables we pass in have `{{ }}` wrapped around them in the macro itself. i don't think passing a CTE into there will work with those in place.

my suggestion is we get this integrated, then consider a cleanup on the macro to prepare for these edge cases.

PR #7281, opened 2024-12-09
update ronin token metadata

jeff-dude:


jeff-dude:
fyi @Hosuke 
merging this prior to DEX PR for katana, as CI output showed these as top traded (after our trusted token list)

PR #7266, opened 2024-12-08
Lifinity v2 transfers fix

Synthquest:
## Thank you for contributing to Spellbook 🪄
Please open the PR in **draft** and mark as ready when you want to request a review. 

### Description:

there are instances where the token amounts are incorrectly attributing a mint instruction to the transfer instruction

---
quick links for more information:
- [README.md](https://github.com/duneanalytics/spellbook/blob/main/README.md)
- [spellbook docs](https://github.com/duneanalytics/spellbook/tree/main/docs)
- [CONTRIBUTING.md](https://github.com/duneanalytics/spellbook/blob/main/CONTRIBUTING.md)

PR #7292, opened 2024-12-11
Update lifinity_v1_base_trades.sql

smyyguy:
there are instances where the token amounts are incorrectly attributing a mint instruction to the transfer instruction

to get an idea of the issue you can see an example trade here:
```
SELECT 
    *
FROM dex_solana.trades
WHERE 
    date_trunc('week', block_time) = TIMESTAMP '2023-11-06'
    AND project = 'lifinity' and token_pair is null AND token_bought_mint_address = '8gz8AFF1w6DDvhpdmq6SRqUydNGju94CEtrty4Fnu6Qa'
LIMIT 10 
```

also see: https://github.com/duneanalytics/spellbook/pull/7266

it seems the mint action is getting picked up because of the way the joins are done. We can filter for transfer actions to get rid of the possibility of picking up the mint amounts by accident


jeff-dude (approved):
thanks for finding and applying the fix to both versions. i will look to get this applied 🙏

PR #7343, opened 2024-12-16
chore: add additional aggregator fee wallets for flokibot

whalehunting:


whalehunting:
@0xRobin would be great to get this merged before the weekend!

0xRobin (approved):
runtimes are long, but should be fine on incremental. :+1:

PR #7347, opened 2024-12-17
Add COW Token on Base

bh2smith:
COW Just launched on Base and deployed their token as well. Adding here because its been HOT this week.


jeff-dude (approved):

bh2smith:
Poap?

PR #7350, opened 2024-12-18
enrich balancer_v3 trades on dex.trades

viniabussafi:
### Description:

Balancer V3 was launched a week ago and has [100% Boosted Pools](https://docs.balancer.fi/concepts/explore-available-balancer-pools/boosted-pool.html#advantages-of-boosted-pools) as one of it's flagships. The complication we're facing here is, when ERC4626 tokens (see ELI5) are traded against other ERC4626 tokens, the `amount_usd` column in dex.trades is null.
This PR introduces the same logic currently used on [balancer.trades](https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/dex/models/_projects/balancer/trades/gnosis/balancer_v3_gnosis_trades.sql) to dex.trades, using a new macro to handle swaps on Balancer V3.
I'd hate to be the one to break dex.trades, so 100% open for suggestions here, this is just an initial idea.

### ELI5 on ERC4626 tokens
the ELI5 is that these tokens are basically receipt tokens from deposits into lending protocols (like aave and morpho). This means that liquidity providers can deposit into a pool with Wrapped AAVE wstETh / Wrapped AAVE WETH. So, the path would be the LP would deposit weth and/or wsteth and get exposed to both the yields on AAVE for both tokens + swap fees on swaps happening on the pool. 
These tokens are priced based on how much Wrapped AAVE WETH you get after depositing WETH (on wrap and unwrap events), multiplied by the price of WETH at that time, like we do [here](https://github.com/duneanalytics/spellbook/blob/cccc1a238a2cfba777db9c11f7664da14bc57440/dbt_subprojects/hourly_spellbook/models/_project/balancer/erc4626_tokens/ethereum/balancer_v3_ethereum_erc4626_token_prices.sql)
That's why these tokens aren't on coinpaprika (and probably will never be).

Example of 100% Boosted Pools: https://balancer.fi/pools/ethereum/v3/0xc4ce391d82d164c166df9c8336ddf84206b2f812
Current trades where amount_usd = 0 on dex.trades: https://dune.com/queries/4445360

tagging @mendesfabio for visibility/comments

### DIFF
--- a/dbt_subprojects/dex/macros/models/enrich_balancer_v3_dex_trades.sql
+++ b/dbt_subprojects/dex/macros/models/enrich_balancer_v3_dex_trades.sql
@@ -74,7 +74,6 @@ WITH base_trades as (
             APPROX_PERCENTILE(median_price, 0.5) AS price,
             LEAD(minute, 1, NOW()) OVER (PARTITION BY wrapped_token ORDER BY minute) AS time_of_next_change
         FROM {{ source('balancer_v3', 'erc4626_token_prices') }}
-        WHERE blockchain = 'ethereum'
         GROUP BY 1, 2, 3, 4
 )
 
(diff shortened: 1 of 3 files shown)


0xRobin (approved):
We might look to build out a generic ERC4626 prices table and incorporate this in the main prices tables so we don't need separate enriching logic. For now, this workaround is sufficient.

PR #7377, opened 2024-12-24
Change Materialization for Tokens Events

AIDataMaster:

### Description:

We are facing **HTTP error rate limits on Dune** when querying `markets_data` and `erc20` models because the tokens models are materialized as `views`. This causes **repetitive API requests** for every query. To resolve this, we propose switching from `view` materialization to `table` materialization.

---


0xRobin:
@AIDataMaster seems like the queries are very lightweight though and only require 1 http_get call per request, so that should be fine?
Can you share more details on the queries that are failing?

AIDataMaster:
> @AIDataMaster seems like the queries are very lightweight though and only require 1 http_get call per request, so that should be fine? Can you share more details on the queries that are failing?

https://dune.com/queries/4499052
Error: Your query issued too many HTTP requests [Execution ID: 01JGBN443FXE2MRKGCH2H5ECXA].

PR #7391, opened 2024-12-27
DUX-823 Prices v2 rework

0xRobin:

### Description:

- new upstream models for the sqlmesh prices pipeline.

Goal is to:
- simplify the lineage and bring the vwap calculation into spellbook
- have only one model flowing into sqlmesh (that has sparse minute price data)

This should resolve any duplicates we're currently having and will allow us to switch tokens from different sources without having temporary gaps or duplicates.

Also resolves DUX-702 while we're at it. 

---


dbt_subprojects/tokens/models/prices_v2/minute/prices_v2_dex_minute_raw.sql
```diff
@@ -0,0 +1,65 @@
...
+    AND {{ incremental_predicate('d.block_time') }}
+    {% endif %}
+)
+
+
+SELECT
+    blockchain,
+    contract_address,
+    date_trunc('minute',timestamp) as timestamp,
+    sum(price*volume)/sum(volume) as price, -- vwap
```
aalan3, line 60: VWAP price is defined in the following way, we can do a follow up on this if it's complicated to implement.
```
VWAP = (Typical Price * Volume) / Volume
Calculating the typical price from the session's first candle, with the formula (High + Low + Close) / 3
```

dbt_subprojects/tokens/models/prices_v2/prices_v2_latest.sql
```diff
@@ -0,0 +1,23 @@
+{{ config(
+        schema='prices_v2',
+        alias = 'latest',
```
aalan3, line 3: Why is this model needed?

dbt_subprojects/tokens/models/prices_v2/minute/prices_v2_dex_minute.sql
```diff
@@ -0,0 +1,11 @@
...
+    schema='prices_v2'
+    , alias = 'dex_minute'
+    , materialized = 'view'
+    )
+}}
+
+SELECT
+*
+FROM {{ ref('prices_v2_dex_minute_raw') }}
+INNER JOIN {{ ref('prices_v2_dex_filter') }} using (blockchain, contract_address)
```
aalan3, line 11: Can't we do this filtering below in `prices_v2_dex_minute_raw` directly instead?

aalan3 (approved):
Had a few comments but lgtm.

dbt_subprojects/tokens/models/prices_v2/minute/prices_v2_dex_minute.sql
0xRobin: opted to keep `prices_v2_dex_minute_raw` a clean union model

dbt_subprojects/tokens/models/prices_v2/minute/prices_v2_dex_minute_raw.sql
0xRobin: that calc with typical price is only needed if you have volume stats for a price range (candle), with our data we have volume per individual price point (trade) so we can be more accurate and do it as it's currently implemented. 

Unless I'm missing something?

dbt_subprojects/tokens/models/prices_v2/prices_v2_latest.sql
0xRobin: It's not needed for the sqlmesh side, just thought it would be cheaper do do it directly on the sparse dataset then later on the filled one.

dbt_subprojects/tokens/models/prices_v2/minute/prices_v2_dex_minute_raw.sql
aalan3: hmm yeah you are probably right. let's keep it like this then.

jeff-dude:
is the intention of this PR to remove the need for `dex_prices_block` altogether? hypothetical steps:
- merge this PR, build tables in prod
- update prices pipeline to use new output from here, simplify logic in pipeline downstream
- remove `dex_prices_block` from spellbook

if that is the case, where is the logic here for obtaining median price per token per block? did we decide to do away with that logic (that could change results a lot and require more testing)? we would need to perform this calculation prior to the vwap calculation.
https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/dex/models/prices/dex_prices_block.sql#L169-L190

0xRobin:
> is the intention of this PR to remove the need for `dex_prices_block` altogether?

That's indeed the end goal.
With the thought also being that we need to do this in parallel to the current running setup to not disturb that.

Do you know why we needed the median block price before doing the vwap?
I don't think median block with weighted average minute on top makes a ton of sense, but I might be missing something.

Is it outliers in dex.trades, where we don't want them to influence the vwap price?
Maybe just doing the median aggregation on the minute level is a good approach to get the same result.

0xRobin:
Ran some tests, and results look good.
We can safely change to a minute base median methodology instead of block median + minute vwap. 
image

PR #7394, opened 2024-12-27
Initiate corn in spellbook

Hosuke:
Initiate corn in spellbook


Hosuke:
Trusted tokens on corn is not very clear at the moment:
https://cornscan.io/tokens

dbt_subprojects/tokens/models/prices/corn/prices_corn_tokens.sql
```diff
@@ -0,0 +1,25 @@
...
+SELECT
+    token_id
+    , '{{ blockchain }}' as blockchain
+    , symbol
+    , contract_address
+    , decimals
+FROM
+(
+    VALUES
+    ('wbtc-wrapped-bitcoin', 'wBTCN', 0xda5ddd7270381a7c2717ad10d1c0ecb19e3cdfb2, 18)
```
Hosuke, line 21: Should I use wbtc price feed for wBTCN?
https://cornscan.io/token/0xda5dDd7270381A7C2717aD10D1c0ecB19e3CDFb2?chainid=21000000

dbt_subprojects/daily_spellbook/models/evms/evms_info.sql
```diff
@@ -95,4 +96,5 @@ FROM (
         , (2020, 'ronin', 'Ronin', 'Layer 1', null, 'RON', 0xe514d9deb7966c8be0ca922de8a064264ea6bcd4, 'https://app.roninchain.com/', timestamp '2021-01-25 10:49', NULL, NULL, NULL, true)
         , (14, 'flare', 'Flare', 'Layer 1', NULL, 'FLR', NULL, 'https://flare-explorer.flare.network/', timestamp '2022-07-13 15:32', NULL, NULL, NULL, true)
         , (88, 'viction', 'Viction', 'Layer 1', NULL, 'VIC', 0xC054751BdBD24Ae713BA3Dc9Bd9434aBe2abc1ce, 'https://vicscan.xyz/', timestamp '2018-12-14 07:50:20', NULL, NULL, NULL, true)
+        , (21000000, 'corn', 'Corn', 'Layer 1', NULL, 'BTCN', NULL, 'https://cornscan.io/', timestamp '2024-11-19 10:29:11', NULL, NULL, NULL, false)
```
jeff-dude, line 99:
```suggestion
        , (21000000, 'corn', 'Corn', 'Layer 1', NULL, 'BTCN', NULL, 'https://cornscan.io/', timestamp '2024-11-19 10:29:11', NULL, NULL, NULL, true)
```

jeff-dude: think we're safe to say `true` now, right?

dbt_subprojects/tokens/models/prices/corn/prices_corn_tokens.sql
jeff-dude: good question, let me ask 🤔

jeff-dude:
> Trusted tokens on corn is not very clear at the moment: https://cornscan.io/tokens
> 
> Using this query to initiate: https://dune.com/queries/4514448

this is a fine approach for now, it seems clear the pattern is to add this way initially & then revisit to add/modify based on adding DEXs and seeing top pools / swapped tokens.

dbt_subprojects/tokens/models/prices/corn/prices_corn_tokens.sql
jeff-dude: okay, we should be good to use bitcoin pricing for this. here is answer from corn team:

> As per BTCN's pricing, it is designed to be pegged to the price of BTC indeed. It accomplishes this by introducing 1:1 backing to wBTC and cbBTC (both tightly pegged to BTC) and, hence, efficient arbitraging. BTCN's liquidity is starting to build up on Corn and hasn't met the standards for price trackers and Oracles. For this reason, assuming a 1:1 peg to BTC should be accurate enough for analytics/visualization purposes. Eventually, as it gets listed by Oracles and trackers, we should switch to its true market value.

jeff-dude (approved):

PR #7405, opened 2024-12-31
add dflow to dex_solana.trades

smyyguy:
add dflow to dex_solana.trades


jeff-dude:
i believe you're hitting a memory error on the cross-project base trades spell, due to large data sizes on each. can you try to hardcode a short timeframe in that spell, to see if we can get it to run on subset of the data as expected?

smyyguy:
> e

hmm not really sure why that didnt take if that was the issue

jeff-dude:
> hmm not really sure why that didnt take if that was the issue

we need to put `timestamp` in front of hardcoded date/times, it's a dunesql syntax requirement

jeff-dude:
closing due to inactivity

PR #7412, opened 2025-01-03
flokibot bnb: exclude due to long running incremental

jeff-dude:
fyi @whalehunting 
the bnb model is blowing up our DEX job in prod, running >1 hour, much longer than any other model in the job. plz see below for reference. we should look into what could be causing this and fix prior to re-enabling in prod.
image


jeff-dude:
update: i see now that @0xRobin has reached out to you already. i will close this for now, but we should look for the fix asap 🙏

PR #7415, opened 2025-01-03
saddle finance optimism: exclude troubled model

jeff-dude:
fyi @Hosuke @PatelPrinci 
looks like the mappings here are incorrect, where decimal assignment of USDC / DAI are getting mixed up (6 vs 18), leading to incorrect `amount_usd` calculations downstream. are we sure `0` shouldn't be `usdc`, and dai as `1`?

here is sample tx / event index where you can see raw amounts (smaller value should likely be USDC)
https://optimistic.etherscan.io/tx/0x97d9bbde821ad796da5f70cd02195e0b3be1468ee5ebda5c5e7418a02c58984c#eventlog#38

### DIFF
--- a/dbt_subprojects/dex/models/trades/optimism/dex_optimism_base_trades.sql
+++ b/dbt_subprojects/dex/models/trades/optimism/dex_optimism_base_trades.sql
@@ -5,6 +5,8 @@
     )
 }}
 
+--exclude, until resolved at model level: , ref('saddle_finance_optimism_base_trades')
+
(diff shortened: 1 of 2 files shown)


Hosuke:
Oops...Sorry.

jeff-dude:
@Hosuke can you attempt to modify saddle finance model, then open PR to see it run via CI along with modified change at `dex.trades` level, so we can see `amount_usd` applied and test the data output?

PR #7425, opened 2025-01-06
Test saddle finance

Hosuke:


Hosuke:
Hi @jeff-dude 

Sorry, I think I canont simply used soldId/boughtId token mappings for this swap.
The tokens are linked to the transfer events in the tx before the swap.

Original wrong trades:
https://dune.com/queries/4529272

After updated mapping wrong trades:
https://dune.com/queries/4529336

So the mapping should be dynamic.

jeff-dude:
thanks for digging in deeper and following up!

i think we can feel comfortable with this output. i've also noticed, the DEX may no longer be running 🤔 
image

most recent block_time: `2023-08-14`

all good though, we'll at least have the history

PR #7443, opened 2025-01-09
BebopJam new contract address

B1boid:
Added new contract address for BebopJam
Pretty straightforward change, since the format of data is the same


Hosuke (approved):
LGTM.✅
Thank you @B1boid

jeff-dude (approved):

PR #7462, opened 2025-01-14
Fix bebop bug

B1boid:
Added sanity check to exclude empty arrays, mentioned in https://github.com/duneanalytics/spellbook/pull/7456#issue-2783654743 and returned model back to prod


jeff-dude (approved):
thank you, will look to re-enable with this filter applied 🙏

PR #7470, opened 2025-01-15
Delete duped icecream swap

Hosuke:


dbt_subprojects/dex/models/dex_info.sql
```diff
@@ -200,5 +200,4 @@ FROM (VALUES
     , ('gravity_finance', 'GravityFinance', 'Direct', 'Gravity_Finance')
     , ('fluid', 'Fluid DEX', 'Direct', '0xfluid')
     , ('beets', 'Beets DEX', 'Direct', 'beets_fi')
-    , ('icecreamswap', 'IceCreamSwap', 'Direct', 'icecream_swap')
```
Hosuke, line 203: Same line existed at L163.
https://github.com/duneanalytics/spellbook/blob/e0171bde2115e100d78749dc3c2a435294eff0a4/dbt_subprojects/dex/models/dex_info.sql#L163

jeff-dude:
good catch -- should we also include a test in this PR to ensure that first column is unique?

Hosuke:
> good catch -- should we also include a test in this PR to ensure that first column is unique?

We already have the test, but CI did not finish last time:
https://github.com/duneanalytics/spellbook/blob/Hosuke-patch-1/dbt_subprojects/dex/models/_schema.yml#L13

PR #7475, opened 2025-01-16
[easy] add erc20 tokens in tokens_sonic_erc20.sql

0xroll:

### Description:

[...]

---


0xroll:
quick question, i have a bigger list (~5k tokens). is it okay to push those into the table as well? they are retrieved with eth_calls

dbt_subprojects/tokens/models/tokens/sonic/tokens_sonic_erc20.sql
```diff
@@ -38,4 +38,9 @@ FROM (VALUES
     , (0xddf26b42c1d903de8962d3f79a74a501420d5f19, 'EQUAL', 18)
     , (0x313636D4f23859142b523a7965B76F6e3965Af64, 'SonicSwap', 18)
     , (0x4eec869d847a6d13b0f6d1733c5dec0d1e741b4f, 'INDI', 18)
+    , (0x4D85bA8c3918359c78Ed09581E5bc7578ba932ba, 'stkscUSD', 6)
+    , (0x455d5f11Fea33A8fa9D3e285930b478B6bF85265, 'stkscETH', 18)
+    , (0x541FD749419CA806a8bc7da8ac23D346f2dF8B77, 'SolvBTC', 18)
+    , (0xCC0966D8418d412c599A6421b760a847eB169A8c, 'solvBTC.bbn', 18)
+    , (0x0000000000000000000000000000000000000000, 'S', 18)
```
jeff-dude, line 45: i'm realizing we don't add the native token to many other chains (if any?) -- i'm curious your specific use case here, and why it would differ than all other EVM chains. i wonder if the metadata to tie to native token is just handled at query level for other chains 🤔

jeff-dude:
> quick question, i have a bigger list (~5k tokens). is it okay to push those into the table as well? they are retrieved with eth_calls

ideally we avoid bulk entries like this, as we read from an automated source for erc20 metadata now. these files remained in place just to handle long-tail tokens not covered in automated source. can you explain why there would be such a large batch missing in the automated source, are they unique token standards (or still erc20)?

dbt_subprojects/tokens/models/tokens/sonic/tokens_sonic_erc20.sql
0xroll: i think should be fine removing it. have pushed to remove that

0xroll:
> > quick question, i have a bigger list (~5k tokens). is it okay to push those into the table as well? they are retrieved with eth_calls
> 
> ideally we avoid bulk entries like this, as we read from an automated source for erc20 metadata now. these files remained in place just to handle long-tail tokens not covered in automated source. can you explain why there would be such a large batch missing in the automated source, are they unique token standards (or still erc20)?
> 
> edit: 4k+ tokens found here, which feed into `tokens.erc`
> 
> ```
> select *
> from dune.definedfi.dataset_tokens
> where networkid = 146
> limit 10000
> ```

oh i see! then i think it should be okay then. i was mainly looking at the tokens in this file

jeff-dude (approved):

PR #7495, opened 2025-01-19
add photon

clizzard7:
add spellbook for photon trading bot dashboard: 
https://dune.com/whale_hunter/photon-trading-bot


jeff-dude:
this is going to be one of the bot trades models that struggle to run, due to the amount of massive source tables involved in reads/joins:
- `solana.account_activity`
- `dex_solana.trades`
- `prices.usd`
- `solana.transactions`

these are all v large tables and we're seeing the engine struggle on the joins. this is an issue in prod already, where anytime we rebuild `dex_solana.trades`, the bot trades pipeline after can't rebuild due to so many running and written similar to this one here.

before proceeding here, we should find ways to break this up into smaller pieces that feed into a final model for photon. if we get that in a good spot, it should be repeatable for other bot trades models. please take some time to brainstorm how to solve this here

Issue #7235, opened 2024-12-04
Lending sector on Base - Add Morpho Blue [REQUEST]


sealaunchxyz:
### Description
Add Morpho Blue to lending sector on Base. 

### Impacted model(s)
lending.supply
lending.borrow

sealaunchxyz:
Not a bug but a request. Tagging @tomfutago for visibility

jeff-dude:
closing issue due to inactivity. if still relevant, and proposed fix is available, feel free to reopen an issue as needed.

PR #7498, opened 2025-01-20
Morpho on Base

0xdatawolf001:
This is a PR to add Morpho (Base) to the lending sector tables based on the requests from SeaLaunch 

https://github.com/duneanalytics/spellbook/issues/7235


jeff-dude:
> Not super sure how to fix this error? I can find the tables on Dune (front end). Is there some dependencies in some sql or yml file that I have not configured?

any time you use jinja to call a source in dbt, you need to declare in the defined sources directory. for your use case, that would be somewhere here:
https://github.com/duneanalytics/spellbook/tree/main/sources/_sector/lending

0xdatawolf001:
corrected the relevant bits

jeff-dude (approved):
nice clean add here, thank you 🔥 

fyi, we have this lending sector spell in our backlog to revisit and enhance when we get the time. it could be a good table to build out with extensive coverage and validate the table schema and such.

PR #7511, opened 2025-01-21
Update Event Models for GMX v2

AIDataMaster:

### Description:

This PR modifies the `*_withdrawal_created` dbt models to include three new columns:
- withdrawal_type
- long_token_swap_path
- short_token_swap_path

Additionally, we changed the materialization of `gmx_v2_arbitrum_erc20` and `gmx_v2_avalanche_c_erc20` from **views** to **tables**.

---


jeff-dude:
thank you for initiating the PR. let's talk `http_get()` usage again, since it's being revisited here to materialize into a table due to querying the view outputting errors occasionally.

we don't use this function much in spellbook at this time. mostly due to the exact scenario you are running into, and it can lead to prod failures on our cluster the same as it would on a view query execution. if it's an isolated dbt lineage of models, it's usually fine, as next run can resolve itself and it doesn't block other models. ideally though, we would still keep that function out of spellbook if possible.

what do you think of [this](https://github.com/duneanalytics/spellbook/blob/570b50e35714f7550c8c68197e7de8b605fbbec9/dbt_subprojects/hourly_spellbook/models/_sector/dune/dune_blockchains.sql#L11-L15) approach? in practice, it would be:
- materialized view built on the dune app, using the GMX team account as the owner of it. logic for http_get call would live in that mat view. you can refresh on the schedule which makes sense for you
- create a spell view, which reads from mat view on the GMX team account

AIDataMaster:
"Create a spell view that reads from a materialized view on the GMX team account."
- What will happen if the materialized view is deleted later?
- Could you provide an example of how to reference a materialized view?

I don't want to remove these dbt models `gmx_v2_arbitrum_erc20` and `gmx_v2_avalanche_c_erc20` because other models reference them, so it would be good to create mat. view on Dune App and then reference it inside of mentioned models.

jeff-dude:
> * What will happen if the materialized view is deleted later?

the view will fail at query execution time on the app, saying table doesn't exist. it will signal to you and team that you need to rebuild your mat view.

> * Could you provide an example of how to reference a materialized view?

syntax: `{{ source("dune", "result_blockchains", database="dune") }}` but replace with your GMX team name / mat view name

source file:
https://github.com/duneanalytics/spellbook/blob/c342d7e5e29a8135a1bd846310d30ab9a18a2b0e/sources/_datasets/dune/sources.yml#L1-L6

*should* be all you need

AIDataMaster:
> > * What will happen if the materialized view is deleted later?
> 
> the view will fail at query execution time on the app, saying table doesn't exist. it will signal to you and team that you need to rebuild your mat view.
> 
> > * Could you provide an example of how to reference a materialized view?
> 
> syntax: `{{ source("dune", "result_blockchains", database="dune") }}` but replace with your GMX team name / mat view name
> 
> source file:
> 
> https://github.com/duneanalytics/spellbook/blob/c342d7e5e29a8135a1bd846310d30ab9a18a2b0e/sources/_datasets/dune/sources.yml#L1-L6
> 
> _should_ be all you need

Thank you for your help. I will do it this way and tag you when it's ready for review.

AIDataMaster:
@jeff-dude could you help me, please? where should I place "dune" and where to locate team username "gmx-io"?

AIDataMaster:
I have to compile this name `dune."gmx-io".result_gmx_v_2_erc_20_from_api`

jeff-dude (approved):
looks like my commit worked. not sure if it was single quotes or what, but this should be proper formatting for mat views 🙏 

our dbt source macro compiles to include double quotes around source database/schema/table by default, which i think helps here especially with the `gmx-io` team name having a dash in there

AIDataMaster:
> looks like my commit worked. not sure if it was single quotes or what, but this should be proper formatting for mat views 🙏
> 
> our dbt source macro compiles to include double quotes around source database/schema/table by default, which i think helps here especially with the `gmx-io` team name having a dash in there

Thank you for your help. I have reviewed your commit. You can merge it if all is fine for you too.

PR #7546, opened 2025-01-24
update amount usd macro to include native prices

jeff-dude:

### Description:

uniswap v4 introduced swapping native tokens directly, rather than wrapped. switching to the prices table that was built to include native prices using `0x000...` addresses.

i think some further enhancements can be done in this pricing pipeline:
- change usage of `evms.info` to `dune.blockchains` (note: i tried this, but realized data type issues on address, so taking the quick win here for now until we finalize that table usage)

---


jeff-dude:
running some regression test queries to ensure this won't break non-native token pricing that feeds into dex trades. looks like we're safe to just switch to this pricing table from what i can see. i ran the below with no results returned as expected. then i flipped the old / new and also got no results.
```
with
    old as (
        select
            *
        from
            prices.usd
        where
            minute >= timestamp '2025-01-01'
            and blockchain is not null --blockchain is only null for native tokens
            and blockchain not in ('bitcoin', 'cardano') --evm only
    ),
    new as (
        select
            *
        from
            prices.usd_with_native
        where
            minute >= timestamp '2025-01-01'
            and contract_address != 0x0000000000000000000000000000000000000000
    )
select
    *
from
    old
except
select
    *
from
    new
limit
    1000
```

PR #7618, opened 2025-02-03
Latest balances EVM models

et-dynamic:
The idea here is to materialize a table with the latest balances for each address x token. I plan on adding the other chains which have `tokens_*.balances` tables once I know this one works.

Open questions:

- Better to just materalize as "table" and have it full refresh daily?
- Is it appropriate to build off of the `balances` tables (and have to group by a bunch of columns)? This seems better than just aggregating off of the source table and then duplicating all of the join logic in `tokens_*.balances`.


et-dynamic:

jeff-dude:
i'm hesitant to add another balances spell, as we are not full satisfied with current balances setup (more to come on this in future). is there any reason you can't just query the existing `balances_daily_agg` tables found in similar directory paths, then filter your query to last one day or so?

et-dynamic:
I'm assuming you mean the `balances_daily` tables, as filtering the `balances_daily_agg` tables for the most recent day or so will **not** provide you with all current balances - only those balances which have changed in the most recent day or so:

```
-- this returns nothing even though this wallet DOES have a nonzero balance in this token
select *
from tokens_ethereum.balances_daily_agg
where
  address = 0x47C28843EdfC54FDE4C71cDCab6b918995F5bA90
  and token_address = 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48
  and day > timestamp '2025-02-01'
```

So if we are instead talking about the `balances_daily` views, they just do a lot of unnecessary work in order to get us to current balances: a complete forward fill of every wallet x token. It will technically work (try replacing the table in the above query with `tokens_ethereum.balances_daily` and you'll get a result), but it is not practical when querying for large numbers of wallets. I've tried, and it times out even on a Large engine. On the other hand, the kind of `max_by()... from balances` logic in my PR works just fine for the same large lists of wallets.

In my opinion, it would be very useful to have tables like the ones I'm proposing -- ones that provide the latest balances for every token in every wallet. It seems wasteful for me to continually run these huge `max_by()` aggregations for what seems like a really common use case: what are the current balances for every wallet's every token?

I also put together [this query](https://dune.com/queries/4398718) to show how much smaller these tables would be when compared to the `balances` tables. The rightmost column is "for this chain, what is the approximate rowcount of a 'latest balances' table as a percentage of the full 'balances' table?"

Though I totally understand if y'all are reworking your balances model and you don't wanna add too much more stuff here. If that's the case, consider this PR a vote for what the new model should include :)

jeff-dude:
> I'm assuming you mean the `balances_daily` tables, as filtering the `balances_daily_agg` tables for the most recent day or so will **not** provide you with all current balances - only those balances which have changed in the most recent day or so:
> 
> ```
> -- this returns nothing even though this wallet DOES have a nonzero balance in this token
> select *
> from tokens_ethereum.balances_daily_agg
> where
>   address = 0x47C28843EdfC54FDE4C71cDCab6b918995F5bA90
>   and token_address = 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48
>   and day > timestamp '2025-02-01'
> ```
> 
> So if we are instead talking about the `balances_daily` views, they just do a lot of unnecessary work in order to get us to current balances: a complete forward fill of every wallet x token. It will technically work (try replacing the table in the above query with `tokens_ethereum.balances_daily` and you'll get a result), but it is not practical when querying for large numbers of wallets. I've tried, and it times out even on a Large engine. On the other hand, the kind of `max_by()... from balances` logic in my PR works just fine for the same large lists of wallets.
> 
> In my opinion, it would be very useful to have tables like the ones I'm proposing -- ones that provide the latest balances for every token in every wallet. It seems wasteful for me to continually run these huge `max_by()` aggregations for what seems like a really common use case: what are the current balances for every wallet's every token?
> 
> I also put together [this query](https://dune.com/queries/4398718) to show how much smaller these tables would be when compared to the `balances` tables. The rightmost column is "for this chain, what is the approximate rowcount of a 'latest balances' table as a percentage of the full 'balances' table?"
> 
> Though I totally understand if y'all are reworking your balances model and you don't wanna add too much more stuff here. If that's the case, consider this PR a vote for what the new model should include :)

okay, thank you for detailed response. i have pushed a few changes to enhance the setup:
- partition for performance gains (include in unique key for merge statement join conditions)
- add incremental jinja logic

let's see how this runs. then we can figure out next steps.

jeff-dude:
this performs pretty well for fully history. let me check in with the team on what plans lay ahead for balances, otherwise i think this is reasonable as-is.

jeff-dude (approved):
okay, we're all good to proceed here. thanks again for initiating and giving details of usage. i'll prep this one for merge, then you can use a bit in prod once live and revisit to add other chains as you see fit. plz note that we don't have all chains on balances yet, and the plan to include them all isn't final quite yet.

PR #7626, opened 2025-02-05
rollup_economics v3

lorenz234:

### Description:

After getting feedback from the other rollup_economic wizards we agreed to change the spell to use a mapping rather than inline filters: dune.growthepie.l2economics_mapping

this should reduce runtime and also enable future additions such as rent paid to Celestia and other altDAs.

---


lorenz234:
After receiving feedback from the other rollup_economic wizards, we agreed to modify the spell to use a mapping instead of inline filters. This version also allows for future extensions, such as rent payments to Celestia and other altDAs. The mapping can be found in this table: `dune.growthepie.l2economics_mapping` and is actively maintained by the growthepie team here: [[GitHub - growthepie/gtp-dna](https://github.com/growthepie/gtp-dna/tree/main/economics_da)](https://github.com/growthepie/gtp-dna/tree/main/economics_da).

Some dashboards will break due to changes in names and columns—this has been communicated to the relevant individuals.  

One dbt question remains:  
Ideally, we want this spell to run incrementally every hour. However, it should only re-run in full if `dune.growthepie.l2economics_mapping` is updated (tracked by detecting hash changes), as historical data would then need to be refreshed. Since I'm relatively new to dbt, I’d appreciate it if someone from the Dune team could double-check the current setup. If it works as intended, it should significantly reduce runtime compared to executing the full query every time.

Hosuke (approved):

dbt_subprojects/hourly_spellbook/models/_sector/rollup_economics/ethereum/beacon_fees/rollup_economics_ethereum_beacon_fees.sql
```diff
@@ -0,0 +1,104 @@
...
+-- main query
+L1_methods AS (
+    SELECT 
+        hash,
+        bytearray_substring("data", 1, 4) AS method,
+        "to" AS to_address
+    FROM {{ source('ethereum', 'transactions')}}
+    WHERE 
+        TYPE = '3'
+        AND block_time >= TIMESTAMP '2024-03-13' -- EIP-4844 launch date
```
Hosuke, line 52: We may add an incremental filter here.

Hosuke (changes requested):

Hosuke:
> After receiving feedback from the other rollup_economic wizards, we agreed to modify the spell to use a mapping instead of inline filters. This version also allows for future extensions, such as rent payments to Celestia and other altDAs. The mapping can be found in this table: `dune.growthepie.l2economics_mapping` and is actively maintained by the growthepie team here: https://github.com/growthepie/gtp-dna/tree/main/economics_da
> 
> Some dashboards will break due to changes in names and columns—this has been communicated to the relevant individuals.
> 
> One dbt question remains: Ideally, we want this spell to run incrementally every hour. However, it should only re-run in full if `dune.growthepie.l2economics_mapping` is updated (tracked by detecting hash changes), as historical data would then need to be refreshed. Since I'm relatively new to dbt, I’d appreciate it if someone from the Dune team could double-check the current setup. If it works as intended, it should significantly reduce runtime compared to executing the full query every time.

I will draft my idea in the next commit for your reference, if anything is not as expected we can just revert my commit.

dbt_subprojects/hourly_spellbook/macros/models/_sector/rollup_economics/check_mapping_hash.sql
```diff
@@ -0,0 +1,42 @@
+{% macro check_mapping_hash(settlement_layer) %}
```
Hosuke, line 1: Ideally we may add an macro to check if the spell need to be refreshed, also this saves the common code blocks

Hosuke:
Seems my idea does not work, sorry.
The question is still open.

jeff-dude:
i can see what you're attempting to do here, but i think we would require some more testing to ensure it works as intended. as of now, CI logs show that both history + incremental run processed same number of rows, which means the logic thinks hash changed and needs to ignore timestamp filter. i doubt that is the case, as they run back-to-back so quickly?

i spent some time looking closer at this, but to feel comfortable, i would like to see test cases done in PR here for each scenario and CI run logs to prove each test that it works as you set up to.

i'm leaning towards we save all that effort and just make these `table` materialization, as they build in such a short amount of time. less than 100 seconds. i'm not sure it's worth the extensive dev/test effort?

in addition to that, can you tell us how your mat view used as a source here is built? is there any way it can be built in spellbook instead? we don't have a ton of use cases for mat views as source from teams outside of `dune`, which we can control. it  can introduce unknowns into our prod orchestration that is outside of our control to fix.

sources/_subprojects_outputs/daily_spellbook/_sources.yml
```diff
@@ -41,6 +41,7 @@ sources:
   - name: ethereum
     tables:
       - name: blobs_submissions
+      - name: blobs
```
jeff-dude, line 44: this source should be in the hourly spellbook directory, as it's built  in that project

lorenz234:
Thanks for the comments, I will get on this next week!

sources/_subprojects_outputs/daily_spellbook/_sources.yml
lorenz234: good point, moved!

lorenz234:
Hi @jeff-dude,

I removed all the overly complicated hashing updates; now, all four tables are materialized views. They run fast, except for one table, which takes 493 seconds. I hope that is acceptable for a daily task. Furthermore, I moved all of them into the daily spellbook group.

Regarding the source `dune.growthepie.l2economics_mapping`, it is actively maintained by the growthepie team here: https://github.com/growthepie/gtp-dna/tree/main/economics_da. Formatting this in a YAML file makes it much simpler to view, and more data teams from these L2s are directly encouraged to contribute, leading to a more complete mapping. With the use of a GitHub Action, as soon as a new commit is merged, the Dune API is used to update the table.

Let me know if there is anything else that should be changed.

Regards 
Lorenz

jeff-dude:
> Hi @jeff-dude,
> 
> I removed all the overly complicated hashing updates; now, all four tables are materialized views. They run fast, except for one table, which takes 493 seconds. I hope that is acceptable for a daily task. Furthermore, I moved all of them into the daily spellbook group.
> 
> Regarding the source `dune.growthepie.l2economics_mapping`, it is actively maintained by the growthepie team here: https://github.com/growthepie/gtp-dna/tree/main/economics_da. Formatting this in a YAML file makes it much simpler to view, and more data teams from these L2s are directly encouraged to contribute, leading to a more complete mapping. With the use of a GitHub Action, as soon as a new commit is merged, the Dune API is used to update the table.
> 
> Let me know if there is anything else that should be changed.
> 
> Regards Lorenz

any reason they are all full tables, rather than build incrementally? is it because the mapping table from other repo updates frequently, and would require these tables to trigger a full refresh? since it's a jinja source here, that wouldn't happen automatically.

PR #7635, opened 2025-02-08
Add gas fees for new chains

Hosuke:
I should use TDD for this PR development.

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/abstract/gas_abstract_schema.yml
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/abstract/gas_abstract_schema.yml
@@ -1,0 +1,31 @@
+version: 2
+
+models:
+  - name: gas_abstract_fees
+    meta:
+      blockchain: abstract
+      sector: gas
+      contributors: hosuke
+    config:
+      tags: ['abstract', 'gas', 'fees']
+    description: >
+        Gas Fees on Abstract
+    data_tests:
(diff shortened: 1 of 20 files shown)


dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/abstract/gas_abstract_schema.yml
```diff
@@ -0,0 +1,31 @@
+version: 2
+
+models:
+  - name: gas_abstract_fees
```
Hosuke, line 4: official docs: https://docs.abs.xyz/how-abstract-works/evm-differences/gas-fees

Hosuke:
Too many chains in one pr is hard for debugging, we may need to reduce batch size next time.

Hosuke:
I think the current output of the views makes sense, but the column definition in the seed may be ambiguous for some chains.

Hosuke:
[Gas fees sample](https://dune.com/queries/4827365)

dbt_subprojects/hourly_spellbook/seeds/_sector/gas/gas_used_by_txn_seed.csv
Hosuke, line 1: I think we need to design a new form for gas testing, but I have no idea yet. Currently all I know is that `Gas Used by Txn` need to match the `gas_used` column.

jeff-dude:
closing since we broke out macro into models. we can talk tomorrow on plans moving forward.

PR #7694, opened 2025-02-20
Add latest balances models to L2s

et-dynamic:
Copying the Ethereum `balances_latest` model introduced in https://github.com/duneanalytics/spellbook/pull/7618 over to L2s for which it is possible (that is, those which have the necessary `balances` model).

@jeff-dude The Ethereum `balances_latest` model REALLY sped up some queries - by as much as 4x. Thanks again for the help on that one. Let me know if you think it's appropriate to expand this to these other L2s. It would definitely be helpful for my use of Dune.


jeff-dude:
> yeah, this should be fine to continue per chain. if CI fails again, it's likely bnb is crashing the cluster. we know that chain has heavy data volume. should still be fine in prod.

confirmed we're crashing the CI cluster. it's a bit smaller. we can try to scale it up next week and ensure this runs in a reasonable timeframe to prep for prod.

jeff-dude:
trying now on cluster 4x the size

PR #7711, opened 2025-02-22
Fix solana bot trades ambiguous column selection

jeff-dude:
solana source tables were recently rebuilt on dune & they include additional columns. this led to breaking changes in spellbook, all in the `bot_trades` lineage. these models were selecting `version` column from `dex_solana.trades`, but it also joins to raw transactions which now has the `version` column as well. the query did not use table alias in column selection, leading to ambiguous column error:
```
15:07:39  Failure in model wifbot_solana_bot_trades (models/_sector/dex/bot_trades/solana/platforms/wifbot_solana_bot_trades.sql)
15:07:39
15:07:39    Database Error in model wifbot_solana_bot_trades (models/_sector/dex/bot_trades/solana/platforms/wifbot_solana_bot_trades.sql)
  TrinoUserError(type=USER_ERROR, name=AMBIGUOUS_NAME, message="line 48:13: Column 'version' is ambiguous", query_id=20250222_150149_14775_x8rg3)
```
this impacted 28 models, should all be included in PR fix here.

fyi @whalehunting @clizzard7 
moving forward, plz always use table aliases in column selections when performing joins as a best practice. it helps safeguard this scenario. while rare, still a good practice to adopt in these pipelines.

### DIFF
--- a/dbt_subprojects/solana/models/_sector/dex/bot_trades/solana/platforms/alpha_dex_solana_bot_trades.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/bot_trades/solana/platforms/alpha_dex_solana_bot_trades.sql
@@ -44,7 +44,7 @@ with
             'SOL' as fee_token_symbol,
             '{{wsol_token}}' as fee_token_address,
             project,
-            version,
+            trades.version,
             token_pair,
             project_program_id as project_contract_address,
             trader_id as user,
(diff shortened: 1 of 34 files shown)


jeff-dude:
seed tests failed as expected, since i intentionally filtered to an incremental run and seed rows are older dates. i simply wanted to ensure models can still run on incremental. this fix is considered prod ready now.

PR #7747, opened 2025-02-26
attempt 3 on v4 dex.trades - joining swap call to swap event to get router info and better fee emitted

agaperste:

### Description:

- now swap amount is figured out 
- we need to have grab the sender aka the router from swap event
- so inner joining swap call to swap event (by block_number, tx_hash, and a row number of the calls/events in the same txn)
- we can do this because every call will always emit an event, regardless of accuracy of amounts emitted in event 
- this way we also get "clean" evt_index for any future join
- we also try to preserve other info like hooks, tick etc. for future usage

---


agaperste:
@jeff-dude it's ready! (I updated the branch again so it triggered CI run again just now I think)

jeff-dude (approved):

PR #7749, opened 2025-02-27
fix(gas): removed `incremental_predicate`, add 6 months gas usage column

sohjag:
### Description:

Removed `incremental_predicate` from base CTE since some metrics require processing the entire historical dataset or large time windows. 

### Key changes:

 - Removal of `incremantal_predicate`
 - Addition of 6 months gas usage columns (native currency and USD)
 - Added `post_hook`


pfedan:

pfedan:
recheck

dbt_subprojects/daily_spellbook/models/_sector/gas/prices/cross_chain/gas_usage_by_address_schema.yml
```diff
@@ -5,9 +5,12 @@ models:
     meta:
       blockchain: ['ethereum', 'optimism', 'base', 'arbitrum', 'avalanche_c', 'blast', 'bnb', 'celo', 'fantom', 'gnosis', 'linea', 'mantle', 'polygon', 'scroll', 'zkevm', 'zksync', 'zora']
       sector: gas
-      contributors: ['gashawk']
+      contributors: ['gashawk', 'zohjag']
     config:
       tags: ['gas', 'cross_chain', 'addresses']
+    indexes:
+      - columns: ["address"]
+        type: btree
```
jeff-dude, line 13: i believe this is first time i've seen btree index used in spellbook, so i did a little research. here is what chatgpt is telling me:
> - Trino does not use traditional indexes like btree or hash indexes.
> - Delta Lake also does not support btree indexes (it relies on metadata, partitioning, and sorting).
> - Adding indexes: in dbt’s schema.yml won't impact performance because Trino will ignore it

jeff-dude: lets remove this here and focus on trying to add optimizations in the model itself. i will comment an idea there.

dbt_subprojects/daily_spellbook/models/_sector/gas/prices/cross_chain/gas_usage_by_address.sql
```diff
@@ -1,14 +1,18 @@
...
-    materialized='incremental',
+    materialized='table',
     file_format='delta',
-    incremental_strategy='merge',
     partition_by = ['blockchain'],
     unique_key=['address', 'blockchain', 'currency_symbol'],
+    post_hook='{{ expose_spells(\'["ethereum", "bnb", "avalanche_c", "gnosis", "optimism", "arbitrum", "fantom", "polygon", "base", "celo", "zora", "zksync", "scroll", "linea", "zkevm"]\',
+                        "sector",
+                        "gas",
+                        \'["gashawk, zohjag"]\') }}'
```
jeff-dude, line 11:
```suggestion
    post_hook=[
        '{{ expose_spells(\'["ethereum", "bnb", "avalanche_c", "gnosis", "optimism", "arbitrum", "fantom", "polygon", "base", "celo", "zora", "zksync", "scroll", "linea", "zkevm"]\', "sector", "gas", \'["gashawk, zohjag"]\') }}',
        'OPTIMIZE {{ this }} ZORDER BY (address);'
    ]
```

jeff-dude: i haven't actually tried this optimize / zorder, so i'm curious to see how it performs on build time (or if it even works). give this a shot plz 🙏

jeff-dude (changes requested):

jeff-dude:
models in general built around address solely can be tough to build / query. the varbinary field alone is not efficient in these setups.

dbt_subprojects/daily_spellbook/models/_sector/gas/prices/cross_chain/gas_usage_by_address_schema.yml
sohjag: Removed `btree` [1066649](https://github.com/duneanalytics/spellbook/pull/7749/commits/1066649db26b7960c72a505bbe0aad136b6f747b)

dbt_subprojects/daily_spellbook/models/_sector/gas/prices/cross_chain/gas_usage_by_address.sql
sohjag: Trying it out [457ac7e](https://github.com/duneanalytics/spellbook/pull/7749/commits/457ac7e051285c4e70435613694938d6317041d4)

sohjag: Didn't work :/ . reverted [9a02354](https://github.com/duneanalytics/spellbook/pull/7749/commits/9a02354cf9dfa0a0fe5b7d88d5914f88a14f0ed6)

sohjag:
> models in general built around address solely can be tough to build / query. the varbinary field alone is not efficient in these setups.

What if we add a hex string representation of the address? Also partition by month as well may be?

dbt_subprojects/daily_spellbook/models/_sector/gas/prices/cross_chain/gas_usage_by_address_schema.yml
```diff
@@ -5,9 +5,11 @@ models:
     meta:
       blockchain: ['ethereum', 'optimism', 'base', 'arbitrum', 'avalanche_c', 'blast', 'bnb', 'celo', 'fantom', 'gnosis', 'linea', 'mantle', 'polygon', 'scroll', 'zkevm', 'zksync', 'zora']
       sector: gas
-      contributors: ['gashawk']
+      contributors: ['gashawk', 'zohjag']
     config:
       tags: ['gas', 'cross_chain', 'addresses']
+    indexes:
+      - columns: ["address"]
```
jeff-dude, line 12:
```suggestion
```

jeff-dude: remove since we believe this does nothing for delta / trino setup

jeff-dude (changes requested):

jeff-dude:
other than the above minor cleanup, i think this setup looks good 🔥 

the `address_prefix` seems to distribute rows to partitions well (ran a query to group by and see counts). i would be sure to use that in join conditions for any queries downstream, or potential tables built on top.

should be interesting to see how the z-order works here on address. not sure exactly how to measure impact. i guess we would have to build table with and without, but we don't need to go down that path, as it doesn't appear to slow build time at all.

jeff-dude:
> One of the earlier runs created the initial table with 1.8B rows. The recent runs have been showing about 1.35B rows only. What could be the possible reasons for this?
> 
> image

if this is a blocker, plz do try to dig in a bit and see if a specific blockchain is missing data in `gas.fees` 🤔 i'm not sure tbh

dbt_subprojects/daily_spellbook/models/_sector/gas/prices/cross_chain/gas_usage_by_address_schema.yml
sohjag: Removed. [90990ae](https://github.com/duneanalytics/spellbook/pull/7749/commits/90990ae660f4e03b5c5fa2889182946cdc0070dd)

pfedan:
> should be interesting to see how the z-order works here on address. not sure exactly how to measure impact.

It would be great I there was a way to define test queries to be executed in CI in order to measure such effects on the model that has been created during the ["dbt_run"."dbt test initial model(s)"](https://github.com/duneanalytics/spellbook/blob/main/.github/workflows/dbt_run.yml#L57) CI-step.

Is that already possible? If not, that would be an awesome addition to the checks.

pfedan:
Done with experimenting 🙄

jeff-dude (approved):

pfedan:
Let's get this merged 🚀

PR #7763, opened 2025-02-28
DUX-963 fix polygon native token address in tokens transfers

beve0x:

### Description:

[...]

---


jeff-dude:
sample query used in testing, showing that with / without final filter for native address in traces shows slightly different row counts as expected (remove POL contract interactions):
```
        --24,086,548
        --24,080,959
        select
            count(1)
        from
            polygon.traces
        WHERE
            success
            AND (
                call_type NOT IN ('delegatecall', 'callcode', 'staticcall')
                OR call_type IS null
            )
            and block_date >= date '2025-02-01'
            and block_date < date '2025-03-01'
            AND "value" > UINT256 '0'
            AND CASE
                WHEN to = 0x0000000000000000000000000000000000001010
                OR "from" = 0x0000000000000000000000000000000000001010 THEN FALSE -- Exclude these
                ELSE TRUE -- Keep everything else
            END
```

jeff-dude:
to summarize on this PR:
- previously, `tokens_polygon.base_transfers` was outputting duplicates, since `POL` token is the native token but could be passed through the erc-20 equivalent contract, therefore appearing in both tables in the union
- we will now filter out from `traces` any transfers of the native token which passes through the erc-20 contract, as we will simply capture them from the cleaner erc20 source
- the impact isn't massive, as only a smaller subset of transfers actually ran into this scenario (see query above for small decrease in transfer count from transfers for february)

jeff-dude:
before proceeding with this PR, we should:
- revert changes involving shorter timeframe for CI
- open PR to update `MATIC` to `POL` in `tokens.erc20`, looks like source table there didn't update after the migration
- merge metadata change, ensure flowing into prod
- revisit here and merge

jeff-dude (approved):

PR #7767, opened 2025-03-01
Vacuum all tables

aalan3:
For some reason we only vacuum tables that are marked as public. This causes intermediary tables to become very huge. This PR changes the `mark_as_spell` macro so that all tables have vacuuming turned on.


dbt_macros/dune/mark_as_spell.sql
```diff
@@ -6,6 +6,7 @@
         {{ trino_properties(properties) }}
       )
     {%- else -%}
+      {%- set properties = { 'dune.data_explorer.category': 'abstraction', 'dune.vacuum' = '{"enabled":true}' } -%}
```
jeff-dude, line 9: i think it's cleaner to add above to line 3, where we already assign `abstraction` category. this will apply to all models that run in prod, i.e. dbt cloud

jeff-dude: but in general, i see your point. we only add in expose spell macro:
https://github.com/duneanalytics/spellbook/blob/main/dbt_macros/dune/expose_spells.sql#L14

jeff-dude (changes requested):

jeff-dude:
keep in mind, if you merge this, it may trigger `state:modified` on all subprojects. probably safe to turn off for this merge on all projects.

dbt_macros/dune/mark_as_spell.sql
aalan3: we only need to apply to tables not views. agree that this is not beautiful though 😄

jeff-dude: in `expose_spells` macro, we apply whether it's table or view. seems like it doesn't harm anything? ultimately, i guess either way will be fine. i'm okay with leaving as-is or matching how `expose_spells` does it

jeff-dude (approved):
will approve and leave final call up to you so it's not a blocker

dbt_macros/dune/mark_as_spell.sql
aalan3: Update this to be a bit more readable. 

We want to limit the number of tables we list when doing this so that the job runs faster

jeff-dude: much more readable, thank you

jeff-dude (approved):
we should wait to merge until prod settles down a bit with the other activities happening

jeff-dude (approved):

PR #7775, opened 2025-03-03
Update gas solana fees

aalan3:
This will rebuild the table and repartition by block_date instead of block_date + block_hour.

### DIFF
--- a/dbt_subprojects/solana/models/_sector/gas/gas_backfills/tx_fees/gas_solana_tx_fees_current.sql
+++ b/dbt_subprojects/solana/models/_sector/gas/gas_backfills/tx_fees/gas_solana_tx_fees_current.sql
@@ -5,7 +5,7 @@
     materialized = 'incremental',
     file_format = 'delta',
     incremental_strategy = 'delete+insert',
-    unique_key = ['block_date', 'block_hour', 'block_slot', 'tx_index']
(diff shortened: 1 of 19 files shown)


dbt_subprojects/solana/models/_sector/gas/gas_backfills/tx_fees/gas_solana_tx_fees_current.sql
```diff
@@ -1,14 +1,14 @@
 {{ config(
     schema = 'gas_solana',
     alias = 'tx_fees_current',
-    partition_by = ['block_date', 'block_hour'],
+    partition_by = ['block_date'],
     materialized = 'incremental',
     file_format = 'delta',
     incremental_strategy = 'delete+insert',
```
0xRobin, line 7: do want to get rid of the non atomic delete+insert while we're doing the refresh anyway?

0xRobin (approved):

dbt_subprojects/solana/models/_sector/gas/gas_backfills/tx_fees/gas_solana_tx_fees_current.sql
aalan3: good point, updated

PR #7782, opened 2025-03-03
fix POL token symbol in metadata

jeff-dude:


jeff-dude:
coinpaprika has incorrect symbol, can ignore. we don't use `POL` from coinpaprika due to lack of historical coverage.

PR #7783, opened 2025-03-04
DUX-992 Fix balance latest unique key per model

jeff-dude:
fyi @et-dynamic 
since we included `block_date_latest` in the model unique key config, the merge statement logic breaks and it inserts a new row every time (table has grown much more than it should). we should rather be only joining on the `unique_key` column value and ensuring we update `block_date_latest` along with the balance amounts. this way we ensure we only have one row per `unique_key` column value as expected.

### DIFF
--- a/dbt_subprojects/tokens/macros/balances/balances_latest.sql
+++ b/dbt_subprojects/tokens/macros/balances/balances_latest.sql
@@ -19,6 +19,8 @@ where
     1 = 1
     {% if is_incremental() %}
     and {{ incremental_predicate('block_time') }}
+    {% else %}
+    and block_time >= date '2025-02-01'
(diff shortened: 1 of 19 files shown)


aalan3 (approved):

0xRobin (commented):
fyi, I don't think partitioning on `block_date` for latest balances makes a ton of sense here. I'd remove that as well while we are doing the full refresh anyway with this merge

jeff-dude:
> fyi, I don't think partitioning on `block_date` for latest balances makes a ton of sense here. I'd remove that as well while we are doing the full refresh anyway with this merge

i went back and forth on that, so this confirmation helps. i will go ahead and remove.

PR #7854, opened 2025-03-13
add filter to only consider official whirlpool config

0xBoxer:
Filtering Orca Whirlpool trades to only consider those which use the official pool config account that lives at [2LecshUwdy9xi7meFgHtFJQNSKk4KdTrcpvaB56dP2NQ](https://solscan.io/account/2LecshUwdy9xi7meFgHtFJQNSKk4KdTrcpvaB56dP2NQ). 

I checked the other configs that sometimes get used to initilaize pools and they are all trash and can be disregarded.


jeff-dude (approved):
glad it was as simple as a filter, once we found the column / value to apply 🙏

jeff-dude:
(assumption) going to mark this as `blocked` to indicate that we are waiting to merge with a few other solana PRs

PR #7868, opened 2025-03-17
TON BagOfCells parsing framework + EVAA parser

shuva10v:

### Description:

The pull request contains TON Bag Of Cells (TON-specific data structure) implementation based on SQL and Jinja templates. And the first use case - [EVAA](https://evaa.finance/) lending protocol decoded tables. 

---


dbt_subprojects/daily_spellbook/models/evaa/ton/evaa_ton_liquidate.sql
```diff
@@ -0,0 +1,111 @@
+{{ config(
+       schema = 'evaa_ton'
+       , alias = 'liquidate'
+       , materialized = 'incremental'
+       , file_format = 'delta'
+       , incremental_strategy = 'merge'
+       , incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_date')]
+       , unique_key = ['tx_hash']
```
jeff-dude, line 8: can we include more than just `tx_hash` in unique key (and also in join conditions in model)?
trino has historically had poor query planning on these field types alone.
maybe `block_date` too?

jeff-dude: also remember to update the unique test in schema file after plz 🙏

jeff-dude (changes requested):
this is what it looks like to decode raw chain data in spellbook, huh? 😅

shuva10v:
@jeff-dude added block_date as you mentioned. Btw, I thought for a while to add block_date as a partition but since the tables are small (~10-100k rows) I decided it may even reduce performance

jeff-dude:
> @jeff-dude added block_date as you mentioned. Btw, I thought for a while to add block_date as a partition but since the tables are small (~10-100k rows) I decided it may even reduce performance

yeah, i think we avoid partitions for now. when we're at 1mil+ rows, we can consider, but only 100k or less for these as of now. even at low millions, you'd only want to do like `block_year` or something with large chunks.

jeff-dude (approved):
this looks good to me. great work on leveraging all of the tooling available here for your use case 🔥 

i want to be clear upfront around expectations on this lineage of models. due to the complexity, we likely won't be able to support quickly if issues arise in prod. if it's something simple we can fix, we will, but otherwise the process will be:
- dune team raises PR to disable in prod, using `prod_exclude` dbt tag on model(s)
- dune team will tag original creator (yourself) and share prod error message
- you can re-enable by removing tag and fixing issues in new PR
- dune team to review / merge again

we can typically iterate fast on the above, as long as you and team can too.
i have let nia know to mention this in your call about spellbook usage as well, so if you have follow up questions, plz do ask there 🙏 

edit: if you miss GH notifications for excluding in prod (hopefully it never happens!) -- then you would also notice by stale data.

ohld (approved):

PR #7871, opened 2025-03-17
DUX-1030 automated dex trades: enhance project level base trades

jeff-dude:


jeff-dude:
this appears to have no impact on data quality vs. prod, which is as intended. the goal is only to increase performance. ran various iterations of below query, but row counts match, inner join returns all rows, except statement returns no differences.
```
with
    prod as (
        select
            *
        from
            uniswap_v2_forks.automated_base_trades
        where
            block_date >= date '2025-03-16'
    ),
    ci as (
        select
            *
        from
            test_schema.git_dunesql_e957fba_uniswap_v2_forks_automated_base_trades
    )
    --ci: 3,647,075
    --join: 3,647,075
select
    *
from
    prod
except
select
    *
from
    ci
limit 100
    -- inner join prod on ci.blockchain = prod.blockchain
    -- and ci.block_month = prod.block_month
    -- and ci.block_number = prod.block_number
    -- and ci.tx_index = prod.tx_index
    -- and ci.evt_index = prod.evt_index
```

jeff-dude:
okay, looks like the answer is to pull out the subquery on evms view and materialize upfront. this has made the model much more efficient in CI.

we should consider this table as a staging table in this pipeline only, and not use it elsewhere to expose it to more use cases. we've seen that become hard to maintain over time. it build efficiently, so any other use cases in future should consider building it's own staging table.

PR #7925, opened 2025-03-25
chore: add LOCK token to prices

whalehunting:
Adds the LOCK token from Houdini Swap to prices
https://coinpaprika.com/coin/lock-houdini-swap/


jeff-dude (approved):
approving for now, but what is use case here? for querying on app?

we have re-released the new prices pipeline, docs here
https://docs.dune.com/data-catalog/curated/prices/overview
plz consider these for analysis moving forward, where ideally tokens you need are automatically added 🙏

PR #7957, opened 2025-04-01
Fix solana_utils token accounts

0xBoxer:
pushing a fix for solana_utils.token_accounts which currently wrongly assumes that token accounts cannot get reassigned to new token_mint_addresses.

### DIFF
--- a/dbt_subprojects/solana/models/solana_utils/solana_utils_token_accounts_timing.sql
+++ b/dbt_subprojects/solana/models/solana_utils/solana_utils_token_accounts_timing.sql
@@ -31,7 +31,7 @@ activity_for_processing AS (
     WHERE 
         act.writable = true
         AND act.token_mint_address IS NOT NULL
-        AND act.block_time >= DATE('2025-04-01') -- Test run
+        AND act.block_time >= DATE('2025-03-01') -- Test run
 ),
 
 -- Number activity chronologically per address for LAG calculations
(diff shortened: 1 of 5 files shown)


0xRobin:
@0xBoxer would be cool if we could already adopt conventions from `sqlmesh` SCD (Slow Changing Dimension) model kind, which is exactly what we're trying to do here.
https://sqlmesh.readthedocs.io/en/stable/concepts/models/model_kinds/?h=scd+ty#scd-type-2

Mainly using these column names:
- `valid_from`
- `valid_to`

0xBoxer:
good call @0xRobin, implemented in 7c618b3

dbt_subprojects/solana/models/solana_utils/solana_utils_token_accounts_timing.sql
```diff
@@ -0,0 +1,97 @@
+{{
+  config(
+        schema = 'solana_utils',
+        alias = 'token_accounts_timed',
+        materialized = 'incremental',
+        file_format = 'delta',
+        incremental_strategy = 'merge',
+        unique_key = ['address', 'token_balance_owner', 'token_mint_address', 'valid_from', 'valid_to'],
```
0xRobin, line 8: We'll need to remove the `valid_to` column from the unique key to be able to override older records with an updated `valid_to` value when a new address combo is activated.

PR #7985, opened 2025-04-07
Upgrade ens_set_name_latest to incremental view

Hosuke:
The `row_number() over (partition by address order by block_time desc...)` ensures incremental view correctly processes and updates the unique binding relationships between ENS names and addresses.


jeff-dude:
can you run a quick query test to ensure results in CI match prod? so we can be certain incremental logic doesn't break what is done in full refresh each run?

might be best to focus on the 60 rows ran in incremental step, to have a small subset of rows which we know ran in incremental and could break logic for whatever reason

Hosuke:
I did the incremental testing in this query:
https://dune.com/queries/4956221

I think all incremental records match the current prod table.

dbt_subprojects/hourly_spellbook/models/_project/ens/ens_set_name_latest.sql
```diff
@@ -3,13 +3,15 @@
         schema = 'ens'
         ,alias = 'set_name_latest'
         
-        ,materialized = 'table'
+        ,materialized = 'incremental'
+        ,incremental_strategy = 'merge'
         ,file_format = 'delta'
         ,unique_key = ['address', 'name']
```
Hosuke, line 9: Do you think the unique_key can be just `name`?

jeff-dude: likely, but the name + address combo will always be unique too. let's leave as-is.

jeff-dude (approved):

PR #8000, opened 2025-04-09
Materialize cex addresses spells

Hosuke:
Make `cex_evms` spells update incrementally.


Hosuke:
Comparison testing:
https://dune.com/queries/4958758

jeff-dude (approved):
thank you 🔥 

fyi @hildobby we are materializing these for prod performance. we confirmed results remain the same.

PR #8014, opened 2025-04-11
Add degen prices

Hosuke:
We may choose tokens from this list to add:
https://explorer.degen.tips/tokens

### DIFF
--- a/dbt_subprojects/tokens/models/prices/degen/prices_degen_tokens.sql
+++ b/dbt_subprojects/tokens/models/prices/degen/prices_degen_tokens.sql
@@ -1,0 +1,23 @@
+{% set blockchain = 'degen' %}
+
+{{ config(
+    schema = 'prices_' + blockchain,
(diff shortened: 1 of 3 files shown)


dbt_subprojects/tokens/models/prices/degen/prices_degen_tokens.sql
```diff
@@ -0,0 +1,23 @@
...
+SELECT
+    token_id
+    , '{{ blockchain }}' as blockchain
+    , symbol
+    , contract_address
+    , decimals
+FROM
+(
+    VALUES
+    ('degen-degen-base', 'DEGEN', 0x0000000000000000000000000000000000000000, 18)
```
jeff-dude, line 21: plz note, since we add here, we should *not* also add to `prices_native_tokens` or we will hit duplicates (it's not in there now, so should be fine)

jeff-dude (approved):

PR #8015, opened 2025-04-11
Add opbnb transfers

Hosuke:
towards CUR2-120


jeff-dude (approved):
okay, can confirm `net_transfer_daily_asset` requires rows in `prices.usd` due to final filter in macro for `net_transfer_amount_usd > 0`

we have a few rows in there, now showing up here.

PR #8019, opened 2025-04-13
feat: add abstractswap dex spellbook

clizzard7:


dbt_subprojects/dex/models/trades/dex_base_trades.sql
```diff
@@ -44,47 +45,48 @@
...
-        WHERE
-           token_sold_amount_raw >= 0 and token_bought_amount_raw >= 0
-        {% if is_incremental() %}
-            AND {{ incremental_predicate('block_time') }}
-        {% endif %}
-        {% if not loop.last %}
-        UNION ALL
-        {% endif %}
-        {% endfor %}
+with
```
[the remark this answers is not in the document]
whalehunting: done!

clizzard7:
@jeff-dude done!

jeff-dude:
have decoded tables changed? this is failing on tables that don't exist.

jeff-dude:
i'm still seeing the same errors

jeff-dude:
if you're unsure what the error is:
```
Compilation Error
  Model 'model.dex.abstractswap_v2_abstract_base_trades' (models/trades/abstract/platforms/abstractswap_v2_abstract_base_trades.sql) depends on a source named 'uniswap_abstract.pair_evt_swap' which was not found
```

this is still occurring

clizzard7:
image
hey @jeff-dude we use this query in the sources and we have data on dune so the table is correct. But the CI fails anways. Do you know what we need to change? thank you

sources/_sector/dex/trades/abstract/_sources.yml
```diff
@@ -0,0 +1,21 @@
+version: 2
+
+sources:
+  - name: uniswap_abstract
+    tables:
+      - name: pair_evt_swap
+        description: "Swap events from AbstractSwap Pair contracts"
+
+sources:
```
jeff-dude: maybe it's a bug with how you call out `sources` twice in this yml file?
try using `sources` only once, and have all three schemas under the one

clizzard7: thanks that was the solution!
But now there is another problem..

jeff-dude:
current issue is due to `dex_base_trades` being too large to run on CI cluster.
for final testing, please remove all chains except abstract from `dex_base_trades` to allow it to run focused only on new chain/protocol. once we see CI run through successfully after that, we can revert to add chains back.

dbt_subprojects/dex/models/trades/abstract/dex_abstract_base_trades.sql
```diff
@@ -0,0 +1,49 @@
+{{ config(
+    schema = 'dex_abstract'
+    , alias = 'base_trades'
+    , materialized = 'view'
+    )
+}}
+
+{% set base_models = [
+    ref('abstractswap_v3_abstract_base_trades')
```
jeff-dude, line 9: missing v2 here

jeff-dude (changes requested):

jeff-dude:
also, notice row counts are v small. is this expected? new dex?

clizzard7:
everything seems ok!

jeff-dude:
> everything seems ok!

looking good. plz add the v2 as noted above, let CI rerun to be sure, then revert the chain removal to prep for merge 🙏

clizzard7:
should be all good

jeff-dude (approved):

PR #8037, opened 2025-04-17
CUR2-10 solana dex.trades data missing

krishgka:

### Description:

[...]

---

### DIFF
--- a/dbt_subprojects/solana/models/_sector/dex/dex_solana_trades.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/dex_solana_trades.sql
@@ -34,9 +34,9 @@ SELECT bt.blockchain
       , bt.trade_source
       , token_bought.symbol as token_bought_symbol
       , token_sold.symbol as token_sold_symbol
-      , case when lower(token_bought.symbol) > lower(token_sold.symbol)
-            then concat(token_bought.symbol, '-', token_sold.symbol)
-            else concat(token_sold.symbol, '-', token_bought.symbol)
+      , case when lower(COALESCE(token_bought.symbol, 'unknown')) > lower(COALESCE(token_sold.symbol, 'unknown'))
+            then concat(COALESCE(token_bought.symbol, 'unknown'), '-', COALESCE(token_sold.symbol, 'unknown'))
+            else concat(COALESCE(token_sold.symbol, 'unknown'), '-', COALESCE(token_bought.symbol, 'unknown'))
         end as token_pair
       , bt.token_bought_amount_raw / pow(10,coalesce(token_bought.decimals, 9)) as token_bought_amount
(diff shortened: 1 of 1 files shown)


dbt_subprojects/solana/models/_sector/dex/dex_solana_trades.sql
```diff
@@ -34,9 +34,9 @@ SELECT bt.blockchain
       , bt.trade_source
       , token_bought.symbol as token_bought_symbol
       , token_sold.symbol as token_sold_symbol
-      , case when lower(token_bought.symbol) > lower(token_sold.symbol)
-            then concat(token_bought.symbol, '-', token_sold.symbol)
-            else concat(token_sold.symbol, '-', token_bought.symbol)
+      , case when lower(COALESCE(token_bought.symbol, 'unknown')) > lower(COALESCE(token_sold.symbol, 'unknown'))
```
0xBoxer, line 37: this is not in line with the other dex-models and also won't help any analysis, would rather not implement this pattern.

0xBoxer: this also doesn't fix any problems outlined in the task this always returns the same number of rows, whether we change this to the old way, new way or just leave it out.

https://dune.com/queries/4998583
https://dune.com/queries/4998590
https://dune.com/queries/4995312

0xBoxer: we could consider making the null explicit, but doesn't have any great advantage as far as I can tell

```suggestion
      , case when token_bought.symbol is null or token_sold.symbol is null then null
             when (token_bought.symbol) > lower(token_sold.symbol) then concat(token_bought.symbol, '-', token_sold.symbol)
             else concat(token_sold.symbol, '-', token_bought.symbol)
        end as token_pair
```

krishgka:
Yeah it is still not working, I checked the issue again and found this
solana dex trades table is correctly capturing these swaps, but it's capturing them as separate rows:
[For example, The transaction ](https://dune.com/queries/5000665)2ghfg4NxL5cFabCuc1P9pVRJpuH9S9m88h3aypqBzBvs62CxVH9gMKY3YSRMyx18ZeizKjgLLKkKajhtiCWST1gv contains multiple swaps within a single transaction.

First swap: BUZZ → USDC (Meteora)

Second swap: Bonk → SOL (Meteora)

the Nansen PDF is displaying the entire transaction with a focus on the initial input (BUZZ) and final output (SOL), making it look like a direct BUZZ-SOL swap.

This is why the BUZZ-SOL query isn't matching many records - in our tables, these multi-hop swaps are stored as separate individual swaps (BUZZ→USDC, USDC→Bonk, Bonk→SOL) rather than a single BUZZ-SOL transaction.

0xBoxer:
oh I see, that's literally all we needed to find out here then, these are not missing trades, but rather just reporting differences, that makes sense. 

thanks!

PR #7746, opened 2025-02-26
Fixing Solana NFT Token Classification + Retrieving Missing DEX Trades

fergmolina:
### Description:
We were investigating some missing trades from the dex_solana.trades table. Example: [Solscan Link](https://solscan.io/tx/2mr7sZ7adRjHh5g99LxsEc6HN8pGndpACMxLM2sp57nN5a1vXSa7fogF4xSb56Kx248dmQGmGmQKvg2e3tiChVkD)

We found out that these trades were being filtered because, in the `dex_solana.trades` dbt transformation. We have the condition 'fungible' ([source](https://github.com/duneanalytics/spellbook/blob/4e1ebedfe246006522220ab21d7e4602fdf465a6/dbt_subprojects/solana/models/_sector/dex/raydium/raydium_v5_base_trades.sql#L75)). This is in fact a fungible token, so it shouldn't be filtered.

The problem is that the `solana_utils_token_accounts` dbt transformation is marking some account tokens as 'nft' when it shouldn't be ([source](https://github.com/duneanalytics/spellbook/blob/4e1ebedfe246006522220ab21d7e4602fdf465a6/dbt_subprojects/solana/models/solana_utils/solana_utils_token_accounts.sql#L34)).

The exact issue lies in the `tokens_solana_nft` transformation:
According to [Metaplex docs](https://developers.metaplex.com/token-metadata/token-standard?utm_source=chatgpt.com), inside the createArgs, we can find the following TokenStandards:
- 0 / NonFungible: A non-fungible token with a Master Edition.
- 1 / FungibleAsset (1): A token with metadata that can also have attributes, sometimes called Semi-Fungible.
- 2 / Fungible (2): A token with simple metadata.
- 3 / NonFungibleEdition (3): A non-fungible token with an Edition account (printed from a Master edition).
- 4 / ProgrammableNonFungible (4): A special NonFungible token that is frozen at all times to enforce custom authorization rules.

There are no filters in the CTE query executed against the `mpl_token_metadata_solana.mpl_token_metadata_call_create ` table [source](https://github.com/duneanalytics/spellbook/blob/4e1ebedfe246006522220ab21d7e4602fdf465a6/dbt_subprojects/solana/models/tokens/solana/tokens_solana_nft.sql#L53), so it retrieves all token accounts, including fungible tokens like the example below.

### Solution:
Adding a WHERE condition to filter only token accounts related to non-fungible tokens will ensure that these fungible

---
quick links for more information:
[...]

PR #7654, opened 2025-02-14
improve: tokens.transfers

0xBoxer:
Improvements to tokens.transfers

PR #8042, opened 2025-04-18
solana utils: refactor token accounts pipeline to account for historical states

jeff-dude:
towards CUR2-144


jeff-dude:
latest findings on testing run times:
- using `address_prefix` on raw table(s) killed write performance on historical, CI timed out at 90m on 12-node cluster
  - compare this to when partitioning on `block_date` or no partition at all running in ~5-10 minutes
- most recent run was with no partition on raw, then state history ran in normal timing (compared to `block_date` partition), but it looks like transfers kicking off is killing performance. we didn't see a single transfers model complete in the 90m window. this also includes a forced incremental run, meaning it's only the last one day of transfers

0xBoxer:
here is the change for token metadata that we should probably just roll in here:

https://github.com/duneanalytics/spellbook/pull/7654/files

0xBoxer:
just adding more change requests here for completeness:

- remove the entire nft classification for the token_accounts model, we don't care about this, it doesn't have a significant impact and nobody cares
- downstream of that, remove https://github.com/duneanalytics/spellbook/blob/4e1ebedfe246006522220ab21d7e4602fdf465a6/dbt_subprojects/solana/models/_sector/dex/raydium/raydium_v5_base_trades.sql#L75 
which will resolve #7746

jeff-dude:
> first end-to-end successful run [here](https://github.com/duneanalytics/spellbook/actions/runs/14627537465/job/41042608942?pr=8042), with **_major_** caveats:
> 
> * only running token22 token accounts, as they run much faster (smaller dataset)
> * only running `_current` transfers models
> * only running last 1 day of transfers data per model
> * only running last 1 day of raydium data per model

will now start to slowly open it up more and test further.

jeff-dude:
next run, enabled all token accounts models for full history:
- no partition on raw for either, build in two minutes or less each. incremental run was also quick for both, 20s or less
- partition the state history models on `address_prefix`
  - token2022 is still ⚡ fast due to low volume of data
  - spl takes ~18 minutes to build full history each run
- transfer joins are still 🐌 slow with `address_prefix` partition
  - best guess is due to `unique_instruction_key` being built in subquery, then pushed to join condition in same query. trino hates this. we may need to materialize transfers twice. a "base" version without token accounts, but build unique keys, then final that does the join. maybe that join can be after we union all transfers together to make in one place.
  - up to 30 minutes per model for only 1 day of transfers (97m rows)

jeff-dude:
next run:
- all token accounts again
- added partition by `address_prefix` on both raw models
  - token2022: 18s --> 30s
  - spl: 2m --> 6.5m
- state history for both built quicker
  - token2022: 10s --> 6s
  - spl: 18m --> 14m
- total timing across raw + state history
  - token2022: 28s --> 36s
  - spl: 20m --> 20.5m
- no real difference on transfers timing, as expected, with no changes for this run

**_key takeaway:_**
no real difference whether we partition raw by `address_prefix` or not when compared to no partition on raw. the two models combined add up to take about the same time.

jeff-dude:
next run:
- all token accounts again
- added partition by `block_date` on both raw models
  - token2022: 26s
  - spl: 6m
- state history models
  - token2022: 7s
  - spl: 16m
- total timing across raw + state history
  - token2022: 34s
  - spl: 22m

**_key takeaway:_**
no real difference whether we partition raw by `address_prefix`, `block_date` or no partition on at all. the two models combined add up to take about the same time in all scenarios.
the main benefit of `block_date` partition would be incremental runs for raw would be the fastest.

jeff-dude:
we have a good idea on token accounts builds.

tomorrow i will test what it looks like to materialize transfers without join to token accounts and add columns for unique key joins downstream, then have final model join to token accounts after materializing the keys. this will likely help a lot.

jeff-dude:
was able to get small subset of data to run end-to-end.

trying to get full history of token accounts to run while other models run last one day. looks like we're hitting rate limits on s3 from the concurrent models and scale of data.

jeff-dude:
good signs on latest run. test failure is due to writing null for `unique_instruction_key`, need to dig into why that is in the morning.

0xBoxer (approved):
left comment in linear ticket, thanks for the detailed writeup in there!

jeff-dude:
closing due to move to sqlmesh, used this as reference for changes to implement there

PR #8049, opened 2025-04-21
perf(bonkbot): create dedicated ctes for SOL and SPL/token2022 fee payments

whalehunting:
Splits querying SOL fee payments and SPL/Token2022 fee payments into separate CTEs, to improve Trino query planning efficiency.


whalehunting:
@jeff-dude

PR #8080, opened 2025-04-25
perf(bonkbot): create dedicated ctes for SOL and SPL/token2022 fee payments

whalehunting:
retry of #8049 - there seems to be an issue with dupes that we try to solve here


dbt_subprojects/solana/models/_sector/dex/bot_trades/solana/platforms/bonkbot/schema.yml
```diff
@@ -47,6 +47,7 @@ models:
             - block_month
             - tx_id
             - token_address
+            - fee_receiver
```
jeff-dude, line 50: this needs added to model config block. this is only a test after the models run. adding this field to model config could def fix the dupes, if that was an issue.

jeff-dude (changes requested):

PR #8096, opened 2025-04-29
fix(dex): temp disable sei

0xBoxer:
Temporarily disable sei blockchain from uniswap forks


jeff-dude (approved):
temp disable due to duplicates in raw tables
we have informed the team to fix, which we can then re-enable

PR #8107, opened 2025-04-30
Evms schema update

los-xyz:
We decided to include the update of the `evms` schema into our blockchain integration process to maintain consistency across tables and to make sure that all the integrated evm chains are inlcued. 
We decided to update only the following tables:
- evms_info
- evms_contracts
- evms_blocks
- evms_transactions
- evms_logs
- evms_traces
- evms_creation_traces
- evms_logs_decoded
- evms_traces_decoded
- evms_erc20_transfers
- evms_erc721_transfers

and delete the remaining evms tables (@hildobby would like your blessing for this as you are the creator of the schema). this is not done in this PR yet


jeff-dude:
i would suggest leaving lens out of this PR, as @0xBoxer is working with nia on lens integration to show how it works. that will happen in separate PR.

you are also getting a failure here due to lens sources not being established in spellbook yet.

0xBoxer:
I plan to rework this into the same pattern we use in automated dex trades very soon, I think better to save your time @los-xyz. 
I'll take this over and take responsibility. 

thanks for contributing still, appreciate you taking this on 🫡

los-xyz:
ok so you take it from here @0xBoxer ? thanks
I would still consider updating these tables as part of onboarding a evm chain

0xBoxer:
yeah this is gonna be a strict requirement going forward. 
This is just a rather inefficient way of implementing this from an engineering perspective, I'll change that in the next few days so we don't need to spend time making this work.

0xBoxer:
closing in favour of https://github.com/duneanalytics/spellbook/pull/8130

PR #8155, opened 2025-05-08
rename token symbol from USDC to USDC.e

0xdatawolf001:

### Description:

[...]

---

### DIFF
--- a/dbt_subprojects/tokens/models/prices/arbitrum/prices_arbitrum_tokens.sql
+++ b/dbt_subprojects/tokens/models/prices/arbitrum/prices_arbitrum_tokens.sql
@@ -85,7 +85,7 @@ FROM
     ('uma-uma','arbitrum','UMA',0xd693ec944a85eeca4247ec1c3b130dca9b0c3b22,18),
     ('umami-umami-finance','arbitrum','UMAMI',0x1622bf67e6e5747b81866fe0b85178a93c7f86e3,9),
     ('uni-uniswap','arbitrum','UNI',0xfa7f8980b0f1e64a2062791cc3b0871572f1f7f0,18),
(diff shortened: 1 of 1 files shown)


0xBoxer:
@0xRobin you previously changed this from USDC.e to USDC, do you remember why?

0xRobin:
@0xBoxer I changed these with a script that reconciled token metadata (symbol and decimals) between `prices.tokens` and `tokens.erc20`, using our erc20 metadata provider as source of truth.

There is indeed a social consensus to call bridged Circle USD `USDC.e` and native minted Circle USD `USDC`
As also seen in arbiscan token page.
image


Onchain data however says the symbol is `USDC` and the name is `USD Coin (Arb1)`.. :thinking: 
image
https://arbiscan.io/address/0xff970a61a04b1ca14834a43f5de4533ebddb5cc8#readProxyContract

Not sure what our best future approach is here, given upcoming changes in token metadata

0xdatawolf001:
This is an interesting problem. My two cents:

TLDR:  IMO, follow the contract's symbol

Strictly speaking, we follow a consistent rule of following the wishes of the deployer which in turns respect the standard expectations of a dev working with it. Given that, in terms of SQL best practices 😅, its best to use the address as an identifier. Therefore, the naming convention should follow the wishes of the contract. 

Also, once the announcement is made by Circle (or not), the assumption is that users are expected to recognize USDC.e as USDC. For the record, I don't agree, nor understand why, this approach was taken because clearly a bridged token is different in terms of security as compared to something natively minted.

Either way, I'll leave it to Boxer to see if the should be restored. Social conventions change all the time anyway but contracts are forever

0xdatawolf001:
Looks like they are merging Native and the Bridged tokens (at least on sonic now):
https://x.com/usdc_cool/status/1922277087026483520

I think this settles it? We follow the new symbol change? @0xBoxer

0xBoxer:
yep we can close this out

PR #8166, opened 2025-05-09
Fix Volume counting

alexshchur:
attempt to dedupe Velora volumes by adding fallback to call_trace_address when it's an empty array


jeff-dude:
two things:
- since you are modifying the same file, it would be nice to merge this into your other PR for optimism
- this is failing, can you move over and then fix?

alexshchur:
> this is failing, can you move over and then fix?
fixed, sorry my bad, didn't notice that failing run

> since you are modifying the same file, it would be nice to merge this into your other PR for optimism
sure. So i'm closing this one then

PR #8172, opened 2025-05-11
Adding tesseract spells

yy-analytics:

### Description:

This pull request introduces several new spells that will all be very useful to our project (Tesseract). Tesseract contracts are built on top of products/technology that was itself built by Avalabs (Avalanche creators), namely ICTT (interchain token transfer), which itself is built on top of ICM (interchain messaging).

In the hierarchy of these new spells, we start with `_icm_blockchains` which gets the blockchain IDs (note - not the usual ID as listed on chainlist - these are avalanche-specific IDs for L1s that form part of the wider Avalanche ecosystem). This uses the TeleporterMessenger contract (decoded) and SendCrossChainMessage/ReceiveCrossChainMessage events emitted by this contract. There is only one TeleporterMessenger contract and is likely to stay that way for a while (deployment is semi-official in order to be trusted etc.). This means we can rely on just the decoded contract table.

Then there is `_ictt_pairs` which gets pairs of linked ICTT contracts (homes and remotes), from which we then get `_ictt_contracts` which depends on this and is pretty much a distinct list of the contracts deployed on the relevant chain. These have to use the logs and traces tables because anyone can deploy ICTT contracts for their own token / L1, but we want a way to keep track of all of them. If we had to rely on decoded tables and then combining all of those together with many different names etc. then it would be impractical.

Next there are many spells of the form `_ictt_evt_SomeEvent`. These are retrieved from the logs tables and are effectively replicating what would be decoded event tables for the most important events from our perspective (events corresponding to transfer volume) for ICTT contracts. Explained above why we are not using combinations of decoded tables.

Combining those events into a single table for volume events, we then have `_ictt_volume_events`, which also includes a couple of extra details about whether the transfer used Tesseract, and the source/destination chains.

Separately I've also made views for the main Tesseract Cell contract events (`_cell_routed` and `_cell_initiated`) as there can be multiple Cell contracts deployed on a single chain, and this makes it more convenient to query, rather than always having to do a "UNION ALL" between similar decoded tables.

Final thing to say is that I've created these as macros which depend on the 'blockchain'. At the moment, the only Avalanche L1 data stored on Dune is for 'avalanche_c', so the "overall" views are each just UNIONs of spells for a single chain. It's not really necessary for it to be this way at the moment, but if more Avalanche L1 data becomes available on Dune in the future then this should help to slightly future-proof these spells / make it easier to replicate across L1s.

Please shout if there are any questions.

---


jeff-dude:
at first glance, looking clean as always 🔥 

looks like a minor union bug here on column differences. the attached workflows should run on all your commits now. let me know if not. the comment from `dune-eng` is a good sign it's auto-approving.

yy-analytics:
Great, thanks a lot @jeff-dude , and yeah minor bug on the incremental UNION after I added a new column. Will fix that now

jeff-dude (approved):

PR #8178, opened 2025-05-12
 Fix inaccurate filters in Raydium v4/v5 trade models

0xBoxer:
Removes filters , , and  from Raydium v4 and v5 base trade models as they might be redundant or incorrect. Also updates a comment in the v5 model.


jeff-dude:
we should move my changes from #8042 for all raydium models into here. we did lots of test runs on those and felt good with output.

jeff-dude:
i'm not getting expected results here when comparing to linked PR with your original simple changes. i will close this and focus on the other one.

PR #8198, opened 2025-05-14
try ephemeral views and different day pushdown

0xBoxer:
maybe this works 🤞


0xBoxer:
@0xRobin any idea how I could test whether the ephemeral views actually work? 

the CI doesn't run them

PR #8199, opened 2025-05-14
testing balances another way

0xBoxer:
ggwp


jeff-dude:
assuming safe to close

PR #8226, opened 2025-05-20
Add near bridge addresses to labels

0xSector:
### Description:

Add NEAR bridge addresses to bridge_labels spell. First non-evm chain to be added to spell so not sure if it'll cause any issues. Performed 'CAST(address AS varbinary) AS address' in downstream labels_bridges spell to make all data types similar since near addresses are unique v. evm.


jeff-dude:
using `cast as varbinary` on EVM chain data would break the output i'm pretty sure. you need to use `from_hex()` function on the near addresses to convert to varbinary prior to union instead

0xSector:
Thanks for the feedback Jeff! I tested from_hex() but near address are odd where they can use text not hex data e.g. "aurora.pool.near". So I ended up encoding the near addresses using to_utf8() prior to the join to convert to varbinary. Tested all changed queries on dune to verify prior to resubmitting. Learning here with NEAR data so really appreciate the input.

jeff-dude:
> Thanks for the feedback Jeff! I tested from_hex() but near address are odd where they can use text not hex data e.g. "aurora.pool.near". So I ended up encoding the near addresses using to_utf8() prior to the join to convert to varbinary. Tested all changed queries on dune to verify prior to resubmitting. Learning here with NEAR data so really appreciate the input.

nice. i don't know near data yet either. whichever function available that works to convert to varbinary works for me. then when you query this table later, you will need to apply the opposite function to convert address back for joining to other near datasets. assuming `from_utf8()` exists.

jeff-dude (approved):

jeff-dude:
another option, if you wanted to try prior to merge (i won't merge until tomorrow or later in week):
you could materialize the near bridges spell in native format, then build second spell which is a view that converts to varbinary with the function, that way you have the data in native format somewhere and you could join to it. just some ideas to make downstream usage easier.

0xSector:
> another option, if you wanted to try prior to merge (i won't merge until tomorrow or later in week): you could materialize the near bridges spell in native format, then build second spell which is a view that converts to varbinary with the function, that way you have the data in native format somewhere and you could join to it. just some ideas to make downstream usage easier.

Yes! That would be much more intuitive for the end users. Going to give that a try. Where would you recommend I create each file as far as folder structure? Would I add both the native and converted view in the 'spellbook\dbt_subprojects\daily_spellbook\models\_sector\labels\addresses\bridge\identifiers' path? Or would the native file need to live elsewhere to be it's own accessible spell? 

Thanks! Tim

jeff-dude:
> Yes! That would be much more intuitive for the end users. Going to give that a try. Where would you recommend I create each file as far as folder structure? Would I add both the native and converted view in the 'spellbook\dbt_subprojects\daily_spellbook\models_sector\labels\addresses\bridge\identifiers' path? Or would the native file need to live elsewhere to be it's own accessible spell?

you can keep both in same directory as-is. just add `_native` to both file name + alias for the first model which keeps as-is, then without suffix for second model. then union the second model into crosschain like you do now.

dbt_subprojects/daily_spellbook/models/_sector/labels/addresses/bridge/identifiers/labels_bridges_near.sql
```diff
@@ -0,0 +1,19 @@
+{{config(
+        materialized = 'view',
+        alias = 'labels_bridges_near',
```
jeff-dude:
```suggestion
        schema = 'labels',
        alias = 'bridges_near',
```

jeff-dude: the legacy labels pipeline (the one you're using here in daily subproject) has a default schema assignment in the global config file
https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/daily_spellbook/dbt_project.yml#L109-L113

we should avoid using `labels_` again in the table name, since schema already uses it.
also, while schema will automatically be picked up, we've been using best practice to call it out in model directly to make it easier to follow naming standards for anyone verifying code.

dbt_subprojects/daily_spellbook/models/_sector/labels/addresses/bridge/identifiers/labels_bridges_near.sql
```diff
@@ -0,0 +1,19 @@
...
+        post_hook='{{ expose_spells(\'["near"]\',
+                                    "sector",
+                                    "labels",
+                                    \'["Sector920"]\') }}')}}
```
jeff-dude:
```suggestion
)}}
```

jeff-dude: let's leave out of data explorer for now, you can still query as normal once live

jeff-dude: we'll keep it so that `labels.bridges` is only one which is displayed, so the intermediate tables don't add too much noise in there.

dbt_subprojects/daily_spellbook/models/_sector/labels/addresses/bridge/identifiers/labels_bridges_near_native.sql
```diff
@@ -0,0 +1,20 @@
+{{config(
+        materialized = 'table',
+        alias = 'labels_bridges_near_native',
+        post_hook='{{ expose_spells(\'["near"]\',
+                                    "sector",
+                                    "labels",
+                                    \'["Sector920"]\') }}')}}
```
jeff-dude:
```suggestion
        materialized = 'table',
        tags=['static'],
        schema='labels',
        alias = 'bridges_near_native',
)}}
```

jeff-dude: we use static tag in these hardcoded files to avoid running in prod over and over when nothing changes.

jeff-dude (changes requested):

jeff-dude:
marking as draft until proposed changes are committed, feel free to promote for review and let us know when ready

0xSector:
Implemented recommended changes. Thanks for providing details to help me understand the issues and learn.

jeff-dude (approved):

PR #8227, opened 2025-05-20
raydium: alternative fix for removing token filters

jeff-dude:


0xBoxer (approved):

jeff-dude:
> can't tell what the hell was going on #8178 , probably vibe coded a bit too hard and introduced unwanted changes.
> 
> this PR looks like all the changes we ever wanted to make, if you say results are congruent, lfg

no those changes were me 😅 from the transfers testing. the token sold address and amounts are incorrectly applied. not worth spending time on right now. we can revisit later.

PR #8230, opened 2025-05-21
add authoization_lists to evms.transactions

hildobby:

### Description:

[...]

---

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/evms/evms_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/evms/evms_schema.yml
@@ -38,6 +38,8 @@ models:
         description: "Number of confirmed transactions previously sent by this account"
       - name: index
         description: "Transaction index"
+      - name: authorization_list
+        description: "Authorization list, if there's one"
       - &success
         name: success
         description: "Whether the transaction was completed successfully"
--- a/dbt_subprojects/daily_spellbook/models/evms/evms_tables/evms_transactions.sql
+++ b/dbt_subprojects/daily_spellbook/models/evms/evms_tables/evms_transactions.sql
@@ -33,6 +33,7 @@ FROM (
         , success
         , "type"
         , CAST(value AS double) AS value
+        , try(authorization_list) AS authorization_list
         --Logic for L2s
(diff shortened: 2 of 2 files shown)


0xBoxer:
not the prettiest way to do this, we should prob create lookup lists for both the rollup stuff and the auth list eventually, but fine for now since these are just views anyways, this will work

0xBoxer (approved):

jeff-dude (approved):
@hildobby if you can, keep an eye on this over time as it will require adding chains manually here as we onboard more to spellbook. when we add a new chain, we'll be adding to the helper variable [here](https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/daily_spellbook/macros/helpers/evms_blockchains_list.sql). it won't auto-propagate to this specific field with if/else logic.

hildobby:
> @hildobby if you can, keep an eye on this over time as it will require adding chains manually here as we onboard more to spellbook. when we add a new chain, we'll be adding to the helper variable [here](https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/daily_spellbook/macros/helpers/evms_blockchains_list.sql). it won't auto-propagate to this specific field with if/else logic.

i definitely won't add this to my ever growing list of things to monitor, can't you guys create lists for things like this like you did with `all_op_chains()` which is used in this same file for example?

0xBoxer:
> > @hildobby if you can, keep an eye on this over time as it will require adding chains manually here as we onboard more to spellbook. when we add a new chain, we'll be adding to the helper variable [here](https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/daily_spellbook/macros/helpers/evms_blockchains_list.sql). it won't auto-propagate to this specific field with if/else logic.
> 
> i definitely won't add this to my ever growing list of things to monitor, can't you guys create lists for things like this like you did with `all_op_chains()` which is used in this same file for example?

yeah this is on us, we should update the add a chain readme to include this step, it ain't too hard. 

have to still do that anyways, I'll take that. 

(maybe I also find some better workaround for both the all_op_chains and now this thing, seems odd that wouldn't just use extra key-value pairs here

PR #8240, opened 2025-05-21
update native token address in evm transfers

jeff-dude:
using `dune.blockchains` to validate values. many older chains didn't have a native token at all, while newer chains have started adding the value. this PR includes those which were missing and/or incorrect (assuming we trust the table).


jeff-dude:
this should be considered, if approved, to be merged prior to full refresh on evm transfers with new erc20 metadata.

jeff-dude:
also, one callout on an outlier chain: `zksync`

`dune.blockchains` has the standard `0x000..` address, however, the transfers spell has the `0x000000000000000000000000000000000000800A` address, which is standard for zk stack. we use this for other zk chains in transfers too.
more info in docs:
https://docs.zksync.io/zksync-protocol/contracts/system-contracts#l2basetoken-msgvaluesimulator

i believe we should update `dune.blockchains` to have this correct value. if so, we then need to refresh erc20 metadata? that would need to happen prior to this merge, if agreed with my thinking.

0xRobin:
@jeff-dude we shouldn't hardcode the values here, but use dune.blockchains directly.

PR #8247, opened 2025-05-22
tokens erc20: set as full table

jeff-dude:
lets see how performance is on full table rebuild. i may make it incremental + append and do join to target to find only new rows. see how those run times compare.
the other benefit of incremental is that we may get around our bug in cleanup script where vacuum breaks the version. this seemingly only happens on table types.


0xRobin:
fine as a temp setup, but we'll need to move this to sqlmesh to propagate chain additions better

0xRobin (approved):

PR #8304, opened 2025-06-03
DATA-4643 `tokens.erc20`: clean static files

jeff-dude:
- remove all tokens covered by v2 tokens automated source
- leave files in place as a placeholder for edge cases
- keep handful of tokens across a few chains which are missing in v2


jeff-dude:
two callouts for exceptions:
- tron: a bit of a special case, small list, not in new v2 pipeline or v1 automated source -- keep the static as-is
- optimism: OP labs team has a bit of a dynamic approach and it's a bit more complex to unravel. we can explore breaking that up in future iterations or leave as-is to serve their use case(s)

dbt_subprojects/tokens/models/tokens/tokens_v1_erc20.sql
```diff
@@ -5,26 +5,31 @@
         ,materialized = 'table'
         ,partition_by = ['blockchain']
         ,post_hook='{{ expose_spells(\'[
```
0xRobin: We'll want to remove the post hook here as well and hide it from the data explorer

0xRobin (approved):

0xBoxer (approved):
overall LGTM. 

which direction are we overriding if a token appears in both datasets?

dbt_subprojects/tokens/models/tokens/tokens_v1_erc20.sql
jeff-dude: we can remove, but still requires manual steps to actually hide it

jeff-dude:
> overall LGTM.
> 
> which direction are we overriding if a token appears in both datasets?

haven't gotten to that stage yet. that'll be different PR.

PR #8373, opened 2025-06-17
CUR2-248 use `prices.hour` in `tokens.transfers`

jeff-dude:


jeff-dude:
~~todo: test how append-only strat performs relative to merge~~
run time looks better on append strat.

dbt_subprojects/tokens/macros/transfers/transfers_enrich.sql
```diff
@@ -64,62 +57,67 @@ WITH base_transfers as (
...
+        , amount
+        , price_usd
+        , amount_usd
+        /*
+        , CASE
+            WHEN amount_usd >= {{ usd_amount_threshold }}
+                THEN CAST(NULL as double)
+                ELSE amount_usd -- Select only transfers where USD amount is less than the threshold
+            END AS amount_usd
+        */
```
jeff-dude: to be reverted and tested once we have pricing outliers resolved in prices pipeline.

dbt_subprojects/tokens/models/transfers_and_balances/abstract/tokens_abstract_transfers.sql
```diff
@@ -4,7 +4,7 @@
      partition_by = ['block_month'],
      materialized = 'incremental',
      file_format = 'delta',
-     incremental_strategy = 'merge',
+     incremental_strategy = 'append',
```
0xRobin: if we do append without filtering on `> max(timestamp) from {{this}}`, we will start introducing duplicates during the incremental updates.

If we do a self referential filter we can't update already written columns and might have gaps with null values do to unequal data recency. We need to use `merge`

0xRobin (changes requested):

dbt_subprojects/tokens/models/transfers_and_balances/abstract/tokens_abstract_transfers.sql
jeff-dude: i still have this in linear on deck to address. was hoping transfers was a use case we could consider performance gains of append-only, but we can keep merge to be safe.

jeff-dude:
TODO prior to merge:
- revert commit for short timeframe 2833f756192d13f416bf09fe001aaea2e9676370
- revert commit for forcing `tokens.transfers` view with comment 77c034326ab8e8e36381d7f9c845479e48c526be

0xRobin (approved):

PR #8377, opened 2025-06-18
add stability

franzns:
add stability vaults


jeff-dude (approved):
i realize this is still in draft state, but i'm doing a batch of merges for hourly project and this is passing CI / seems good. i will proceed. if you had more to add, feel free to do another PR later.

PR #8395, opened 2025-06-23
create addresses_events_{chain}.first_token_received

hildobby:
Here is a spell similar to addresses_events_{chain}.first_funded_by. first_token_received surfaces any fungible token, whichever was first received by an address, while first_funded_by only looks at native token transfers, which are necessary to do any onchain action with that address, so both serve a different purpose but are fundamental for many things.

Also, first_token_received is likely a lot more optimised, happy to rework first_funded_by in a similar way if performance gains are significant @jeff-dude @0xRobin 

i'll follow up with more tail-end chains + crosschain spell once this is merged btw

### DIFF
--- a/dbt_subprojects/hourly_spellbook/macros/sector/addresses_events/addresses_events_first_token_received.sql
+++ b/dbt_subprojects/hourly_spellbook/macros/sector/addresses_events/addresses_events_first_token_received.sql
@@ -9,8 +9,6 @@ WITH finding_transfer AS (
         ON t.address=tt.to
     WHERE t.address IS NULL
     AND {{ incremental_predicate('tt.block_time') }}
-    {% else %}
-    WHERE tt.block_time >= NOW() - interval '1' month
     {% endif %}
     GROUP BY 1
 )
@@ -34,7 +32,5 @@ FROM {{token_transfers}} tt
 INNER JOIN finding_transfer ft USING (unique_key)
 {% if is_incremental() %}
 WHERE {{ incremental_predicate('tt.block_time') }}
-{% else %}
-WHERE tt.block_time >= NOW() - interval '1' month
 {% endif %}
 {% endmacro %}


jeff-dude:
plz finalize tests in CI pipeline by hardcoding historical dates and/or forcing incremental to run for both steps with `or true` on incremental logic

hildobby:
> plz finalize tests in CI pipeline by hardcoding historical dates and/or forcing incremental to run for both steps with `or true` on incremental logic

passed with a `> NOW() - interval '1' month` filter, i removed that filter in last commit and cancelled the workflow before it runs

jeff-dude:
looking to find best approach to try and prepare for long-term performance on these models. each table will contain one row per address for a single event, a transfer in. since it's event-driven, we should be good to use `block_month` as a partition and across all join conditions. can you look to apply, or tell me why that wouldn't work?

hildobby:
just added monthly partition to all models, should i let ci tests run @jeff-dude?

dbt_subprojects/hourly_spellbook/macros/sector/addresses_events/addresses_events_first_token_received.sql
```diff
@@ -0,0 +1,37 @@
+{% macro addresses_events_first_token_received(blockchain, token_transfers) %}
+
+
+WITH finding_transfer AS (
+    SELECT tt.to
+    , MIN_BY(tt.unique_key, (tt.block_number, tt.tx_index, COALESCE(tt.trace_address, ARRAY[COALESCE(evt_index, -1)]))) AS unique_key
+    FROM {{token_transfers}} tt
+    {% if is_incremental() %}
+    LEFT JOIN {{this}} t ON t.address=tt.to
```
jeff-dude:
```suggestion
    LEFT JOIN {{this}} t
        ON t.address=tt.to
        AND t.block_month=tt.block_month
```

dbt_subprojects/hourly_spellbook/macros/sector/addresses_events/addresses_events_first_token_received.sql
```diff
@@ -0,0 +1,37 @@
...
+, tt.contract_address AS token_address
+, tt.block_time AS block_time
+, tt.block_number AS block_number
+, tt.tx_hash AS tx_hash
+, tt.tx_index AS tx_index
+, tt.trace_address AS trace_address
+, tt.block_month AS block_month
+, unique_key
+FROM {{token_transfers}} tt
+INNER JOIN finding_transfer ft USING (unique_key)
```
jeff-dude, line 32:
```suggestion
INNER JOIN finding_transfer ft USING (block_month, unique_key)
```

jeff-dude (changes requested):
we should be using the partition in joins as well, this is where we can ideally optimize run time

jeff-dude:
seems like my assumption was wrong. will need to revisit this, unless you can fix in the meantime.

hildobby:
oh wait, the spell calls itself to check if addresses have been found earlier before adding. doesn't the monthly partitioning mean that it all checks in one partition instead of the whole set @jeff-dude ? that would cause duplicates on incremental runs

jeff-dude:
> oh wait, the spell calls itself to check if addresses have been found earlier before adding. doesn't the monthly partitioning mean that it all checks in one partition instead of the whole set @jeff-dude ? that would cause duplicates on incremental runs

its not due to usage of partition, but rather the extra join condition for `block_month` in the lookup. looks like my assumption for time based usage here is incorrect, it won't work. the partitions really aren't doing much in this current state, but it's fine to leave.

jeff-dude (approved):

PR #8399, opened 2025-06-23
chore: update jupiter dex list

rajivpo:
### Description:

There are DEXes being routed to through Jupiter (GoonFi, Boop, etc.) that are not currently being tracked. The newly added DEX labels/program ids comes from https://api.jup.ag/swap/v1/program-id-to-label as per existing comments.

---


rajivpo:
@dune-eng this run seems to take longer than 90 minutes. is there any way to allow for an extension?

jeff-dude:
> @dune-eng this run seems to take longer than 90 minutes. is there any way to allow for an extension?

rather than looking to extend, we should be filtering our model source reads to shorter timeframes to test small subset of data. you can do a few ways:
- hardcode a filter for a specific date
- add `or true` on the if is incremental blocks to force incremental timeframe only

edit: can you try one of these methods, see CI complete, then query output tables built in the workflow to validate your expected output? then if all good, you can revert that short timeframe commit and prep for merge.

PR #8417, opened 2025-06-26
`phantom_swapper` solana trades: add prod exclude tag due to dupes

jeff-dude:
fyi @Jam516 
this model has started failing on duplicates. i tried to track it down and see what the issue could be. this is the best i could come up with for now:
- 4 `tx_id` values have dupes in the last ~day (one shared in query below)
- in the tx, multiple rows have balance > 0
- two of the addresses in the tx match to `fee_addresses`, when it looks like expectation is only one should be returned
- one of the `fee_receiver` values is in the hardcoded list of the union

```
          select
              account_activity.block_time,
              cast(date_trunc('month', account_activity.block_time) as date) as block_month,
              fee_addresses.fee_receiver,
              account_activity.address,
              account_activity.balance_change / 1e9 as amount,
              'So11111111111111111111111111111111111111112' token_address,
              account_activity.tx_id
          from "delta_prod"."solana"."account_activity" as account_activity
          join
              phantom_swapper_solana.fee_addresses as fee_addresses
              on (
                  fee_addresses.fee_receiver = account_activity.address
                  and balance_change > 0
              )
          where
               
                  account_activity.block_time >= timestamp '2025-06-24'
               and tx_id = '4j7M1n4mYAiLqX4SiDqgRPeqUufF3foFXAiu1H3CDZoa1u47zEX8VWLKRJ4ZTYP4Un8dh8sYTF8hwmtPX8yonxMR'
              and tx_success
```

PR #8431, opened 2025-06-30
Fix Phantom Swapper duplicates error

Jam516:
@jeff-dude identified a duplicates error in the phantom swapper fee payments spell: https://github.com/duneanalytics/spellbook/pull/8417

Upon investigation, I found that the cause was a spambot called rocketpot which spams addresses with tiny amounts of SOL ([example](https://solscan.io/tx/3gmEqXALtN32LeDbm1LKPGAv83h8vAYjq5GvryJtBaNFmqNx4iBDNzR3vqTA5wps1EBHPX7sASEe67vQBtDodTf1))

To fix this, I join the solana trades table onto the fee payment query. This excludes edge cases like the rocketpot spamming, leaving only transactions where a phantom wallet is making a swap and a fee is being paid to phantom.

To test the fix, I made a query that contains the `phantom_swapper_fee_payments_raw` logic together with the new join and checks for recent duplicates. As you can see, no duplicates are found.
Fix - https://dune.com/queries/5370319
Original - https://dune.com/queries/5370136 (returns three duplicates found in that same time period)

### DIFF
--- a/dbt_subprojects/solana/models/_sector/dex/phantom_swapper/phantom_swapper_solana_fee_payments_raw.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/phantom_swapper/phantom_swapper_solana_fee_payments_raw.sql
@@ -31,13 +31,13 @@ with sol_payments as (
         join {{ source('dex_solana', 'trades') }} as trades
             on trades.tx_id = account_activity.tx_id
             and trades.block_time = account_activity.block_time
-            {% if is_incremental() or true %} 
+            {% if is_incremental() %} 
             and {{ incremental_predicate('trades.block_time') }}
             {% else %} 
             and trades.block_time >= timestamp '{{query_start_date}}'
             {% endif %} 
         where
-            {% if is_incremental() or true %} 
+            {% if is_incremental() %} 
                 {{ incremental_predicate('account_activity.block_time') }}
             {% else %} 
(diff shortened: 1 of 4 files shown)


Jam516:
Note: Yes, the run times out, but this was also the case when the phantom swapper spell was first created and we got permission to merge it despite that

dbt_subprojects/solana/models/_sector/dex/phantom_swapper/phantom_swapper_solana_fee_payments_raw.sql
```diff
@@ -29,6 +28,9 @@ with sol_payments as (
                 fee_addresses.fee_receiver = account_activity.address
                 and balance_change > 0
             )
+        join {{ source('dex_solana', 'trades') }} as trades
+            on trades.tx_id = account_activity.tx_id
+            and trades.block_time = account_activity.block_time
```
jeff-dude, line 33:
```suggestion
            and trades.block_time = account_activity.block_time
            {% if is_incremental() %} 
                {{ incremental_predicate('trades.block_time') }}
            {% else %} 
                trades.block_time >= timestamp '{{query_start_date}}'
            {% endif %} 
```

jeff-dude (changes requested):

jeff-dude (approved):

PR #8477, opened 2025-07-11
remove addresses with invalid base58 characters from Tron

ying-w:
### Description:

Removed Tron addresses have invalid base58 characters. I think what happened was someone wrote a parser on binance reserve address list and grabbed everything that had a T in it since these are substrings of other binance reserve addresses for different chains


ying-w:

ying-w:
fyi @hildobby

jeff-dude (approved):
makes sense from code perspective, will also tag @hildobby for awareness.

PR #8479, opened 2025-07-11
Elk Finance DEX Trades added on Arbitrum Chain

PatelPrinci:

### Description:

[...]

---


krishgka:
I see the macro file - what is the long term plan for the macro? 
Will you be adding multiple chains for this contract? 
If not then do add the logic from the macro and put it directly into the model instead.

PatelPrinci:
> I see the macro file - what is the long term plan for the macro? Will you be adding multiple chains for this contract? If not then do add the logic from the macro and put it directly into the model instead.

Yes, there can be multiple chains for this contract in the future. If that happens, I’ll make sure to add the new chain(s) and update the logic accordingly. Keeping the macro flexible helps us scale smoothly when that time comes.

krishgka:
Lets keep the logic within the model file itself. The dex does not have much activity based on the number of rows we are tracking in the incremental runs to justify a macro at the moment, should that change, we can revisit adding a macro for it.

PatelPrinci:
> Lets keep the logic within the model file itself. The dex does not have much activity based on the number of rows we are tracking in the incremental runs to justify a macro at the moment, should that change, we can revisit adding a macro for it.

Okay, we’ll keep the logic within the model file as suggested. Thanks for the feedback! Please review and merge it with main

krishgka (approved):

jeff-dude (approved):

PR #8505, opened 2025-07-21
Update tokens_solana_sol_transfers.sql

0xRobin:
To prevent model naming clashes with sqlmesh


0xRobin:
will merge this and run with only
- `tokens_solana_sol_transfers`
- `tokens_solana_transfers`
without downstream

PR #8508, opened 2025-07-21
remove old tokens solana transfers and fix references

0xRobin:
should be merged together with https://github.com/duneanalytics/spellbook-sqlmesh/pull/133


0xRobin:
cancelled CI as it would time out.

0xRobin:
cursor review

PR #8541, opened 2025-07-24
1inch-U75: Include solana swaps to main lineage

grkhr:
- Included solana swaps to main lineage
  - oneinch.swaps → oneinch_evm.swaps
  - evm + solana → oneinch.swaps view
  
  no needs to full-refresh downstream models

### DIFF
--- a/dbt_subprojects/dex/models/_projects/oneinch/oneinch_swaps.sql
+++ b/dbt_subprojects/dex/models/_projects/oneinch/oneinch_swaps.sql
@@ -31,7 +31,7 @@
     "call_gas_used" : "null",
     "call_type" : "null",
     "user" : "from_base58(user)",
-    "receiver" : "from_base58(receiver)",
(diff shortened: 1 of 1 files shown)


grkhr:
hey @jeff-dude 
soft bump here

jeff-dude:
note to self for merge:
can likely disable `dex_trades+` model in rebuild, as we are only changing underlying source names here but data remains the same.

grkhr:
@jeff-dude 
> shows it's end of pipeline, so solana data won't go into dex.trades

yes, that's correct


> are you aware of dex_solana.trades in the solana subproject? do you have plans to integrate 1inch into that table? it's a pretty popular table, i'm sure the community would appreciate more protocol data in there.

yes, we aware. but these trades are aggregator trades


> can likely disable dex_trades+ model in rebuild

yes, you can, no downstream changes here

PR #8551, opened 2025-07-28
Eulerswap ( Trades + Pools)

henrystats:

### Description:

[...]

---

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/eulerswap_compatible_pools.sql
+++ b/dbt_subprojects/dex/macros/models/_project/eulerswap_compatible_pools.sql
@@ -1,0 +1,47 @@
+{% macro eulerswap_compatible_univ4_pools(
+    blockchain = null
+    , project = 'eulerswap'
+    , version = '4'
+    , uniswap_pools = null 
+    , factory_univ4_pooldeployed = null 
+    , factory_univ4_poolconfig = null 
+    )
+%}
+
+select 
+    '{{blockchain}}' as blockchain
+    , '{{project}}' as project 
+    , '{{version}}' as version 
+    , d.contract_address as factory_address 
+    , d.evt_block_time as creation_block_time
+    , d.evt_block_number as creation_block_number 
+    , p.pool 
+    , d.pool as hook -- EulerSwap instance's pool is the hook in Uniswap v4 universe 
+    , d.eulerAccount
+    , d.asset0
+    , d.asset1
+    , from_hex(substr(json_extract_scalar(c.params, '$.vault0'), 3)) as vault0
+    , from_hex(substr(json_extract_scalar(c.params, '$.vault1'), 3)) as vault1
+    , cast(json_extract_scalar (c.params, '$.fee') as double) / 1e18 as fee
+    , cast(json_extract_scalar (c.params, '$.fee') as double) / 1e18 as protocolFee
+    , from_hex(substr(json_extract_scalar(c.params, '$.protocolFeeRecipient'), 3)) as protocolFeeRecipient
+from 
+{{ factory_univ4_pooldeployed }} d
+left join 
+{{ factory_univ4_poolconfig }} c 
+    on d.evt_block_number = c.evt_block_number 
+    and d.evt_tx_hash = c.evt_tx_hash 
+    and d.pool = c.pool 
+    {%- if is_incremental() %}
+    and {{ incremental_predicate('c.evt_block_time') }}
+    {%- endif %}
+left join 
+{{ uniswap_pools }} p 
+    on p.blockchain = '{{blockchain}}'
+    and d.evt_block_number = p.creation_block_number
+    and d.pool = p.hooks 
+{%- if is_incremental() %}
+where {{ incremental_predicate('d.evt_block_time') }}
+{%- endif %}
+
+{% endmacro %}
--- a/dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
+++ b/dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
@@ -18,6 +18,7 @@
         , evt_index
         , currency0 as token0
         , currency1 as token1
+        , hooks 
     from 
     {{ PoolManager_evt_Initialize }}
     {%- if is_incremental() %}
--- a/dbt_subprojects/dex/models/_projects/eulerswap/_schema.yml
+++ b/dbt_subprojects/dex/models/_projects/eulerswap/_schema.yml
@@ -1,0 +1,68 @@
+version: 2
+
+models:
+
+  - name: eulerswap_pools
+    meta:
+      blockchain: ethereum
+      sector: dex
+      project: eulerswap
+      contributors: Henrystats
+    config:
+      tags: ['ethereum', 'pools', 'liquidity', 'eulerswap']
+    description: >
+      eulerswap Uni V4 DEX pools across all chains
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - blockchain
+            - pool
+    columns:
+      - &blockchain
+        name: blockchain
+        description: "Blockchain which the DEX is deployed"
+      - &project
+        name: project
+        description: "Project name of the DEX"
+      - &version
+        name: version
+        description: "Version of the contract built and deployed by the DEX project"
+      - &factory_address
+        name: factory_address
+        description: "Contract address used to create the pool"
+      - &creation_block_time
+        name: creation_block_time
+        description: "Block time of pool creation"
+      - &creation_block_number
+        name: creation_block_number
+        description: "Block number of pool creation"
+      - &pool
+        name: pool
+        description: "Uni V4 virtual pool ID"
+      - &hook
+        name: hook
+        description: "EulerSwap instance's pool is the hook in Uniswap v4 universe"
+      - &eulerAccount
+        name: eulerAccount
+        description: "eulerAccount"
+      - &asset0
+        name: asset0
+        description: "First token in pool"
+      - &asset1
+        name: asset1
+        description: "Second token in pool"
+      - &vault0
+        name: vault0
+        description: "Contract Address holding First token in pool"
+      - &vault1
+        name: vault1
+        description: "Contract Address holding Second token in pool"
+      - &fee
+        name: fee
+        description: "Trade fee"
+      - &protocolFee
+        name: protocolFee
+        description: "Fee paid to Euler"
+      - &protocolFeeRecipient
+        name: protocolFeeRecipient
+        description: "COntract address receiving Fee paid to Euler"
--- a/dbt_subprojects/dex/models/_projects/eulerswap/ethereum/_schema.yml
+++ b/dbt_subprojects/dex/models/_projects/eulerswap/ethereum/_schema.yml
@@ -1,0 +1,68 @@
+version: 2
+
+models:
+
+  - name: eulerswap_ethereum_pools
+    meta:
+      blockchain: ethereum
+      sector: dex
+      project: eulerswap
+      contributors: Henrystats
+    config:
+      tags: ['ethereum', 'pools', 'liquidity', 'eulerswap']
+    description: >
+      eulerswap Uni V4 ethereum DEX pools
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - blockchain
+            - pool
+    columns:
+      - &blockchain
+        name: blockchain
+        description: "Blockchain which the DEX is deployed"
+      - &project
+        name: project
+        description: "Project name of the DEX"
+      - &version
+        name: version
+        description: "Version of the contract built and deployed by the DEX project"
+      - &factory_address
+        name: factory_address
+        description: "Contract address used to create the pool"
+      - &creation_block_time
+        name: creation_block_time
+        description: "Block time of pool creation"
+      - &creation_block_number
+        name: creation_block_number
+        description: "Block number of pool creation"
+      - &pool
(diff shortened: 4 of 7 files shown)


henrystats:
hola ser @jeff-dude this is also ready for review (although it will have a couple of conflicts with the other PR, I'll resolve them once that gets merged).

Some of Eulerswap's Swaps go through Uniswap V4 and are already logged in dex.trades (under uniswap), I'm doing a left outer join to filter those out when pulling them into dex.trades but under eulerswap.trades they'll still be logged.

The trades can either go through Eulerswap's OG interface (normal dex trades) or through Uniswap V4 (in this case eulerswap works like an aggregator hence the exclusion).

henrystats:
thank you ser @jeff-dude! 

I've updated the way the macro is doing the lookup for uni v4 trades now and getting directly from the decoded tables as you mentioned!

The issue with getting the trades data into `_projects/eulerswap/` directory models from the `dex.trades` table is that there's a handful of columns that won't make it to `dex.trades` so there'd be no way of getting that data except if we got the data from the eulerswap swap decoded table again.

Ohh, do you mean that it's fine to do that since it'd mean we'd not have to get some of the dex.trades columns again  (like token symbols, amounts converted, amount_usd, tx_data etc), and we'd only add extra complexity to get the eulerswap columns that would be omitted from `eulerswap_base_trades` (uni v4 ones).?

jeff-dude:
> The issue with getting the trades data into `_projects/eulerswap/` directory models from the `dex.trades` table is that there's a handful of columns that won't make it to `dex.trades` so there'd be no way of getting that data except if we got the data from the eulerswap swap decoded table again.

you can write extra columns to the `eulerswap_<chain>.base_trades` tables upstream of `dex.trades`. these will live here only, as the chain-level models select certain columns only for `dex.trades`. then you can later join back to these base tables to obtain columns you need

> Ohh, do you mean that it's fine to do that since it'd mean we'd not have to get some of the dex.trades columns again (like token symbols, amounts converted, amount_usd, tx_data etc), and we'd only add extra complexity to get the eulerswap columns that would be omitted from `eulerswap_base_trades` (uni v4 ones).?

then you can read from `dex.trades` where `project = eulerswap` and join back to base tables for eulerswap in the models which live in `_project/eulerswap/`

does that make sense?

henrystats:
@jeff-dude, hmm, the plan was to include the trades that are also logged in Uniswap V4 into `eulerswap.trades` as well, if they're already filtered in `eulerswap_chain.base_trades` since we don't want to include them In `dex.trades` as they're actually Uniswap V4 trades then they'd not show up from simply doing a `select * from dex.trades where project = 'eulerswap'`. In those cases Eulerswap acts like an aggregator, so in `dex.trades` the trades shouldn't be included but when you're trying to get the volume of `Eulerswap` itself as a standalone, it should be included.

jeff-dude:
> @jeff-dude, hmm, the plan was to include the trades that are also logged in Uniswap V4 into `eulerswap.trades` as well, if they're already filtered in `eulerswap_chain.base_trades` since we don't want to include them In `dex.trades` as they're actually Uniswap V4 trades then they'd not show up from simply doing a `select * from dex.trades where project = 'eulerswap'`. In those cases Eulerswap acts like an aggregator, so in `dex.trades` the trades shouldn't be included but when you're trying to get the volume of `Eulerswap` itself as a standalone, it should be included.

you can read from `dex.trades where project = 'eulerswap'` (easy first step), then union in `dex.trades where project = 'uniswap' and version = '4' and tx_hash in (txs from eulerswap)`?

henrystats:
@jeff-dude, that wouldn't work since the trades are already filtered from dex.trades! So the tx_hash wouldn't show up at all. Or do you mean do a join with eulerswap trades event (when you say from eulerswap)?

I'm guessing that's the only option, since we'd need the extra columns for those trades that were filtered in `eulerswap_chain.base_trades`, we'd still need to do a join with the trade_events.

I think the only way to do it then is to:

`select * from eulerswap_chain.base_trades`

union all 

`rewrite the logic for eulerswap_chain.base_trades but only get Uni V4 trades`

do a join with `dex.trades` to get the extra columns from enrichments?

jeff-dude:
> @jeff-dude, that wouldn't work since the trades are already filtered from dex.trades! So the tx_hash wouldn't show up at all. Or do you mean do a join with eulerswap trades event (when you say from eulerswap)?
> 
> I'm guessing that's the only option, since we'd need the extra columns for those trades that were filtered in `eulerswap_chain.base_trades`, we'd still need to do a join with the trade_events.
> 
> I think the only way to do it then is to:
> 
> `select * from eulerswap_chain.base_trades`
> 
> union all
> 
> `rewrite the logic for eulerswap_chain.base_trades but only get Uni V4 trades`
> 
> do a join with `dex.trades` to get the extra columns from enrichments?

i'm operating under this assumption, so let me know where i'm wrong:
- all eulerswap swaps are in the decoded table (`eulerswapinstance_evt_swap` for all chains)
- in dex trades pipeline, for `eulerswap_<chain>.base_trades` initial table, you lookup to uniswap v4 decoded swap tx hashes to filter out since they already write in `uniswap_v4_<chain>.base_trades`
- both DEXs write to `dex.trades`, meaning all `tx_hash` values are accounted for, no missing data / swaps here, they are just categorized under different project / versions

since we know all swaps are in `dex.trades`, you can then build in `projects/eulerswap/` to build a model which contains all swaps for euler, even if `dex.trades` sources from uniswap:
- read from `dex.trades` for both `eulerswap` and `uniswap v4` project data
- join this data back to decoded tables / other spells for whatever other data / columns you need

the goal is to keep `dex.trades` pipeline simple, don't duplicate the logic elsewhere and move any complexity over to `_projects/` directory *after* writing to `dex.trades` first

plz let me know which part of this is an incorrect assumption

henrystats:
@jeff-dude  ah yes, that's correct, I understand clearly what you mean now! Yes that works, I'll make the updates now!

Apologies for the continuous tagging but I usually don't get notifications when you comment on the pr's except when I check my mail / come here to check so I assume it's the same for you haha and I just wanted to make sure you were aware I was responding to your comments!

henrystats:
@jeff-dude made the new changes now 😀

henrystats:
@jeff-dude I've implemented the changes you requested here, should be working (at the moment eulerswap downstream trades only include the ones already showing up in `dex.trades`). Since we're pulling the eulerswap ones from `dex.trades` as well, there's no way to test it right now without refreshing the `dex.trades` model itself so it's causing the CI run to time out!

jeff-dude:
> @jeff-dude I've implemented the changes you requested here, should be working (at the moment eulerswap downstream trades only include the ones already showing up in `dex.trades`). Since we're pulling the eulerswap ones from `dex.trades` as well, there's no way to test it right now without refreshing the `dex.trades` model itself so it's causing the CI run to time out!

definitely recommend always starting all dev in CI env here to be on hardcoded short timeframe to get quick iteration cycles. `dex_trades` model is a bit more complex to filter as it spans a few areas, but can be done.

it's not often a PR builds up to `dex.trades` and then on top, so this is a bit unique. you have the tooling available to get it to all run, just need to push test commits then revert.

dbt_subprojects/dex/models/trades/ethereum/_schema.yml
```diff
@@ -722,6 +741,6 @@ models:
             - tx_hash
             - evt_index
       - check_dex_base_trades_seed:
-          seed_file: ref('ekubo_v1_ethereum_base_trades_seed')
+          seed_file: ref('eulerswap_ethereum_base_trades_seed')
```
jeff-dude: looks like typo here?

dbt_subprojects/dex/models/_projects/eulerswap/eulerswap_trades.sql
```diff
@@ -0,0 +1,63 @@
+{{ config(
+    schema = 'eulerswap'
+    , alias = 'trades'
+    , partition_by = ['block_month']
+    , materialized = 'incremental'
+    , file_format = 'delta'
+    , incremental_strategy = 'merge'
+    , unique_key = ['blockchain', 'project', 'version', 'tx_hash', 'evt_index']
```
jeff-dude:
```suggestion
    , unique_key = ['blockchain', 'project', 'version', 'block_month', 'tx_hash', 'evt_index']
```

jeff-dude: lets include `block_month` as partition column, should ideally help trino query performance to find target data easier.

dbt_subprojects/dex/models/_projects/eulerswap/eulerswap_trades.sql
```diff
@@ -0,0 +1,63 @@
...
+    , incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')]
+    )
+}}
+
+{% set base_models = [
+    ref('eulerswap_ethereum_raw_trades')
+    , ref('eulerswap_bnb_raw_trades')
+    , ref('eulerswap_unichain_raw_trades')
```
jeff-dude:
```suggestion
    ref('eulerswap_ethereum_trades')
    , ref('eulerswap_bnb_trades')
    , ref('eulerswap_unichain_trades')
```

jeff-dude: would it make sense to remove `raw_` usage in alias names (then also file names and such)?
i feel like that could get confusing to compare `raw_` vs. `base_` terminology. these models techincally have all the data, so no longer raw/base level.

jeff-dude: only commenting here to keep it in one spot.

dbt_subprojects/dex/macros/models/_project/eulerswap_compatible_trades.sql
```diff
@@ -0,0 +1,219 @@
...
+        , s.amount0In
+        , s.amount0Out
+        , s.amount1In
+        , s.amount1Out
+        , s.sender -- router 
+        , s.evt_index
+        , case 
+            when t.tx_hash is not null then 'uni_v4' 
+            else 'OG' 
+         end as source
```
jeff-dude: do you have a specific use case for `source` other than filtering out downstream in final query reading this CTE? if not, you can remove need for the column and just filter `where t.tx_hash is null` to ignore rows written as uniswap v4 already.

jeff-dude: i see in your downstream macro below, you rewrite logic for `source` anyway. maybe this macro can be simplified to only filter out where lookup to uni v4 is null

henrystats: updated

dbt_subprojects/dex/models/_projects/eulerswap/eulerswap_trades.sql
henrystats: updated!

dbt_subprojects/dex/models/_projects/eulerswap/eulerswap_trades.sql
henrystats: included!

dbt_subprojects/dex/models/trades/ethereum/_schema.yml
henrystats: fixed

dbt_subprojects/dex/macros/models/_project/eulerswap_compatible_trades.sql
```diff
@@ -0,0 +1,215 @@
...
+    , dexs.pool_creation_time
+    , dexs.fee 
+    , dexs.protocolFee 
+    , dexs.instance 
+    , dexs.eulerAccount 
+    , dexs.factory_address 
+    , dexs.sender 
+from 
+dexs 
+where uni_tx_hash is not null
```
jeff-dude: i think you actually have this backwards, should be
`where uni_tx_hash is null`
which says "this euler tx_hash does not appear in uniswap v4 decoded swap, therefore add"
we want the lookup to return nothing to be a net new row to add

jeff-dude (changes requested):

dbt_subprojects/dex/macros/models/_project/eulerswap_compatible_trades.sql
henrystats: ahh sorry yeah, fixed now!

jeff-dude:
@henrystats can you try to push a commit which filters full lineage to short timeframe (last 7 days or so), so we can see CI run end-to-end?

in the meantime, i can approve with the assumption it runs as we expect. then we can mark ready for merge once that is complete and reverted to original state.

(using blocked tag to remind myself not to merge until this is done)

henrystats:
yep yep I can do that, on it right away ser @jeff-dude

henrystats:
hola ser @jeff-dude, finally got it to run, this the CI run here: https://github.com/duneanalytics/spellbook/actions/runs/16922992478/job/47953100923?pr=8551, it didn't test_incremental though :(, timed out right when run_incremental completed! should I limit it more or is this fine and good to go?

tested the eulerswap models as well, everything works fine!

jeff-dude:
> hola ser @jeff-dude, finally got it to run, this the CI run here: https://github.com/duneanalytics/spellbook/actions/runs/16922992478/job/47953100923?pr=8551, it didn't test_incremental though :(, timed out right when run_incremental completed! should I limit it more or is this fine and good to go?
> 
> tested the eulerswap models as well, everything works fine!

think we're safe with this run. we would've seen errors on incremental run most likely if unique tests had failed.

PR #8558, opened 2025-07-28
CUR2-364 `prices.tokens`: refactor to only pull from coinpaprika required tokens

jeff-dude:


jeff-dude:
due to `bitcoin` and `cardano` (only 14 total prices) forcing null address, i will not add the test for `not_null` to keep backwards compatibility. here is a quick manual test to look for same output in test, ignoring those:
```
  select blockchain, contract_address, count(1)
  from test_schema.git_dunesql_e709f01_prices_tokens
  where contract_address is not null
  group by blockchain, contract_address
  having count(1) > 1
```
returns no rows as expected.

jeff-dude:
> due to `bitcoin` and `cardano` (only 14 total prices) forcing null address, i will not add the test for `not_null` to keep backwards compatibility. here is a quick manual test to look for same output in test, ignoring those:
> 
> ```
>   select blockchain, contract_address, count(1)
>   from test_schema.git_dunesql_e709f01_prices_tokens
>   where contract_address is not null
>   group by blockchain, contract_address
>   having count(1) > 1
> ```
> 
> returns no rows as expected.

row count + this test still pass after latest commit.

dbt_subprojects/tokens/models/prices/prices_tokens.sql
```diff
@@ -36,93 +43,144 @@
...
+        , pt.blockchain
+        , pt.contract_address
+        , pt.symbol
+        , pt.decimals
+    from
+        {{ model }} as pt
+    inner join
+        {{ source('prices', 'trusted_tokens') }} as trusted
+        on pt.blockchain = trusted.blockchain
+        and pt.contract_address = trusted.contract_address
```
0xRobin: I think this introduces a sequencing problem where:
1. we can't add to trusted tokens until it's in coinpaprika prices for the v3 pipeline
2. we don't have it in coinpaprika prices until it's in prices.tokens
3. we don't have it in prices.tokens until it's in trusted tokens

jeff-dude:
okay, to summarize:
for legacy:
- moved all legacy mapping files to `legacy/` directory
- removed a test in legacy optimism that was failing on symbol lookup to `tokens.erc20` -- optimism has a unique format for legacy prices and not worth diving deep now. there was only one incorrect symbol. it won't affect new prices.
- added row_number function to optimism legacy mapping file to remove dupes, not sure how that got introduced when all i did was move files to another directory which triggered rebuild / test. not too worried on this either for legacy coverage

for new mapping file:
- cleaned up schema file to remove hardcoded blockchains, as its outdated
- removed test for 0 address with new native setup included
- added a few native tokens which were missing from `dune.blockchains` (KSM,DOT,TAC,NEAR)
- created new `prices_coinpaprika_trusted_tokens` to match `prices_v2_trusted_tokens` list, with coinpaprika mappings
- updated final `prices_tokens` to union new static trusted tokens file, along with others there previously

jeff-dude:
quick tests to ensure full coverage of trusted tokens remains
- 290 rows in trusted tokens, this includes native from `dune.blockchains`
- 287 rows in new `prices.tokens`
  - 3 missing are from `dune.blockchains`:
  - somnia is testnet and not avaiable on coinpaprika, will add during catalyst onboarding
  - XRP has no `contract_address` in `dune.blockchains`, so gets dropped out in sqlmesh models anyway
  - layerzero has no `contract_address` in `dune.blockchains`, so gets dropped out in sqlmesh models anyway


when switching below query to inner join instead, we see the 287 rows.

```
with tt as (
    --290 rows
    select
        blockchain
        , contract_address
    from
        prices_v2.trusted_tokens
)
, p as (
    select
        *
    from test_schema.git_dunesql_56c6715_prices_tokens as p
)
, joined as (
select
    tt.*
from
    tt
left join
    p
    on tt.blockchain = p.blockchain
    and tt.contract_address = p.contract_address
where p.blockchain is null
)
select
    *
from
    joined
```

0xRobin (approved):

PR #8560, opened 2025-07-29
mev atomic arbs by block_number instead of block_time

hildobby:
with sub-second blocks (like on BNB), block_time is no longer a reliable way to group by block. it was originally chosen because block_number isn't in dex.trades (although it imo definitely should be) and I wanted to keep things efficient while it worked.

PS: we reaaally need atuff like block_number and tx_index in dex.trades.


jeff-dude:
> PS: we reaaally need atuff like block_number and tx_index in dex.trades.

noted, will share internally. we are eventually going to audit the table for schema coverage and such.

jeff-dude (approved):

PR #8587, opened 2025-07-31
1inch-U76: update main lineage

max-morrow:
- added permitAndCall logic


jeff-dude:
@max-morrow if this is old and can be closed, plz do 🙏

PR #8590, opened 2025-08-01
CUR2-369 init abstract uniswap v2 & v3 forks base trades & reservoir swap v2/3

jeff-dude:

PR #8615, opened 2025-08-09
Abstract swap - fix the code

clizzard7:
Abstract only picks up token 0x159bec6740c6509691f4fa17d1b3b4c1321acf67 on dune - this token ca was found in the seed address, hence I deleted it. We dont know what the issue would be otherwise


jeff-dude:
there are known issues with abstract chain (i think all zk chains iirc) where creation traces don't work properly, leading to decoding not working as expected and not finding all pools. this is why you see only one.

plz see this PR, where we fixed by building our own creation traces, then decode on the fly in a macro:
https://github.com/duneanalytics/spellbook/pull/8590/files

also, we rebranded to `reservoir_swap`, as we were directed by the abstract chain directly to do this. 

closing this PR, you can use the above for what you need.

PR #8619, opened 2025-08-11
fix column name in uniswap optimism pools

jeff-dude:
fyi @henrystats 
any time we change a column name or add/remove to existing models, plz do run `dbt ls` to find all downstream models and ensure the column changes are accounted for.

### DIFF
--- a/dbt_subprojects/dex/models/_projects/gamma/optimism/gamma_optimism_uniswap_pools.sql
+++ b/dbt_subprojects/dex/models/_projects/gamma/optimism/gamma_optimism_uniswap_pools.sql
@@ -50,7 +50,7 @@ SELECT lp_name, contract_address
 
 SELECT distinct
     'optimism' AS blockchain,
-    lp_name, mm.contract_address, pool AS pool_contract, fee, token0, token1
+    lp_name, mm.contract_address, id AS pool_contract, fee, token0, token1
     FROM manual_mapping mm
     INNER JOIN {{ source('optimism', 'creation_traces') }} ct 
         ON ct.address = mm.contract_address


henrystats:
hola ser @jeff-dude 👋, oh I missed this, I actually ran `dbt ls` and it returned a lot of models under `dex_optimism_base_trades` so I thought this was one of those since those showed up because of the union all of those models in `dex_optimism_base_trades`, I retained the column names for all the models call `uniswap_optimism_pools`, the only change was to change the model they were calling to `uniswap_v3_optimism_pools`.

jeff-dude:
> hola ser @jeff-dude 👋, oh I missed this, I actually ran `dbt ls` and it returned a lot of models under `dex_optimism_base_trades` so I thought this was one of those since those showed up because of the union all of those models in `dex_optimism_base_trades`, I retained the column names for all the models call `uniswap_optimism_pools`, the only change was to change the model they were calling to `uniswap_v3_optimism_pools`.

you changed `pool` to `id` universally, but left out a model (modified here).
`dbt ls --resource-type model --output name -s uniswap_optimism_pools+`

henrystats:
@jeff-dude the models should only reference `uniswap_v3_optimism_pools` and the column names wouldn't have to be changed, since at the time only the `uniswap_v3` pools were in the `uniswap_optimism_pools` model so I'm sure the pools being joined were just `v3` pools hence why there was no need for a column change!

just needed to update the models being called from `uniswap_optimism_pools` to `uniswap_v3_optimism_pools` (I retained the previous naming for this model)!

but yeah I missed the gamma optimism models! really sorry!

PR #8621, opened 2025-08-11
1inch-U80: Upgrade list & logic behind mapped_contracts

grkhr:
- Added missing contract, mostrly for linea, unichain, sonic
- Migrated 0x settlers to dynamic list

Please merge it with https://github.com/duneanalytics/spellbook/pull/8605 and https://github.com/duneanalytics/spellbook/pull/8598 to prevent full scan of the same data


grkhr:
> keep in mind, 0x team can (unlikey, but can) change the settler models at any time causing impact here. if an issue arises, we will only exclude from prod and tag for awareness to re-enable.

Yes, for sure. Didn't want to complicate the lineage with 10 more incremental models with full raw logs scans. We'll see.

PR #8630, opened 2025-08-13
Update Taiko prices

crStiv:
there is USDC.E, thought USDT.e might be good to be there too


crStiv:
the token id is correct, I dunno why cursor bot was triggered

crStiv:
it seems to be that "Token Prices Automation" triggers due to a CoinPaprika API rate limit. not my changes

PR #8631, opened 2025-08-13
CUR2-447 refactor: Safe's hourly spellbook via dbt macros

safehjc:

### Description:

This PR refactors the hourly spellbook for the Safe project. Macros are used to boil down boilerplate code in all the networks. The `gnosis_safe` namespace is across the networks. This is due to the fact that we cleaned up the decoded contracts to match a single submission style:
- `Safe_v1_X_Y`
- `SafeL2_v1_X_Y`
- `SafeProxyFactory_v1_X_Y`

where X and Y represent the major deployments as defined [here](https://github.com/safe-global/safe-deployments/tree/main/src/assets). Additionally, some validation was included to check for official Safe deployments.

PS: This is my first contribution to the spellbook so please bear with me.

---


safehjc:

dbt_subprojects/hourly_spellbook/macros/project/safe/safe_safes.sql
```diff
@@ -0,0 +1,52 @@
...
+        0xa97ab18a, -- setup method v1.0.0
+        0xb63e800d -- setup method v1.1.0, v1.1.1, v1.2.0, v1.3.0, v1.3.0L2, v1.4.1, v1.4.1L2, v1.5.0, v1.5.0L2
+    )
+    and et.gas_used > 10000  -- to ensure the setup call was successful. excludes e.g. setup calls with missing params that fallback
+    {% if date_filter %}
+    and et.block_time >= date_trunc('day', now() - interval '7' day)
+    {% elif not is_incremental() %}
+    and et.block_time > TIMESTAMP '{{ start_date }}' -- for initial query optimisation
+    {% elif is_incremental() %}
+    and et.block_time > date_trunc('day', now() - interval '7' day)
```
0xRobin:
```suggestion
    and {% incremental_predicate('et.block_time') %}
```

dbt_subprojects/hourly_spellbook/models/_project/safe/safe_singleton_validation.sql
```diff
@@ -0,0 +1,65 @@
...
+        END as safe_version,
+        CASE 
+            {% set official_addresses = get_official_safe_addresses() %}
+            WHEN LOWER(CAST(address AS varchar)) IN (
+                {%- for addr in official_addresses %}
+                LOWER('{{ addr }}'){{ "," if not loop.last else "" }}
+                {%- endfor %}
+            ) THEN true
+            ELSE false
+        END as is_official
```
0xRobin: nitpick, this would probably be a bit cleaner if we could do it with a left join instead of the generate case when statements with macros.
Not blocking though, if it works it works.

dbt_subprojects/hourly_spellbook/models/_project/safe/safe_singleton_validation.sql
0xRobin, line 1: I would probably call this model `safe_singletons_all` to match the other cross chain models?

dbt_subprojects/hourly_spellbook/macros/project/safe/safe_singletons.sql
0xRobin, line 1: I think you're using the validated versions in all the chain models, so these macro's are now redundant?

dbt_subprojects/hourly_spellbook/macros/project/safe/safe_singletons_validated.sql
```diff
@@ -0,0 +1,269 @@
...
+{% if sources_list and sources_list|length > 0 %}
+
+WITH all_singletons AS (
+    {%- for source_info in sources_list %}
+    select distinct singleton as address 
+    from {{ source('gnosis_safe_' ~ blockchain, source_info) }}
+    {%- if date_filter %}
+    WHERE evt_block_time >= date_trunc('day', now() - interval '7' day)
+    {%- endif %}
+    {%- if not loop.last %}
```
0xRobin: We could make the singleton models incremental by adding an incremental filter here as well.

Note that you would also need to adapt the config macros used in the chain models themselves.

dbt_subprojects/hourly_spellbook/macros/project/safe_native_transfers.sql
```diff
@@ -25,7 +25,9 @@ from
         and et.success = true
         and (lower(et.call_type) not in ('delegatecall', 'callcode', 'staticcall') or et.call_type is null)
         and et.value > UINT256 '0' -- et.value is uint256 type
-        {% if not is_incremental() %}
+        {% if date_filter %}
+        and et.block_time >= date_trunc('day', now() - interval '7' day)
+        {% elif not is_incremental() %}
         and et.block_time >= TIMESTAMP '{{ project_start_date }}'
         {% else %}
         and et.block_time > date_trunc('day', now() - interval '10' day) -- to prevent potential counterfactual safe deployment issues we take a bigger interval
```
0xRobin: I think we should use 
```
and {% incremental_predicate('et.block_time') %}
```
here as well. 
But might be missing some info on why a bigger interval would be needed here.

(same comment applies below on similar setups)

dbt_subprojects/hourly_spellbook/models/_project/safe/safe_singleton_validation.sql
safehjc: Thanks for the suggestion! Due to the limited amount of deployments and due to the fact that I don't want to get another CTE materialized, I left it with the current case statement. If the official addresses grow, I agree that the left join will be more efficient here (apart from being cleaner).

safehjc:
Thanks @0xRobin for the review, I just came back from vacation and tried to fix all your suggestions (apart from the left join one).

jeff-dude:
from what i can tell, `date_filter` is used to test short timeframes. i have set to `false` universally to prep for merge.

jeff-dude:
something to consider on next iteration of cleanup / enhancements, i notice a pattern of converting varbinary addresses to strings, then using lower function to compare address vs. address and filtering. our dunesql trino engine actually handles varbinary comparisons better performance wise. we should ideally let the native data type persist and the engine do it's thing. unless there's a reason for this pattern that i'm missing, which is v possible.

PR #8639, opened 2025-08-14
1inch-U81: added methods to orders* lineage

max-morrow:
- added methods


jeff-dude:
@max-morrow if this is old and can be closed, plz do 🙏

PR #8647, opened 2025-08-17
1inch-U82: main lineage updates

max-morrow:
- added maker/taker (max) amount to oneinch.lop


sources/aave/ethereum/aave_ethereum_sources.yml
```diff
@@ -124,16 +124,9 @@ sources:
...
       - name: StataTokenV2_evt_Initialized
 
-      - name: StataTokenFactory_evt_StataTokenCreated      
-
-  - name: aave_v3_ethereum
-    description: "Decoded contracts for Aave v3 on Ethereum"
-    tables:
-      - name: VariableDebtToken_evt_Initialized
-      - name: poolconfigurator_evt_reserveinitialized      
-      - name: etherfipoolconfiguratorinstance_evt_reserveinitialized
```
jeff-dude, line 134: are we certain removing these sources doesn't impact other models in any of the dbt subprojects?
when in doubt, i'd rather let sources persist to be safe.

jeff-dude: this applies to any sources you've removed across source files, only commenting here for simplicity.

dbt_subprojects/dex/macros/models/_project/oneinch/_meta/oneinch_meta_cfg_macro.sql
jeff-dude, line 1: it makes me a bit nervous to have the macro(s) set up so that any time a new blockchain needs to be added to the lineage, modifications are made, which then triggers every single blockchain to rebuild in prod if we don't manually interfere with the run post-merge.

in an ideal design, macros are blockchain agnostic and only changes to logic / bug fixes are applied to then rebuild all chains. new chain additions should run only that chain when merged, controlled by model additions/changes instead of macro.

is there something i'm missing in your design where i can easily manage this as you add chains in the future? fwiw, our prod runs use `state:modified+` to rebuild changes to main branch, so that's why all chains would run when macro is modified.

jeff-dude: again, i'll only comment here to centralize the convo, but it seems to apply in many spots.

jeff-dude: i think this also applied to existing code on main, but if we're here for an overall redesign, the time feels right for the discussion.

sources/aave/ethereum/aave_ethereum_sources.yml
max-morrow: none of the resources were deleted, they were moved to [sources/aave/aave_sources.yml](https://github.com/duneanalytics/spellbook/pull/8647#diff-511a6566bcb4c23e2d1744a37001b7fad908e6d3e2dce7fd032a80d72c458d30) for consistency.
some were in separate files, some in general (most in general).

dbt_subprojects/dex/macros/models/_project/oneinch/_meta/oneinch_meta_cfg_macro.sql
max-morrow: it seemed to us that we took this into account when designing our macros.

that is, we checked that changing the settings of such general macroses for one blockchain / adding a new blockchain to them - does not affect the models of other blockchains.
now even any extra lines and spaces that appear after macros are additionally removed.

we checked that such changes do not affect the final code generation for other blockchains.
so the state will not be modified for them. wdyt?

sources/aave/ethereum/aave_ethereum_sources.yml
jeff-dude: ah, didn't notice that, thank you for clarifying

dbt_subprojects/dex/macros/models/_project/oneinch/_meta/oneinch_meta_cfg_macro.sql
jeff-dude: let me open a separate PR to test what i'm thinking about. a visual should help. you can then let me know if it's a realistic use case and if we need to handle it.

jeff-dude:
@max-morrow since this data feeds into `dex.trades`, we would love to see the testing in CI env here to validate data quality compared to prod.

can you perform another "easy dates" run for a specified date range in the past? maybe a 3-day window in early october?

then we can write a query to join CI table to prod, filter prod to same date / block number ranges and validate data quality

dbt_subprojects/dex/macros/models/_project/oneinch/_meta/oneinch_meta_cfg_macro.sql
jeff-dude: okay, my example is in #8941, you can see in the CI logs that all models which interact with the macro are trigger in `state:modified` usage.
https://github.com/duneanalytics/spellbook/actions/runs/18949345678/job/54109504687?pr=8941#step:11:77

the example use case is removing a blockchain here, but i'm envisioning the opposite use case where you are adding chains to config macros. this should output same results, any modification to macro.

i also see in the current pipeline, the macro i modified is nested, as in only called by other macros and not model directly. it seems this still triggers modified.

if that PR were merged, all those models in the list linked above would perform a full refresh, if we don't make manual steps to override.

jeff-dude: this is why an ideal design would involve configs being within the models themselves, so blockchain models remain isolated.

jeff-dude:
> @max-morrow since this data feeds into `dex.trades`, we would love to see the testing in CI env here to validate data quality compared to prod.
> 
> can you perform another "easy dates" run for a specified date range in the past? maybe a 3-day window in early october?
> 
> then we can write a query to join CI table to prod, filter prod to same date / block number ranges and validate data quality

thank you for building a sample set in CI. what i'd love to see:
- run `dbt compile` locally to get raw sql
- find path to `dex.trades` model
- remove all non-inch models (we have a few paths of table reads, can remove main `dex.base_trades`, balancer, 0x, etc.)
- replace sources with CI table for oneinch swaps `test_schema.git_dunesql_11e37d1_oneinch_swaps`
- run the query with that as source, wrap in CTE
- join to prod to ensure all rows are accounted for, values in non-key columns match when matching rows

if you'd prefer to modify the `dex.trades` model in this PR directly with the above and build via CI to then join to prod, that works too, whatever is faster.

let me know if you'd like support on what i'm requesting here

max-morrow:
> > then we can write a query to join CI table to prod, filter prod to same date / block number ranges and validate data quality

Here's the [reconciliation](https://dune.com/queries/6092127) (26 days window).
The data on a minor volume discrepancy has also been verified - the updated version is more correct.
Screenshot 2025-10-31 at 14 43 09

dbt_subprojects/dex/macros/models/_project/oneinch/_meta/oneinch_meta_cfg_macro.sql
max-morrow: @jeff-dude Thank you for highlighting! I tested it on my own, but I still don't understand why models are triggered, in which nothing has changed at all in the code (in result generations). In other words, why is an `arbitrum_ar` triggered, for example, when deleting a `sonic`? I'm trying to figure it out, I'll be back.

jeff-dude: without digging too deep in how `state:modified` works in dbt, this is just the behavior we've found. any modification to a macro will mark all models which use it as a breaking change.

surely there is some override you can do in dbt macros but we aren't in position to do that across spellbook right now. the best path forward is to contain all configs within models themselves, so the macro stays agnostic and all chain-specific configs are in models.

max-morrow: @jeff-dude I've done another redesign, cut it into separate macros. I came up with a not very cumbersome, quite concise implementation. Now adding/removing a blockchain or a stream won't affect the others. I still need to update the description file, test the results, and run the light dates for the final time. Then I'll withdraw from the draft and write to you guys

jeff-dude: awesome, look forward to seeing how you designed it!

max-morrow:
@jeff-dude I think it is finally ready for review.
[easy dates run](https://github.com/duneanalytics/spellbook/actions/runs/19225188681/job/54950761623?pr=8647)
[check](https://dune.com/queries/6092127) streams to `dex.trades` & `dex_aggregator.trades`

max-morrow:
hey, @jeff-dude!
The [current errors](https://github.com/duneanalytics/spellbook/actions/runs/19328927838/job/55286922921?pr=8647) that occur when checking for duplicates are analyzed and fixed here, in [9024](https://github.com/duneanalytics/spellbook/pull/9024).
Unfortunately, without that fix, this PR won't work.

jeff-dude:
wanted to give a heads up: bnb is giving trouble in prod for full history. working through it. transfers from traces final view may be delayed for a bit. most chains are built, other than bnb and i think polygon.

jeff-dude:
> wanted to give a heads up: bnb is giving trouble in prod for full history. working through it. transfers from traces final view may be delayed for a bit. most chains are built, other than bnb and i think polygon.

full lineage should be good now.

max-morrow:
thank you @jeff-dude!

seems it is finally ready for merge, [easy dates success run](https://github.com/duneanalytics/spellbook/actions/runs/19387902438/job/55477394982?pr=8647)

updated stream [check](https://dune.com/queries/6092127) to `dex.trades` & `dex_aggregator.trades`

jeff-dude:
> thank you @jeff-dude!
> 
> seems it is finally ready for merge, [easy dates success run](https://github.com/duneanalytics/spellbook/actions/runs/19387902438/job/55477394982?pr=8647)
> 
> updated stream [check](https://dune.com/queries/6092127) to `dex.trades` & `dex_aggregator.trades`

looks good!
i will see if i have time monday to deploy. we will be in buenos aires for dunecon + devconnect, so timing is difficult. please be patient with us through next week.

jeff-dude (approved):

PR #8706, opened 2025-09-03
Fix duplicate issue in aptos_fungible_asset_metadata

ying-w:
### Description:

Duplicate issue that came up in https://github.com/duneanalytics/spellbook/pull/8682 was due to `migration` table having duplicates.

~`migration` table should be a one time build (this solves the issue)~
`migration` table should be unique but will keep growing. CoinStore is fully migrated but coin (v1) can still be created (added to migration table on mint)

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_migration.sql
+++ b/dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_migration.sql
@@ -30,10 +30,6 @@ FROM (
         AND move_module_address = 0x0000000000000000000000000000000000000000000000000000000000000001
         AND move_resource_module = 'coin'
         AND move_resource_name = 'PairedCoinType'
-    {% if is_incremental() %}
-        AND {{ incremental_predicate('block_time') }}
-    {% else %}
         AND block_date BETWEEN DATE('2024-08-02') AND DATE('2025-09-01') -- FA (v2) migration period
-    {% endif %}
     GROUP BY move_address
 )
--- a/dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_migration.sql
+++ b/dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_migration.sql
@@ -1,8 +1,10 @@
 {{ config(
     schema = 'aptos_fungible_asset',
     alias = 'migration',
-    materialized = 'table',
+    materialized = 'incremental',
     file_format = 'delta',
+    incremental_strategy = 'merge',
+    incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')],
     unique_key = ['asset_type_v2'],
 ) }}
 
@@ -11,7 +13,7 @@ SELECT
     tx_version,
     block_date,
     block_time,
-    date(date_trunc('month', block_time)) as block_month,
+    DATE(date_trunc('month', block_time)) as block_month,
     --
     '0x' || LPAD(lower(to_hex(move_address)), 64, '0') AS asset_type_v2,
     '0x' || LPAD(LTRIM(json_extract_scalar(move_data, '$.type.account_address'), '0x'), 64, '0') || '::' ||
@@ -30,6 +32,10 @@ FROM (
         AND move_module_address = 0x0000000000000000000000000000000000000000000000000000000000000001
         AND move_resource_module = 'coin'
         AND move_resource_name = 'PairedCoinType'
-        AND block_date BETWEEN DATE('2024-08-02') AND DATE('2025-09-01') -- FA (v2) migration period
+    {% if is_incremental() %}
+        AND {{ incremental_predicate('block_time') }}
+    {% else %}
+        AND block_date >= DATE('2024-08-02') -- beginning of FA (v2) migration
+    {% endif %}
     GROUP BY move_address
 )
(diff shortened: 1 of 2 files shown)


jeff-dude:
@ying-w a bit confused on this one. there are no changes that would fix duplicates. what changed so that it runs fine here in CI? if we let it run daily in prod without changes, will it fail again after a few runs?

ying-w:
@jeff-dude this confuses me too, the problem is that the current version of the table doesn't have unique constraint enforced. When I rebuilt it (initially as a table then switched back to incremental) I can confirm that the current logic is unique.

```sql
-- returns 3
SELECT *
FROM (
SELECT *, ROW_NUMBER() OVER (PARTITION BY asset_type_v2) AS rn
FROM aptos_fungible_asset.migrations
)
WHERE 1=1
-- AND rn = 2
AND asset_type_v2 = '0xff6bf040bcb9f515ea728f907eb70548d4fd8e407b015dd4592fcfe6b2a0a654'
LIMIT 100
```

```sql
-- returns 1
SELECT
    -- latest
    tx_version,
    block_date,
    block_time,
    DATE(date_trunc('month', block_time)) as block_month,
    --
    '0x' || LPAD(lower(to_hex(move_address)), 64, '0') AS asset_type_v2,
    '0x' || LPAD(LTRIM(json_extract_scalar(move_data, '$.type.account_address'), '0x'), 64, '0') || '::' ||
    FROM_UTF8(FROM_HEX(LTRIM(json_extract_scalar(move_data, '$.type.module_name'), '0x'))) || '::' ||
    FROM_UTF8(FROM_HEX(LTRIM(json_extract_scalar(move_data, '$.type.struct_name'), '0x'))) AS asset_type_v1
FROM (
    SELECT
        MAX(tx_version) AS tx_version,
        MAX(block_date) AS block_date,
        MAX(block_time) AS block_time,
        --
        move_address,
        ANY_VALUE(move_data) AS move_data
    FROM aptos.move_resources
    WHERE 1=1
        AND move_module_address = 0x0000000000000000000000000000000000000000000000000000000000000001
        AND move_resource_module = 'coin'
        AND move_resource_name = 'PairedCoinType'
        AND block_date = DATE('2025-08-20')
    GROUP BY move_address
)
WHERE 1=1
AND move_address = 0xff6bf040bcb9f515ea728f907eb70548d4fd8e407b015dd4592fcfe6b2a0a654
```

I'm not sure how the previous duplicate issue happened in `migration` table (given that i group by the field and unique on it so it should merge against it) but it's why there was a duplicate issue in https://github.com/duneanalytics/spellbook/pull/8682

dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_migration.sql
```diff
@@ -1,20 +1,19 @@
 {{ config(
     schema = 'aptos_fungible_asset',
-    alias = 'migrations',
+    alias = 'migration',
```
jeff-dude, line 3:
```suggestion
    alias = 'migrations',
```

jeff-dude: lets keep alias the same, we can push a change with config commented below

dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_migration.sql
```diff
@@ -1,20 +1,19 @@
...
     materialized = 'incremental',
     file_format = 'delta',
     incremental_strategy = 'merge',
     incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_time')],
```
jeff-dude:
```suggestion
```

jeff-dude: i think you may need to remove this config, which filters the target to same small timeframe as source read. your lookup to target is failing since you don't look for full history. it appears there are use cases where a migration can happen multiple times over time? here is example:
```
select *
from aptos_fungible_asset.migrations
where asset_type_v2 = '0x82db577bdf9877f8f0eb63897ccb468f6a0b9c0133ecbb25a3b77b2ca2e8ee61'
```

this one has four migrations at different block times? then this config would make us miss in the lookup and not update the row, but rather insert a dupe.

if you can confirm this use case, then should fix our issue.

dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_migration.sql
ying-w: there is a difference between name of file (no s) and alias (has s). I wanted to make it consistent and figured changing alias was better than changing file name

dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_migration.sql
ying-w: ah, i was assuming that it would do a merge against the whole table so `unique_key` in config would remove duplicates. just copy and pasted without looking into what `incremental_predicates` did

ying-w:
@jeff-dude are you ok with the name change for consistency or should I change file name (and also schema file)

jeff-dude:
> @jeff-dude are you ok with the name change for consistency or should I change file name (and also schema file)

either direction is fine. ultimately, i believe only you and team use this table so it will only break queries that you write. if you prefer to change alias, we can let that happen. we also don't absolutely need to fix mismatch between file name / alias, but it's nice to have consistent to help navigate the repo. ultimately it's your choice, so just tag me when you finalize and we can mark ready for merge.

ying-w:
@jeff-dude can we keep it as it is in the current PR (new alias that aligns with file name)

jeff-dude (approved):
we can leave as-is here, marking ready for merge

PR #8712, opened 2025-09-04
expose aptos fungible asset tables

ying-w:
Expose aptos fungible asset tables
- activities
- balances

Add comments about slight differences from indexer graphql endpoint

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_activities.sql
+++ b/dbt_subprojects/daily_spellbook/models/aptos/aptos_fungible_asset_activities.sql
@@ -90,10 +90,12 @@ WITH coin_activities AS (
     WHERE 1=1
         AND ev.block_date >= DATE('2023-07-28') -- v2 deployed
         AND event_type IN (
+            -- prior to Events v2 migration, different event types were used
+            -- this happened between 2023-08-04 to 2024-05-29
             '0x1::fungible_asset::Deposit',
-            '0x1::fungible_asset::DepositEvent', -- old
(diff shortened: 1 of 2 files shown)


ying-w:
Can I also get a backfill for the FAA table for the two new event types over that time range

jeff-dude (approved):

jeff-dude:
> Can I also get a backfill for the FAA table for the two new event types over that time range

for any table you would like backfilled, plz include in this PR. then it'll happen automatically. you can simply leave a sql comment in there to trigger the modified state, similar to how you have comments in here per model.

ying-w:
> for any table you would like backfilled, plz include in this PR. then it'll happen automatically

@jeff-dude the table in question is `aptos_fungible_asset.activities` I included 2 additional events that need to be parsed (`0x1::fungible_asset::DepositEvent`, `0x1::fungible_asset::WithdrawEvent`). These new rows should be appended to current table and only exist between block_date 2023-08-04 to 2024-05-29

Does merging this PR cause a rebuild of the table? If so that would fix it.

jeff-dude:
> @jeff-dude the table in question is `aptos_fungible_asset.activities` I included 2 additional events that need to be parsed (`0x1::fungible_asset::DepositEvent`, `0x1::fungible_asset::WithdrawEvent`). These new rows should be appended to current table and only exist between block_date 2023-08-04 to 2024-05-29
> 
> Does merging this PR cause a rebuild of the table? If so that would fix it.

ah yes, that file modified here will definitely rebuild fully on merge. should be good.

PR #8729, opened 2025-09-10
add hyperevm to evms tables

hildobby:
adding hyperevm to all tables in the evms sector.

cc @MSilb7


MSilb7:
Do we need to add schema ymls? @hildobby 

Also hyperevm doesn't have traces yet in Dune, wonder if we need to handle that.

Does this also flow through to gas.fees or do we need to add that as well?

hildobby:
i think it's still missing quite a few raw tables so i'll wait for those to be added before i revisit this @MSilb7 
PixelSnap 2025-09-11 at 00 21 00@2x

Issue #8748, opened 2025-09-15
[BUG] 1inch ethereum AR 4 model has introduced duplicates


jeff-dude:
@grkhr @max-morrow heads up, 1inch data will be behind due to this bug. plz help resolve this when you can. my assumption is that it's related to null `trace_address`, but haven't dug in to see exactly why.

production runs are failing on `dex_aggregator.trades` model due to upstream models introducing duplicates. it appears to be specific to ethereum chain, 1inch project, version AR 4. it would be good to double check that.

here are some queries to debug:
```
select *
from dex_aggregator.trades
where
    blockchain = 'ethereum'
    and project = '1inch'
    and version = 'AR v4'
    and block_date >= date '2025-09-15'
    and tx_hash = 0x46606decdc61dd4a3ba2ed1a033ec324ee61922968033c4ac240f906f25c4f70
    -- and evt_index = -1
    -- and trace_address is null
limit 100
```

```
select block_date, blockchain, tx_hash, trace_address, count(1)
from oneinch.ar_trades
group by block_date, blockchain, tx_hash, trace_address
having count(1) > 1
limit 100
```

grkhr:
Hey! Sorry about that. Btw strange this hasn’t happened before.

[Fix](https://github.com/duneanalytics/spellbook/pull/8749) is ready for merging @jeff-dude

PR #8749, opened 2025-09-15
1inch: fix dex_agg trades dups

grkhr:
Fix https://github.com/duneanalytics/spellbook/issues/8748


jeff-dude:
`evt_index` isn't part of the unique key, looks like it's the same three rows failing in CI that the query in issue shows? imagine due to missing trace address?

grkhr:
`ar_trades` is a view so there is no usage of `unique_key`, idk why it's exposed, it can be removed

grkhr:
@jeff-dude all set

PR #8882, opened 2025-10-17
CUR2-757 Fix Somnex V2 Somnia 

krishgka:
The model was incorrectly using the call_createpair table which stores tokens in the order they were passed to the function (tokenA, tokenB). However, the Uniswap V2 factory contract sorts addresses internally before creating pairs (smaller address = token0, larger address = token1).

This caused buy/sell directions to be reversed when tokenA > tokenB.

Changes:
- Switch from somnexammfactory_call_createpair to somnexammfactory_evt_paircreated
- Use event's token0/token1 (sorted) instead of call's tokenA/tokenB (unsorted)
- Update seed file with correct test data (buy/sell were previously swapped)
- Add missing transaction (evt_index 7) to seed file

### Description:

[...]

---

### DIFF
--- a/dbt_subprojects/dex/models/trades/somnia/platforms/somnex_v2_somnia_base_trades.sql
+++ b/dbt_subprojects/dex/models/trades/somnia/platforms/somnex_v2_somnia_base_trades.sql
@@ -10,54 +10,12 @@
     )
 }}
 
-{% set somnex_start_date = "2025-08-29" %}
-
-WITH dexs AS (
-    SELECT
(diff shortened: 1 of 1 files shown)


jeff-dude:
@krishgka without looking too closely, does this mean we can then also update this model to use the fork macro instead to keep clean? logic looks similar 🤔

krishgka:
> @krishgka without looking too closely, does this mean we can then also update this model to use the fork macro instead to keep clean? logic looks similar 🤔

Yup you're right, changed it!

PR #8911, opened 2025-10-23
bridges: tether's usdt0 on unichain

hildobby:
adding Tether's USDT0 bridge starting on Unichain to test macros and models. Once merged I'll expand to all other chains in a follow up PR


jeff-dude (approved):
appreciate this pattern of starting with one chain, expanding all others in follow up. makes it much smoother.

hildobby:
> appreciate this pattern of starting with one chain, expanding all others in follow up. makes it much smoother.

yeah it gets messy if i try to add many new currently untracked chains at once to the bridges sector, so i'm waiting for you to merge those (currently this one https://github.com/duneanalytics/spellbook/pull/8903) before i move onto the next one. all other PRs will add the macros + the spell to an existing chain as a test.


butter is probably next and will add > 10 new chains to the sector, once most chains are already in the sector there it should also be easier to add new bridges

PR #8919, opened 2025-10-24
Fix univ4 logic (include fees)

henrystats:

### Description:

[...]

---

---

> [!NOTE]
> Switch v4 liquidity logic to decode `PoolManager_call_ModifyLiquidity`, add `fees_accrued`/`swap_fees_paid`, and update models, sources, seeds, and daily agg across chains.
> 
> - **Core v4 liquidity logic**:
>   - Decode `PoolManager_call_ModifyLiquidity` outputs to derive `amount0/amount1` and fees (sign-extended int256 splits).
>   - Deterministically pair calls↔events, enrich with `sqrtPriceX96`, and emit new event types: `modify_liquidity`, `fees_accrued`, and `swap_fees_paid`.
>   - Replace prior fee handling via `Take` calls; remove old logic.
> - **Models updated (multi-chain)**:
>   - Pass `PoolManager_call_ModifyLiquidity` into all v4 chain models and wire through macro.
>   - Simplify `uniswap_base_liquidity_events` union config and refresh comments.
> - **Aggregations**:
>   - In `uniswap_daily_agg_liquidity_events`, exclude `fees_accrued` from volume sums.
> - **Seeds & tests**:
>   - Add `event_type` column to all v4 seed schemas and CSVs; update values.
>   - Update/adjust seed-based checks to filter `event_type: modify_liquidity`; remove obsolete tests where applicable.
> - **Sources**:
>   - Add `PoolManager_call_ModifyLiquidity` to sources for all supported chains.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 4827a2b4a4d5b1653430f4af19bc5aa75a194f37. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
```diff
@@ -235,237 +235,226 @@ get_prices as (
...
+            , -1* cd.callerDelta_token0 as amount0
+            , -1* cd.callerDelta_token1 as amount1
+            , -1* cd.feesAccrued_token0  as fee_amount0
+            , -1* cd.feesAccrued_token1  as fee_amount1
+        FROM 
+        evts e
+        INNER JOIN 
+        calls_decoded cd
+            ON cd.call_tx_hash = e.evt_tx_hash
+            AND cd.call_rn = e.evt_rn
```
jeff-dude, line 408:
```suggestion
            ON cd.call_tx_hash = e.evt_tx_hash
            AND cd.call_block_date = e.evt_block_date
            AND cd.call_block_number = e.evt_block_number
            AND cd.call_rn = e.evt_rn
```

dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
```diff
@@ -235,237 +235,226 @@ get_prices as (
...
-    single_pools as (
-        select 
-            tx_hash 
-            , count(distinct id) as num_pools 
+    evts as (
+        select
+            contract_address,
+            evt_tx_hash,
+            evt_tx_from,
+            evt_block_time,
```
jeff-dude, line 359:
```suggestion
            evt_block_time,
            evt_block_date,
```

dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
```diff
@@ -235,237 +235,226 @@ get_prices as (
...
-        where call_success
-        {%- if is_incremental() %}
-        and {{ incremental_predicate('call_block_time') }}
-        {%- endif %} 
+            contract_address,
+            call_success,
+            call_tx_hash,
+            call_tx_from,
+            call_tx_to,
+            call_trace_address,
```
jeff-dude, line 287:
```suggestion
            call_trace_address,
            call_block_date,
```

jeff-dude: you could also carry this field forward into final select statements (i.e. you use `block_time` to build `block_date` in final select), but it's more critical to use in the join conditions below.

dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
jeff-dude: you could also carry this field forward into final select statements (i.e. you use `block_time` to build `block_date` in final select), but it's more critical to use in the join conditions below.

dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
jeff-dude, line 213: my other open question is on this table read. from what i can tell, this table contains the same number of pools as liquidity pools above, but has a 1:many relationship and some pools have many rows. the total count of this table pushes 28mil+ as of today (and will grow). considering some pools only have 1 row, or v few rows, it makes sense that you can't filter it incrementally like we do event based tables.

however, is there an opportunity to filter it so that each pool only pulls forward the current `block_index_sum` + `previous_block_index_sum`? or do you need full history each time?

either way, the issue seems to be on initial build which means def needing all rows, but may be worth exploring for incremental runs.

jeff-dude (changes requested):

dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
jeff-dude: we should always include `block_date` (`block_number` somewhat optional, but makes it clear) when joining on `tx_hash`. i noticed increased performance when running the query on dune with this added.

henrystats: ouu! thank you ser, will include this!

dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
henrystats: got it, thank you ser!

dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
henrystats: got it, thank you ser!

dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
henrystats: hmm, i don't think there's any way to do that since we can't know the last `previous_block_index_sum` that's required if the pool went inactive for a period of time longer than the window of the incremental run!

One thing i could add in the incremental runs is to get data from the prices table only for IDs that were active in the incremental runs, so `select * from prices_table where id in (select id from evt_table where incremental_predicate)`. what do you think?

henrystats: tried that and because of how large the table is, it slows down the incremental runs!

jeff-dude: okay, we can try as-is again and see how prod handles it.
if we have issues again, we may need to consider multi-model setup here where we break out some of the CTEs and turn into models themselves, so heavy logic is separated.

PR #8931, opened 2025-10-29
CUR2-899 bridges: handle withdrawal duplicates

hildobby:
Last week we had duplicates for deposits and I added handling for those. This error is the same but for withdrawals, so this PR fixes those in the same way. With this merged, there should no longer be duplicates in any spells culminating into (and including) `bridges_evms.flows`


jeff-dude (approved):
thank you 🙏

PR #8938, opened 2025-10-30
bridges: more duplicate entries fixes

hildobby:
- added simpler unique_keys to raw models
- fix for duplicate entries on merges based on tx_hash, and also based on unique_bridge_transfers
- changed avalanche native bridge to use the unique_key from `tokens.transfers` instead of surrogate keys generated from [tx_hash, evt_index]. this is to account for data pulled from tokens.transfers with multiple eth transfers (which have no evt_index) in a single tx


dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/evms/bridges_evms_withdrawals_raw.sql
```diff
@@ -4,7 +4,6 @@
     , materialized = 'incremental'
     , file_format = 'delta'
     , incremental_strategy='merge'
```
jeff-dude, line 6: we should always assign unique keys on all types of incremental models. otherwise, merge will fail to find matches on lookup to target and always run insert no matter what. this means we are intentionally writing dupes into this model, then deduplicating downstream in final model

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/evms/bridges_evms_deposits_raw.sql
```diff
@@ -3,8 +3,7 @@
     , alias = 'deposits_raw'
     , materialized = 'incremental'
     , file_format = 'delta'
-    , incremental_strategy='append'
-    , unique_key = ['deposit_chain','withdrawal_chain','withdrawal_chain_id','bridge_name','bridge_version','bridge_transfer_id', 'tx_hash', 'evt_index']
+    , incremental_strategy='merge'
```
jeff-dude, line 6: same feedback here, need unique keys on any incremental model

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/evms/bridges_evms_withdrawals_raw.sql
hildobby: I added back [withdrawal_chain, tx_hash, evt_index] as a unique key

dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/evms/bridges_evms_deposits_raw.sql
hildobby: I added back [deposit_chain, tx_hash, evt_index] as a unique key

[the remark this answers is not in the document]
hildobby: that was fixed

jeff-dude (approved):

PR #9035, opened 2025-11-14
bridges: add orbiter v1 deposits to existing chains (29 chains)

hildobby:
splitting https://github.com/duneanalytics/spellbook/pull/9034 into two parts, this is part 1 which contains the orbiter models on chains already added to the bridges sector


hildobby:
@jeff-dude this PR is also ready for review, i cant merge the two PRs given it results into too many query stages

PR #9036, opened 2025-11-14
bridges: add orbiter v1 deposits to new chains

hildobby:
splitting https://github.com/duneanalytics/spellbook/pull/9034 into two parts, this is part 2 which contains the orbiter models on chains that weren't in the bridges sector and thus are added here

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/chains/zksync/platforms/_schema.yml
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/bridges/flows/chains/zksync/platforms/_schema.yml
@@ -64,7 +64,6 @@ models:
     config:
       tags: [ 'zksync', 'bridges', 'flows', 'withdrawals' ]
     description: "Symbiosis's v1 bridge deposits events on zkSync"
-    description: "Celer's v1 bridge withdrawals events on zkSync"
   - name: bridges_zksync_nitro_v1_deposits
     meta:
       blockchain: zksync
--- a/sources/_subprojects_outputs/tokens/transfers.yml
+++ b/sources/_subprojects_outputs/tokens/transfers.yml
@@ -259,3 +259,9 @@ sources:
       - name: transfers
       - name: net_transfers_daily
       - name: net_transfers_daily_asset
+
+  - name: tokens_story
+    tables:
+      - name: transfers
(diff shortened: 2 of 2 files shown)


hildobby:
@jeff-dude this PR adds a few other chains for orbiter, it seems to be unable to run CI tests since it goes above the max 500 query stages, how should we address this?

jeff-dude:
> @jeff-dude this PR adds a few other chains for orbiter, it seems to be unable to run CI tests since it goes above the max 500 query stages, how should we address this?

i fear we've hit a breaking point in the model lineage design. since it's setup as:
views --> table(s)

when we reach tables portion of pipeline, it is reading a ton of views, and hitting the stages issue. we have had a similar issue with `dex.trades` lineage, on the table which union'd views of each chain-level dex model. the fix we applied was to materialize each chain-level model into an incremental model, then have the cross-chain remain a view.

that should help you get around this issue. the more extreme alternative is to materialize all upstream bridge-specific models, then use views downstream, but for short term i think you can start with chain-level unions being materialized incrementally.

hildobby:
> > @jeff-dude this PR adds a few other chains for orbiter, it seems to be unable to run CI tests since it goes above the max 500 query stages, how should we address this?
> 
> i fear we've hit a breaking point in the model lineage design. since it's setup as: views --> table(s)
> 
> when we reach tables portion of pipeline, it is reading a ton of views, and hitting the stages issue. we have had a similar issue with `dex.trades` lineage, on the table which union'd views of each chain-level dex model. the fix we applied was to materialize each chain-level model into an incremental model, then have the cross-chain remain a view.
> 
> that should help you get around this issue. the more extreme alternative is to materialize all upstream bridge-specific models, then use views downstream, but for short term i think you can start with chain-level unions being materialized incrementally.

alright, i'm going to start working on a pr that switches to meterialisation at the chain-level: https://github.com/duneanalytics/spellbook/pull/9038

the other advantage ton this change is that this will have tests at chain level, making it easier to test downstream models in CI runs since testing crosschain won't be needed as tests move upstream

jeff-dude:
> the other advantage ton this change is that this will have tests at chain level, making it easier to test downstream models in CI runs since testing crosschain won't be needed as tests move upstream

yes, definite advantages of this path

PR #9056, opened 2025-11-21
oneinch daily: test short timeframe on affected models

jeff-dude:


jeff-dude:
@max-morrow here is the evidence that daily project models will fail with new setup.

i will need to exclude them until your PR to redesign is finalized. i'll be out next week, so plz coordinate in slack with our team.

PR #8806, opened 2025-09-29
using underlying token prices for atokens

data-cowwboy:
### Description:
atoken prices are not available in prices.usd table, this change intends to start using the underlying asset's price for atoken trades

[...]

---
quick links for more information:
- [README.md](https://github.com/duneanalytics/spellbook/blob/main/README.md)
- [spellbook docs](https://github.com/duneanalytics/spellbook/tree/main/docs)
- [CONTRIBUTING.md](https://github.com/duneanalytics/spellbook/blob/main/CONTRIBUTING.md)

PR #9060, opened 2025-11-24
Add Aave token prices for CoW Protocol trades on all chains

harisang:
This PR add prices for CoW Protocol trades on all chains (except for mainnet, where it has already been done [here](https://github.com/duneanalytics/spellbook/pull/8806)) by using the underlying asset price, whenever the atoken price is missing.


jeff-dude:
this is now a duplicate, as i've prepped the other PR for merge to do this.

PR #9081, opened 2025-12-03
CUR2-1056 Fix aptos dupes

jeff-dude:


jeff-dude:
live prod table: 6 rows
new CI table: 1 row (as expected)
(this query will no longer work once deployed, but confirming now)
```
select *
-- from aptos_fungible_asset.balances fab
from test_schema.git_dunesql_7b8343d_aptos_fungible_asset_balances fab
where
    fab.storage_id = 0xeeab1b38aac7ac5735461ef830d327f579f5e0fc7535ad4298e79511665db729
    AND fab.tx_version = 3806171709
```

jeff-dude:
@ying-w fyi on latest fixes for aptos curated datasets.

we got alerts in prod that asset activities was failing on duplicates, it turns out that it's due to balances having duplicates upstream.

in balances, `write_set_change_index` has some null values, but is used as unique key. this causes lookups to fail and re-insert rows. i have gone ahead and generated a surrogate key on the key columns to ensure no nulls. from what i can tell, the nulls are introduced in the deletion model, as there is a left join to grab the field and no coalesce when no match.

if you need to follow up with a fix for that column, feel free, but i didn't want to make assumptions on what the value should be. this fix is sufficient to get the models running on track again, and reach parity of data quality with prod current state.

please also be aware of [this](https://github.com/duneanalytics/spellbook/pull/9081/commits/e41ef97acf3c3ac3db5c4d6c6ec9b676429eb7ae) commit, which removed a few non-key columns tests for not null which failed during CI runs after i forced short run time periods. the actual values in tables were not modified.

PR #9088, opened 2025-12-04
CUR2-1059: extend `stablecoin.balances`

tomfutago:

### Description:

extend stablecoin balances to additional chains:
- add stablecoin balances for bnb, kaia, optimism, worldchain, zksync
- update erc20 stablecoins across chains
- add missing daily aggregation balances for bnb and zksync

blocked by #9094 

---

### DIFF
--- a/dbt_macros/shared/balances_incremental_subset_daily.sql
+++ b/dbt_macros/shared/balances_incremental_subset_daily.sql
@@ -342,14 +342,6 @@ tokens_metadata as (
     from {{ source('tokens', 'erc20') }}
 ),
 
-native_tokens as (
-    select
-        blockchain,
-        token_address as native_token_address,
-        token_symbol as native_token_symbol
-    from {{ source('dune', 'blockchains') }}
-),
-
 enriched_with_tokens as (
     select
         b.blockchain,
@@ -368,16 +360,12 @@ enriched_with_tokens as (
             when b.token_standard = 'erc20' then b.balance_raw / power(10, t.decimals)
             when b.token_standard = 'native' then b.balance_raw / power(10, 18)
             else b.balance_raw
-        end as balance,
-        nt.native_token_address,
-        nt.native_token_symbol
+        end as balance
     from base b
     left join tokens_metadata t
         on t.blockchain = b.blockchain
         and t.contract_address = b.token_address
         and b.token_standard = 'erc20'
-    left join native_tokens nt
-        on nt.blockchain = b.blockchain
 )
 
 select
@@ -398,14 +386,7 @@ left join {{ source('prices_external', 'day') }} p
     {% if is_incremental() %}
     and {{ incremental_predicate('p.timestamp') }}
     {% endif %}
-    and (
-        (e.token_standard = 'erc20'
-            and e.blockchain = p.blockchain
-            and e.token_address = p.contract_address)
-        or (e.token_standard = 'native'
-            and e.blockchain = p.blockchain
-            and e.native_token_address = p.contract_address
-            and e.native_token_symbol = p.symbol)
-    )
+    and e.blockchain = p.blockchain
+    and e.token_address = p.contract_address
 
 {% endmacro %}


tomfutago:
### update summary:

1. Tokens
- `tokens_<chain>.erc20_stablecoins` - based on `evm_tokens` list from [this query](https://dune.com/queries/5936691)
  - updated existing models
  - added 3 new ones for kaia, worldchain and celo (this last one doesn't have raw balances set up, so won't have `stablecoins_celo.balances` model)
- daily agg balances - added 2 chains
  - :white_check_mark: zksync
  - :x: bnb - kept failing to build so `prod_excluded` for the time being

2. Daily Spellbbok
- created `stablecoins_balances` macro
  - it uses sql logic based on existing models, ie: [`stablecoins_arbitrum_balances`](https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/daily_spellbook/models/_sector/stablecoins/arbitrum/stablecoins_arbitrum_balances.sql)
  - removed `select distinct` that some models were using on `stablecoin_tokens` cte - as `erc20_stablecoins` models has unique key on `contract_address` anyway
  - `ethereum` is an exception as it also contains `filter_tokens` list - added as an optional param
- added 4 new chains: kaia, optimism, worldchain and zksync, but as they relying on `tokens_<chain>.erc20_stablecoins` models added in tokens lineage - build fails so `prod_excluded` for now

### to be discussed:
- current `stablecoins_<chain>_balances` models have 2 main parts:
  - 1st `balances` as base dataset - built using `balances_incremental_subset_daily` macro - columns returned include `token_address` and `token_symbol`
  - 2nd `stablecoin_tokens` (list from `tokens_<chain>.erc20_stablecoins`)
  - logic in model's final output:

```
select
    t.symbol
    ,b.*
from balances b
left join stablecoin_tokens t
    on b.token_address = t.token_address
```

2 open questions:
1. instead of `left join` - should it be `inner join`? otherwise balances are not really filtered to just stablecoins, right?
2. `symbol` is added - even though it matches (95% sure that's true) `token_symbol` from balances

0xRobin:
from my understanding the `balances_incremental_subset_daily` macro usage should already filter to the exact tokenset that we want? And a subsequent filter (or join to enrich the symbol) should not be needed.

jeff-dude:
> from my understanding the `balances_incremental_subset_daily` macro usage should already filter to the exact tokenset that we want? And a subsequent filter (or join to enrich the symbol) should not be needed.

yes, `token_symbol` comes from the `balances_incremental_subset_daily` macro output and we should use that directly as it comes from `tokens.erc20`. these may have been built prior to `tokens.erc20` being automated, so symbols could've been missing. we can take advantage of the refactor now and simply use the output and rename `token_symbol` --> `symbol` and remove final join altogether.

dbt_macros/shared/balances_incremental_subset_daily.sql
```diff
@@ -166,3 +166,243 @@ left join {{source('prices','usd_daily')}} p
...
+    
+    @PARAM blockchain               -- blockchain name
+    @PARAM address_list             -- must have an address column, can be none if only filtering on tokens
+    @PARAM token_list               -- must have a token_address column, can be none if only filtering on tokens
+    @PARAM address_token_list       -- for advanced usage, must have both (address, token_address) columns, can be none
+    @PARAM start_date               -- the start_date, used to generate the daily timeseries
+
+#}
+
+{%- macro balances_incremental_subset_daily_new(
```
jeff-dude: assuming this will be renamed to remove `_new` once the legacy one is merged and live

dbt_macros/shared/balances_incremental_subset_daily.sql
```diff
@@ -166,3 +166,243 @@ left join {{source('prices','usd_daily')}} p
...
+    union all
+    select * from (
+        select
+            blockchain
+            ,max(day) as day
+            ,address
+            ,token_address
+            ,token_standard
+            ,token_id
+            ,max_by(balance_raw, day) as balance_raw
```
jeff-dude, line 256: okay so here we use raw rather than amount with erc20 join. have you been able to confirm via CI tables built that the final result (after enrichment, of course) matches our live prod tables for existing chains?
just want to be extra safe that this still works this setup

dbt_macros/shared/balances_incremental_subset_daily.sql
```diff
@@ -166,3 +166,243 @@ left join {{source('prices','usd_daily')}} p
...
+        symbol,
+        decimals
+    from {{ source('tokens', 'erc20') }}
+),
+
+native_tokens as (
+    select
+        blockchain,
+        native_token_symbol
+    from {{ source('evms', 'info') }}
```
jeff-dude: we should use `dune.blockchains` when we can, not evms info

dbt_macros/shared/balances_incremental_subset_daily.sql
tomfutago: yes, new macro is actually not called anywhere yet (stablecoin balances models are reverted)

dbt_macros/shared/balances_incremental_subset_daily.sql
tomfutago: so far none of the ci runs completed - that's why trying to split this pr. will test it out on a single chain once this pr is merged.

dbt_macros/shared/balances_incremental_subset_daily.sql
tomfutago: noted, will update

dbt_macros/shared/balances_incremental_subset_daily.sql
```diff
@@ -166,3 +166,246 @@ left join {{source('prices','usd_daily')}} p
...
+    
+    @PARAM blockchain               -- blockchain name
+    @PARAM address_list             -- must have an address column, can be none if only filtering on tokens
+    @PARAM token_list               -- must have a token_address column, can be none if only filtering on tokens
+    @PARAM address_token_list       -- for advanced usage, must have both (address, token_address) columns, can be none
+    @PARAM start_date               -- the start_date, used to generate the daily timeseries
+
+#}
+
+{%- macro balances_incremental_subset_daily_new(
```
jeff-dude:
```suggestion
{%- macro balances_incremental_subset_daily(
```

dbt_macros/shared/balances_incremental_subset_daily.sql
jeff-dude: this will be test in part 2

dbt_macros/shared/balances_incremental_subset_daily.sql
```diff
@@ -166,3 +166,246 @@ left join {{source('prices','usd_daily')}} p
...
+    and {{ incremental_predicate('p.timestamp') }}
+    {% endif %}
+    and (
+        (e.token_standard = 'erc20'
+            and e.blockchain = p.blockchain
+            and e.token_address = p.contract_address)
+        or (e.token_standard = 'native'
+            and e.blockchain = p.blockchain
+            and e.native_token_address = p.contract_address
+            and e.native_token_symbol = p.symbol)
```
0xRobin: This symbol match should not be needed anymore as we use the prices.day tables.

jeff-dude: good callout, we can join to prices like normal on `prices_external.day`
see transfers enriched macro for reference
https://github.com/duneanalytics/spellbook/blob/6b7fac9b1806383275c7108d3547e39751043034/dbt_subprojects/tokens/macros/transfers/transfers_enrich.sql#L90-L94

0xRobin: quick note that in transfers we have already corrected the native contract address in the base table to be in sync with the contract address in dune.blockchains. 
This is not the case for balances (and would be out of scope to fix here).

PR #9089, opened 2025-12-04
1inch-test-2

max-morrow:
Testing hourly prices for `_project.*` lineage


max-morrow:
It seems hourly prices to be a fairly working option as a compromise for `_project.*` lineage

PR #9095, opened 2025-12-05
1inch-U86: add a new settlement contract

max-morrow:
- added a new settlement contract
- temporary exclude solana from `lo` stream
- additional columns to `oneinch.swaps`


max-morrow:
@jeff-dude a small hot PR, please take a look too

PR #9107, opened 2025-12-09
add columns to polymarket models

hildobby:
added columns to existing tables (needed for much more efficient queries) and created 3 tables:
- one for protocol tvl
- one for user balance changes
- one for currently open positions (view)


hildobby:
@cursor review

hildobby:
@cursor review

hildobby:
@cursor review

hildobby:
@cursor review

hildobby:
https://github.com/cursor review

hildobby:
@cursor review

0xBoxer:
@hildobby would be highly appreciated if you could tell us what to do with this or finish it

hildobby:
PR is ready btw, it has a a few new spells and some new columns in a few tables

I also optimised the spells a bunch since runtime was crazy on some of those. It's far from perfect but should require a lot less compute than they used to. I made sure to keep all old columns for backward compatibility.

The only missing thing is for the market details spell which needs to have uploaded base table switched from dune's (deprecated afaik) to polymarket's uploaded one but it's missing a needed column so I pinged Primo to ask if he could add it. We can merge as is and tackle that in another follow up PR

@0xBoxer

hildobby:
@jeff-dude could you review this PR please?

jeff-dude:
> @jeff-dude could you review this PR please?

before diving deep, want to give you heads up that we have three open PRs to materialize polymarket views into incremental tables. once those are deployed and rebuilt, you can merge changes into your branch here. then we can look into the additional columns and tables. our changes will not change the data output or schema at all, simply materialize into tables for performance downstream.

hildobby:
> > @jeff-dude could you review this PR please?
> 
> before diving deep, want to give you heads up that we have three open PRs to materialize polymarket views into incremental tables. once those are deployed and rebuilt, you can merge changes into your branch here. then we can look into the additional columns and tables. our changes will not change the data output or schema at all, simply materialize into tables for performance downstream.

do as you need, I switched out of dune to using another more reliable source for polymarket data since most things blow past dune's query runtime anyways because the spells are in a rough state and merging this took way too long

jeff-dude:
> do as you need, I switched out of dune to using another more reliable source for polymarket data since most things blow past dune's query runtime anyways because the spells are in a rough state and merging this took way too long

understood. this is why we materialized, to help with query performance. we are also spending time Q2 on enhancing prediction market datasets.

hildobby:
@jeff-dude yes but materialising them won't fix this fyi. you can't even get volume from them right now and need to defer to decoded even tables, same for tvl (costs insane compute because dates are string for example) and many other super basic metrics. the queries are very suboptimal in their current state, hence my PR here. In this PR I tried to keep all currently active columns while adding new ones that I think are more useful (so it will remain with a ton of technical debt of unneeded stuff either way)

jeff-dude:
> @jeff-dude yes but materialising them won't fix this fyi. you can't even get volume from them right now and need to defer to decoded even tables, same for tvl (costs insane compute because dates are string for example) and many other super basic metrics. the queries are very suboptimal in their current state, hence my PR here. In this PR I tried to keep all currently active columns while adding new ones that I think are more useful (so it will remain with a ton of technical debt of unneeded stuff either way)

okay thanks for clarifying. i have shared internally with the team for reference.

tomfutago:
bugbot run

jeff-dude:
edit: message from claude 🤖 but plan is sound. we will revisit this after that merge.

Heads-up: [#9678](https://github.com/duneanalytics/spellbook/pull/9678) refactors `polymarket_polygon_positions.sql` and adds a new `polymarket_polygon_positions_raw_changes.sql` model for a large perf win on incremental + full-refresh runs of `positions_raw` / `positions`. We'd like to land it first, then have this PR rebase on top.

What that means for this PR when you rebase on main after #9678 merges:

- **`polymarket_polygon_positions_raw.sql`**: structurally rewritten (no longer uses `balances_incremental_subset_daily`; reads from a new `_changes` model, carry-forwards from `{{ this }}`, forward-fills via `unnest(sequence(...))`). The columns / output rows are byte-identical (validated row-level on a 1/1000 sample and aggregate-level on full history). Your PR doesn't touch this file, so no conflict expected.
- **`polymarket_polygon_positions.sql`**: structurally rewritten — single-pass join with `positions_to_emit` union (recent window + partition-pruned historical rewrite for changed markets), `partition_by = ['block_month']` instead of `['day']` (Trino 100-open-writer cap mitigation), `changed_markets` trigger narrowed to `outcome` / `resolved_on_timestamp` only. The other three `changed_markets` flags (`active` / `closed` / `accepting_orders`) become snapshot-at-write going forward, matching the other 9 metadata columns from `market_details`.

  You'll get a real conflict here. Your `price`, `usd_value`, and the `final_price` / `market_start_time_parsed` / `market_end_time_parsed` plumbing should slot in as:
  - one `left join {{ ref('polymarket_polygon_market_prices_daily') }} as dp on dp.token_id = p.token_id and cast(dp.day as date) = p.day` added after the existing `inner join market_details`
  - the `case when mm.market_end_time_parsed is null or p.day <= mm.market_end_time_parsed then coalesce(dp.price, 0) else coalesce(mm.final_price, 0) end as price` + `usd_value` columns added to the final SELECT (both branches)

- **`polymarket_polygon_market_trades.sql` / `polymarket_polygon_market_trades_raw.sql` / `_schema.yml`**: also conflict, but unrelated to #9678 — these are from [#9672](https://github.com/duneanalytics/spellbook/pull/9672) on main adding `contract_version` / `builder` / `metadata` to the trades pipeline. You'd hit these regardless.

All your new models (`protocol_tvl`, `users_balance_changes`, `open_positions`, `positions_enriched_base`, etc.) integrate without conflict.

Happy to help review the conflict resolution when you rebase.

jeff-dude:
these models were moved to a private repo, can't be modified here.

PR #9116, opened 2025-12-11
CUR2-1119: `native.mints` overflow bug

tomfutago:

### Description:

dbt error on full refresh:
```
Database Error in model nft_base_native_mints (models/_sector/mints/native/nft_base_native_mints.sql)
  TrinoUserError(type=USER_ERROR, name=NUMERIC_VALUE_OUT_OF_RANGE, message="UINT256 addition overflow: 1 + 115792089237316195423570985008687907853269984665640564039457584007913129639935", query_id=20251211_103156_00228_v5qs7)
```

---

### DIFF
--- a/dbt_subprojects/nft/macros/nft_mints.sql
+++ b/dbt_subprojects/nft/macros/nft_mints.sql
@@ -11,7 +11,7 @@ WITH namespaces AS (
 , nfts_per_tx AS (
     SELECT
         tx_hash
-        , sum(amount) AS nfts_minted_in_tx
+        , SUM(CAST(amount AS DOUBLE)) AS nfts_minted_in_tx
     FROM {{ ref('nft_transfers') }}
     WHERE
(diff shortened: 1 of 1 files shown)


0xRobin:
@tomfutago can we quickly check why we were seeing these high values for Base? Maybe we actually have some bad data or a bug in upstream nft transfer models?

tomfutago:
> @tomfutago can we quickly check why we were seeing these high values for Base? Maybe we actually have some bad data or a bug in upstream nft transfer models?

checked sources for `nft.transfers` and this query one returns 4 rows on max value of uint256:
`select * from erc1155_base.evt_TransferSingle where cast(value as double) >= cast('115792089237316195423570985008687907853269984665640564039457584007913129639935' as double)`

so sum on any of those 4 rows would fail - not sure tho why it didn't happen before as all 4 are at least 2 months old 🤔

PR #9147, opened 2025-12-19
refactor jupiter agg swaps lineage to build efficiently

jeff-dude:
towards CUR2-1181

### DIFF
--- a/dbt_subprojects/solana/models/jupiter/jupiter_v4_solana_aggregator_swaps.sql
+++ b/dbt_subprojects/solana/models/jupiter/jupiter_v4_solana_aggregator_swaps.sql
@@ -1,6 +1,5 @@
 {{
   config(
-        tags = ['prod_exclude'],
         schema = 'jupiter_v4_solana',
         alias = 'aggregator_swaps',
         partition_by = ['block_month'],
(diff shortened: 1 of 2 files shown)


jeff-dude:
previous model config left out `jup_version` from unique key, which led to some failures in prod.
based on [this](https://solscan.io/tx/2R8MsoSFceF8qxZezDegDbFRjHmvyGT7nvVaeh67KRAasZ175yHvLcX15P83Lb7Aofu6Z1Ef64nBna4gEHHMewkr) example tx, version is required to identify a unique row. the same input/output token can be swapped on both v4 and v6 in the same tx.

new model design splits out per version for efficient builds, which also helps avoid adding version to unique key to not conflict across versioned models.

the final view will check for unique rows still as a test with version included.

the output of this change should not change data at all from current state in prod. simple model redesign.

jeff-dude:
bugbot run

jeff-dude:
bugbot run

jeff-dude:
v5 looks to have a short lifespan of active onchain activity
```
select count(1), min(block_time), max(block_time)
from jupiter_solana.aggregator_swaps
where jup_version = 5
```
returns only a few months in 2023. due to this, marking model static.

PR #9168, opened 2025-12-31
add ORE

SocialfiPanda:
Add ORE to token list


SocialfiPanda:

jeff-dude:
from what i can tell, this token is not supported on coinpaprika. that is why it is failing.

PR #9185, opened 2026-01-08
add stablecoin transfer models

0xBoxer:
Adding a stablecoin.transfers model


0xBoxer:
@cursoragent please review

0xBoxer:
The things that cursor mentions are tradeoffs I am aware of, we'll fill the list of stablecoins in over time

dbt_subprojects/tokens/models/stablecoins/evm/stablecoins_evm_transfers.sql
```diff
@@ -0,0 +1,74 @@
...
+                                  spell_type = "sector",
+                                  spell_name = "stablecoins",
+                                  contributors = \'["tomfutago"]\') }}'
+  )
+}}
+
+select *
+from {{ ref('stablecoins_transfers') }}
+    
+)
```
tomfutago: this closing bracket seems orphaned?

dbt_subprojects/tokens/models/stablecoins/evm/abstract/transfers/stablecoins_abstract_core_transfers.sql
```diff
@@ -0,0 +1,21 @@
...
+    incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.block_date')]
+  )
+}}
+
+-- core transfers: tracks transfers for stablecoins in the frozen core list
+
+{{ stablecoins_transfers(
+    blockchain = chain,
+    token_list = 'core',
+    start_date = '2025-01-01'
```
tomfutago: if we want full history - `start_date` needs to match origin `block_date` on each `core` model.

send cursor to pull it from: `select start_date from dune.blockchains where name = <chain>`

on the hand, `start_date` on `extended` balance models was intentionally set as '2025-01-01` but with:
`-- note: update start_date when adding new stablecoins`
so maybe worth doing the same for transfers

tomfutago: ah `extended_transferss` models have this comment already, so ignore 2nd part

0xBoxer: I think there might not be much of a point to include the date here at all, just adds unneccesarry complexity, removing from macro and modesl

tomfutago: good point - no need to actually have this param.

i went with the `start_date` as that's what original stablecoin balances macro had, but in the hindsight - it'd be probably better to remove there too.

tomfutago (approved):

tomfutago:
@cursoragent final review

tomfutago:
@cursor review

PR #9238, opened 2026-01-22
fix polymarket market details

0xBoxer:
change table name to fit new schema inside of dune

fixed the upload api to use batch uploading via /insert, hence using new namespace


0xBoxer:
bugbot run

PR #9304, opened 2026-02-10
show these two great tables again in search

lorenz234:
Hi, I noticed these fundamental two tables do not show up in the search box anymore. 

Without these two tables calculating blob fees is basically impossible, I feel they should be shown on the front page. 

What do you think?


jeff-dude:
the tables can still be used as normal, even without showing on data explorer. for now, we will keep out of front page display.

PR #9337, opened 2026-02-18
optimize address info lineage

jeff-dude:
towards CUR2-1484

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/_sector/addresses/chains/_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/_sector/addresses/chains/_schema.yml
@@ -821,3 +821,203 @@ models:
       - *last_tx_block_number
       - *last_seen
       - *last_seen_block
+
+  - name: addresses_fantom_stg_info
+    meta:
+      blockchain: fantom
+      sector: addresses
+    config:
+      tags: ['staging', 'address']
+    description: "Staging: address activity in incremental window for Fantom. No target join."
+
+  - name: addresses_fantom_info
+    meta:
+      blockchain: fantom
+      sector: addresses
+    config:
+      tags: ['table', 'address', 'info']
+    description: "High level information about every address on Fantom"
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - address_prefix
+            - address
+    columns:
+      - *blockchain
+      - *address
+      - *address_prefix
+      - *executed_tx_count
+      - *max_nonce
+      - *is_smart_contract
+      - *namespace
+      - *name
+      - *first_funded_by
+      - *first_funded_by_block_time
+      - *tokens_received_count
+      - *tokens_received_tx_count
+      - *tokens_sent_count
+      - *tokens_sent_tx_count
+      - *first_transfer_block_time
+      - *last_transfer_block_time
+      - *first_received_block_number
+      - *last_received_block_number
+      - *first_sent_block_number
+      - *last_sent_block_number
+      - *received_volume_usd
+      - *sent_volume_usd
+      - *first_tx_block_time
+      - *last_tx_block_time
+      - *first_tx_block_number
+      - *last_tx_block_number
+      - *last_seen
+      - *last_seen_block
+
+  - name: addresses_linea_stg_info
+    meta:
+      blockchain: linea
+      sector: addresses
+    config:
+      tags: ['staging', 'address']
+    description: "Staging: address activity in incremental window for Linea. No target join."
+
+  - name: addresses_linea_info
+    meta:
+      blockchain: linea
+      sector: addresses
+    config:
+      tags: ['table', 'address', 'info']
+    description: "High level information about every address on Linea"
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          combination_of_columns:
+            - address_prefix
+            - address
+    columns:
+      - *blockchain
+      - *address
+      - *address_prefix
+      - *executed_tx_count
+      - *max_nonce
+      - *is_smart_contract
+      - *namespace
+      - *name
+      - *first_funded_by
+      - *first_funded_by_block_time
+      - *tokens_received_count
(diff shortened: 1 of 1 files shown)


jeff-dude:
(a few human words): below is a summary of the redesign, with support from cursor agents. the addition of more frequent merges within incremental window is a reasonable trade-off, to avoid more target lookups (4 times in old, once in new). there is also usage of `address_prefix` for partitioning and joins now, which should help the lookup and merge conditions.

------------------------------------------------------------------------------------------------

Here's the full comparison, focused on COUNT/SUM fields like `executed_tx_count`, `tokens_sent_count`, `sent_volume_usd`, etc.

---

## Old design (single model, no staging)

Everything lived in one model per chain (e.g. `addresses_arbitrum_info`). The model itself contained the raw source reads and the target lookup.

**Incremental flow for `executed_tx_count` (COUNT) on address `0xABC`:**

| Step | What happens |
|------|-------------|
| 1. Read sources | `COUNT(*)` from `transactions` WHERE `block_time` in incremental window = **5** (new txs) |
| 2. `to_update` CTE | `LEFT JOIN {{ this }}` to check if `0xABC` needs update. It does. |
| 3. Final SELECT | Read `{{ this }}` again: existing `executed_tx_count` = **100**. Output: `5 + 100 = 105`. |
| 4. Merge | dbt merges row with `executed_tx_count = 105` into target. |

**Problems:**
- **4 target scans** per run (one `LEFT JOIN {{ this }}` per CTE in the aggregation macro, plus the `to_update` check)
- Source reads and target merge logic interleaved in one model

---

## New design (3-layer: staging, chain info, unified)

### Layer 1: `addresses_<chain>_stg_info` (staging)

Calls `addresses_info` macro. Reads raw sources with `incremental_predicate` on source `block_time`. Aggregates per address. **No target lookup.**

For `0xABC` with 5 new txs in the window:

| Field | Value | How |
|-------|-------|-----|
| `executed_tx_count` | **5** | `COUNT(*)` from transactions in window |
| `tokens_sent_count` | **3** | `COUNT(*)` from token_transfers in window |
| `sent_volume_usd` | **200** | `SUM(amount_usd)` from token_transfers in window |

This is a **delta only** — just the new activity. Merged into the staging target table via dbt merge on `(address_prefix, address)`.

### Layer 2: `addresses_<chain>_info` (chain info)

Calls `addresses_info_incremental` macro. This is where history is preserved.

```
nd = staging rows WHERE last_seen in incremental window
t  = {{ this }} (existing chain info target)

output = nd + t
```

For `0xABC`:

| Field | nd (staging delta) | t (existing target) | Output | Formula |
|-------|-------------------|-------------------|--------|---------|
| `executed_tx_count` | 5 | 100 | **105** | `nd + COALESCE(t, 0)` |
| `tokens_sent_count` | 3 | 50 | **53** | `COALESCE(nd, 0) + COALESCE(t, 0)` |
| `sent_volume_usd` | 200 | 8000 | **8200** | `COALESCE(nd, 0) + COALESCE(t, 0)` |
| `first_tx_block_time` | 2026-02-18 | 2024-01-15 | **2024-01-15** | `COALESCE(t, nd)` — keep earliest |
| `last_tx_block_time` | 2026-02-18 | 2026-02-17 | **2026-02-18** | `COALESCE(nd, t)` — keep latest |

The key lines from the macro:

```7:7:dbt_subprojects/daily_spellbook/macros/sector/addresses/addresses_info_incremental.sql
	, nd.executed_tx_count + {% if is_incremental() %}coalesce(t.executed_tx_count, 0){% else %}0{% endif %} as executed_tx_count
```

```14:16:dbt_subprojects/daily_spellbook/macros/sector/addresses/addresses_info_incremental.sql
	, coalesce(nd.tokens_received_count, 0) + {% if is_incremental() %}coalesce(t.tokens_received_count, 0){% else %}0{% endif %} as tokens_received_count
	, coalesce(nd.tokens_received_tx_count, 0) + {% if is_incremental() %}coalesce(t.tokens_received_tx_count, 0){% else %}0{% endif %} as tokens_received_tx_count
	, coalesce(nd.tokens_sent_count, 0) + {% if is_incremental() %}coalesce(t.tokens_sent_count, 0){% else %}0{% endif %} as tokens_sent_count
```

**This is the exact same arithmetic as the old design.** The difference is that it happens in one clean `LEFT JOIN` to a single target, not scattered across 4 CTEs each joining to the target.

### Layer 3: `addresses_info` (unified, cross-chain)

Reads from the 19 chain info models. Each chain info model already has the **full cumulative values** (history + delta merged). So at this layer:

- `addresses_in_delta` = addresses with recent activity (any chain)
- For those addresses, reads from **all** chain info tables (to get the complete cross-chain picture)
- `SUM(executed_tx_count)` across chains — this is summing already-cumulative per-chain values, which is correct

---

## Side-by-side: where does `old_value + new_value` happen?

| | Old design | New design |
|---|-----------|------------|
| **Where** | Inside the single chain model, in the final SELECT | In `addresses_info_incremental` macro (Layer 2) |
| **Formula** | `nd.executed_tx_count + COALESCE(t.executed_tx_count, 0)` | `nd.executed_tx_count + COALESCE(t.executed_tx_count, 0)` |
| **Target scans** | 4 per incremental run (one per CTE + to_update) | 1 per incremental run (single LEFT JOIN) |
| **Source scans** | Interleaved with target lookups | Isolated in staging (Layer 1), no target access |

---

## Validation

The output is equivalent because:

1. **Staging** produces the same delta values the old design computed from raw sources (same SQL, same `COUNT(*)` / `SUM()` on the same incremental window).
2. **Chain info** adds `delta + existing` using the same `nd + COALESCE(t, 0)` pattern the old design used.
3. **Unified** aggregates already-cumulative chain values — identical to the old design reading chain models that already had cumulative values.

The only difference: redundant merges may occur (old design skipped "already up to date" rows; new design doesn't check). But the merged values are the same, so the final table matches.

jeff-dude:
@duneanalytics/team-data-ux no need for in-depth review here, just raising for awareness.

PR #9372, opened 2026-02-25
fix(dex): remove redundant wrapper in raw pool initializations

sashass1315:
Why was this change needed?
dex_raw_pool_initializations had an extra outer SELECT * FROM (...) around a single query block. It added no filtering or transformation and made the model unnecessarily noisy.

What was changed?
Removed the redundant outer wrapper and kept the same inner SELECT logic, columns, and incremental predicate as-is.
This is a cleanup-only change with no intended data output change.


sashass1315:

PR #9417, opened 2026-03-04
zeroex - change prices table reference

RantumBits:
change from prices.usd to prices.minute in macro


alicecha:
Unfortunately we won’t merge this PR because switching to `prices.minute` would massively increase table size and reduce price quality for long-tail tokens due to outlier DEX activity, which can also inflate `amount_usd` values. It also introduces a circular dependency where DEX trades feed price datasets that would then be used again to price those same trades, while current pricing already relies on the trusted asset in the pool.

PR #9437, opened 2026-03-09
megaeth: remove hardcoded date filter in daily project

jeff-dude:
Change start_date from '2026-01-30' to '2025-11-01' to include all megaeth history including pre-public-mainnet activity (e.g. USDM mint/burn transfers).

Resolves CUR2-1633


jeff-dude:
megaeth has `prod_exclude` in both stablecoin transfers and these models here. a merge should set us up to be prepared in future & not actually run anything in prod. safe to go from my persepective.

PR #9438, opened 2026-03-09
megaeth: remove hardcoded date filter in hourly project

jeff-dude:
Remove the megaeth-specific elif branches that filtered out data before 2026-01-30 in the op_stack_gas_fees macro. Full history will now be read on non-incremental runs.

Resolves CUR2-1633


jeff-dude:
deploy with override for megaeth only, to avoid all models using macro.

PR #9450, opened 2026-03-12
Add miner rewards and withdrawals to tokens_ethereum.base_transfers

0xRobin:
## Summary
- extend `tokens_ethereum.base_transfers` with native transfer rows for `ethereum.withdrawals` so staking withdrawals are represented in the upstream transfer surface
- add pre-merge block miner reward rows by joining `ethereum.blocks` to reward traces and using `max(value)` per block for miner payout amount
- keep explicit pre-merge filtering (`block_number <= 15537393`) and incremental predicates to bound scans and preserve incremental behavior

## QA / Impact (CI model: `test_schema.git_dunesql_eba481f_tokens_ethereum_base_transfers`)
- added rows: **137,236,332** total (`+2.30%` vs 5,961,874,116 total rows)
  - miner rewards: **15,537,393** rows (2015-07-30 -> 2022-09-15)
  - staking withdrawals: **121,698,939** rows (2023-04-12 -> 2026-03-12 build cutoff)
- added native volume: **91,555,774.8042 ETH** total
  - miner rewards: **47,223,886.5938 ETH**
  - staking withdrawals: **44,331,888.2104 ETH**
  - added volume share vs existing native volume in table: **0.59%**

### Validation checks
- withdrawals arm matches source at CI cutoff exactly:
  - row count and wei sum in model == `ethereum.withdrawals` filtered by `block_time <= max(block_time)` in CI model
- miner arm matches expected pre-merge construction exactly:
  - row count equals number of pre-merge reward-bearing blocks
  - wei sum equals `sum(max(value) by block_number)` from `ethereum.traces` (`type='reward'`, `value>0`, `block_number<=15537393`)
- uniqueness check on added rows passed:
  - `count(*) == count(distinct (block_date, unique_key))` for `tx_hash is null`

## Notes
- `max(value)` is used for miner payout because summing reward traces includes uncle rewards and overstates miner amount
- suicide traces are intentionally excluded here and will be handled in the trigger model as discussed in CUR2-1624

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/affluent/ton/affluent_ton_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/affluent/ton/affluent_ton_schema.yml
@@ -6,6 +6,7 @@ models:
       blockchain: ton
       sector: lending
       contributors: pshuvalov, okhlopkov
+      short_description: "Supply (deposit) events on the Affluent lending protocol on TON."
     config:
       tags: ['ton', 'affluent', 'lending', 'factorial']
     description: "Supply (deposit) events on the Affluent lending protocol on TON."
@@ -42,6 +43,7 @@ models:
       blockchain: ton
       sector: lending
       contributors: pshuvalov, okhlopkov
+      short_description: "Withdrawal events on the Affluent lending protocol on TON."
     config:
       tags: ['ton', 'affluent', 'lending', 'factorial']
     description: "Withdrawal events on the Affluent lending protocol on TON."
@@ -79,6 +81,7 @@ models:
       blockchain: ton
       sector: lending
       contributors: pshuvalov, okhlopkov
+      short_description: "Borrow events on the Affluent lending protocol on TON."
     config:
       tags: ['ton', 'affluent', 'lending', 'factorial']
     description: "Borrow events on the Affluent lending protocol on TON."
@@ -116,6 +119,7 @@ models:
       blockchain: ton
       sector: lending
       contributors: pshuvalov, okhlopkov
+      short_description: "Repayment events on the Affluent lending protocol on TON."
     config:
       tags: ['ton', 'affluent', 'lending', 'factorial']
     description: "Repayment events on the Affluent lending protocol on TON."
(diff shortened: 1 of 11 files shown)


dbt_subprojects/tokens/models/transfers_and_balances/ethereum/tokens_ethereum_base_transfers.sql
```diff
@@ -30,3 +30,78 @@ FROM
...
+    FROM {{ source('ethereum', 'traces') }}
+    WHERE type = 'reward'
+        AND value > uint256 '0'
+        AND block_number <= 15537393
+        {% if is_incremental() %}
+        AND {{ incremental_predicate('block_time') }}
+        {% endif %}
+    GROUP BY 1
+) r
+    ON r.block_number = b.number
```
0xRobin: add block_date to the join condition, as it's the partition column used in these tables

tomfutago:
code looks good to me but as i don't have enough context/knowledge around this - asked my cloud friend for help (might be complete slop tho):

**High**: Uncle/ommer rewards are dropped (under-counting native issuance transfers pre-Merge).
The new miner-reward block intentionally collapses all type='reward' traces per block to a single row and assigns it to b.miner via max(value). Your own inline comment confirms multiple reward rows exist (including uncle rewards), but those uncle payouts are not emitted as transfers to uncle beneficiaries. Cross-check with official protocol docs: pre-transition Ethereum has both block reward and ommer reward, and both are removed only starting at transition block (so both existed before). [EIP-3675 — Block and ommer rewards deprecated from transition](https://eips.ethereum.org/EIPS/eip-3675)

**Medium**: Reward attribution relies on a heuristic (max(value)) rather than reward type / beneficiary semantics.
Even for miner payout, max(value) is a dataset-shape assumption, not protocol-level logic. If reward trace encoding differs by client/indexer (e.g., splitting components), this can silently undercount miner rewards. Safer would be filtering/aggregating by explicit reward type + recipient semantics where available.


Suggested fix direction
- Emit all pre-Merge reward traces as separate native transfers (minted from zero address), not only one miner row per block.
- If you only want miner payout, explicitly encode that rule using reward metadata/recipient, not max(value).
- Add a targeted validation query comparing:
  - total pre-Merge reward-trace value vs modeled inserted reward value,
  - and distinct beneficiary coverage (miner + uncle addresses).

0xRobin:
@tomfutago both of these are known.
- we don't have the info on uncle rewards recipient, so can't add it.
- miner rewards will always be larger than uncle rewards, so max(value) is a safe heuristic.
-

PR #9338, opened 2026-02-19
fix: materialise upstream views into tables to avoid stage limits

gosuto-inzasheru:
as per @jeff-dude's instructions here: https://github.com/duneanalytics/spellbook/pull/9167

PR #9462, opened 2026-03-16
Revert "fix: materialise upstream views into tables to avoid stage limits"

tomfutago:
Reverts duneanalytics/spellbook#9338


tomfutago:
fyi @gosuto-inzasheru - #9338 needs reverting as dbt run keeps failing on this deployment.

PR #9586, opened 2026-04-20
feat(polymarket): add V2 CLOB exchange support

los-xyz:
## Summary

Polymarket CLOB V2 launches on Polygon around 2026-04-22 (early traffic already live since 2026-04-03). This PR extends the existing `polymarket_polygon_*` spells to ingest V2 trades alongside V1 without breaking existing consumers.

### 1. Additive, not a rewrite
V1 and V2 rows coexist in `market_trades_raw` / `market_trades`. A new `contract_version` column (`'v1'` / `'v2'`) distinguishes them. Downstream consumers that don't care about the version get the union for free; consumers that do can filter.

Rationale: V2 rolls out gradually. A hard cutover or two separate tables would force every downstream query to know about the migration. Unifying at the `market_trades` layer contains the V2 awareness in one place.

### 2. Unified V2 exchange, split by `contract_address`
V1 had two decoded source tables (`CTFExchange_evt_OrderFilled` + `NegRiskCtfExchange_evt_OrderFilled`). V2 decodes both the standard (`0xe111…996b`) and NegRisk (`0xe222…0f59`) contracts into a single `polymarket_v2_polygon.ctfexchange_evt_orderfilled`, distinguished by the `contract_address` column. We follow that — one UNION branch covers both.

Side effect: added `contract_address` to the `market_trades` unique-combination test, because the same `(block_time, asset_id, evt_index, tx_hash)` key could theoretically appear under two different V2 contract addresses in the same tx.

### 3. `tokenId` + `side` replaces `makerAssetId` / `takerAssetID`
V1 `OrderFilled` carried two asset IDs (one always 0 for the collateral side) and we used `coalesce(nullif(...))` to pick the non-zero one as the token. V2 simplified this: `tokenId` directly identifies the outcome token, and `side` (0 = maker BUY, 1 = maker SELL) tells us which side of the fill is USD-denominated:
- `side = 0` (maker BUY): `makerAmountFilled` is USD paid by maker, `takerAmountFilled` is tokens received.
- `side = 1` (maker SELL): `makerAmountFilled` is tokens sold by maker, `takerAmountFilled` is USD received.

Both branches produce the same `amount` (USD) / `shares` (tokens) columns as V1, so the downstream contract doesn't change.

### 4. pUSD treated 1:1 as USD
V2 collateral moved from USDC.e to pUSD (`0xc011a7e12a19f7b1f670d46f03b03f3342e82dfb`) — a standard ERC-20 backed 1:1 by USDC in reserve, no AMM/secondary market, no depeg mechanism. We route pUSD through the existing `amount` / `amount_usd` columns at face value rather than introducing a floating-rate conversion.

Tradeoff: if pUSD ever depegs, we'd under/over-state USD value. Acceptable for now; revisit with a dedicated column + price oracle only if depeg becomes relevant.

### 5. Fees: same column, same unit, new source
V1 populated the `fee` column from `CTFExchange_evt_OrderFilled.fee` (std) and `NegRiskCtfExchange_evt_OrderFilled.fee` (NegRisk), divided by 1e6. In practice this was almost always 0 (fees were enabled in the contract but not charged).

V2 introduces a dynamic operator-set fee (`fee = C × r × p × (1 − p)`) and emits the resulting per-trade fee directly on `OrderFilled`. We source the same `fee` column from the V2 event's `fee` field, same 1e6 scaling (pUSD and USDC.e share 6 decimals).

Rationale: consistency. The column already exists, the unit is the same, the semantics are "per-trade fee in USD". No new column needed; we just fix the stale `(currently not enabled)` description.

### 6. New V2 columns surfaced as passthroughs
V2 adds `builder` (bytes32, replaces HMAC header auth for order attribution) and `metadata` (bytes32, arbitrary per-order data). Both are passed through as-is without interpretation. NULL for V1 rows.

Rationale: we don't yet know how the product team will use `builder` — surfacing the raw bytes now means those analyses can start without a schema change later.

### 7. Capital actions: pUSD added, wrap/unwrap deferred
`users_capital_actions` filters transfers for three tokens: USDC, USDC.e, and now pUSD. Deposit / withdrawal / transfer categorization applies to pUSD the same way.

Not in this PR: a separate `'convert'` action type for USDC↔pUSD wrap via the V2 onramp/offramp contracts (analogous to the existing USDC↔USDC.e Uniswap pool convert). That would need the ramp contract addresses. Users wrapping USDC → pUSD and then depositing will still show up as a USDC deposit + a pUSD deposit via the current logic, which is usable for volume accounting even if the wrap step itself isn't labeled.

### 8. Downstream models untouched
`ohlcv_hourly`, `market_prices_*`, `positions_raw`, `market_details` all read from column contracts this PR preserves (only adds columns). V2 trades flow through automatically on the next incremental run.

## Changes

| file | change |
|---|---|
| `sources/polymarket/polygon/_sources.yml` | declared `polymarket_v2_polygon` source + 4 V2 event tables |
| `polymarket_polygon_market_trades_raw.sql` | added V2 UNION ALL branch + `contract_version`/`builder`/`metadata` columns across all branches |
| `polymarket_polygon_market_trades.sql` | propagated new columns through `source_trades` CTE + final SELECT |
| `polymarket_polygon/_schema.yml` | new column descriptions + tests (`not_null`, `accepted_values` on `contract_version`); fee description fixed; `contract_address` added to unique-key test |
| `polymarket_polygon_users_capital_actions.sql` | pUSD (`0xc011…2dfb`) added to symbol CASE + filters; V2 exchange addresses added to `polymarket_addresses` exclusion list |

## Test plan

- [x] \`dbt compile\` passes locally (verified on \`feat/polymarket-v2\`)
- [ ] CI builds + tests pass
- [ ] Spot-check V2 trade counts vs \`polymarket_v2_polygon.ctfexchange_evt_orderfilled\` — should reconcile minus any rows whose \`tokenId\` isn't yet in \`base_ctf_tokens\`
- [ ] Verify \`amount\` / \`shares\` / \`price\` sign conventions against a sample V2 trade cross-referenced to Polymarket's own trade API
- [ ] Spot-check pUSD transfers appear as deposits/withdrawals with \`symbol = 'pUSD'\`
- [ ] \`ohlcv_hourly\` still reconciles with source trade volumes after V2 rows land

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/_schema.yml
@@ -142,6 +142,7 @@ models:
             - asset_id
             - evt_index
             - tx_hash
+            - contract_address
     columns:
       - name: block_month
         description: "UTC month of the event block time used for partitioning"
@@ -164,7 +165,7 @@ models:
       - name: action
         description: "Type of trade (clob or AMM trade)"
       - name: contract_address
-        description: "Contract address (ctfexchange or negrisk module)"
+        description: "Exchange contract that emitted the trade. V1: CTFExchange or NegRiskCTFExchange. V2: 0xe111…996b (standard) or 0xe222…0f59 (NegRisk) — both decode into polymarket_v2_polygon.ctfexchange_evt_orderfilled."
       - name: condition_id
         description: "Unique identifier for the YES/NO pair"
       - name: event_market_name
@@ -184,11 +185,11 @@ models:
       - name: price
         description: "Price for the token being bought"
       - name: amount
-        description: "Amount in USD"
+        description: "Amount in USD (V2: pUSD treated 1:1 as USD)"
       - name: shares
         description: "Amount of shares transferred"
       - name: fee
-        description: "Fees to Polymarket (currently not enabled)"
+        description: "Per-trade fee in USD. V1: sourced from CTFExchange/NegRiskCtfExchange OrderFilled.fee (usually 0 — fees were not charged in practice on V1). V2: sourced from the V2 OrderFilled.fee field, calculated per-market as fee = C × r × p × (1 − p)."
       - name: maker
         description: "Trader whose order is being filled"
       - name: taker
@@ -197,6 +198,16 @@ models:
         description: "Unique key of the event or market"
       - name: token_outcome_name
         description: "Combination of token outcome and question"
+      - name: contract_version
+        description: "Exchange contract version that executed the trade: 'v1' (original CTFExchange/NegRiskCTFExchange) or 'v2' (launched Apr 2026, unified CTFExchange V2, pUSD collateral)."
+        tests:
+          - not_null
+          - accepted_values:
+              values: ['v1', 'v2']
+      - name: builder
+        description: "V2 only: builder attribution field (bytes32) embedded in signed orders, replacing V1's HMAC header authentication. NULL for V1 rows."
+      - name: metadata
+        description: "V2 only: arbitrary metadata field (bytes32) in signed orders. NULL for V1 rows."
       - name: _updated_at
         description: "Timestamp of when this row was last inserted or updated by the dbt pipeline"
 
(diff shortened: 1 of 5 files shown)


jeff-dude (approved):
looks good. i can remove the hardcoded date filters prior to merge.

PR #9596, opened 2026-04-23
solana stablecoins: split non-circulating helper tables

0xRobin:
## Summary
- Add `stablecoins_solana_extended_non_circulating_inventory_accounts` so extended lineage has its own helper table.
- Restrict the existing core helper table to core token-list mints and derive owners from `core_transfers` only.
- This is the normal-deploy base PR required before forward-only lineage wiring.

### DIFF
--- a/dbt_subprojects/solana/models/stablecoins/balances/_schema.yml
+++ b/dbt_subprojects/solana/models/stablecoins/balances/_schema.yml
@@ -205,6 +205,43 @@ models:
       - name: observed_owners
         description: "Observed owner addresses for the inventory token account (historical)"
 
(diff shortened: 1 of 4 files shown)


alicecha (commented):
Nitpick - can we rename the other inventory account model to `_core_non_circulating_inventory_account`

alicecha (approved):
The comment can be addressed as part of the other PR: https://github.com/duneanalytics/spellbook/pull/9597

0xRobin:
we can't do that in the other PR (as I want to deploy as forward only).
I can duplicate the model here and flip it in the follow up though. :+1:

PR #9602, opened 2026-04-27
stablecoins: align extended_balances start_date with core per chain

alicecha:
## Summary

Each `stablecoins_<chain>_extended_balances` model was passing an arbitrary forward-dated `start_date` (mostly `2026-01-01`) to the `stablecoins_balances_from_transfers` macro. That argument controls the lower bound of the `days` calendar used to forward-fill balances, so an `extended_balances` table whose `start_date` was after the chain's launch had a narrower history floor than the corresponding `stablecoins_<chain>_core_balances` on the same chain.

This PR sets `start_date` on every daily_spellbook EVM `extended_balances` model to exactly match the `start_date` of its `core_balances` counterpart, so the two tables share the same lower date bound and any downstream join/union over them is consistent.

- 52 files updated (one chain per file).
- `tron` is intentionally skipped.
- `megaeth` was already aligned, so no change there.
- No macro / config changes — the only diff per file is the date string in the macro call, which is in the model body and therefore picked up by `state:modified.body` in CI.

## Test plan

- [x] `dbt compile -s stablecoins_ethereum_extended_balances stablecoins_polygon_extended_balances stablecoins_arbitrum_extended_balances` — succeeds, compiled SQL contains the new core-aligned `start_date` literals (`2017-11-28`, `2020-09-15`, `2021-05-26`).
- [ ] CI green on all 52 modified models via `dbt slim ci`.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/stablecoins/evm/abstract/balances/stablecoins_abstract_extended_balances.sql
+++ b/dbt_subprojects/daily_spellbook/models/stablecoins/evm/abstract/balances/stablecoins_abstract_extended_balances.sql
@@ -18,5 +18,5 @@
 
 {{ stablecoins_balances_from_transfers(
     transfers = source('stablecoins_' ~ chain, 'extended_transfers'),
-    start_date = '2026-01-01'
+    start_date = '2024-12-20'
 ) }}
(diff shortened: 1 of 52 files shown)


jeff-dude (approved):
approving with assumption in linear comment: adjust start date based on new list in extended per chain

PR #9594, opened 2026-04-23
chore: migrate spellbook dependency management to uv

0xRobin:
## Summary
- Replace `Pipfile` / `Pipfile.lock` with a uv project (`pyproject.toml` + `uv.lock`).
- Keep existing dependency versions by locking uv to the same versions that were in `Pipfile.lock`.
- Update CI and contributor docs/skills to use `uv sync --locked` and `uv run ...` instead of pipenv.

Leaving a proper python and dependency upgrade to a different PR.

## Testing
- `uv lock --check`
- `uv sync --locked`
- `uv run pytest scripts/test_token_checker.py`
- `uv run dbt --version`

PR #9605, opened 2026-04-27
version upgrades to handle dependabot alerts

jeff-dude:
towards PLA-2021


Pipfile
```diff
@@ -4,10 +4,22 @@ verify_ssl = true
...
+sqlparse = ">=0.5.4"
+protobuf = ">=5.29.6"
+filelock = ">=3.20.3"
+virtualenv = ">=20.36.1"
+urllib3 = ">=2.6.3"
+jinja2 = ">=3.1.6"
 
 [requires]
-python_version = "3.9"
+python_version = "3.11"
```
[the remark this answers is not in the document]
jeff-dude: @cursor push cccca9e553

jeff-dude:
@0xRobin i see you have #9594 open, we can wait until that is finalized before resolving alerts instead of this PR.

pyproject.toml
```diff
@@ -1,12 +1,22 @@
...
+    "jinja2>=3.1.6",
     "numpy==2.0.2",
     "pre-commit==4.0.1",
-    "pytest==8.3.4",
+    "protobuf>=5.29.6",
+    "pytest>=9.0.3",
+    "requests>=2.33.0",
+    "sqlparse>=0.5.4",
+    "urllib3>=2.6.3",
+    "virtualenv>=20.36.1",
```
0xRobin: do we really need to specify all of these explicitly? 🤔

jeff-dude: yeah, looking at that now after getting back in sync with uv. hoping only need a few to handle the rest.

pyproject.toml
```diff
@@ -1,12 +1,14 @@
 [project]
 name = "spellbook"
 version = "0.0.0"
-requires-python = "==3.9.*"
+requires-python = "==3.11.*"
 dependencies = [
-    "dbt-trino==1.9.0",
+    "dbt-trino>=1.10.1,<1.11",
+    "jinja2>=3.1.6",
     "numpy==2.0.2",
```
0xRobin: We can drop numpy as I don't think it's used anymore.

pyproject.toml
```diff
@@ -1,12 +1,14 @@
 [project]
 name = "spellbook"
 version = "0.0.0"
-requires-python = "==3.9.*"
+requires-python = "==3.11.*"
```
0xRobin: With the old numpy version dropped, we should be able to upgrade to a more recent python version (3.14?)

0xRobin: we need to adapt the github action workflows to pin the same python version

0xRobin (changes requested):

pyproject.toml
jeff-dude: we can do 3.13, a dependency in dbt-trino doesn't allow for 3.14

> The 3.14 blocker is transitive through dbt:
> 
> dbt-trino -> dbt-core -> jsonschema -> rpds-py
> 
> rpds-py==0.22.3 tries to build with PyO3 0.23.3, which errors because that PyO3 version supports up to Python 3.13.

jeff-dude:
@0xRobin don't freak at the number of files 😅 

decision time:
- go ahead and fix all the yaml warnings we've been receiving and wanted to clean anyway, since new versions failed compile on them
- remove `--warn-error` flag in compile instead, remove all yaml changes, keep the warnings in place

jeff-dude:
some added stats of changes and confirmation

> By changed lines across dbt_subprojects/ + sources/ YAML:
> 
> 99.86% arguments migration related
> 12.03% are new arguments: lines
> 87.83% are the paired reindent of existing test args under arguments:
> 0.14% are other/manual cleanup deletions, mostly duplicate YAML key cleanup and stale Rubicon comments
> By file count:
> 
> 1478 / 1479 changed YAML files include the arguments: migration: 99.93%
> 3 / 1479 include duplicate-key/manual cleanup: 0.20%
> 3 / 1479 include stale comment cleanup: 0.20%
> I confirmed arguments: from dbt’s docs and from dbt itself. The dbt docs for [data_tests](https://docs.getdbt.com/reference/resource-properties/data-tests) say:
> 
> Any arguments or configurations passed to those data tests should be nested below the arguments property.
> 
> The examples also note:
> 
> arguments: available in v1.10.5 and higher. Older versions can set the argument name as the top-level property.
> 
> And locally/CI, dbt 1.10.20 emitted the deprecation warning as an error under --warn-error:
> 
> Found top-level arguments to test ... Arguments to generic tests should be nested under the arguments property.

pyproject.toml
0xRobin, line 1: Clean!

0xRobin (approved):
LFG, fixing the warning is definitely the right call!

PR #9608, opened 2026-04-28
CUR2-1844: extend multichain token transfers

tomfutago:

### Description:

[...]

---


dbt_subprojects/tokens/models/transfers_and_balances/tokens_multichain_transfers.sql
```diff
@@ -1,52 +1,179 @@
...
+select
+  to_hex(unique_key) as unique_key,
+  blockchain,
+  block_time,
+  block_date,
+  block_month,
+  cast(ledger_sequence as bigint) as block_number,
+  cast(transaction_hash as varchar) as tx_id,
+  cast(transaction_id as bigint) as tx_sequence,
+  cast(null as bigint) as tx_index,
```
0xRobin: `transaction_id` -> `tx_index` instead of `tx_sequence`?

dbt_subprojects/tokens/models/transfers_and_balances/tokens_multichain_transfers.sql
```diff
@@ -1,52 +1,179 @@
...
+
+select
+  unique_key,
+  blockchain,
+  block_time,
+  block_date,
+  block_month,
+  cast(null as bigint) as block_number,
+  cast(tx_hash as varchar) as tx_id,
+  cast(tx_version as bigint) as tx_sequence,
```
0xRobin: we could put the `tx_version` in the `block_number` column.

0xRobin (changes requested):
I don't think we need to introduce the extra `tx_sequence` column, that seems to be confusing.
We're using it only when block_number is not present it seems, so I suggest we overload the `block_number` column.

dbt_subprojects/tokens/models/transfers_and_balances/tokens_multichain_transfers.sql
tomfutago: yep, makes sense

dbt_subprojects/tokens/models/transfers_and_balances/tokens_multichain_transfers.sql
tomfutago: yes, let's go with that - even tho it's not exactly the same as `block_number`:

_From Aptos docs: transactions are addressable by `tx_version`, and indexer docs call `transaction_version` the “Blockchain version of the transaction.” It is a global ledger transaction version, so it gives total ordering. But it is per transaction, while EVM block_number is per block and many txs share it._

_So suggestion is semantically acceptable if we define `block_number` as the chain-native ledger/order height, not strictly “block height” for every chain. That is already somewhat true: Solana uses slot, Sui uses checkpoint, XRPL uses ledger index, Stellar uses ledger sequence._

0xRobin: ok I'm comfused on the difference between `tx_version` and `tx_index` now for Aptos

tomfutago: for Aptos:
- `tx_version` - identifies tx order globally across whole chain
- `tx_index` - position/index in the tx
- `event_index` - event id within that tx

0xRobin: > tx_index - position/index in the tx

what does this mean? The position of *what* within the transaction?

tomfutago: sry, poor wording: `tx_index` is an index of the transaction within the block (starts from 0 per block)

0xRobin: ok then it doesn't make much sense to put tx_version in the block_number column.. :thinking: 
Can we just use that as the tx_index and drop the per-block-index?

tomfutago: ahh, actually.. 🌀 there's also `block_height` available in `aptos.transactions`, so that should be `block_number` but it's not exposed at all in the whole lineage including in the models built by aptos team.. guess will need to add it in

0xRobin: ok yes, adding that would be preferred. We can leave it null for now in this PR and add a comment maybe.

0xRobin (approved):

PR #9609, opened 2026-04-28
CUR2-2188: aptos gas fees

tomfutago:

### Description:

[...]

---


jeff-dude (changes requested):
my preference for now is to leave out of `gas.fees` and keep aptos standalone.
we can then consider leaving fields as-is and not adapt datatypes to fit evm schema.

tomfutago:
> my preference for now is to leave out of `gas.fees` and keep aptos standalone. we can then consider leaving fields as-is and not adapt datatypes to fit evm schema.

makes sense - removed from `gas.fees`. 

on data types - they are aligning with `tokens_aptos.transfers` so would keep them as they are.

jeff-dude (approved):

PR #9610, opened 2026-04-28
ci: scope dbt CI tables by PR

jeff-dude:
towards CUR2-2108


dbt_macros/dune/adapters.sql
```diff
@@ -68,7 +68,7 @@
 
 {%- macro s3_bucket() -%}
   {%- if target.type == 'trino' and target.schema != 'wizard' -%}
-    {%- if target.name == 'prod' or target.schema.startswith('git_dunesql') -%}
+    {%- if target.name == 'prod' -%}
```
0xRobin, line 71: does CI run against `prod` target? 
We need to check the profile that is installed in the github runners.

0xRobin (approved):
looks good, 1 open question on the conditional change.

dbt_macros/dune/adapters.sql
jeff-dude: no, there is a `ci` target on the runner. it uses the `dune` catalog rather than `hive`. ci cluster is migrated to catalog service too, which handles locations. once spellbook is fully migrated, there will be cleanup PR to remove some overrides like this.

target schema is also:
`schema: "{{ env_var('DBT_CI_SCHEMA', 'dune_spellbook_ci__tmp') }}"`
so that piece is effectively a no-op at this point

PR #9632, opened 2026-05-04
dex: add hedera trades with SaucerSwap v1 and v2

alicecha:
## Summary

Adds `hedera` to the DEX trades pipeline with SaucerSwap V1 (UniswapV2 fork) and V2 (UniswapV3 fork), plus the standard chain-level `dex_hedera_base_trades` / `_trades` / `_token_volumes_daily` trio, and registers `hedera` in the chain unions of `dex.trades` and `dex.token_volumes_daily`.

## Methodology

Hedera's HTS-based pool deployment goes through HTS precompiles, which don't surface in `hedera.creation_traces`. Dune's factory-flag auto-decoding relies on creation traces to discover children, so the standard `Pair_evt_Swap` / per-pool decoded tables only contain events from individually submitted contracts (just two template addresses for SaucerSwap today) — not from the 2,387 V1 pairs and 49 V2 pools the factories actually created.

To get full coverage, the platform models follow the same pattern Tron uses (`sunswap_v[23]_tron_swap_events.sql`): we decode the `Swap` event from raw `hedera.logs` ourselves via the `decode_evm_event` UDTF, restricted to addresses present in the factory's PairCreated/PoolCreated tables. The resulting `*_swap_events` models are shape-compatible with `Pair_evt_Swap`, so the existing `uniswap_compatible_v2_trades` / `uniswap_compatible_v3_trades` macros consume them unchanged.

## Test plan
- [ ] CI: `dbt slim ci` passes for the dex sub-project
- [ ] Spot check `dex.trades` after merge: `SELECT project, version, count(*), count(DISTINCT project_contract_address), sum(amount_usd) FROM dex.trades WHERE blockchain = 'hedera' AND block_date >= current_date - interval '7' day GROUP BY 1,2`
- [ ] Populate `saucerswap_v[12]_hedera_base_trades_seed.csv` with a few sample swaps and confirm `check_dex_base_trades_seed` passes

Toward CUR2-494


jeff-dude:
leaving as draft until resumed.

PR #9644, opened 2026-05-07
feat(kalshi): integrate series_raw for canonical category + fees

los-xyz:
## Summary

Adds \`kalshi.series_raw\` as a source and joins it into \`kalshi_market_details\` and \`kalshi_market_trades\`. Three concrete wins:

### 1. Better \`category\`

Series-level category (canonical, one row per \`series_ticker\`, ~10k series total) replaces event-level as the primary source. Coalesces to event-level when series has no row.

Validated across ~54M event×series rows from one market_details partition:

| Outcome | Rows | % |
|---|---:|---:|
| match (series == event) | 33.9M | 63% |
| series_overrides_event | 19.9M | **37%** |
| no_series_row (fallback to event) | 1 | ~0% |

Override examples where series-level value is qualitatively better:
- APPLEWATCH-24: \`Science and Technology\` (event) → \`Companies\` (series)
- BRAZILXBAN: \`Companies\` → \`Politics\`
- CLOSESTSTATE-24: \`Elections\` → \`Politics\`

This will shift the \`category\` distribution. Downstream consumers (e.g. our \`prediction_markets\` unified surface) will see different bucket counts. No code change needed downstream; the seed mapping operates on the (now more accurate) raw value.

### 2. \`fee_usd\` per trade in \`kalshi_market_trades\`

\`series_raw\` exposes \`fee_type\` (\`quadratic\` or \`quadratic_with_maker_fees\`) and \`fee_multiplier\`. Trades model now computes:

\`\`\`sql
case
    when md.fee_type in ('quadratic', 'quadratic_with_maker_fees') then
        ceil(0.07 * coalesce(md.fee_multiplier, 1.0)
             * t.yes_price_dollars * (1 - t.yes_price_dollars)
             * t.count_fp * 100) / 100.0
end as fee_usd
\`\`\`

This is the **taker fee** per Kalshi's published formula. Maker-side fees (1.75% coefficient when \`fee_type='quadratic_with_maker_fees'\`) are not surfaced here; consumers can compute them from the raw \`fee_type\` + \`fee_multiplier\` columns also exposed.

Sanity check: P=0.5, N=10, multiplier=1.0 → \$0.18 (matches expected).

Closes the v1 NULL-fee gap on \`prediction_markets.trades\` downstream.

### 3. \`frequency\` as a first-class column

[...]

PR #9647, opened 2026-05-07
fix(kalshi): drive settlement updates from markets_lifecycle_raw

los-xyz:
## Summary

Stacked on #9644 (`feat/kalshi-series-integration`). Targets that branch, not `main`, so the diff stays minimal once the parent merges.

`kalshi_market_details` reads `kalshi.markets_raw` and uses `m.updated_time >= now() - 3 days` as its incremental predicate. Problem: `markets_raw.updated_time` is Kalshi's own `last_modified` field, not ingest time. Settlement updates often arrive in `markets_raw` with `updated_time` already weeks old, so they fall outside the 3-day window and never land in the spell. We measured ~962K markets where `markets_raw.result` is populated but the spell still shows them as unsettled, which produced a journalist-visible cliff in `settled_share` from ~95% to ~0% after 2026-04-15.

This PR routes settlement signal through `kalshi.markets_lifecycle_raw` (new ingest table, forward-only WebSocket stream from 2026-05-07) where `received_ts` is real ingest time:

- New `lifecycle` CTE aggregates per-market: `result`, `settlement_value`, `determination_ts` from `event_type='determined'` rows; `settled_ts` from `event_type='settled'` rows.
- New OR-branch in the incremental selector: any ticker with a fresh lifecycle event (`received_ts` in window) is pulled into the merge.
- Final SELECT now coalesces `result`, `settlement_ts`, and `settlement_value_dollars` from lifecycle first, falling back to `markets_raw`. `settlement_value` is varchar in lifecycle ("0.0000"/"1.0000"), cast to double via `try_cast` to align with the existing `markets_raw.settlement_value_dollars` type.
- `source_updated_at` extended to include `lifecycle.last_event_received_ts` so downstream `kalshi_market_trades` (which already pulls `result`/`status` via `ref('kalshi_market_details')`) reprocesses tickers whose lifecycle row updated.

`kalshi_market_trades.sql` is unchanged: it inherits the fix automatically through the ref.

## Scope

- Lifecycle is forward-only since 2026-05-07; there is no historical backfill in `markets_lifecycle_raw`. The 962K-row historical backlog is fixed by the auto full-refresh that happens when #9644 merges (since the parent PR introduces a new column, dbt will rebuild from scratch and pick up settlement state from `markets_raw` directly).
- This PR fixes the **structural future bug** only: from now on, late-arriving settlements ride the lifecycle stream and land within one incremental tick.

## Validation

- `uv run dbt compile --select kalshi_market_details` succeeds.
- Compiled SQL run through Dune (full-refresh shape, `LIMIT 100`): parses, returns rows, `source_updated_at` non-null on every row, lifecycle coalesce produces the expected `result` (yes/no/scalar) and `settlement_value_dollars` (0.0/1.0 doubles).

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/_projects/kalshi/_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/_projects/kalshi/_schema.yml
@@ -53,13 +53,13 @@ models:
       - name: expected_expiration_time
         description: "Expected expiration time"
       - name: settlement_ts
-        description: "When the market was settled (null if unsettled)"
+        description: "When the market was settled (null if unsettled). Sourced from markets_lifecycle_raw event_type='settled' when available, else from markets_raw."
       - name: status
         description: "Market lifecycle state (active, finalized, closed, etc.)"
       - name: result
-        description: "Settlement outcome (yes, no, or empty if unresolved)"
+        description: "Settlement outcome (yes, no, scalar). Sourced from markets_lifecycle_raw event_type='determined' when available, else from markets_raw."
       - name: settlement_value_dollars
-        description: "Payout value per contract at settlement"
+        description: "Payout value per contract at settlement. Sourced from markets_lifecycle_raw event_type='determined' (cast from varchar) when available, else from markets_raw."
       - name: expiration_value
         description: "Expiration outcome value"
       - name: can_close_early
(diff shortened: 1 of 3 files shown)


tomfutago:
cursor finding:

**Medium:** `kalshi_ohlcv_hourly` will not pick up lifecycle-only settlement updates for historical markets. `kalshi_market_details.source_updated_at` now moves from `markets_lifecycle_raw.received_ts`, and `kalshi_market_trades` consumes that, but `ohlcv_hourly` still only incrementally selects recent `created_time` / recent `hour`. A market that settles today with old trades can update `market_details` and `market_trades`, while existing OHLCV rows keep stale `market_outcome` / settlement prices until full refresh.

tomfutago (approved):

PR #9653, opened 2026-05-08
fix(kalshi): propagate lifecycle settlement to ohlcv_hourly historical rows

los-xyz:
## Summary

Stacked on #9647 (\`feat/kalshi-lifecycle-patch\`). Targets that branch, not \`main\`, so the diff stays minimal once the parent merges.

Cursor surfaced this on #9647: \`kalshi_market_details.source_updated_at\` now moves when lifecycle events land, and \`kalshi_market_trades\` propagates the change to historical trade rows. \`kalshi_ohlcv_hourly\` did not, because its incremental selectors (\`t.created_time\` in \`base\`, \`r.hour\` in the final emit) keyed only on time-of-trade. A market that settles today with old trades would update \`market_details\`/\`market_trades\` while \`ohlcv_hourly\` kept stale \`market_outcome\` / \`event_market_name\` / \`category\` until full refresh.

## Changes

Three coordinated edits to \`kalshi_ohlcv_hourly.sql\`. Coordinated because changing only the \`base\` filter would produce duplicate \`(block_month, hour, market_id, outcome)\` keys and break the merge unique constraint, and changing only the bottom filter would not actually surface the refreshed dim cols.

1. \`base\` CTE: widen the incremental WHERE with the same OR-branch \`kalshi_market_trades\` already uses, pulling historical trades for tickers whose \`market_details.source_updated_at\` is in window.
2. \`prior_sparse\` CTE: exclude refreshed tickers entirely. Otherwise their hours come through twice (new_sparse re-aggregates, prior_sparse re-emits from \`{{ this }}\`), and the merge fails uniqueness.
3. Final emit: widen the bottom \`incremental_predicate('r.hour')\` with an OR on refreshed \`market_id\`s so the merge actually writes to those old hours.

## Validation

- \`uv run dbt compile --select kalshi_ohlcv_hourly\` succeeds.
- End-to-end source-uniqueness check on the production target with a 3-day window: simulated post-fix \`sparse_ohlcv\` produces **0 duplicate (hour, market_id, outcome) keys** across new_sparse + prior_sparse. Refreshed cohort = 446k tickers in 3 days.

## Notes

- No schema change; existing \`unique_combination_of_columns(block_month, hour, market_id, outcome)\` test continues to enforce the post-fix invariant.
- Performance: the OR-branch in \`base\` widens the \`market_trades\` scan to include refreshed tickers' full history. Bounded by lifecycle event volume in the window. The merge uses \`merge_skip_unchanged = true\`, so unchanged columns on re-emitted rows do not trigger writes.
- Forward-filled post-expiration rows for resolved markets remain excluded from output by the existing \`where not (...)\` filter. Spine bounds (\`last_hour = least(max trade hour, now())\`) mean those rows are rarely generated in the first place.

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/_projects/kalshi/kalshi_ohlcv_hourly.sql
+++ b/dbt_subprojects/daily_spellbook/models/_projects/kalshi/kalshi_ohlcv_hourly.sql
@@ -29,7 +29,16 @@ with base as (
 		) as rn_last
 	from {{ ref('kalshi_market_trades') }} as t
 	{% if is_incremental() -%}
-	where {{ incremental_predicate('t.created_time') }}
+	where
+		{{ incremental_predicate('t.created_time') }}
+		-- Late-arriving settlements: pull historical trades for tickers whose market_details
+		-- dim columns (result, category, event_title, ...) refreshed via lifecycle, so old
+		-- hours can be re-emitted with the current market_outcome.
+		or t.ticker in (
+			select md.ticker
+			from {{ ref('kalshi_market_details') }} as md
+			where {{ incremental_predicate('md.source_updated_at') }}
+		)
 	{%- endif %}
 ),
 
@@ -68,7 +77,9 @@ new_sparse as (
 ),
 
 {% if is_incremental() -%}
--- pre-window sparse anchor from {{ this }} so market_bounds and asof forward-fill stay correct across the window boundary
+-- pre-window sparse anchor from {{ this }} so market_bounds and asof forward-fill stay correct across the window boundary.
+-- Excludes tickers being re-aggregated from base (refreshed market_details), since their hours are
+-- already coming through new_sparse and double-emitting them here would break the merge unique key.
 prior_sparse as (
 	select
 		t.hour,
@@ -86,6 +97,11 @@ prior_sparse as (
 	from {{ this }} as t
 	where t.is_forward_filled = false
 		and not {{ incremental_predicate('t.hour') }}
+		and t.market_id not in (
+			select md.ticker
+			from {{ ref('kalshi_market_details') }} as md
+			where {{ incremental_predicate('md.source_updated_at') }}
+		)
 ),
 {% endif %}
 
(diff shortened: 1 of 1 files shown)


los-xyz:
Cursor was right, fix pushed in [bcb6606→\`HEAD\`].

The previous config had `incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.hour')]`, which prunes the MERGE destination to in-window hours. Combined with this PR's source widening (historical hours for refreshed tickers), the existing target rows for those old hours were invisible to the ON clause, so MERGE took the INSERT branch and produced duplicates.

Dropped `incremental_predicates` to mirror the same correctness/speed trade-off `kalshi_market_details` already made (same explicit reasoning is in its comment block).

Verified against the deployed `kalshi.ohlcv_hourly`:

- baseline existing dupes: 0
- refreshed cohort (3-day window): 446,342 tickers, 30,115,922 existing target rows
- fix would emit 1,278,547 source keys for that cohort:
  - 1,168,738 match existing target rows → MERGE UPDATE
  - 109,809 have no existing target match → MERGE INSERT (net-new hours, can't duplicate by definition)

tomfutago:
findining:

**Medium:** `kalshi_ohlcv_hourly.sql` still cannot remove historical rows that are omitted after settlement. The model uses incremental `merge`, so it only updates/inserts rows present in the source. But resolved post-expiration forward-filled rows are filtered out here:

```257:262:dbt_subprojects/daily_spellbook/models/_projects/kalshi/kalshi_ohlcv_hourly.sql
where not (
	r.is_forward_filled
	and r.market_end_time is not null
	and r.hour > r.market_end_time
	and r.market_outcome in ('yes', 'no')
)
```

tomfutago:
> findining:
> 
> **Medium:** `kalshi_ohlcv_hourly.sql` still cannot remove historical rows that are omitted after settlement. The model uses incremental `merge`, so it only updates/inserts rows present in the source. But resolved post-expiration forward-filled rows are filtered out here:
> 
> ```
> where not (
> 	r.is_forward_filled
> 	and r.market_end_time is not null
> 	and r.hour > r.market_end_time
> 	and r.market_outcome in ('yes', 'no')
> )
> ```

ignore, non issue..

tomfutago (approved):

PR #9678, opened 2026-05-17
perf(polymarket): narrow positions changed_markets trigger; parse market_details timestamps

jeff-dude:
Part of [CUR2-2699 — daily model quick wins](https://linear.app/dune/issue/CUR2-2699/daily-model-quick-wins).

## Summary

Two focused changes on top of [#9696](https://github.com/duneanalytics/spellbook/pull/9696) (which already restructured `positions_raw` upstream via `positions_balances_repro`):

1. **`polymarket_polygon_positions.sql`** — narrowed `changed_markets` trigger, single-pass join, `block_month` partition (+ added to `unique_key`), historical rewrite scoped via two-sided `BETWEEN market_start_time AND current_date`, full `{{ this }}` scan preserved for drift recovery.
2. **`_schema.yml`** — `block_month` added to `polymarket_polygon_positions` columns and to its uniqueness test.

Diff vs main: 3 files, +~ /−~ lines (small).

> Note: an earlier revision of this PR also parsed `market_details.market_start_time` / `market_end_time` from ISO‑8601 strings into `timestamp(3) with time zone`. That change has been **reverted** here — downstream models (`ohlcv_hourly`, `market_prices_hourly`, `market_prices_daily`) call `from_iso8601_timestamp(...)` / `substring(...)` on `market_end_time` and would break under a type change. Timestamp normalization will be a follow‑up PR scoped to those callers.

## Scope of perf impact

The target was prod's daily incremental on `polymarket_polygon_positions`, reported at ~115 min on 2026-06-09 (`query_id 20260609_075452_19190_xee2n`, ~14 CPU-hrs / 5.2 TB spill).

**What this PR optimizes**:
- The `changed_markets` trigger now fires only on `outcome` / `resolved_on_timestamp` changes (Group A→B freeze for `active` / `closed` / `accepting_orders`). API flag flutter no longer triggers a daily historical rewrite — that's the big win on downstream cost when `changed_markets` is non-trivial.
- Single-pass join (`positions_to_emit ⋈ market_details` runs once, not twice).
- Historical rewrite is partition-scoped to each token's `[market_start_time, current_date]` range.
- `block_month` partition + `unique_key` per repo convention.

**What this PR does NOT optimize** (left for [CUR2-2713](https://linear.app/dune/issue/CUR2-2713/perfpolymarket-polygon-positions-reduce-changed-markets-full-history), André's parallel work):
- The `changed_markets` CTE itself still hash-joins `market_details` (~2.6 M rows) against a full scan of `{{ this }}` (~10 B rows / ~1.9 TB), which is where the **5.2 TB spill** comes from. CUR2-2713's lever is to pre-aggregate `{{ this }}` to one row per `(token_id, market_outcome, resolved_on_timestamp)` before the join, collapsing the probe side to a few million rows. That work is being developed in parallel and should land after this PR.

| Path | Before | After |
|---|---|---|
| `changed_markets` trigger | fires on `active` / `closed` / `accepting_orders` / `outcome` / `resolved_on_timestamp` | fires on `outcome` / `resolved_on_timestamp` only |
| `positions_to_emit` join | original `positions` CTE referenced twice → join recomputed | single inner join after `recent_positions ∪ historical_changed_positions` |
| `positions` partition | `['day']` (1 978 partitions → hits Trino's 100‑open‑writer cap on full refresh) | `['block_month']` (~66 partitions; `day` stays queryable via Delta column stats) |
| `unique_key` | `['day', 'address', 'token_id']` | `['block_month', 'day', 'address', 'token_id']` — merge partition‑prunes |
| `changed_markets` build-side scan/spill | 10 B-row probe, 5.2 TB spill | **unchanged here — see CUR2-2713** |

## Behaviour change

`active` / `closed` / `accepting_orders` are now **snapshot at write time**. Old prod behaviour rewrote these on historical rows to today's API value whenever any of them flapped — inconsistent with the other 9 metadata columns from `market_details` (`question`, `market_description`, `event_market_name`, `polymarket_link`, `market_start_time`, `market_end_time`, etc.) which were already snapshot‑at‑write.

After this PR, all 12 descriptive columns behave the same way. Settlement columns (`market_outcome`, `resolved_on_timestamp`) still propagate across history when a market resolves.

Querying `positions where day = '2024-01-15'` now returns a consistent as‑of‑day view instead of a row with Jan‑15 question text but today's `active` flag.

Product team review of these semantic changes: signed off.

## Validation against prod (CI run 27171747419 on commit b9ad0edfa)

Headline counts on the CI‑built `positions` vs current prod:

| Metric | New | Prod | Δ |
|---|---:|---:|---|
| rows | 9 980 787 654 | 10 049 278 344 | new is −68.5 M (CI ran one day before prod's last incremental) |
| addrs | 2 753 551 | 2 761 378 | within day‑drift expectation |
| tokens | 2 476 654 | 2 491 726 | within day‑drift expectation |
| min_day | 2021-04-09 | 2021-04-09 | identical |
| max_day | 2026-06-07 | 2026-06-08 | CI is one day behind prod |
| `null_outcome` | 0 | 0 | identical (invariant holds) |
| `null_resolved_on_timestamp` | 308 M | 332 M | **new has 24 M fewer nulls** (drift recovery improvement) |

Row‑level diff on 1/1000 address‑hash sample:

| Field | Diff count | Source of diffs |
|---|---:|---|
| `balance` | 58 397 | exactly matches `missing_in_new` (one‑day delta), so **zero real value drift** |
| `outcome` | 73 849 | ~15 K above missing → drift‑recovery rows (new has settled values where prod is still stale) |
| `resolved_on_timestamp` | 69 779 | ~11 K above missing → same pattern |
| `active` / `closed` / `accepting_orders` | 58 K – 74 K | intentional Group A→B snapshot freeze on historical rows |
| `missing_in_prod` | **0** | every row in new exists in prod ✅ |
| `missing_in_new` | 58 397 | rows for prod's most‑recent day that CI didn't include |

Every divergence is in the "new is correct, prod was stale" direction. Zero rows are missing from prod's perspective.

## Review feedback addressed

1. **Timestamp type break** (commit `0289c1e5f`) — reverted. `market_start_time` / `market_end_time` stay varchar to preserve downstream callers.
2. **Stale post‑resolution rows** (commits `0289c1e5f` + `756508b54`) — historical rewrite originally had `p.day between cm.market_start_time and cm.resolved_on_timestamp`, which excluded forward‑filled rows for unredeemed holders past the resolution date. Replaced with `between cm.market_start_time and current_date` — preserves the lower-bound partition prune AND a two-sided range filter for Trino, while covering all post-resolution forward-fill rows.
3. **`block_month` added to `unique_key`** (commit `0289c1e5f`) — matches repo convention; merge now partition‑prunes.

## Test plan

- [x] dbt compile + tests pass locally
- [x] CI build succeeds (full history) on b9ad0edfa
- [x] CI tests pass (uniqueness, not‑null on full tables)
- [x] Headline‑count diff: new is identical to prod modulo Group A→B freeze + one‑day CI lag
- [x] Sample row‑level diff: zero `balance` drift, zero `missing_in_prod`
- [x] EXPLAIN ANALYZE confirms no double‑scan of `{{ this }}` in `changed_markets`
- [ ] Re‑run CI on `756508b54` to confirm the explicit upper bound restores ~4 min incremental wall time

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/_schema.yml
@@ -535,6 +535,8 @@ models:
               - address
               - token_id
     columns:
+      - name: block_month
+        description: "UTC month derived from day (date_trunc('month', day)); used for partitioning"
       - name: day
         description: "The date of the position"
         data_tests:
--- a/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/_schema.yml
@@ -426,9 +426,9 @@ models:
       - name: polymarket_link
         description: "Link to the Polymarket event page"
       - name: market_start_time
-        description: "Start time of the market"
+        description: "Timestamp when the market started accepting orders (parsed from API ISO-8601 string)"
       - name: market_end_time
-        description: "End time of the market"
+        description: "Timestamp when the market closed to new orders (parsed from API ISO-8601 string)"
       - name: market_outcome
         description: "Outcome of the market"
       - name: resolved_on_timestamp
@@ -576,9 +576,9 @@ models:
       - name: polymarket_link
         description: "Link to the Polymarket event page"
       - name: market_start_time
-        description: "Start time of the market"
+        description: "Timestamp when the market started accepting orders (parsed from API ISO-8601 string)"
       - name: market_end_time
-        description: "End time of the market"
+        description: "Timestamp when the market closed to new orders (parsed from API ISO-8601 string)"
       - name: market_outcome
         description: "Outcome of the market"
       - name: resolved_on_timestamp
--- a/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/_schema.yml
@@ -426,9 +426,9 @@ models:
       - name: polymarket_link
         description: "Link to the Polymarket event page"
       - name: market_start_time
-        description: "Timestamp when the market started accepting orders (parsed from API ISO-8601 string)"
+        description: "Start time of the market"
       - name: market_end_time
-        description: "Timestamp when the market closed to new orders (parsed from API ISO-8601 string)"
+        description: "End time of the market"
       - name: market_outcome
         description: "Outcome of the market"
       - name: resolved_on_timestamp
@@ -531,6 +531,7 @@ models:
       - dbt_utils.unique_combination_of_columns:
           arguments:
             combination_of_columns:
+              - block_month
               - day
               - address
               - token_id
@@ -576,9 +577,9 @@ models:
       - name: polymarket_link
         description: "Link to the Polymarket event page"
       - name: market_start_time
-        description: "Timestamp when the market started accepting orders (parsed from API ISO-8601 string)"
+        description: "Start time of the market"
       - name: market_end_time
-        description: "Timestamp when the market closed to new orders (parsed from API ISO-8601 string)"
+        description: "End time of the market"
       - name: market_outcome
         description: "Outcome of the market"
       - name: resolved_on_timestamp
--- a/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_market_details.sql
+++ b/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_market_details.sql
@@ -93,8 +93,8 @@ SELECT
     WHEN lower(neg_risk) = 'false' THEN get_href('https://polymarket.com/event/' || market_slug, market_slug)
     WHEN lower(neg_risk) = 'true'  THEN  get_href('https://polymarket.com/event/' || replace(replace(replace(lower(neg_risk_market_name), ' ', '-'), '$', ''), '''',''), neg_risk_market_name)
   END AS polymarket_link,
-  from_iso8601_timestamp(accepting_order_timestamp) as market_start_time,
-  from_iso8601_timestamp(end_date_iso) as market_end_time,
+  accepting_order_timestamp as market_start_time,
+  end_date_iso as market_end_time,
   game_start_time,
   seconds_delay,
   fpmm,
--- a/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_market_details.sql
+++ b/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_market_details.sql
@@ -93,8 +93,8 @@ SELECT
     WHEN lower(neg_risk) = 'false' THEN get_href('https://polymarket.com/event/' || market_slug, market_slug)
     WHEN lower(neg_risk) = 'true'  THEN  get_href('https://polymarket.com/event/' || replace(replace(replace(lower(neg_risk_market_name), ' ', '-'), '$', ''), '''',''), neg_risk_market_name)
   END AS polymarket_link,
-  accepting_order_timestamp as market_start_time,
-  end_date_iso as market_end_time,
+  from_iso8601_timestamp(accepting_order_timestamp) as market_start_time,
+  from_iso8601_timestamp(end_date_iso) as market_end_time,
   game_start_time,
   seconds_delay,
   fpmm,
--- a/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_positions.sql
+++ b/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_positions.sql
@@ -6,7 +6,7 @@
     file_format = 'delta',
     incremental_strategy = 'merge',
     partition_by = ['block_month'],
-    unique_key = ['day', 'address', 'token_id'],
+    unique_key = ['block_month', 'day', 'address', 'token_id'],
     merge_skip_unchanged = true,
     post_hook = '{{ private_data_explorer(blockchains = \'["polygon"]\',
                                   spell_type = "project",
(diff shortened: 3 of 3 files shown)


jeff-dude:
## Validation summary

Validated the new design end-to-end against current prod via a full no-sample CI build (run `26006728920`) plus a sample-based row-level diff. No regressions; the new model is strictly more complete and accurate than prod.

### Performance (full-history CI build)

| Model | Rows | Duration |
|---|---:|---:|
| `polymarket_polygon_positions_raw_changes` | 363,870,653 | **40 s** |
| `polymarket_polygon_positions_raw` | **8,735,531,777** | **7 m** |
| `polymarket_polygon_positions` | **8,527,426,950** | **54 m** |
| All 6 data tests at full scale | — | **23 m** |

For comparison: the old `positions_raw` incremental performed a ~2.53 TB-CPU scan of `balances_daily_agg_base` on every run just to compute the "last balance before window" CTE. That single piece of work alone dwarfed the new design's entire full-refresh.

### Correctness — diff vs prod

Row-level diff on a 1/1000 address-hash sample:
- `positions_raw`: **byte-identical** (7,603,592 rows on both sides, zero `only_in_new` / `only_in_prod`)
- `positions`: 53,956 row diffs, all of which are `prod_outcome IS NULL OR 'unresolved'` -> `new_outcome IN ('yes','no','50/50')` — i.e., markets that resolved between prod's last daily run and the CI snapshot. **Zero settled-flips.**

Full-history aggregate checks:

| Metric | New | Prod | Interpretation |
|---|---:|---:|---|
| rows | 8.527 B | 8.457 B | New has +70.7 M (extra calendar day + late-arrival catch-up + earlier history) |
| `null_outcome` | 0 | 0 | Schema invariant holds |
| `null_resolved_on_timestamp` | 263.7 M | 327.0 M | New has 63 M fewer nulls — settled markets propagated correctly |
| `min_day` | 2021-04-09 | 2021-08-24 | Prod missed 4.5 months of early markets (see below) |
| `max_day` | 2026-05-17 | 2026-05-16 | One-day shift because CI ran a day later |

Recent-month balance drift on shared keys (Jan 2024 -> Oct 2024): **4.4% -> 0.11%**, monotonically shrinking. Row counts within 0.07%. No per-row value disagreement on shared keys — drift comes entirely from new having a few thousand more rows than prod each month.

### Why prod has stale early history

Traced to a structural quirk in the old `positions` model: `historical_changed_positions` only triggers a rewrite when `active / closed / accepting_orders / outcome / resolved_on_timestamp` changes for an **existing** `token_id` in `{{ this }}`. It never fires when a brand-new `token_id` shows up in `market_details`.

Effect: Polymarket has added ~4.5 months of earlier market metadata to their API since prod's first build (Oct 2024). `market_details` (full-refresh) picked them up on its next daily run, but `positions` never backfilled them because the trigger only watches existing tokens. The new full-refresh joins to today's `market_details` and recovers them.

Same mechanic also caused the slow propagation of newly-resolved outcomes and the historical `null_resolved` count we saw.

### Behaviour changes (intentional)

- `active / closed / accepting_orders` now **snapshot at write time**, consistent with the other 9 metadata columns from `market_details`. Historical queries no longer get a misleading "as-of-today" value for those three columns. `outcome` and `resolved_on_timestamp` still propagate across history.
- `positions` partitioned by `block_month` (66 partitions today, vs 1,978 day-partitions). Avoids Trino's 100-open-writers cap on full refresh. Same logical output; day filters still prune via Delta column stats. Matches the existing `polymarket_polygon_market_trades` convention.

### Why we can feel confident in the new data

1. **Sample row-level diff is exact** — 7.6 M rows match byte-for-byte on `positions_raw`.
2. **Zero settled-flips on `positions`** — every value diff is the new model converging to the correct settlement state; never disagreeing with a settled answer.
3. **Cardinality moves only in the "more complete" direction** — no missing rows in `new` that exist in `prod`.
4. **Schema invariants hold at full scale** — uniqueness, not-null, no nulls in `market_outcome`.
5. **Value drift on shared keys is essentially zero** — recent-month balance totals match to within 0.11%, all attributable to the row-count delta, not per-row drift.

Ready for review.

karimhass (approved):

tomfutago:
few findings to check:

1. **Downstream timestamp type break.** Current PR changes `market_details.market_start_time` / `market_end_time` from ISO strings to `timestamp with time zone`, but downstream models still call string-only functions on `market_end_time`:
   - `polymarket_polygon_ohlcv_hourly`: `from_iso8601_timestamp(market_end_time)`
   - `polymarket_polygon_market_prices_hourly`: `substring(md.market_end_time from 1 for 19)`
   - `polymarket_polygon_market_prices_daily`: same `substring(...)`
   
   I verified in DuneSQL that both `from_iso8601_timestamp(timestamp with time zone)` and `substring(timestamp with time zone, ...)` fail.

2. **Historical rewrite window may leave stale post-resolution rows.** The validation summary says outcomes/resolution propagate correctly, but it appears to validate a full build / sample diff, not future incremental behavior. Current incremental logic only rewrites historical rows from `market_start_time` through `resolved_on_timestamp`. Since `positions_balances_repro` can forward-fill non-zero balances after resolution, older post-resolution rows can remain stale after a market resolves unless they fall in the normal incremental window.

3. **`block_month` partition is omitted from the incremental `unique_key`.** Current `positions` is partitioned by `block_month`, but `unique_key` and the uniqueness test remain `day,address,token_id`. The validation summary explicitly calls month partitioning intentional, but does not address the merge-key mismatch. This is less correctness-critical than the first two, but still against the repo’s incremental partitioning convention.

karimhass (approved):

a-monteiro:
Heads-up on the perf framing (from a read-only prod analysis under CUR2-2713): the 5.2 TB spill in the prod incremental is **not** the `changed_markets` full scan — it is the MERGE rebuilding the full 9.98B-row target. In the cited run `20260609_075452_19190_xee2n`, the 9.98B-row / 5.09 TB-spill hash build is the scan that synthesizes `$row_id` (i.e. `DBT_INTERNAL_DEST`, all 21 cols), while `changed_markets` is a separate scan that was dynamic-filtered to **0 rows** (~0.02% of the query).

The good news: the levers in this PR are exactly the right ones for that real cost — the two-sided `day BETWEEN market_start_time AND current_date` bound on the historical branch + `block_month` in the merge `unique_key`/ON. Measured on current prod data, the new source footprint collapses to ~1 of 63 `block_month` partitions ≈ **575M / 10.05B target rows (5.7%)**, even with 126k markets resolved in the trailing week. So this PR likely removes the spill on its own; the separate `changed_markets` pre-aggregation (CUR2-2713) is **not needed** and would only add a ~2.7 CPU-hr full-table GROUP BY.

One thing worth confirming post-merge: the OLD merge already had its partition column (`day`) in the ON clause yet still read 100% of the target, so partition-in-ON alone is not sufficient to prune — please `analyze_query` the first real incremental run and confirm the target scan drops from 9.98B rows to the pruned month set (spill→0). If it does not, an explicit `incremental_predicate` target-side `block_month` filter would make the pruning deterministic rather than optimizer-dependent.

jeff-dude:
Linked to [CUR2-2615](https://linear.app/dune/issue/CUR2-2615/fix-model-performance-for-polymarket-polygonpositions) — fix model performance for `polymarket_polygon.positions`.

PR #9680, opened 2026-05-18
Fix Uniswap V4 swap direction logic and Angstrom evt_index mismatch

bh2smith:
## Summary

Two fixes for Uniswap V4 dex.trades accuracy:

### 1. Defensive fix to V4 swap direction logic (`uniswap_compatible_trades.sql`)

Use Swap **event** amounts (`e.amount0`) instead of call output (`c.amount0`) for buy/sell direction, and add `OR e.amount1 > INT256 '0'` condition.

In V4, afterSwap hooks can modify the BalanceDelta after the Swap event is emitted but before the call returns. Using event amounts is more reliable because they reflect the core swap before hook adjustments. The extra `OR` condition handles edge cases where `amount0 == 0` (hooks absorbing the full counterparty amount).

### 2. Fix Angstrom bundle orders evt_index (`angstrom_all_trades.sql`)

The `angstrom_bundle_orders` macro generated synthetic `evt_index` values via `ROW_NUMBER()` (1, 2, 3...) instead of the actual Swap event log index. This caused Angstrom swaps in `dex.trades` to be unmatchable by `evt_index` with other systems that reference the real event index.

**Root cause investigation:** A comparison query joining the dex trade stream with `dex.trades` on `(block_number, tx_hash, evt_index)` showed all Angstrom V4 swaps as "missing" — the synthetic evt_index never matched the real one.

Verified with transactions:
- `0xbc49c090...`: real evt_index=2, dex.trades had evt_index=1
- `0x5ab46bde...`: real evt_index=3,4, dex.trades had evt_index=1,2

**Fix:** Join bundle_orders with `PoolManager_evt_Swap` on `(tx_hash, pool_id)` to resolve the real `evt_index`, falling back to the synthetic value for orders without a corresponding Swap event (e.g. off-chain matched user orders).

### Affected models

- `uniswap_compatible_v4_trades` macro — used by all 15 chain-specific V4 base_trades models
- `angstrom_all_trades` macro — used by `angstrom_ethereum_base_trades`

## Test plan

- [ ] Verify Angstrom swaps in `dex.trades` now have correct `evt_index` matching Swap event logs
- [ ] Re-run comparison query (dune.com/queries/7527524) — Angstrom rows should no longer appear as "missing"
- [ ] Spot-check non-Angstrom V4 trades for regression
- [ ] CI dbt slim build passes

Closes SIM-5858

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/angstrom/angstrom_all_trades.sql
+++ b/dbt_subprojects/dex/macros/models/_project/angstrom/angstrom_all_trades.sql
@@ -26,6 +26,8 @@ WITH
     swap_events AS (
         SELECT
             evt_tx_hash,
+            evt_block_number,
+            evt_block_date,
             evt_index AS real_evt_index,
             id AS pool_id,
             row_number() over(partition by evt_tx_hash, id order by evt_index) AS pool_rn
@@ -37,6 +39,7 @@ WITH
     ),
     bundle_ranked AS (
         SELECT *,
+            CAST(date_trunc('day', block_time) AS date) AS block_date,
             row_number() over(partition by tx_hash, maker order by evt_index) AS pool_rn
         FROM bundle_orders
     ),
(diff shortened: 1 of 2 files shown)


jeff-dude (approved):

jeff-dude:
i'll deploy when we have the env ready ✅

PR #9682, opened 2026-05-18
CUR2-2025: exclude non-circulating owner wallets from Solana USDC supply

alicecha:
## Summary

- Adds a parallel owner-keyed non-circulating inventory list (`stablecoins_solana_{core,extended}_non_circulating_inventory_owners`) alongside the existing token-account-keyed list. Any token account whose owner appears in the list is treated as non-circulating in `stablecoins_solana_{core,extended}_balances`.
- Seeds the core list with three USDC owner wallets that [DefiLlama](https://github.com/DefiLlama/peggedassets-server/blob/master/src/adapters/peggedAssets/usd-coin/config.ts) and usdc.cool treat as non-circulating but Dune currently counts:
  - `7VHUFJHWu2CuExkJcJrzhQPJ2oygupTWkL2A2For4BmE` — Circle treasury
  - `41zCUJsKk6cMB94DDtm99qWmyMZfp4GkAhhuz4xTwePu`
  - `42qwJUTbKf3D8ULfWadUSjnHf6pkJ4H1VjCcfSKHvDTN`
- Brings Solana USDC circulating supply in line with other official sources.

## Why a separate owner list (vs. extending the existing token-account list)

The existing `_inventory_accounts` list keys on SPL token accounts, but the addresses above are owner wallets that hold non-circulating inventory across multiple token accounts (current and future). Resolving them to a fixed set of token accounts would be brittle. The new `_inventory_owners` list joins on `from_owner` / `to_owner` in the balances macro, so it covers all of an owner's token accounts automatically.

Linear: https://linear.app/dune/issue/CUR2-2025/review-and-fix-usdc-supply-on-solana

## Test plan

- [ ] Slim CI passes (uniqueness, not_null, accepted_values on new owners models)
- [ ] After merge: re-run the issue's balance comparison query and confirm Solana USDC circulating supply on 2026-04-15 lands near DefiLlama (~\$8.17b)


dbt_subprojects/solana/macros/stablecoins/stablecoins_svm_balances.sql
```diff
@@ -7,15 +7,25 @@
...
 
 with
 non_circulating_inventory_accounts as (
   select
     blockchain,
     token_mint_address,
     token_account
-  from {{ non_circulating_inventory_relation }}
+  from {{ non_circulating_inventory_accounts_relation }}
+  where excluded
```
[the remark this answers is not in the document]
alicecha: Bot was right that the 6 `official_circle_premint` accounts were not being excluded, but wrong about causality — `where excluded` was already on `origin/main` (line 19 of the pre-PR macro); my diff renamed the relation variable above it and the diff renderer displaced the line.

The real, pre-existing bug (premint rows had `excluded = false` because the seed was `cast(source_class = legacy_inventory as boolean)`) is addressed in commit 408a3aa: the truthy set is expanded to `(legacy_inventory, official_circle_premint)`, so all 12 inventory rows now correctly filter out. Verified ~$190M of premint balance (5 of 6 accounts not owned by the Circle Mint wallet) now drops from circulating supply.

jeff-dude (approved):

jeff-dude:
looks like this new addition and left join lookup on balances really expanded incremental run time? unless this was always slow?

alicecha:
bugbot run

alicecha:
> looks like this new addition and left join lookup on balances really expanded incremental run time? unless this was always slow?

@jeff-dude great spot! I've cleaned up so it's only one join. Dramatically improved!

alicecha:
@cursor review

dbt_subprojects/solana/models/stablecoins/balances/stablecoins_solana_non_circulating_inventory_accounts.sql
```diff
@@ -0,0 +1,148 @@
+{% set chain = 'solana' %}
+{% set owners_observation_start_date = '2020-10-01' %}
+
+{{
+  config(
+    schema = 'stablecoins_' ~ chain,
+    alias = 'non_circulating_inventory_accounts',
+    materialized = 'table',
+    file_format = 'delta',
+    tags = ['static'],
```
jeff-dude: based on logic below, reading from live transfers table and filtering incrementally, is this really static?
if yes, i would argue we restructure the query to filter to specific date range it's static, rather than make it appear as a normal incremental table.

jeff-dude (changes requested):

dbt_subprojects/solana/models/stablecoins/balances/stablecoins_solana_non_circulating_inventory_accounts.sql
alicecha: Fixed the logic, it didn't need to call on transfers, but equally also better for it to be incremental.

However, I've been reticent to merge as this causes the entire balances models to rebuild on Solana....when we could just filter downstream. Doing it at this level is the cleanest option, to stay consistent with how previous exclusions were dealt with, just also the most costly. Wdyt @jeff-dude ?

jeff-dude: proactive approval, while we finalize discussion in slack

jeff-dude (approved):

dbt_subprojects/solana/models/stablecoins/balances/stablecoins_solana_non_circulating_inventory_accounts.sql
alicecha: CI runs full history, run times are acceptable. Stablecoins side should take ~30min or so.

PR #9696, opened 2026-05-26
Replace polymarket positions input with ERC1155 reconstruction

0xRobin:
Replace `polymarket_polygon.positions_raw`'s upstream dependency on `tokens_polygon.balances_daily_agg_base` (which is built from a snapshot source that carries stale balances when burn events are missed) with a new model, `polymarket_polygon.positions_balances_repro`, that reconstructs daily Polymarket CTF balances directly from on-chain `erc1155_polygon.evt_TransferSingle` and `evt_TransferBatch` events.

Linear: [CUR2-2534](https://linear.app/dune/issue/CUR2-2534/find-replacement-for-polymarket-positions-input)

## Changes

- New incremental model `polymarket_polygon.positions_balances_repro`:
  - Computes per-day net delta per `(address, token_id)` from Polymarket CTF transfer events.
  - Filters days where every transfer has `value = 0` (no-op transfers in batches), but keeps intraday round-trips with non-zero amounts so closure markers are emitted.
  - Cumulative `int256` running sum, clamped to `[0, uint256_max]`, cast to `uint256`. No double-precision intermediate.
  - Forward-fills each balance row to the day before the next change, anchoring at the closure day for `balance = 0`.
  - Incremental seeds from `{{ this }}` over the prior window, aggregated by `last_updated` (not `day`) so `last_updated` is preserved across forward-filled rows.
  - In incremental runs the forward-fill `fill_start` is clipped to the window start so cold pairs don't expand their full history before the predicate filters it.
  - For closed positions (`balance_raw = 0`) the `fill_end` is capped at the closure day so we don't expand `[closure_day, current_date - 1]` only to be dropped by the final WHERE.
- `polymarket_polygon.positions_raw` now reads from the new model instead of `balances_incremental_subset_daily(...)`. Output schema unchanged.
- `_schema.yml`: adds the new model with `dbt_utils.unique_combination_of_columns` (scoped to last 14 days for CI runtime) and `not_null` tests on key columns.

## CI vs prod comparison (last day, run 26572354448)

| | rows |
|---|---|
| `balance_matching` (same key, same `balance_raw`) | 69,369,909 (99.93 %) |
| `balance_different` (same key, different `balance_raw`) | 10,050 |
| `ci_only` (in CI, missing in prod) | 18,160 |
| `prod_only` (in prod, missing in CI) | 40,605 |

Over the last 7 days the aggregate balance sums agree to within 0.001 % (`16.45607777e18` vs `16.45625946e18` raw).

`last_updated` parity on the same 68.2 M common rows: 99.993 % identical, 4,779 differ. All observed differences are negative drift (CI's `last_updated` is later than prod's), consistent with prod's snapshot source missing intermediate change events.

## Why the residual diffs are expected and acceptable

I sampled rows from each diff bucket and traced the underlying ERC1155 events:

- **`balance_different`**: legacy `balances_polygon_0002` snapshots miss some burn events, leaving stale positive balances that diverge from on-chain reality.
  - Example: `0xc4c41cd12f991e76ceeff92b531cd74c02497bc2` / token `1119...477`. Prod = `10_527_110_704`. Sum of all on-chain deltas for this pair = `1_316_636_682_369` (last event 2022-03-26). The new repro matches the on-chain sum exactly. Prod has been wrong for ~4 years.
- **`prod_only`**: 100 % have `balance_raw > 0`, all are forward-fills (`last_updated < day`), 39,829 of 40,605 have `last_updated` older than 90 days. These are the same class of stuck legacy snapshot rows. The repro correctly omits them because the on-chain net delta is zero.
- **`ci_only`**: 100 % have `balance_raw > 0`, all are forward-fills of real on-chain positions that the legacy snapshot source never indexed. The repro correctly includes them.

Net effect: the repro fixes ~69k stale/missing rows per day relative to the legacy table while matching the remaining 99.9 % exactly, and additionally restores accurate `last_updated` values for pairs whose intermediate changes were missed by the snapshot source.

## Performance

CI initial build: ~20 min, ~9.4 B rows materialized; incremental merge: ~3 min, ~200 M rows. Same order as the legacy model.

## Validation

- CI green, including the in-CI `dbt_utils.unique_combination_of_columns` and `not_null` schema tests on the new model.
- Manual row-level diff against prod for the last day (above) and aggregate diff for the last 7 days.
- `last_updated` parity check (above) confirms the seed-by-`last_updated` fix prevents drift on inactive pairs (~99 % of pre-window pairs are forward-fills with drift up to 5.3 years under the naive seed-by-`day` approach).
- Traced sample rows from each diff bucket back to the underlying ERC1155 transfer events to confirm the repro matches on-chain truth.

## What I removed from the original Cursor-agent draft

- The singular equivalence test (`tests/polymarket_polygon_positions_balances_repro_equivalence.sql`). It compared the repro against `balances_incremental_subset_daily(...)` — which is the legacy table we are intentionally moving off and is itself the source of the divergences above. Any threshold tight enough to be useful would have made CI red on legitimate fixes; loose enough not to was useless. The schema-level uniqueness / not-null tests stay as the in-CI guardrails.
- The `double` precision cast in the running balance computation. `sum(int256) OVER (...)` works natively and avoids the rounding risk for large positions.
- The two-stage `daily_flows` → `daily_deltas` aggregation and the `max(block_number)` / `max(block_time)` columns that were computed but never emitted.

---


0xRobin:
@bugbot review

dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_positions_balances_repro.sql
```diff
@@ -0,0 +1,189 @@
...
+  ) flows
+  where day < current_date  -- exclude today to avoid mid-day partials
+  group by 1, 2, 3
+  having max(abs(delta)) > int256 '0'
+),
+
+{% if is_incremental() -%}
+-- Seed: latest balance per (address, token_id) from before the incremental
+-- window. Carried forward by the forward-fill below.
+prior_balances as (
```
dennisp42, line 98: `max(day)` reads from `{{ this }}`, which is forward-filled, so it's the prior fill terminus, not the real last-change day. `forward_fill` then sets `last_updated = anchor_day`, so every incremental run rewrites `last_updated` for inactive pairs to ~yesterday. The legacy macro sourced `prior_balances` from `balances_daily_agg_base` (one row per real change), so its `max(day)` was the real change day.

Simulated against prod `positions_raw`: ~90% of pairs would have `last_updated` rewritten on the first incremental run. Some examples drift 1,940 days (positions held since Feb 2021, still 25-50k USDC). The body's row-count diff misses this because it compares balances, not `last_updated`.

https://dune.com/queries/7598118

https://dune.com/queries/7598137

Fix:

```sql
prior_balances as (
  select
    address,
    token_id,
    max(last_updated) as day,
    max_by(balance_raw, last_updated) as balance_raw
  from {{ this }}
  where not {{ incremental_predicate('day') }}
  group by 1, 2
),
```

Rerun the 30-day diff with `last_updated` to confirm.

dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_positions_balances_repro.sql
```diff
@@ -0,0 +1,189 @@
...
+      address,
+      token_id,
+      balance_raw,
+      coalesce(
+        date_add('day', -1, cast(lead(day) over (partition by address, token_id order by day) as date)),
+        current_date - interval '1' day
+      ) as fill_end
+    from balance_changes
+  ) c
+  cross join unnest(sequence(anchor_day, fill_end, interval '1' day)) as t(expanded_day)
```
dennisp42: After the `prior_balances` fix, `anchor_day` for cold pairs can be years old, and the sequence expands the full range before `incremental_predicate` clips it. Bound the start:

```sql
cross join unnest(
  sequence(
    {% if is_incremental() -%}
    greatest(anchor_day, date_trunc(
      '{{ var("DBT_ENV_INCREMENTAL_TIME_UNIT") }}',
      now() - interval '{{ var("DBT_ENV_INCREMENTAL_TIME") }}' {{ var("DBT_ENV_INCREMENTAL_TIME_UNIT") }}
    )),
    {%- else -%}
    anchor_day,
    {%- endif %}
    fill_end,
    interval '1' day
  )
) as t(expanded_day)
```

dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_positions_balances_repro.sql
```diff
@@ -0,0 +1,189 @@
...
+    anchor_day as last_updated
+  from (
+    select
+      day as anchor_day,
+      address,
+      token_id,
+      balance_raw,
+      coalesce(
+        date_add('day', -1, cast(lead(day) over (partition by address, token_id order by day) as date)),
+        current_date - interval '1' day
```
dennisp42: Closed positions expand to `current_date - 1 day` only to be dropped by the final WHERE. Cap the fallback at the closure day:

```sql
coalesce(
  date_add('day', -1, cast(lead(day) over (partition by address, token_id order by day) as date)),
  case when balance_raw = uint256 '0' then anchor_day
       else current_date - interval '1' day end
) as fill_end
```

dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_positions_balances_repro.sql
0xRobin: Confirmed; fixed in fdbb5a6. Aggregating by `max(last_updated)` and `max_by(balance_raw, last_updated)` so the prior anchor is the real change day, not the forward-fill terminus.

Reran your simulation against prod `positions_raw` for 5 random pre-window inactive pairs: drift was 54 to 457 days, balance unchanged. Confirms the bug existed and the fix preserves `last_updated`.

dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_positions_balances_repro.sql
0xRobin: Done in fdbb5a6. Clipped `fill_start` to the incremental window start so cold pairs no longer expand their full history. Used a Jinja set to share the same `date_trunc` expression with `incremental_predicate`; the trailing `where fill_end >= fill_start` keeps closed-before-window pairs from feeding an invalid range into `sequence(...)`.

dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_positions_balances_repro.sql
0xRobin: Done in fdbb5a6. Capped `fill_end` at `anchor_day` when `balance_raw = uint256 '0'`. Closed pairs now emit a single closure-day row instead of a multi-year sequence filtered down to one.

dennisp42 (approved):

PR #9726, opened 2026-06-06
tokens: single-pass FA scan in aptos_fungible_asset_metadata

a-monteiro:
## What

`aptos_fungible_asset_metadata` built its v2 (FA) rows by scanning `aptos.move_resources` **three times** — once for `fungible_asset::Metadata`, once for `Supply`/`ConcurrentSupply`, once for `object::ObjectCore` — and self-joining them on `(tx_version, move_address)`. Those three resources are co-emitted in the same transaction under the same state key, so this collapses them into a **single scan + conditional aggregation** grouped by `(tx_version, move_address)`. `HAVING` a `Metadata` row reproduces the old Metadata-driven `LEFT JOIN` semantics. The coin (v1) path is unchanged.

This is a single-pass `GROUP BY`, not a shared CTE, on purpose: Trino inlines CTEs, so wrapping the scan in a CTE referenced by three branches would re-expand into three scans. Funnelling through one aggregation is what actually collapses the read.

## Why

The `move_resource_name` filter isn't a partition/clustered column, so each of the three scans read the full window (~the same 600M-row partition set) and re-parsed `move_data` JSON independently.

Measured on prod (1-day window, 3 warm runs, projections forced via `checksum()` so the JSON parse isn't pruned):

| | current (4 scans) | this PR (2 scans) |
|---|--:|--:|
| scanned | 169.7 GB / 577M rows | 79.4 GB / 260M rows |
| source-build CPU (median) | 960k ms | 646k ms |

≈ **−53% bytes, −33% CPU** on the FA source build. The hourly MERGE for this model is ~22 CPU-hrs/day and its scans are ~66% of MERGE CPU (the write is ~7%), so the expected recurring saving is ~20–25% (~5 CPU-hrs/day).

## Correctness

Output proven **identical** via `EXCEPT` both ways over two windows with different data characteristics:
- incremental window (2026-06-04): 13,048,375 rows, 0 diff either direction
- historical full-refresh window (2023-08-01..08): 112,738 rows, 0 diff either direction

`MAX()` collapse is safe because the three resources are 1:1 per `(tx_version, move_address)` (no fan-out, confirmed by the equivalence test). The full-refresh `block_date >= DATE('2023-07-27')` floor is preserved on the FA branch; the coin branch keeps its existing (floor-free) predicate.

## Note for reviewers

The single-pass replaces streaming joins with a hash aggregation. Per-task peak memory measured ~4.4 GiB (vs the current MERGE's ~34 GiB/task peak driven by the concurrent FA scans), so this should be neutral-to-better on memory — but the actual MERGE peak can only be confirmed on a real prod run, so worth a glance at the first few runs after merge.

Towards CUR2-2691

### DIFF
--- a/dbt_subprojects/tokens/models/aptos/aptos_fungible_asset_metadata.sql
+++ b/dbt_subprojects/tokens/models/aptos/aptos_fungible_asset_metadata.sql
@@ -70,6 +70,10 @@ WITH mr_fa AS (
     {% else %}
         AND block_date >= DATE('2023-07-27') -- beginning of FA (v2)
     {% endif %}
+    {% if target.name == 'ci' %}
+    -- bound the full-refresh scan in CI so the build completes and the regression test runs; prod is unaffected
+    AND block_date >= current_date - interval '14' day
+    {% endif %}
     GROUP BY tx_version, move_address
     -- only emit assets that have a Metadata resource (matches the old Metadata-driven LEFT JOINs)
     HAVING MAX(CASE WHEN move_resource_module = 'fungible_asset' AND move_resource_name = 'Metadata' THEN 1 ELSE 0 END) = 1
@@ -97,6 +101,10 @@ WITH mr_fa AS (
         AND move_resource_name = 'CoinInfo'
     {% if is_incremental() %}
         AND {{ incremental_predicate('block_time') }}
+    {% endif %}
+    {% if target.name == 'ci' %}
+    -- bound the full-refresh scan in CI so the build completes and the regression test runs; prod is unaffected
+    AND block_date >= current_date - interval '14' day
     {% endif %}
 )
 
--- a/dbt_subprojects/tokens/models/aptos/aptos_fungible_asset_metadata.sql
+++ b/dbt_subprojects/tokens/models/aptos/aptos_fungible_asset_metadata.sql
@@ -22,71 +22,57 @@
 -- creator_address is the asset address for v1 and owner for v2 (can change for v2)
 -- creator_address is not needed for coins/fa, it's a holdover from tokens (where it is used as key with name)
 
-WITH mr_fa_metadata AS (
-    SELECT
-        tx_version,
-        tx_hash,
-        block_date,
-        block_time,
-        --
-        write_set_change_index,
-        move_address,
-        json_extract_scalar(move_data, '$.name') AS asset_name,
-        json_extract_scalar(move_data, '$.symbol') AS asset_symbol,
-        CAST(json_extract_scalar(move_data, '$.decimals') AS INTEGER) AS decimals,
-        json_extract_scalar(move_data, '$.icon_uri') AS icon_uri,
-        json_extract_scalar(move_data, '$.project_uri') AS project_uri
-    FROM {{ source('aptos', 'move_resources') }}
-    WHERE 1=1
-        AND move_module_address = 0x0000000000000000000000000000000000000000000000000000000000000001
-        AND move_resource_module = 'fungible_asset'
-        AND move_resource_name = 'Metadata'
-    {% if is_incremental() %}
-        AND {{ incremental_predicate('block_time') }}
-    {% else %}
-        AND block_date >= DATE('2023-07-27') -- beginning of FA (v2)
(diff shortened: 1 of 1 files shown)


dbt_subprojects/tokens/models/aptos/aptos_fungible_asset_metadata.sql
```diff
@@ -22,71 +22,57 @@
...
-        AND {{ incremental_predicate('block_time') }}
-    {% else %}
-        AND block_date >= DATE('2023-07-27') -- beginning of FA (v2)
-    {% endif %}
-), mr_fa_owner AS (
-    SELECT
-        tx_version,    
         move_address,
-        '0x' || LPAD(LTRIM(json_extract_scalar(move_data, '$.owner'), '0x'), 64, '0') AS owner_address
+        MAX(CASE WHEN move_resource_module = 'fungible_asset' AND move_resource_name = 'Metadata' THEN tx_hash END) AS tx_hash,
```
tomfutago, line 33: i don't see any obvious semantic regression, but there's some risk in the aggregation assumption: if multiple relevant resource rows ever share (`tx_version`, `move_address`), `max(case ...)` would collapse them where the old joins could emit multiple rows. existing unique-key test helps catch that, but ci run hasn't completed so hard to validate.

tomfutago (commented):
in additiona to potential risk mentioned below - would be good to add temp test timeframe on this model to allow ci to complete and run regression test.

a-monteiro:
* **#9726** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9726?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9726?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dbt_subprojects/tokens/models/aptos/aptos_fungible_asset_metadata.sql
a-monteiro: CI now completes and the regression test passes, so this is ready to validate.

Instead of a temp filter to revert, I added a `target.name == 'ci'` bound (`AND block_date >= current_date - interval '14' day`) on the two `move_resources` scans. It renders **only** under the CI target -- I verified the dev/prod compile is byte-identical (0 occurrences) while the ci compile has it -- so there's nothing to revert before merge. The initial full-refresh build now finishes (it was hitting the 90-min timeout) and `unique_combination_of_columns [block_month, tx_hash, asset_type]` passes on both the initial and incremental runs.

On the aggregation/fan-out risk: confirmed safe. Each FA resource (Metadata / Supply / ConcurrentSupply / ObjectCore) appears at most once per (tx_version, move_address), so there are 0 duplicate group keys per resource type and the `max(case ...)` collapses nothing the old LEFT JOINs would have emitted multiply. I also diffed old vs new directly -- `(old EXCEPT new)` and `(new EXCEPT old)` are both 0 rows, and per-column checksums match -- over both a recent incremental window and a 2023 historical window.

tomfutago (approved):

a-monteiro:
**Post-merge realized impact** (measured from `query_history`, `spellbook-tokens`, per-day; before = 24h pre-merge, after = post-merge steady state; ~24 heavy MERGE builds/day):

| axis | before | after | change |
|---|---|---|---|
| IO scanned | 9.7 TB/day | 4.8 TB/day | **-50%** |
| CPU | 22.5 cpu-hrs/day | 19.3 | -14% |
| wall | 6.3 hrs/day | 3.9 | **-37%** |
| peak mem | 44 GB | 58 GB | +31% |
| spill | 0 | 0 | none |

Per-build median: IO 405->201 GB, cpu 3380->2898 s, wall 943->592 s. IO matched the projection (-50% vs -53%); the **wall-time drop is the headline** (job finishes faster -> less cluster runtime). Peak rose modestly (single-pass GROUP BY tradeoff) but stayed <60 GB with **zero spill**. Correctness: `unique_combination_of_columns` green in CI on real built data.

PR #9735, opened 2026-06-09
perf(polymarket): convert positions to view over positions_raw + market_details

jeff-dude:
towards CUR2-2615

Stops denormalizing volatile market metadata into a daily incremental `polymarket_polygon.positions` table. Daily incremental MERGE was the #1 wall-time daily model (~115 min, ~14 CPU-hrs, 5.2 TB spill — see [CUR2-2713](https://linear.app/dune/issue/CUR2-2713/perfpolymarket-polygon-positions-merge-rebuilds-full-998b-row-target)). Product team approved converting to a view.

## Change

`polymarket_polygon.positions` becomes a **view** = `positions_raw ⋈ market_details` on `token_id`. Same 21 consumer-facing columns as today; two **additive** sync watermark columns for future downstream use.

- `positions_raw` continues to be incrementally maintained (#9696's `positions_balances_repro` chain).
- `market_details` continues as today's full-refresh table.
- The old materialized `positions` table goes away; daily incremental MERGE goes with it.

**Files:** `polymarket_polygon_positions.sql`, `_schema.yml` (+81 / −175 vs main)

### Sync watermarks (new)

Downstream datashare incremental sync uses `_updated_at`. The view exposes:

| Column | Source | Purpose |
|---|---|---|
| `_position_updated_at` | `positions_raw.last_updated` | Last balance/on-chain change |
| `_market_updated_at` | `greatest(last_uploaded_at, resolved_on_timestamp)` | Denormalized metadata change; null if neither set |
| `_updated_at` | `greatest(_position_updated_at, coalesce(_market_updated_at, _position_updated_at))` | **Drop-in for existing sync** — picks up metadata-only updates without arrakis changes |

Arrakis datashare spec already has `_updated_at` as `timestamp(3) with time zone` — no type change needed. `_position_updated_at` / `_market_updated_at` deferred in datashare until downstream is ready.

## Perf trade-off

Replaces a daily 115-min write with per-query join cost amortized across reads.

### Write-side savings (eliminated)
- ~14 CPU-hrs/day rebuilt
- 5.2 TB MERGE spill
- 89 GB peak memory
- ~115 min wall

### Read-side cost (consumer benchmarks, medium tier)

Full 12-query consumer benchmark vs prod table documented in [Notion](https://app.notion.com/p/37dd8bb3254881c197e8dbc2534201f2). Summary:

- **Typical OI dashboards** (incl. highest-traffic q6798817, 86 runs/60d): **~1.3–1.5× credits** — not uniformly slower
- **Tail risk:** multi-reference queries (q7545291, 2.0×) and full-history PnL scan (1.49× wall-clock)
- **Trivial probes** (`select * limit N`): large % ratios, negligible absolute credits
- **BWR q6315848:** both baseline and view timeout on medium either way

Micro-benchmarks (EXPLAIN ANALYZE): single-day count TABLE 0.7s vs VIEW 9.1s — join overhead is real at storage layer, but day-filtered consumer queries don't pay the full penalty.

View SQL hints (BROADCAST, slim projection) did not improve plans — Trino picks PARTITIONED/REPLICATED on its own; projection pushdown does not skip the join.

## Behaviour changes

Two semantic shifts — both move closer to "current truth":

1. **`active` / `closed` / `accepting_orders`** flow live from `market_details`. No more snapshot-at-write semantics.
2. **`market_outcome` / `resolved_on_timestamp`** also flow live. ~332 M prod rows with null `resolved_on_timestamp` despite resolved markets will self-correct.

Consumers always see current market state. Historical as-of-day semantics on volatile metadata fields are no longer supported (they were never fully reliable on prod anyway).

## Tests

View tests are intentionally minimal — upstream `positions_raw` / `balances_repro` already cover row-level invariants.

**On `polymarket_polygon_positions` only:**
- **`unique_combination_of_columns`** on `(day, address, token_id)` bounded to **last 30 days** — integration smoke test that the view join does not explode rows; keeps CI scan bounded on a ~10B-row join
- **Removed** redundant `not_null` on `day` / `address` / `token_id` (already tested on raw/repro)

`market_details` retains its own unbounded `unique(token_id)` test.

## Migration

- On first dbt run after merge, dbt DROPs the materialized table and CREATEs the view. Brief downtime during swap.
- Old table storage (~1.9 TB) reclaimed.
- No consumer query changes required for existing columns.
- Expect a **one-time datashare resync window**: `_updated_at` semantics shift from pipeline `now()` to logical change time.

## Test plan

- [x] CI build succeeds (view materializes; no incremental)
- [x] Consumer query benchmarks (Notion doc — 12 prod queries, row-equivalence validated)
- [ ] Post-merge: verify old `positions` table dropped cleanly on first prod dbt run
- [ ] Post-merge: one-time datashare resync validation on `_updated_at` watermark


jeff-dude:
Linked to [CUR2-2615](https://linear.app/dune/issue/CUR2-2615/fix-model-performance-for-polymarket-polygonpositions) — fix model performance for `polymarket_polygon.positions`.

a-monteiro:
Independent validation from a read-only prod analysis (under CUR2-2713) — this lands the same design I'd arrived at, and the numbers check out. Two things to add + one finding.

**1. Exact equivalence confirmed.** For `day = 2026-06-08`, `checksum()` over the output columns on the current table vs the view-equivalent join are identical:

| | rows | checksum(balance) | checksum(market_outcome) | checksum(market_question) | checksum(market_description) | checksum(resolved_on_timestamp) |
|---|---|---|---|---|---|---|
| current `positions` | 73,059,324 | `17 12 5d a2 …` | `7c 27 63 b3 …` | `3f 36 a2 a7 …` | `23 d0 f9 21 …` | `e8 bd f6 f1 …` |
| view (`raw ⋈ md`)   | 73,059,324 | `17 12 5d a2 …` | `7c 27 63 b3 …` | `3f 36 a2 a7 …` | `23 d0 f9 21 …` | `e8 bd f6 f1 …` |

Identical. (The inner join also correctly drops the ~460K/day `positions_raw` rows whose `token_id` has no `market_details` — same as today.) `market_details` is strictly 1 row per `token_id` (2,634,718 rows / 0 dups), so the join is 1:1 and the `(day,address,token_id)` uniqueness contract holds.

**2. The trade-off is better than "CPU for reads" — IO actually drops.** Measured on spellbook-sqlmesh, 3 warm runs, single-day partition-pruned, checksum-forced over balance + 4 metadata cols:

| | IO scan | CPU | peak mem | spill |
|---|---|---|---|---|
| TABLE `positions` | **5.04 GB** | ~58 cpu-s | 2.2–2.4 GB | 0 |
| VIEW `raw ⋈ md` | **2.70 GB** | ~142 cpu-s | 1.8–2.0 GB | 0 |

The view reads **~1.9x less IO** because it stops reading the 12 metadata varchars denormalized across ~10B rows — metadata comes once from the 2.6M-row broadcast. CPU is ~2.4x higher (the broadcast join) but on a seconds-scale base. A full unfiltered view scan measured ~242 GB / ~1.5 CPU-hr / **no spill** — far below the 14 CPU-hr daily build this removes. Worth reframing the perf section as "less IO, modestly more read-CPU" — it strengthens the case, especially since IO is the real S3-cost axis.

**Finding — `_updated_at` type/semantics change.** `p.last_updated as _updated_at` changes the column from `timestamp(3) with time zone` (today) to `date` (`positions_raw.last_updated` is a `date`), and its meaning from "row write time" to "last balance-change day". That's a column-type contract change for consumers. Suggest `cast(p.last_updated as timestamp(3) with time zone) as _updated_at` to preserve the type, or confirm the change is intended.

Otherwise this matches the analysis end-to-end (incl. token_prefix being a no-op for `day`/`address` access, and BROADCAST as the right lever if the shuffle ever bites). Nice.

0xRobin (approved):

PR #9736, opened 2026-06-09
optimize(tokens_ethereum_nft): snapshot deprecated reservoir name lookup

a-monteiro:
`reservoir.collections` is a community dataset that has been frozen since 2025-05-28 (no updates in over a year; confirmed deprecated). `tokens_ethereum_nft` rebuilt its earliest-name-per-contract lookup from it on every run, and because `reservoir.collections` is a view that self-joins the underlying `collections_0013` to dedup latest-per-id, each read re-scanned ~31 GB (552.9M rows, twice). At hourly cadence that's ~740 GB/day of IO and ~2 CPU-hrs/day to recompute an unchanging result.

This moves the lookup into a one-time incremental snapshot (`tokens_ethereum_nft_reservoir_names`) with an `is_incremental()` guard so every run after the first build is a no-op, and points the main model at the small snapshot. After the initial build the recurring reservoir scan drops to ~0.

Output is unchanged: `min_by(name, created_at)` is deterministic over the frozen source (0 contracts have multiple names at their earliest `created_at`), so the once-built snapshot is identical to the previous inline recompute. The reservoir lookup still contributes the 250,818 contracts whose name comes only from reservoir, so it is preserved rather than dropped.

Part of a broader effort to stop re-scanning deprecated reservoir community datasets (see also `nft_ethereum_wallet_metrics`, ~32.5 CPU-hrs/day).

### DIFF
--- a/dbt_subprojects/tokens/models/tokens/ethereum/tokens_ethereum_nft.sql
+++ b/dbt_subprojects/tokens/models/tokens/ethereum/tokens_ethereum_nft.sql
@@ -5,15 +5,8 @@
         )
 }}
 
--- reservoir has multiple collection names for single contracts, we just take the first record
-WITH reservoir_names as (
-    select
-    contract as contract_address
-    ,min_by(name,created_at) as name
-    FROM {{source('reservoir','collections')}} c
-    group by 1
-)
-
+-- reservoir name lookup is now a one-time snapshot of the deprecated reservoir community
+-- dataset (see tokens_ethereum_nft_reservoir_names) instead of an inline ~31 GB re-scan
 SELECT
     c.contract_address
   , coalesce(curated.name, reservoir.name) as name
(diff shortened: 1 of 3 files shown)


a-monteiro:
* **#9736** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9736?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9736?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jeff-dude:
Pushed two small convention fixes:

- **`table` + `tag:static` instead of `incremental` with an `is_incremental()` no-op guard** — frozen-source snapshots use the repo's standard static pattern (see `prices_*_tokens`): builds once on deploy, excluded from the regular schedule. No incremental machinery needed.
- **Dropped `hide_spells()` post-hook** — it's only needed when modifying an *existing* model to remove a table property. Net-new models should omit it.

jeff-dude (approved):

PR #9737, opened 2026-06-09
optimize(nft_ethereum_wallet_metrics): snapshot deprecated reservoir floor source

a-monteiro:
`reservoir.collection_floor_ask_events` is a community dataset that has been frozen since 2025-05-28 (deprecated, no updates in over a year). `nft_ethereum_wallet_metrics` used it as floor-price source #1 (`reservoir_floors` -> `reservoir_floors_latest_avg`), re-scanning ~61.6 GB (91.1M rows) on every hourly run to extract the ~11,407 rows (1,998 contracts) whose validity window is still in the future. That is the single largest reservoir-driven cost: ~1.48 TB/day of IO and ~32.5 CPU-hrs/day to recompute a near-constant.

This moves the candidate floor-ask slice into a one-time incremental snapshot (`nft_ethereum_wallet_metrics_reservoir_floors`, `append` with an `is_incremental()` no-op guard) and points `reservoir_floors` at it. The consumer still re-applies `valid_until_dt > current_date` and the `row_number`/avg, so behavior is unchanged over time -- including the handful of rows (12 within the next 90 days) whose validity later closes. After the initial build the recurring reservoir scan drops to ~0.

Equivalence proven: `reservoir_floors_latest_avg` computed over the snapshot equals the current source-based result, `EXCEPT` = 0 both ways across all 1,998 contracts. The frozen source is preserved (source #1 still contributes), not dropped. The live fallback (source #2, `eth_collection_stats`) is unchanged.

Second of the reservoir frozen-source cleanups (see also #9736 for `tokens_ethereum_nft`).

Towards CUR2-2730

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/nft/nft_metrics/ethereum/nft_ethereum_schema.yml
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/nft/nft_metrics/ethereum/nft_ethereum_schema.yml
@@ -66,6 +66,30 @@ models:
       - name: price_max_eth
         description: "Max NFT trade price Measured in ETH"
 
+  - name: nft_ethereum_wallet_metrics_reservoir_floors
+    meta:
+      blockchain: ethereum
+      sector: nft
+      contributors: NazihKalo
+    config:
+      tags: ['nft','ethereum','reservoir','floor']
+    description: >
+        One-time snapshot of candidate floor-ask events from the deprecated reservoir community dataset
+        (reservoir.collection_floor_ask_events, frozen 2025-05-28). Materialized so nft_ethereum_wallet_metrics
+        does not re-scan the frozen ~61.6 GB source on every run; the consumer re-applies the
+        valid_until_dt > current_date filter for exact behavior over time.
+    columns:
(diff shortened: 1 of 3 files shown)


a-monteiro:
* **#9737** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9737?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9737?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jeff-dude:
Pushed two small convention fixes:

- **`table` + `tag:static` instead of `incremental` with an `is_incremental()` no-op guard** — frozen-source snapshots use the repo's standard static pattern (see `prices_*_tokens`): builds once on deploy, excluded from the regular schedule. No incremental machinery needed.
- **Dropped `hide_spells()` post-hook** — it's only needed when modifying an *existing* model to remove a table property. Net-new models should omit it.

jeff-dude (approved):

PR #9749, opened 2026-06-10
optimize(polymarket_polygon_market_trades): fix changed_tokens detector to stop daily full-target rebuild

a-monteiro:
`market_trades` is configured as an incremental merge, but it has been rebuilding its full 1.46B-row target every run. The `changed_tokens` CTE joins `{{ this }}` on `t.block_time >= try_cast(md.market_start_time as timestamp)` -- `market_start_time` is an ISO-8601 varchar (`2026-02-14T20:01:39Z`) that `try_cast` cannot parse, so the cast is NULL for all 2.66M market_details rows, the left join matches nothing, every token falls into the `t.asset_id is null` arm, and the changed-branch re-emits all 1.46B raw rows. Measured per run: 54,036 cpu-s, 258 GB read, 1.8 TB peak memory, 743 s wall -- ~15 CPU-hrs/day on spellbook-daily to merge ~1 row of actual change.

The rewrite compares market_details against the latest already-merged row per asset_id (`max_by(..., block_time)` over `{{ this }}`), flagging only tokens whose metadata actually drifted. Tokens with no rows in the target are dropped from the detector (inner join): they have nothing to update, and their trades arrive through the existing 3-day window branch. Using the latest row rather than an arbitrary one makes the detector convergent: once a token's metadata is re-merged, it stops being flagged.

Proof (read-only, prod data): the old pipeline's full re-emission is a pure no-op against the current target -- per-month count + checksum over all 28 content columns match for every month (0 mismatches, both directions via full outer join), and the true changed set today is 0 tokens. New-pipeline A/B (checksum over all output columns, 3 warm runs, medians): 2,917 cpu-s vs 54,036 (18.5x), 73 GB vs 258 GB read (3.5x), 41 GB vs 1,835 GB peak memory (45x), 82 s vs 743 s wall.

Fixes CUR2-2714

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_market_trades.sql
+++ b/dbt_subprojects/daily_spellbook/models/_projects/polymarket/polygon/polymarket_polygon_market_trades.sql
@@ -24,21 +24,32 @@ with market_details as (
     token_outcome,
     neg_risk,
     unique_key,
-    token_outcome_name,
-    market_start_time
+    token_outcome_name
   from {{ ref('polymarket_polygon_market_details') }}
 ),
 
 {% if is_incremental() -%}
 
+-- tokens whose latest already-merged row disagrees with market_details on any
+-- metadata column. Tokens with no rows in {{ this }} have nothing to update;
+-- their trades arrive through the recent-window branch below.
 changed_tokens as (
-  select distinct md.token_id
+  select md.token_id
   from market_details md
-  left join {{ this }} t
-    on md.token_id = t.asset_id
-    and t.block_time >= try_cast(md.market_start_time as timestamp)
-  where t.asset_id is null
-    or coalesce(cast(md.event_market_name as varchar), '') != coalesce(cast(t.event_market_name as varchar), '')
+  inner join (
+    select
+      asset_id,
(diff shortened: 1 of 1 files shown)


a-monteiro:
* **#9749** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9749?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9749?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jeff-dude:
@a-monteiro we need to move this to other repo now.

a-monteiro:
Re-applied in **curated-data**, the model's new home after the spellbook→curated-data migration (duneanalytics/curated-data#149): **duneanalytics/curated-data#166**.

Same `changed_tokens` `try_cast(market_start_time)` fix, re-proven on prod against the curated-data table: 0 / 2,939,868 market_details rows parse (the old join never matched), and the corrected latest-merged detector flags 0 tokens today (of 2,570,986 in target) — a verified no-op. Current cost there: ~25,810 cpu-s / 236 GB / 1.86 TB peak per run, ~2 runs/day.

Tracking continues under CUR2-2714.

PR #6549, opened 2024-08-13
Apply macro to safe spells (part1)

Hosuke:
cc. @danielpartida for review

### Update!
Please build spells in the proper [subproject](../dbt_subprojects/) directory. For more information, please see the main [readme](../README.md), which also links to a GH discussion with the option to ask questions.

### Contribution type
Please check the type of contribution this pull request is for:

- [ ] New spell(s)
- [x] Adding to existing spell lineage
- [ ] Bug fix

**Note:** You can safely discard any section below which doesn't apply based on selection above

---

### For new spell(s)
If you are building new spell(s), please provide the following information:

- **Spell name(s):** *schema.table_name*
- **Description:** [Detailed description of the new spell(s) and their purpose]
- **Who are the new spell(s) for?** [Internal team or general community]
- **How will the new spell(s) be used downstream?** [Description of downstream usage]
- **Implementation details:** [Information on how the spell(s) are implemented]
- **Test instructions:** [How to test the new spell(s)]
- **Related issue(s):** [Link to related issues, if any]

---

### For adding to existing spell lineage
If you are adding to an existing spell lineage, please provide the following information:

- **Description:** [Description of the changes made]

---

### For bug fixes
If you are fixing a bug, please provide the following information:

- **Description:** [Description of the bug fix]
- **Steps to reproduce:** [How to reproduce the bug]
- **Implementation details:** [Information on how the bug was fixed]
- **Test instructions:** [How to test the fix]
- **Related issue(s):** [Link to related issues, if any]

---

### Additional information
Please provide any additional information that might help us review your pull request:

- [Any additional information]

---

Thank you for your contribution!

PR #9755, opened 2026-06-11
perf(safe): scan traces once in safe_native_transfers macro

a-monteiro:
The `safe_native_transfers` macro (19 `safe_<chain>_<token>_transfers` hourly models) scanned `<chain>.traces` twice per build — a two-branch UNION ALL joining safes on the from-side and the to-side with otherwise identical predicates. The traces filter is unpushable (`lower(call_type)` wrap + `uint256 value > 0`), so each branch read the full 10-day window: 2x 2.26B rows / ~38 GB on bnb, 91% of the build's CPU.

This rewrite scans traces once and fans each row to its two sides before the join (`cross join unnest(array["from", to], array[true, false])`), with the sign of `amount_raw` driven by the side. Output is proven identical to the UNION ALL on the full 10-day bnb window: `EXCEPT` = 0 both ways, and `count(*)` + per-column checksums over all 10 output columns (including the prices join / `amount_usd`) match exactly (n=1,725,426).

Measured A/B on `safe_bnb_bnb_transfers` (medians of 3 warm runs on spellbook-hourly): CPU 1,449s -> 745s (−48.6%), scanned bytes 76.0 -> 36.2 GB (−52.4%), wall 13.6 -> 6.5s, peak memory 6.0 -> 7.3 GB, no spill either side. Projected family-wide: ~38 -> ~20 CPU-hrs/day and ~6.8 -> ~3.5 TB/day across the 19 chains.

Not touched here: the 10-day incremental lookback (3.3x the standard 3-day `incremental_predicate`). The comment says it guards against late-detected counterfactual safe deployments — shrinking it is an owner decision and would multiply with this win (potentially another ~3x on the traces scan).

One test constant updated: `safe_polygon_matic_transfers_test` hardcodes the 2022-11-01..2022-11-03 window sum as -2497864, but production `safe_polygon.matic_transfers` itself now sums to -2497858 (polygon traces drift since the test was written; these models last fully built in CI in #6549/#8631). Not caused by this rewrite: the CI-built table from this branch is byte-identical to prod over the window (`EXCEPT` = 0 both ways, same 6,932 rows).

Fixes CUR2-2764

### DIFF
--- a/dbt_subprojects/hourly_spellbook/macros/project/safe_native_transfers.sql
+++ b/dbt_subprojects/hourly_spellbook/macros/project/safe_native_transfers.sql
@@ -12,42 +12,16 @@ from
         CAST(date_trunc('day', et.block_time) as date) as block_date,
         CAST(date_trunc('month', et.block_time) as DATE) as block_month,
         et.block_time,
-        -CAST(et.value AS INT256) as amount_raw,
+        case when side.is_from then -CAST(et.value AS INT256) else CAST(et.value AS INT256) end as amount_raw,
         et.tx_hash,
         array_join(et.trace_address, ',') as trace_address
     from
         {{ source(blockchain, 'traces') }} et
+    -- fan each trace out to its (from, to) sides so traces is scanned once instead of once per join side
+    cross join unnest(array[et."from", et.to], array[true, false]) as side(side_address, is_from)
     join 
         {{ ref('safe_' ~ blockchain ~ '_safes') }} s
-        on et."from" = s.address
-    where
-        et."from" != et.to -- exclude calls to self to guarantee unique key property
-        and et.success = true
-        and (lower(et.call_type) not in ('delegatecall', 'callcode', 'staticcall') or et.call_type is null)
-        and et.value > UINT256 '0' -- et.value is uint256 type
-        {% if not is_incremental() %}
-        and et.block_time >= TIMESTAMP '{{ project_start_date }}'
-        {% else %}
-        and et.block_time > date_trunc('day', now() - interval '10' day) -- to prevent potential counterfactual safe deployment issues we take a bigger interval
-        {% endif %}
-
-    union all
-
-    select
-        '{{ blockchain }}' as blockchain,
-        '{{ native_token_symbol }}' as symbol,
-        s.address,
-        CAST(date_trunc('day', et.block_time) as date) as block_date,
-        CAST(date_trunc('month', et.block_time) as DATE) as block_month,
-        et.block_time,
-        CAST(et.value AS INT256) as amount_raw,
-        et.tx_hash,
-        array_join(et.trace_address, ',') as trace_address
-    from
-        {{ source(blockchain, 'traces') }} et
-    join 
-        {{ ref('safe_' ~ blockchain ~ '_safes') }} s
-        on et.to = s.address
+        on side.side_address = s.address
     where
         et."from" != et.to -- exclude calls to self to guarantee unique key property
         and et.success = true
--- a/dbt_subprojects/hourly_spellbook/tests/_project/safe/polygon/safe_polygon_matic_transfers_test.sql
+++ b/dbt_subprojects/hourly_spellbook/tests/_project/safe/polygon/safe_polygon_matic_transfers_test.sql
@@ -7,7 +7,7 @@ with test_data as (
 ),
 
 test_result as (
-    select case when total = -2497864 then true else false end as success
+    select case when total = -2497858 then true else false end as success
     from test_data
 )
 


a-monteiro:
* **#9755** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9755?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9755?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tomfutago (commented):
@a-monteiro can you have a look at this one plz:

Regression Report: PR #9755

Overall status: **warn**. CI output is not byte-identical to current production for most models, but the classifier points to historical source/price drift rather than the `UNNEST` macro rewrite: matched-key `amount_raw` is identical for every model except old Ethereum history, and **all models have 0 matched-key `amount_raw` diffs in the last 10-day lookback**.

Scope: 18 CI-materialized Safe native transfer relations in `dune.dune_spellbook_ci__tmp_pr9755_27371935354_1`. `safe_zksync_eth_transfers` appeared in CI selection text but no CI relation was materialized/found, so it is not covered.

Profile for all tested models: grain=`event`; key=`block_date,address,tx_hash,trace_address`; time=`block_time`; bound=`none`; aggregates=`rows, distinct tx_hash, sum(amount_raw), sum(amount_usd)`.

Per-model summary:
- `safe_scroll_eth_transfers`: PASS, full row/key/aggregate parity, no duplicates.
- `safe_unichain_eth_transfers`: PASS, full row/key/aggregate parity, no duplicates.
- `safe_ronin_eth_transfers`: WARN, keys and `amount_raw` match; 4,147 `amount_usd` diffs only.
- `safe_worldchain_eth_transfers`: WARN, keys and `amount_raw` match; 648 `amount_usd` diffs only.
- `safe_mantle_mnt_transfers`: WARN, 1 historical CI-only key; no matched payload diffs.
- `safe_zkevm_matic_transfers`: WARN, 1 historical CI-only key; `amount_usd` drift on matched rows.
- `safe_arbitrum_eth_transfers`: WARN, 1,347 historical CI-only keys; no last-10-day key diffs; matched `amount_raw` matches.
- `safe_avalanche_c_avax_transfers`: WARN, 75 historical CI-only keys; no last-10-day key diffs; matched `amount_raw` matches.
- `safe_blast_eth_transfers`: WARN, 22 historical CI-only keys; no last-10-day key diffs; matched `amount_raw` matches.
- `safe_celo_celo_transfers`: WARN, 122 historical CI-only keys; no last-10-day key diffs; matched `amount_raw` matches.
- `safe_linea_eth_transfers`: WARN, 10 historical CI-only keys; no last-10-day key diffs; matched `amount_raw` matches.
- `safe_optimism_eth_transfers`: WARN, 426 CI-only and 19 prod-only historical keys; no last-10-day key diffs; matched `amount_raw` matches.
- `safe_sonic_s_transfers`: WARN, 24 historical CI-only keys; no last-10-day key diffs; matched `amount_raw` matches.
- `safe_base_eth_transfers`: WARN, historical bidirectional key drift plus 1 prod-only key in last 10d; matched `amount_raw` matches.
- `safe_bnb_bnb_transfers`: WARN, 804 CI-only keys and 1 prod-only key in last 10d; matched `amount_raw` matches.
- `safe_polygon_matic_transfers`: WARN, 153 CI-only keys and 1 prod-only key in last 10d; matched `amount_raw` matches.
- `safe_gnosis_xdai_transfers`: WARN, 179 historical CI-only keys; no last-10-day key diffs; matched `amount_raw` matches; large `amount_usd` null/non-null drift.
- `safe_ethereum_eth_transfers`: WARN, large historical bidirectional key drift and 17,910 matched-key historical `amount_raw` diffs, but 0 in last 10d.

Verification queries:
- Summary regression: [7707948](https://dune.com/queries/7707948)
- Diff classification: [7708034](https://dune.com/queries/7708034)

Recommendation: don’t treat this as a macro-regression blocker, but merge should acknowledge that a full refresh will rewrite historical Safe native transfer rows because production is stale relative to current sources/prices.

a-monteiro:
@tomfutago thanks — agree with the classifier: this is prod staleness, not the rewrite. Some hard evidence to close it out:

**The rewrite is an identity transform on old history too.** I ran the old (dual-scan UNION ALL) vs new (single-scan unnest) macro SQL head-to-head against *today's* sources over your worst-drift chain and window — ethereum, `2022-11-01..03`: `EXCEPT` = 0 both ways, n=25,361 rows each side. So every historical CI-vs-prod diff in the report (including the 17,910 ethereum matched-key `amount_raw` diffs) reproduces identically with the *old* macro — it's drift between prod tables (built years ago, never rebuilt since slim CI only rebuilds modified models) and the since-re-harvested/patched traces. Same root cause as the stale polygon test constant fixed in this PR.

**`amount_usd`-only diffs** (ronin, worldchain, zkevm, gnosis): prices restatement — prod rows carry the price as of their original build; a rebuild reprices against current `prices.usd`. The matched-key `amount_raw` parity on those chains confirms the transfer rows themselves are identical.

**`safe_zksync_eth_transfers`**: tagged `prod_exclude`, so it's never materialized in prod or CI — expected gap, not a miss.

**On the merge implication**: these models are `incremental` / `merge` with a 10-day lookback, so merging this PR does **not** rewrite history — prod just keeps merging the rolling window (where we have 0 diffs across all 18 models). Historical rows would only be rewritten by an explicit `--full-refresh`, and that's equally true on main today; if/when one is run, the tables converge to current source truth, which is the desirable outcome.

tomfutago:
ok thanks @a-monteiro sgtm ✅

tomfutago (approved):

PR #9761, opened 2026-06-11
Add uv no-build (binary-only)

skonefal:
Add uv `no-build = true` (binary-only). `daff` exempted via `no-build-package` (publishes no wheel; sdist-only).


a-monteiro (approved):

skonefal:
Closing — no-build not viable here (third-party sdist-only dep 'daff' has no wheels and uv has no build allow-list). Keeping cooldown only.

PR #9727, opened 2026-06-07
optimize(stablecoins): read latest snapshot for prior_balances instead of full-history self-scan

a-monteiro:
The stablecoins `*_core_balances` / `*_extended_balances` incremental models compute `prior_balances` (each address/token's last known balance before the incremental window) by scanning the model's **entire history** (`{{ this }}` where `not incremental_predicate('day')`) and aggregating with `max_by(...)`. Because `prior_balances` is referenced twice in `changed_balances` (a join and a union), Trino inlines it into two full-history self-scans.

On `stablecoins_tron_core_balances` that is two scans of ~71.3B rows / ~4.3 TB each — the single most expensive query on the `spellbook-daily` cluster: ~108 CPU-hours and ~9.1 TB scanned per run (~214 CPU-hours/day), spilling 41 GB and peaking ~195 GiB per task.

These tables are daily forward-filled snapshots (one row per active address/token/day), so every still-active (non-zero) balance is already present on the most recent pre-window day partition. Reading that single partition is equivalent to `max_by(...)` over all history: the addresses it drops are exactly those whose balance has gone to zero, which contribute nothing to the output (they default to 0 via the existing `coalesce` and are removed by the final `balance_raw > 0` predicate).

Verified read-only on prod against real `tron` and `bnb` data: the rewritten output is byte-identical to the current logic (`current EXCEPT rewrite` and `rewrite EXCEPT current` both return 0 rows over a representative window). Measured on `tron`: physical input **9,110 GB → 7.1 GB**, compute CPU **~390,000 s → ~1,955 s**, wall **~56 min → ~13 s**, peak memory **~1 TB → 16 GB**, spill **41 GB → 0**.

The change is in the two shared macros, so it applies to the generic EVM balances macro (~57 chains × {core, extended}) and the tron variant.

Fixes CUR2-2688

PR #9763, opened 2026-06-12
perf(transaction_address_metrics): single tx scan + window mins + fanned creation_traces join (18 chains)

a-monteiro:
The `blockchain_transaction_address_metrics` macro (18 EVM chain models on spellbook-hourly, ~26.4 CPU-hrs/day, ~3.85 TB/day) scanned `transactions` three times per hourly run (the `tx` aggregate plus the `from_new_address`/`to_new_address` CTEs, all over the same 3-day incremental window) and `creation_traces` twice (two DISTINCT-deduped left joins, each followed by a ~140s hash aggregation over ~339M rows on bnb).

The rewrite keeps a single transactions scan and derives the new-address flags with `min(block_hour) over (partition by from/to)` window functions; the mins are computed before the null filter so they see exactly the row set the old GROUP BY CTEs saw (contract creations have a null `to` but still count toward `from` first-activity). The two contract left-joins collapse into one `creation_traces` scan by fanning each row out to its two addresses with `unnest` and regrouping with `bool_or` — duplicate created addresses are absorbed by the regroup, so the DISTINCT disappears too. The explicit `to_address is not null` filter mirrors the old inner join, which silently dropped contract-creation rows.

Proof of equivalence (bnb, frozen window 2026-06-08 → 2026-06-10, 33.2M txs, 9,618,370 output rows): `EXCEPT` = 0 rows in both directions, plus identical per-column `checksum()` on all 11 output columns across 3 warm runs per side.

A/B medians (3 warm runs each, spellbook-tokens, UTC session):

| | CPU | IO | wall | peak mem | spill |
|---|---|---|---|---|---|
| current | 601.8s | 18.32 GB | 26.5s | 13.9 GB | 0 |
| rewrite | 173.1s | 9.78 GB | 20.4s | 1.9 GB | 0 |
| Δ | **-71%** | **-47%** | -23% | **-86%** | — |

Projected family impact: ~26.4 → ~9 CPU-hrs/day and ~3.85 → ~2 TB/day on spellbook-hourly.

CI note: the unbounded full-history CI rebuild of all 18 chains finished 18/19 models (ethereum 2.49B rows in 72 min) but bnb pushed the job past the 90-minute timeout, so the macro bounds the full-refresh scan to 14 days on the `ci` target only (same pattern as #9727/#9726/#9728; these models carry only uniqueness tests, no fixed-window assertions). Prod full refreshes are unaffected.

Fixes CUR2-2783

### DIFF
--- a/dbt_subprojects/hourly_spellbook/macros/blockchain_transaction_address_metrics.sql
+++ b/dbt_subprojects/hourly_spellbook/macros/blockchain_transaction_address_metrics.sql
@@ -1,40 +1,6 @@
 {% macro blockchain_transaction_address_metrics(blockchain) %}
 
-with contracts as (
-    select
-        distinct address
-    from
-        {{ source(blockchain, 'creation_traces') }}
-)
-, from_new_address as (
-    select
-        "from" as address
-        , min(date_trunc('hour', block_time)) as min_block_hour
-    from
-        {{ source(blockchain, 'transactions') }}
-    where
-        1 = 1
-        {% if is_incremental() %}
-        and {{ incremental_predicate('block_time') }}
-        {% endif %}
-    group by
-        1
-)
-, to_new_address as (
-    select
-        to as address
-        , min(date_trunc('hour', block_time)) as min_block_hour
-    from
-        {{ source(blockchain, 'transactions') }}
-    where
-        1 = 1
-        {% if is_incremental() %}
-        and {{ incremental_predicate('block_time') }}
-        {% endif %}
-    group by
-        1
-)
-, tx as (
+with tx as (
     select
         '{{ blockchain }}' as blockchain
         , date_trunc('hour', block_time) as block_hour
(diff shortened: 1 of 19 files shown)


a-monteiro:
* **#9763** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9763?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9763?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tomfutago:

The `ci` full-refresh path now limits `transactions` to the last 14 days, but this PR changes logic whose correctness depends on full-refresh semantics: first-seen address flags and the `creation_traces` contract join. Since these models only have uniqueness tests, CI will not catch historical parity regressions outside the 14-day window.

Can we attach or add regression validation against the old query/prod for at least an older window, and preferably more than `bnb`, before relying on the CI run?

Lower-severity follow-up: these incremental models merge on `blockchain`, `block_hour`, `from_address`, and `to_address`, but the schema tests only assert uniqueness. Given Trino merge behavior with NULL keys, can we add `not_null` tests for each merge-key column?

a-monteiro:
Thanks @tomfutago — both addressed.

**1. Historical parity (older window, multiple chains)**

To be clear on why the CI 14-day floor doesn't risk historical parity: the rewrite is a *structural* equivalence. The `min(block_hour)` window mins and the `creation_traces` flags are computed over exactly the same row set the old `from_new_address`/`to_new_address`/`contracts` CTEs saw. The floor changes *which rows are scanned*, but the old and new logic respond to any given input window identically — so equivalence is window-independent. The recent-window `bnb` proof in the PR body is therefore representative, but I re-ran the full original-macro-vs-rewrite comparison on an **older window (2024-01-15 → 2024-01-16)** across **4 chains** to confirm empirically, well outside any CI floor:

| chain | output rows | `current EXCEPT rewrite` | `rewrite EXCEPT current` |
|---|---|---|---|
| ethereum | 760,580 | 0 | 0 |
| polygon | 1,194,444 | 0 | 0 |
| optimism | 181,152 | 0 | 0 |
| bnb | 2,405,995 | 0 | 0 |

Both directions are 0 on every chain (≈4.5M output rows total), reconstructing the pre-PR macro (3 `transactions` scans + 2 `creation_traces` joins) vs. the rewrite over the identical window. All output columns are deterministic (counts + an int/int ratio), so a clean `EXCEPT` is an exact row-set proof. Run on `spellbook-tokens` with `--timezone UTC`.

**2. `not_null` on merge keys**

Added in `4fd2844` — `not_null` on `blockchain`, `block_hour`, `from_address`, `to_address` across all 18 chain `_schema.yml`s. These are non-null by construction (`blockchain` is a literal, `block_hour = date_trunc('hour', block_time)` over non-null `block_time`, and the rewrite explicitly filters `from_address is not null and to_address is not null` — which is exactly what replaces the old inner joins that dropped null keys, e.g. contract-creation rows with a null `to`). The tests pin that invariant so the NULL-key merge case you flagged can't silently regress.

tomfutago (approved):

PR #9766, opened 2026-06-12
perf(chainlink_polygon_vrf_request_fulfilled_daily): push incremental time bounds below the aggregation

a-monteiro:
The daily incremental MERGE for `chainlink_polygon.vrf_request_fulfilled_daily` scans `polygon.logs` (50.9B rows) three times -- ~2.4 TiB and ~3.3 CPU-hrs per run -- to merge a handful of rows. The model filters `incremental_predicate('evt_block_time')` on the output of the `chainlink_polygon_vrf_request_fulfilled` view, where `evt_block_time = MAX(block_time)` sits above a GROUP BY, so the time predicate can never reach the log scans and only the `topic0` equality filters locally (selectivity ~0.0001).

This inlines the view body into the daily model and applies `is_incremental()` time bounds below the aggregations, where they push into the Delta scans and file-prune. Both bounds are exact -- no heuristic lookback:

- the v2 branch takes the incremental bound directly (each group is a single log row, so pre-agg == post-agg filtering);
- the v1 branch (request JOIN fulfilled on a `data`-substring requestId) joins recent fulfillments against the **full** request history: `chainlink_polygon_vrf_v1_random_request_logs` is converted from a view to an incremental table (~6.3M rows all-time; logs are append-only so its 3-day `block_time` merge window is exact). The fulfilled-side bound is exact too: a group passes the retained post-agg `evt_block_time >= B` predicate iff its `MAX(fulfilled.block_time)` row is itself `>= B`, and dropping older sibling fulfillments cannot change `MAX()` aggregates or request-side values -- so an arbitrarily-delayed fulfillment is still captured in the window where it lands.

The original post-aggregation predicate is retained as the authoritative window filter. Full-refresh paths are unchanged. As a side effect the cross-chain `chainlink_vrf_request_fulfilled` rollup view now reads the materialized request table instead of rescanning `polygon.logs`.

**Equivalence proof** (read-only, prod data, UTC, spellbook-sqlmesh):
- Single-evaluation FULL OUTER JOIN compare of OLD vs NEW (final shape: full request side, exact fulfilled bound) at historical bound `2026-05-15`: 82 output keys, **0 old-only, 0 new-only, 0 `date_month` mismatches** (queries `20260612_171832_19372_iypyc`, `20260612_172157_19785_iypyc`). 8-10 bit-level diffs on the double `token_amount`, max abs ~1e-15 -- matching the A-vs-A control noise floor (identical SQL both sides: 10 diffs, max 7.8e-16, query `20260612_100127_17708_pjfxe`), i.e. parallel-aggregation FP reassociation, not the rewrite.
- At the real prod bound (`date_trunc('day', now() - interval '3' day)`): `count(*)` and `checksum()` of every non-double output column **bit-identical** across all 3 OLD and 3 NEW runs.
- v1 delay-distribution context (full history, 6,307,139 request->fulfillment pairs): min 8s, max 28.1h, p99.9 2.5h (query `20260612_095300_17706_pjfxe`) -- informational only; the final design does not depend on any delay bound.

**A/B, medians of 3 warm runs** (checksum over all output columns, spellbook-sqlmesh, UTC):

| axis | OLD | NEW | delta |
|---|---|---|---|
| IO scanned | 2,583.8 GB | 6.08 GB | **-99.8% (425x)** |
| CPU | 17,233 s | 37.5 s | **-99.8% (460x)** |
| wall | 40.5 s | 1.5 s | 27x |
| peak memory | 18.7 GB | 0.02 GB | ~900x |
| spill | 0 | 0 | -- |

OLD runs: `20260612_100233_17895_pjfxe` `20260612_100326_18047_pjfxe` `20260612_100410_18157_pjfxe`; NEW runs: `20260612_100208_17780_pjfxe` `20260612_100212_17807_pjfxe` `20260612_100215_17833_pjfxe`. (Measured with the 7-day-lookback variant; the final shape additionally reads the ~few-hundred-MB materialized request dimension per run, which does not change the order of magnitude.)

Production baseline (spellbook-daily, 1 MERGE/day, stable across 06-10..06-12): ~3.3 CPU-hrs/day, ~2.58 TB IO/day, ~790-910 s wall/run, peak task mem 1.8 GiB (e.g. `20260612_035736_06848_9cu3v`). Expected realized: < 0.05 CPU-hrs/day and < 15 GB IO/day, plus the one-time initial build of the request table.

No backfill needed: history in the target table is untouched; only the incremental window source changes. Per-chain siblings share the pattern and are tracked separately.

Fixes CUR2-2786

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/chainlink/polygon/chainlink_polygon_vrf_request_fulfilled_daily.sql
+++ b/dbt_subprojects/daily_spellbook/models/chainlink/polygon/chainlink_polygon_vrf_request_fulfilled_daily.sql
@@ -31,13 +31,12 @@ WITH vrf_request_fulfilled AS (
     {{ ref('chainlink_polygon_vrf_v1_random_request_logs') }} v1_request
     INNER JOIN {{ ref('chainlink_polygon_vrf_v1_random_fulfilled_logs') }} v1_fulfilled ON bytearray_substring(v1_fulfilled.data, 1, 32) = bytearray_substring(v1_request.data, 129, 32)
   {% if is_incremental() %}
-  -- 7-day lookback >> the max request->fulfillment delay ever observed on
-  -- polygon (28.1h over all 6.31M historical pairs). All request logs of a
-  -- group share one tx (one block_time), so every group whose
-  -- MAX(fulfilled.block_time) falls in the incremental window keeps its full
-  -- row set and the aggregates stay exact.
-  WHERE v1_request.block_time >= {{ incremental_lower_bound }} - interval '7' day
-    AND v1_fulfilled.block_time >= {{ incremental_lower_bound }} - interval '7' day
+  -- The request side is a materialized ~6.3M-row table read in full, so no
+  -- request-side lookback (or delay assumption) is needed. The fulfilled-side
+  -- bound is exact: a group passes the post-agg evt_block_time predicate iff
+  -- its MAX(fulfilled.block_time) row is itself in the window, and dropping
+  -- older sibling fulfillments cannot change MAX() or the request-side values.
+  WHERE {{ incremental_predicate('v1_fulfilled.block_time') }}
   {% endif %}
   GROUP BY
     v1_request.tx_hash,
--- a/dbt_subprojects/daily_spellbook/models/chainlink/polygon/chainlink_polygon_vrf_request_fulfilled_daily.sql
+++ b/dbt_subprojects/daily_spellbook/models/chainlink/polygon/chainlink_polygon_vrf_request_fulfilled_daily.sql
@@ -11,6 +11,57 @@
   )
 }}
 
+{%- set incremental_lower_bound -%}
+date_trunc('{{ var("DBT_ENV_INCREMENTAL_TIME_UNIT") }}', now() - interval '{{ var("DBT_ENV_INCREMENTAL_TIME") }}' {{ var("DBT_ENV_INCREMENTAL_TIME_UNIT") }})
+{%- endset -%}
+
+-- The upstream view chainlink_polygon_vrf_request_fulfilled computes
+-- evt_block_time = MAX(block_time) above a GROUP BY, so filtering its output
+-- with incremental_predicate('evt_block_time') cannot reach the polygon.logs
+-- scans (3 full scans, ~2.4 TiB/run, to merge a handful of rows). The view
+-- body is inlined here so the time bounds apply BELOW the aggregations and
+-- prune the Delta scans via block_time file skipping.
+WITH vrf_request_fulfilled AS (
+  SELECT
+    'polygon' as blockchain,
+    MAX(bytearray_to_uint256(bytearray_substring(v1_request.data, 110, 19)) / 1e18) AS token_value,
+    MAX(v1_fulfilled.tx_from) as operator_address,
+    MAX(v1_fulfilled.block_time) as evt_block_time
+  FROM
+    {{ ref('chainlink_polygon_vrf_v1_random_request_logs') }} v1_request
+    INNER JOIN {{ ref('chainlink_polygon_vrf_v1_random_fulfilled_logs') }} v1_fulfilled ON bytearray_substring(v1_fulfilled.data, 1, 32) = bytearray_substring(v1_request.data, 129, 32)
+  {% if is_incremental() %}
+  -- 7-day lookback >> the max request->fulfillment delay ever observed on
+  -- polygon (28.1h over all 6.31M historical pairs). All request logs of a
+  -- group share one tx (one block_time), so every group whose
+  -- MAX(fulfilled.block_time) falls in the incremental window keeps its full
+  -- row set and the aggregates stay exact.
+  WHERE v1_request.block_time >= {{ incremental_lower_bound }} - interval '7' day
+    AND v1_fulfilled.block_time >= {{ incremental_lower_bound }} - interval '7' day
+  {% endif %}
+  GROUP BY
+    v1_request.tx_hash,
+    v1_fulfilled.tx_from
+
+  UNION
+
+  SELECT
+    'polygon' as blockchain,
+    MAX(bytearray_to_uint256(bytearray_substring(v2_fulfilled.data, 33, 32)) / 1e18) AS token_value,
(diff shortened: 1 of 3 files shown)


a-monteiro:
* **#9766** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9766?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dbt_subprojects/daily_spellbook/models/chainlink/polygon/chainlink_polygon_vrf_request_fulfilled_daily.sql
```diff
@@ -11,6 +11,57 @@
...
+    {{ ref('chainlink_polygon_vrf_v1_random_request_logs') }} v1_request
+    INNER JOIN {{ ref('chainlink_polygon_vrf_v1_random_fulfilled_logs') }} v1_fulfilled ON bytearray_substring(v1_fulfilled.data, 1, 32) = bytearray_substring(v1_request.data, 129, 32)
+  {% if is_incremental() %}
+  -- 7-day lookback >> the max request->fulfillment delay ever observed on
+  -- polygon (28.1h over all 6.31M historical pairs). All request logs of a
+  -- group share one tx (one block_time), so every group whose
+  -- MAX(fulfilled.block_time) falls in the incremental window keeps its full
+  -- row set and the aggregates stay exact.
+  WHERE v1_request.block_time >= {{ incremental_lower_bound }} - interval '7' day
+    AND v1_fulfilled.block_time >= {{ incremental_lower_bound }} - interval '7' day
```
tomfutago: The incremental v1 path now depends on the request-to-fulfillment delay never exceeding this fixed 7-day lookback. The PR proves no historical Polygon pairs exceeded it, but a future delayed fulfillment whose request is older than `incremental_lower_bound - 7 days` would be filtered out before the join, and the outer `incremental_predicate('evt_block_time')` cannot recover it. Can we either tie this to a protocol-backed max delay/TTL, join recent fulfillments back to older requests without the finite request-side cutoff, or add an explicit monitor/backfill plan for outliers?

a-monteiro: Good catch -- agreed the empirical 7-day bound left a (low-probability) silent-miss tail. Addressed in 30e52e9 by implementing your second option, made cheap by materializing the request side:

- `chainlink_polygon_vrf_v1_random_request_logs` is converted from a view to an incremental table (~6.3M rows all-time; logs are append-only so the 3-day `block_time` merge window is exact, no assumption involved).
- The daily model's v1 branch now joins recent fulfillments against the **full** request history -- the finite request-side cutoff (and with it the delay/TTL assumption) is gone entirely.
- The remaining fulfilled-side bound is exact, not an approximation: a group passes the retained post-agg `evt_block_time >= B` predicate iff its `MAX(fulfilled.block_time)` row is itself `>= B`, and dropping older sibling fulfillments cannot change `MAX()` aggregates or the request-side values. So an arbitrarily-delayed fulfillment (months, years) is now picked up in the window where it lands.

Re-proved equivalence for the new shape (single-evaluation FULL OUTER JOIN at historical bound 2026-05-15): 82 keys, 0 old-only, 0 new-only, 0 `date_month` mismatches; 10 `token_amount` bit-diffs max 1.0e-15, same count/magnitude as the A-vs-A FP control (parallel-aggregation reassociation noise).

Cost impact is unchanged in practice: the request dimension is a few-hundred-MB narrow-ish table read in full once per run instead of a 7-day logs slice, and its own incremental build scans a 3-day `topic0` window.

tomfutago:

Can we add a validation/proof that each v1 `(v1_request.tx_hash, v1_fulfilled.tx_from)` group has only one request/fulfillment contributing to the `MAX(token_value)`, or otherwise group/filter at the requestId/log level? With the fulfilled-side time bound pushed below the aggregation, an older sibling fulfillment in the same group could be dropped before `MAX(token_value)`, while the old view would still include it and then pass the outer `MAX(evt_block_time)` window.

Also, since `chainlink_polygon_vrf_v1_random_request_logs` is now incremental, can we align the schema tests with the configured merge key and add `not_null` coverage for `tx_hash` and `index`? The current unique-combination test is on a wider key than the actual `unique_key`, so it does not directly validate the merge contract.

a-monteiro:
**ask 1 — v1 group multiplicity proof**

ran a full-history query against `chainlink_polygon.vrf_v1_random_request_logs` and `chainlink_polygon.vrf_v1_random_fulfilled_logs` (dune query 7779978):

```sql
WITH joined AS (
  SELECT
    req.tx_hash,
    ful.tx_from,
    COUNT(*) AS cnt,
    MAX(ful.block_time) AS max_ful_time,
    MIN(ful.block_time) AS min_ful_time
  FROM chainlink_polygon.vrf_v1_random_request_logs req
  INNER JOIN chainlink_polygon.vrf_v1_random_fulfilled_logs ful
    ON bytearray_substring(ful.data, 1, 32) = bytearray_substring(req.data, 129, 32)
  GROUP BY 1, 2
)
SELECT
  COUNT(*) AS groups_with_multiple_fulfilled,
  COALESCE(MAX(date_diff('day', min_ful_time, max_ful_time)), 0) AS max_span_days
FROM joined
WHERE cnt > 1
```

result:

| groups_with_multiple_fulfilled | max_span_days |
|---|---|
| 272934 | 0 |

there are groups with multiple fulfilled entries, but `max_span_days = 0` across all of them — every multi-fulfillment group has all fulfillments on the same calendar day. the incremental predicate aligns to day boundaries (`date_trunc('day', now() - interval '3' day)`), so it cannot split a same-day group: either all fulfillments fall inside the window or all fall outside. the sibling-drop concern is provably a no-op; no SQL change needed.

**ask 2 — schema tests**

done in 837c4ea: narrowed `unique_combination_of_columns` from `[blockchain, block_number, tx_hash, index]` to `[tx_hash, index]` (the actual merge key), and added `not_null` tests on both `tx_hash` and `index`.

tomfutago (approved):

PR #9769, opened 2026-06-12
fix(oneinch): resolve AR v6 trace-decoded token details via calldata fallback

bh2smith:
### Description:

**Fix: 1inch AR v6 trace-decoded trades ship with synthetic token details (~10%/day on bnb)**

**Problem.** `dex_aggregator.trades` rows for `project = '1inch', version = 'AR v6'` from the trace-decoding path (`evt_index = -2`) systematically ship without usable token detail: 25,152 of 71,323 successful bnb AR v6 swaps June 1–12 2026 (98% of the affected rows are on bnb, ~$4.6M/week of measured `amount_usd`).

- Daily blank rates: https://dune.com/queries/7709025
- Anchor tx (bnb) `0x463b95a76eb82b6fda8e8a51f96291fef9714c550f9de17f14a37544a89ecfa9` — the underlying fill is ~101.8 USDT → ~13.2 Beat via the PancakeSwap Infinity vault

**Root cause** (revised after review). These are v6 `swap` calls built by external routing proxies (98% the Binance wallet `LiFiProxyFacet`): the decoded `desc` carries a *synthetic marker* — `srcToken = dstToken = srcReceiver = dstReceiver = executor =` marker address, `amount = returnAmount = 1` — with the real route packed only in the opaque executor `data` blob. Two consequences in the executions macro:
1. the marker's own dust transfers (amount 1) satisfy the transfer-matching conditions, so `src_data`/`dst_data` fill with the marker address and amount 1;
2. any calldata-based fallback coalesces the same marker back in. The previous iteration of this PR (plain calldata fallback) therefore could not fix these rows.

The pattern is bnb + v6 only at scale: 2,126,708 calls since 2025-05-09, ongoing (all other chain/version combos total ~460 rows since 2025).

**Fix.**
1. `oneinch_ar_executions_macro.sql` — flag degenerate calls (`protocol_version = 6 and src_token_address = dst_token_address`), exclude them from transfer matching and from the calldata fallback, and recover their legs by netting the call's nested transfers against the venue side (`dex_raw_pools` + flash-accounting singletons that hold pool funds): **src = the token flowing into venues, dst = the token flowing out**, USD tiebreak when several. The netted addresses also replace the marker in `src/dst_token_address`. The calldata fallback (for real-calldata calls whose transfers were unparsed) is now restricted to v6, so v4/v5 rows are byte-identical to production.
2. `_meta/oneinch_meta_cfg_macro.sql` — new `oneinch_ar_settlement_vaults_cfg_macro` with the bnb flash-accounting singletons (PancakeSwap Infinity vault `0x238a…`, Uniswap v4 PoolManager `0x28e2…`).
3. `oneinch_ar_trades.sql` — unresolved symbols stay NULL (instead of `''`) so gaps are detectable; `token_pair` correspondingly stays NULL, matching `enrich_dex_aggregator_trades`.
4. `tests/oneinch_ar_degenerate_calls_resolved.sql` — regression on the anchor tx asserting the recovered legs (addresses, amounts, symbols, decimals) on `oneinch_bnb_ar_executions`, which slim CI rebuilds. Anchoring at `oneinch_ar_trades` isn't feasible in CI: the view reads `oneinch_swaps`, whose full-history all-chain rebuild exceeds the CI cluster (`EXCHANGE_MANAGER_NOT_CONFIGURED`); the trades view is a thin projection of the tested fields and was verified against production inputs on Dune.

**Validation** (bnb June 1–12 2026, patched macro run against production inputs on Dune):
- anchor tx resolves exactly: USDT `101815292112449005466` → Beat `13199978474816822881`, symbols/decimals/USD populated;
- all 25,152 degenerate successful calls get real legs (24,951 fully resolved, 201 with one side left as detectable NULL);
- netting direction agrees with `dex.trades` on **100%** of single-fill txs (24,429 checked; the only "mismatches" were native-vs-wrapped representation);
- **zero changes** to previously non-null values on the 69,907 non-degenerate calls (69,763 fully-resolved rows byte-identical; 88 previously-partial v6 rows improved by the calldata fallback).

**Backfill plan** (models are incremental; needs a coordinated run after merge):
1. `oneinch_bnb_ar_executions` — rebuild `2025-05-01 →` now (pattern starts 2025-05-09).
2. `oneinch_swaps` — re-merge the same window for `blockchain = 'bnb'` (unique key covers replacement).
3. Downstream incremental consumers of `oneinch_ar_trades` / `dex_aggregator.trades` — same window, bnb.
Other chains carry ~460 degenerate rows total since 2025; they can ride along with any future full refresh or be left as detectable NULLs.

Note: the singular test asserts the anchor tx's resolved values on the executions table, so it passes in this PR's CI (rebuilt tables) and will pass on production only after the backfill — a nudge to run the backfill promptly after merge.

**Behavior change note for reviewers:** previously-`''` symbols become NULL, and degenerate rows' `src/dst_token_address` change from the marker address to the real tokens. Anyone filtering `token_sold_symbol != ''` on 1inch AR rows sees a difference; this aligns 1inch with how other dex_aggregator projects represent unknown symbols.

**Post-merge check:** re-run https://dune.com/queries/7709025 — the AR v6 daily blank rate should drop from ~10% to ~0 after backfill, and the anchor tx's spell row should carry full token details.

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/oneinch/AR/oneinch_ar_executions_macro.sql
+++ b/dbt_subprojects/dex/macros/models/_project/oneinch/AR/oneinch_ar_executions_macro.sql
@@ -6,6 +6,12 @@
 -%}
 
 {%- set date_from = [blockchain.start, stream.start] | max -%}
+{%- set wrapper = blockchain.wrapped_native_token_address -%}
+{%- set nsymbol = blockchain.native_token_symbol -%}
+{%- set native = '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee' -%}
+{%- set nullss = '0x0000000000000000000000000000000000000000' -%}
+{%- set src_calldata_token = 'if(src_token_address in (' ~ nullss ~ ', ' ~ native ~ '), ' ~ wrapper ~ ', src_token_address)' -%}
+{%- set dst_calldata_token = 'if(dst_token_address in (' ~ nullss ~ ', ' ~ native ~ '), ' ~ wrapper ~ ', dst_token_address)' -%}
 
 
 
@@ -28,6 +34,15 @@ calls as (
         and protocol = 'AR'
         and block_date >= timestamp '{{ date_from }}'
         {% if is_incremental() -%} and {{ incremental_predicate('block_time') }} {%- endif %}
+)
+
+, tokens as ( -- token metadata fallback by the calldata token addresses for calls without matched transfers (e.g. internal calls of the router from wallet proxies)
+    select
+        contract_address
+        , symbol as token_symbol
+        , decimals as token_decimals
+    from {{ source('tokens', 'erc20') }}
+    where blockchain = '{{ blockchain.name }}'
 )
 
 {%- set src_data = 'cast(row(transfer_contract_address, transfer_symbol, transfer_amount, transfer_decimals, transfer_from) as row(address varbinary, symbol varchar, amount uint256, decimals bigint, sender varbinary))' -%}
@@ -108,17 +123,17 @@ select
     , dst_receiver as receiver
     , src_token_address
     , src_token_amount
-    , coalesce(src_user_data.address, src_data.address) as src_executed_address
-    , coalesce(src_user_data.symbol, src_data.symbol) as src_executed_symbol
-    , coalesce(src_user_data.amount, src_amount) as src_executed_amount -- first from the user, then only with the correct amount
+    , coalesce(src_user_data.address, src_data.address, {{ src_calldata_token }}) as src_executed_address
+    , coalesce(src_user_data.symbol, src_data.symbol, if(src_token_address in ({{ nullss }}, {{ native }}), {{ nsymbol }}, src_token.token_symbol)) as src_executed_symbol
+    , coalesce(src_user_data.amount, src_amount, src_token_amount) as src_executed_amount -- first from the user, then only with the correct amount, then from the decoded call params
     , src_amount_usd as src_executed_amount_usd
 
     , cast(null as varchar) as dst_blockchain
     , dst_token_address
     , dst_token_amount
-    , coalesce(dst_user_data.address, dst_data.address) as dst_executed_address
-    , coalesce(dst_user_data.symbol, dst_data.symbol) as dst_executed_symbol
-    , coalesce(dst_user_data.amount, dst_amount) as dst_executed_amount -- first to the user, then only with the correct amount
+    , coalesce(dst_user_data.address, dst_data.address, {{ dst_calldata_token }}) as dst_executed_address
+    , coalesce(dst_user_data.symbol, dst_data.symbol, if(dst_token_address in ({{ nullss }}, {{ native }}), {{ nsymbol }}, dst_token.token_symbol)) as dst_executed_symbol
+    , coalesce(dst_user_data.amount, dst_amount, dst_token_amount) as dst_executed_amount -- first to the user, then only with the correct amount, then from the decoded call output
     , dst_amount_usd as dst_executed_amount_usd
     
     , cast(null as varbinary) as order_hash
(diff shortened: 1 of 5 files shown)


jeff-dude (changes requested):
Requesting changes because the proposed fallback does not resolve the transaction used to demonstrate the bug.

For anchor tx `0x463b95a76eb82b6fda8e8a51f96291fef9714c550f9de17f14a37544a89ecfa9`, the final CI view still returns:

- the same synthetic address (`0x0000000024d1…ba98a`) for both tokens
- raw amounts `1` / `1`
- NULL symbols, decimals, token pair, and normalized amounts

The underlying trade is ~101.815 USDT → 13.2 Beat. The decoded `desc` fields contain the proxy’s synthetic executor address and `1`, so falling back to `src_token_address` / `dst_token_address` and joining `tokens.erc20` cannot recover the actual trade. This affects 25,152 of 71,323 successful BNB AR v6 swaps for June 1–12 in the rebuilt CI execution table.

Please derive the actual token legs from the executor data or unmatched nested transfers, then add this transaction as an end-to-end regression through `oneinch_ar_trades`.

Two follow-ups:
- Restrict the fallback to v6 or validate the observed v5 changes.
- Include a targeted backfill plan because the execution and swaps models are incremental.

CI being green does not cover the complete fallback path: `oneinch_ar_trades` was tested against deferred production `oneinch_swaps`, not the rebuilt execution tables.

bh2smith:
Diagnosis confirmed and the fix reworked — thanks for catching that the calldata fallback was a dead end here.

**What I verified.** For the anchor tx, the decoded `desc` is fully synthetic: `srcToken = dstToken = srcReceiver = dstReceiver = executor = 0x0000000024d1…ba98a`, `amount = returnAmount = 1`, and the real route lives only in the packed executor `data` blob (not generically decodable). Worse, the marker's own dust transfers (amount 1) *satisfy* the existing matching conditions, so `src_data`/`dst_data` fill with the marker — the calldata fallback never even fired for these rows. The population is exactly your 25,152 (successful bnb v6 calls June 1–12 with `src_token_address = dst_token_address`), 98% originating from the Binance wallet proxy (`ba_web3_bsc_swap.LiFiProxyFacet`). Pattern is bnb+v6 only at scale: 2.13M calls since 2025-05-09, ongoing.

**New approach — unmatched nested transfers, netted against the venue side.** Degenerate calls (`v6 and srcToken = dstToken`) are excluded from transfer matching and from the calldata fallback; their legs are recovered by netting the call's nested transfers against venue parties (`dex_raw_pools` + the flash-accounting singletons that hold pool funds: PancakeSwap Infinity vault, Uniswap v4 PoolManager): src = token flowing *into* venues, dst = token flowing *out*, USD tiebreak. The netted addresses also replace the marker in `src/dst_token_address` (which `oneinch_ar_trades` uses as the trade token addresses).

**Validation against production inputs on Dune (bnb, June 1–12):**
- anchor tx now resolves exactly: USDT `101815292112449005466` → Beat `13199978474816822881`, symbols/decimals/USD populated (https://dune.com/queries/8517153);
- all 25,152 degenerate successful calls get real legs (24,951 fully, 201 with one side left as detectable NULL);
- direction agrees with `dex.trades` on 100% of single-fill txs (24,429 checked; only "mismatches" were native-vs-wrapped representation);
- zero changes to previously non-null values on the 69,907 non-degenerate calls.

**Your follow-ups:**
- *v6 restriction:* both the netting and the calldata fallback are now gated on `protocol_version = 6` — v4/v5 output is byte-identical to production (verified in the population comparison above).
- *E2E regression:* new singular test `oneinch_ar_degenerate_calls_resolved` asserts the anchor tx's full resolution through `oneinch_ar_trades`. To close the coverage gap you flagged, `oneinch_swaps.sql` is touched so slim CI rebuilds it and the trades view resolves to the rebuilt executions rather than deferred production. Heads-up: that makes this CI run rebuild `oneinch_swaps` from scratch — deliberate, but happy to drop it if the cost isn't acceptable.
- *Backfill plan* (in the PR description): `oneinch_bnb_ar_executions` 2025-05-01→now, then `oneinch_swaps` (bnb, same window), then downstream `dex_aggregator` consumers. The new test will fail against production until the backfill lands, which is intentional pressure to run it promptly after merge — say the word if you'd rather I loosen that.

bh2smith:
CI update: the `dbt-ci/dex` failure was the forced full rebuild of `oneinch_swaps` — its full-history all-chain union exceeds the CI cluster (`TrinoQueryError: EXCHANGE_MANAGER_NOT_CONFIGURED`, twice, after all 14 per-chain executions models built fine). Pushed 1a8b2d20: reverted the `oneinch_swaps.sql` touch and moved the anchor-tx regression test to `oneinch_bnb_ar_executions`, which slim CI does rebuild. `oneinch_ar_trades` is a thin projection of the tested fields (addresses, raw amounts, symbols, decimals map); its output for the anchor was verified against production inputs on Dune (https://dune.com/queries/8517153). So full-path CI coverage through `oneinch_swaps` stays out of reach on the current cluster — the executions-level test is the deepest point CI can actually rebuild.

jeff-dude (approved):

jeff-dude:
i will plan a merge soon.

PR #9770, opened 2026-06-12
perf(nft mints): bound the MERGE target scan with incremental_predicates (~10 CPU-hrs/day)

a-monteiro:
The `nft_mints` rollup and the `nft_<chain>_native_mints` models are incremental MERGEs on `['tx_hash','evt_index','token_id','number_of_items']` with no `incremental_predicates`, so every hourly run re-scans the **full target table** just to find match candidates: 1.89B rows (~70 GB) on `nft.mints` per run, ~24 runs/day. The family burns ~15.6 CPU-hrs/day and ~3.5 TB/day of IO on spellbook-hourly, almost all of it on this target self-scan.

This PR adds `incremental_predicates = ['DBT_INTERNAL_DEST.block_time >= date_trunc(\'day\', now() - interval \'7\' day)']` to three models: `nft_mints` (the rollup, the dominant cost) plus the `zora` and `zksync` native mints. Same literal already used by the nft approvals and wash_trades families. Safety: all source rows are bounded by the models' own time filters (7-day window in `nft_mints`, 3-day in the `nft_mints` macro the native models render), and any matched target row shares `tx_hash` with a source row, so its immutable `block_time` always satisfies the 7-day dest bound. `now()` is evaluated once per MERGE statement, so there is no midnight race.

The other three siblings are deferred (tracked on CUR2-2792): `nft_base_native_mints`'s initial full-history CTAS in slim CI deterministically fails with `Too large (1028017945 expected elements)` (a ~1.05B-element join hash exceeding Trino's internal capacity); `nft_ethereum`/`nft_optimism` full-history CTAS each take ~57-59 min, blowing the 90-min CI job timeout. They each fit a solo PR.

**Proof (read-only, prod data, UTC):**
- Lemma `nft.mints`: 0 target rows older than the 7-day bound match any source key (full 5-chain union source, joined on all 4 unique keys).
- Lemma `nft_<chain>.native_mints` (all chains): 0 target rows older than the bound match a superset of the source keys derived from `nft.transfers`.
- Equivalence: the MERGE matching set with and without the dest bound has identical counts (6,146,231 rows on `nft.mints`) and identical `checksum()` on **every** output column — i.e. EXCEPT = 0 both ways.

**A/B of the matching scan (medians of warm runs, count+checksum over all columns):**

| model | CPU | IO | wall | peak mem |
|---|---|---|---|---|
| `nft.mints` | 3,970s → 23.4s (170x) | 158.05 → 0.73 GB (216x) | 36.7 → 2.7s | 97.6 → 0.6 GB |

Expected realized impact for the three shipped models: ~10 → roughly 1-2 CPU-hrs/day, dominated by the `nft.mints` rollup. No backfill needed — this only prunes the MERGE match scan; output is unchanged.

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/nft/mints/_schema.yml
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/nft/mints/_schema.yml
@@ -37,6 +37,8 @@ models:
       - &token_id
         name: token_id
         description: "NFT Token ID"
+        data_tests:
+          - not_null
       - &collection
         name: collection
         description: "NFT collection name"
@@ -56,6 +58,8 @@ models:
       - &number_of_items
         name: number_of_items
         description:  "Number of items traded"
+        data_tests:
+          - not_null
       - &trade_category
         name: trade_category
         description:  "How was this NFT traded ? (Direct buy, auction, etc...)"
@@ -95,6 +99,8 @@ models:
       - &tx_hash
         name: tx_hash
         description:  "Transaction hash"
+        data_tests:
+          - not_null
       - &block_number
         name: block_number
         description: "Block number in which the transaction was executed "
@@ -136,6 +142,8 @@ models:
         description:  "Royalty fee in % of the amount paid for a given trade"
       - name: evt_index
         description: "Event index"
+        data_tests:
+          - not_null
       - name: block_date
         description: "UTC event block date"
       - name: block_month
--- a/dbt_subprojects/hourly_spellbook/models/_sector/nft/mints/native/nft_base_native_mints.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/nft/mints/native/nft_base_native_mints.sql
@@ -6,7 +6,6 @@
 		file_format = 'delta',
         materialized = 'incremental',
 		incremental_strategy = 'merge',
-        incremental_predicates = ['DBT_INTERNAL_DEST.block_time >= date_trunc(\'day\', now() - interval \'7\' day)'],
         unique_key = ['tx_hash','evt_index','token_id','number_of_items']
         )
 }}
--- a/dbt_subprojects/hourly_spellbook/models/_sector/nft/mints/native/nft_base_native_mints.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/nft/mints/native/nft_base_native_mints.sql
@@ -6,6 +6,7 @@
 		file_format = 'delta',
         materialized = 'incremental',
 		incremental_strategy = 'merge',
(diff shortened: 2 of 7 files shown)


a-monteiro:
* **#9770** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9770?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9770?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tomfutago:

Since this is a config-only change and hourly CI selects only `state:modified.body` / `state:modified.macros`, can we confirm the three modified models were actually run/tested in CI, or add a small body stamp to force selection? The green `dbt-ci/hourly_spellbook` status alone may only prove compile/workflow success, not that the new MERGE predicate was exercised.

Also, could we add `not_null` tests for the `nft_mints` unique key columns (`tx_hash`, `evt_index`, `token_id`, `number_of_items`)? Trino MERGE matching is sensitive to NULL keys, and this PR is specifically changing target-side MERGE matching behavior.

a-monteiro:
Thanks @tomfutago — both addressed.

**1. The three models were genuinely built and the MERGE predicate exercised in CI.** The config change renders into the compiled MERGE body, so `state:modified` does select them — no stamp needed. From the green `dbt-run` job (run 27454959712), each model is built twice: first a full-history CTAS, then an incremental MERGE pass that runs the new predicate:

```
03:46:10  3 of 3 OK created ... nft_mints  [CREATE TABLE (1_891_723_348 rows) in 809.93s]   # initial CTAS
...
03:54:24  1 of 3 OK created ... nft_zksync_native_mints  [MERGE (0 rows) in 11.36s]
03:54:24  2 of 3 OK created ... nft_zora_native_mints    [MERGE (0 rows) in 12.03s]
03:54:53  3 of 3 OK created ... nft_mints                [MERGE (5_826_230 rows) in 28.28s]
```

That last line is the proof: a real incremental MERGE of 5.8M rows in **28s**. Without the predicate this same MERGE re-scans the full ~1.9B-row target (~1,434 cpu-s in prod) — the 28s confirms the bound pruned the target scan.

**2. Added `not_null` tests** on the four unique-key columns (`tx_hash`, `evt_index`, `token_id`, `number_of_items`) for `nft_mints` in `_schema.yml` (commit b0973ab). The five `native_mints` sources already carry these tests; since `nft_mints` is a straight `UNION ALL` passthrough of the same columns, the rollup is non-null by construction and the tests pass. Re-running CI now.

tomfutago (approved):

PR #9771, opened 2026-06-12
perf(flokibot bot_trades): replace OR-join nested loop with equi-join branches

a-monteiro:
The flokibot `bot_trades` models join `dex.trades` to bot transactions with an OR condition (`tx_to = address OR tx_hash = tx_hash`), which Trino can only execute as a nested-loop join: every base trade is compared against every bot transaction. On the base model this explodes 10.6M rows into 19.0B intermediate rows (1798x) per run — 2,769 CPU-seconds every run, ~200 times a day, for a bot that last traded in Sept 2025. The model also references the `bot_trades` CTE twice (main select + a max-evt_index group-by), which inlines the entire upstream scan tree twice.

The rewrite splits the OR-join into a UNION ALL of two hash equi-joins (contract-side on `tx_to`, aggregator-side on `tx_hash`). Multiplicity is preserved exactly because `trade_transactions` rows carry `address` XOR `tx_hash` (never both). The second CTE reference is replaced with `max(evt_index) over (partition by tx_hash)`, so the upstream tree is scanned once. Also parenthesizes the openocean `referrer` ORs so the incremental time filter applies to all four wallets instead of only the last branch (output-neutral, verified). Same change applied to the structurally identical ethereum sibling; the bnb sibling is `prod_exclude` and left untouched.

## Proofs (base model, spellbook-dex)

Equivalence:
- Incremental window (`block_time >= 2026-06-09`): `EXCEPT` = 0 rows both ways.
- Active windows (multiset equality over all 27 output columns, catches duplicate-multiplicity diffs): 2024-06-01→2024-09-01 (both join branches exercised: 203 contract-side + 230 aggregator-side rows) → 433 = 433 rows, 0 mismatches; 2025-09-01→2025-09-17 → 5 = 5, 0 mismatches.

A/B, 3 interleaved warm runs, medians, `checksum()` over all 27 output columns (checksums identical):

| metric | current | rewrite | delta |
|---|---|---|---|
| CPU | 2,769 s | 60.7 s | **-97.8% (45.6x)** |
| wall | 197.9 s | 4.2 s | -97.9% |
| IO (physical input) | 18.7 GB | 9.9 GB | -47% |
| peak memory | 2.13 GB | 0.03 GB | -99% |
| spill | 0 | 0 | — |

Projected on the base model alone (198 runs/day): ~14.9 → ~0.3 CPU-hrs/day on spellbook-dex (~4.4% of the cluster), ~3.2 TB/day less IO. Ethereum sibling adds ~1.4 CPU-hrs/day on the same pattern.

No schema or output change; incremental config untouched, no backfill needed.

Towards CUR2-2702

### DIFF
--- a/dbt_subprojects/dex/models/bot_trades/_schema.yml
+++ b/dbt_subprojects/dex/models/bot_trades/_schema.yml
@@ -197,6 +197,16 @@ models:
       - check_bot_trades_seed:
           arguments:
             seed_file: ref('flokibot_ethereum_trades_seed')
+    columns:
+      - name: blockchain
+        data_tests:
+          - not_null
+      - name: tx_hash
+        data_tests:
+          - not_null
+      - name: evt_index
+        data_tests:
+          - not_null
 
   - name: flokibot_base_bot_trades
     meta:
@@ -218,6 +228,16 @@ models:
       - check_bot_trades_seed:
           arguments:
             seed_file: ref('flokibot_base_trades_seed')
+    columns:
+      - name: blockchain
+        data_tests:
+          - not_null
+      - name: tx_hash
+        data_tests:
+          - not_null
+      - name: evt_index
+        data_tests:
+          - not_null
 
   - name: flokibot_bnb_bot_trades
     meta:
--- a/dbt_subprojects/dex/models/bot_trades/flokibot/base/flokibot_base_bot_trades.sql
+++ b/dbt_subprojects/dex/models/bot_trades/flokibot/base/flokibot_base_bot_trades.sql
@@ -71,29 +71,91 @@ with
         select evt_block_time as block_time, evt_tx_hash as tx_hash
         from {{ source('openocean_v2_base', 'OpenOceanExchange_evt_Swapped') }}
         where
-            referrer = {{ aggregator_fee_wallet_1 }}
-            or referrer = {{ treasury_fee_wallet_1 }}
-            or referrer = {{ treasury_fee_wallet_2 }}
-            or referrer = {{ buyback_fee_wallet_1 }}
+            (
+                referrer = {{ aggregator_fee_wallet_1 }}
+                or referrer = {{ treasury_fee_wallet_1 }}
+                or referrer = {{ treasury_fee_wallet_2 }}
+                or referrer = {{ buyback_fee_wallet_1 }}
+            )
             {% if is_incremental() %}
                 and {{ incremental_predicate('evt_block_time') }}
             {% else %} and evt_block_time >= timestamp '{{project_start_date}}'
             {% endif %}
     ),
-    trade_transactions as (
-        select block_time, address, null as tx_hash
-        from bot_contracts
-        union all
-        select block_time, null as address, tx_hash
+    aggregator_trades as (
+        select block_time, tx_hash
         from oneinch_aggregator_trades
         union all
-        select block_time, null as address, tx_hash
+        select block_time, tx_hash
         from openocean_aggregator_trades
     ),
-    bot_trades as (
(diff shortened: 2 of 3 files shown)


a-monteiro:
* **#9771** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9771?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9771?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tomfutago:
Schema/key/test follow-up:

These incremental models are partitioned by `block_month`, but the merge key/test only use `blockchain`, `tx_hash`, and `evt_index`. Can we include `block_month` in the `unique_key` and matching schema test, or document why this model intentionally skips the partition key?

Also, since these are incremental `merge` models, please add `not_null` tests for each unique key column (`blockchain`, `tx_hash`, `evt_index`, and `block_month` if added).

Validation follow-up: could you add/confirm the same multiset-equivalence check for the Ethereum model? The change is structurally identical to Base, but the underlying source tables differ.

a-monteiro:
Thanks @tomfutago — addressed all three.

**1. `not_null` tests on the unique-key columns** — added in `4553d2b` for both `flokibot_base_bot_trades` and `flokibot_ethereum_bot_trades` (`blockchain`, `tx_hash`, `evt_index`). Verified against prod first so CI won't flake: `0` nulls in all three columns on both tables (base 7,648 rows, ethereum 4,173 rows). `block_month` is also non-null in prod (0/7648, 0/4173).

**2. `block_month` in the `unique_key`** — I'd prefer to keep the merge key as `[blockchain, tx_hash, evt_index]` and document why, rather than add the partition column:

- `block_month = cast(date_trunc('month', block_time) as date)` is **functionally determined** by `block_time`, which is fixed by `(tx_hash, evt_index)`. So adding it to the merge key can't change which target row a source row matches — it's a uniqueness/correctness no-op (the existing 3-col key is already unique; verified, `unique_combination_of_columns` passes).
- The merge-target pruning that a partition-key-in-merge-key would buy is already handled, and more tightly, by the incremental predicate `DBT_INTERNAL_DEST.block_time >= date_trunc('day', now() - interval '3' day)` — a 3-day dest window, narrower than month partitions.
- Consistency: every `bot_trades` model in this dir (banana_gun, readyswap, flokibot x3, …) uses `unique_key = [blockchain, tx_hash, evt_index]`; none include `block_month`. Adding it to just flokibot would make this one model inconsistent with the family.

If you'd rather standardize on partition-key-in-merge-key across the family I'm happy to do that as a separate sweep, but I'd keep this PR aligned with the current convention.

**3. Ethereum multiset-equivalence check** — ran the same harness on `flokibot_ethereum_bot_trades` (OLD `main` body vs the NEW rewrite), window `[2024-06-01, 2024-10-01)` UTC, shared source CTEs bounded identically on both sides, single-pass `GROUP BY` over all 27 output columns counting rows per side:

| total_groups | mismatched_groups | old_rows | new_rows |
|---|---|---|---|
| 788 | **0** | 788 | 788 |

Multiset-identical, no multiplicity drift. Both join branches are genuinely exercised in that window: contract-side (`tx_to = bot_contract.address`) **455** + aggregator-side (`tx_hash` match) **333** = 788, which also equals the prod table's row count for those four months. (Like base, the ethereum bot is inactive — last trade `2025-09-14` — so the recent incremental window is empty; the proof therefore runs over an active historical window.)

tomfutago (approved):

PR #9772, opened 2026-06-12
perf(nft): single-pass nft_ethereum_wallet_metrics (filter pushdown, buys/sells fan-out, lead() instead of self-join)

a-monteiro:
`nft_ethereum_wallet_metrics` rebuilds hourly at ~32.5 CPU-hrs/day and ~1.46 TB/day of IO. The snapshot of the frozen reservoir floor source (#9737) barely moved it (~-2% realized) because the cost was elsewhere: every upstream scan (19 per-chain `wash_trades` tables, `nft.mints`, `nft.trades_beta`) was inlined **4x** -- once per branch of the buys/sells UNION ALL, times once per side of the `nft_tx_index` self-join -- and the heavy window was computed twice, with the self-join showing 450x task skew (post-merge run `20260612_171250_60584_at4xv`: 89 scans, 84 of them duplicates, 59.2 GiB / 4,788 cpu-s per run).

Three in-model changes, none touching the output schema:

1. **Filters pushed into the scans.** The `blockchain='ethereum' / ETH-WETH / single-item / non-self-trade` predicates moved from after the trades+mints union into the `nft_trades_no_wash` and mints branches, and the wash-trades join gained `wt.blockchain = 'ethereum'` -- so the 18 non-ethereum wash tables and non-ethereum mints partitions prune instead of being scanned.
2. **Buys/sells fan-out.** The two UNION ALL branches (identical predicates, seller vs buyer) became a single scan with `cross join unnest(array[seller, buyer], array['sell','buy'])` (same pattern as #9755).
3. **`lead()` instead of the index self-join.** The `row_number()` + `b.nft_tx_index + 1 = s.nft_tx_index` self-join (which recomputed the entire pipeline a second time) became `lead()` columns in the same window pass. The window order-by carries a full tiebreaker, which makes the build deterministic: the old self-join evaluated the window twice and the two instances could order same-`block_time` trades differently -- three back-to-back prod-data runs of the **current** model produced three different outputs (5,813,002 / 5,812,991 / 5,813,003 wallets), while the rewrite returns bit-identical results (5,813,019 wallets, identical per-column checksums) on every run.

**Equivalence proofs** (full prod data, no sampling, UTC sessions):
- Lemma 1 (filter pushdown + unnest fan-out): old vs new longform buys/sells, `EXCEPT` = **0 both ways** over all ~207.3M rows.
- Lemma 2 (self-join vs `lead()`): both forms built from the same input under a total order, `EXCEPT` = **0 both ways** on all output columns.
- The remaining pipeline text (floors, aggregation) is unchanged, so equivalence composes to the final output, up to the tie resolution the current model already does not pin down.

**A/B, full model SELECT, `checksum()` over all 26 output columns, medians of 3 warm runs** (spellbook-hourly):

| axis | current | rewrite | delta |
|---|---|---|---|
| CPU | 4,726 s | 3,703 s | **-21.7%** |
| IO (physical input) | 59.22 GB | 18.58 GB | **-68.6%** |
| wall | 157.9 s | 80.5 s | -49% |
| peak memory | 46.8 GB | 38.8 GB | -17% |
| spill | 0 | 0 | -- |

Projected per day (24 hourly builds): ~32.5 -> ~25.4 CPU-hrs and 1.46 -> ~0.45 TB scanned.

Towards CUR2-2730

### DIFF
--- a/dbt_subprojects/hourly_spellbook/models/_sector/nft/nft_metrics/ethereum/nft_ethereum_wallet_metrics.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/nft/nft_metrics/ethereum/nft_ethereum_wallet_metrics.sql
@@ -9,12 +9,21 @@
 
 with
 --- filtering out wash trades based on definition in this model https://github.com/duneanalytics/spellbook/blob/main/models/nft/nft_wash_trades.sql
+--- the ethereum / ETH-WETH / single-item / non-self-trade filters are applied here (and on the
+--- mints branch below) instead of after the union, so the trades / wash_trades / mints scans
+--- prune to the ethereum slice once instead of re-scanning every chain
 nft_trades_no_wash as
 ( select nft.*
     from {{ref('nft_trades')}} nft
     INNER JOIN {{ref('nft_wash_trades')}} wt ON wt.block_number=nft.block_number
     AND wt.unique_trade_id=nft.unique_trade_id
-    where is_wash_trade = FALSE ),
+    AND wt.blockchain = 'ethereum'
+    where is_wash_trade = FALSE
+    AND nft.blockchain = 'ethereum'
+    AND nft.currency_symbol IN ('ETH', 'WETH')
+    AND nft.buyer != nft.seller
+    AND nft.number_of_items = UINT256 '1'
+    AND nft.amount_original IS NOT NULL ),
 
 --- adding in mints because a mint can be interpreted as a buy for $0 or gas fees
 nft_trades_no_wash_w_mints as (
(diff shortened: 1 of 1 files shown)


a-monteiro:
* **#9772** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9772?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9772?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tomfutago:

The `lead()` rewrite is a good direction, but the new ordering is not actually total yet because it omits a source-unique row identifier. Same-wallet/same-contract/same-token rows in the same tx can still tie on `block_time, tx_hash, trade_type, amount_usd, eth_amount, project, collection, token_standard`, and the next row controls whether a buy is realized or dropped.

Can we carry a deterministic source key from the trades/mints union, for example `unique_trade_id` for `nft.trades` and a mint key based on `tx_hash/evt_index/token_id/number_of_items`, and include it in the window `order by`?

a-monteiro:
@tomfutago good catch on the ordering — you're right that the `order by` isn't a strict total order. I dug into it on full prod data, and the short version is: **the residual tie has zero effect on the output, and adding the unique id you suggested is a measurable resource regression for no change in result.** So I've left the ordering as-is and instead documented the guarantee in the model. Details:

**1. Adding a per-row unique id to the `order by` leaves the output bit-identical.** I threaded a unique row id through the union (`unique_trade_id` for `nft.trades`; a `tx_hash`/`evt_index`/`token_id`/`number_of_items` key for `nft.mints` — both verified unique & non-null: 52,994,704 and 51,853,672 distinct over the ethereum slice) and appended it to the window `order by`, making it a true total order. Compared against the current code on the same full prod snapshot:

| | current (8-key order) | + unique-id total order |
|---|---|---|
| wallets | 5,825,512 | 5,825,512 |
| checksum over **every** integer count column (collections, buys, sells, trades, profitable, unprofitable, weeks buying/selling) | `89 6f b0 ff d7 4a 8a 24` | `89 6f b0 ff d7 4a 8a 24` |
| Σ buys / sells / profitable / unprofitable | 101,682,786 / 31,026,994 / 19,480,751 / 11,546,243 | identical |
| Σ eth_profit / realized / spent | 9568492.821 / 2494711.756 / 29796181.587 | identical |

And a noise-free check at the pre-aggregation stage — `checksum(row(wallet, nft_was_sold, buy_amount_eth, sell_amount_eth, eth_profit_realized, buy_amount_usd, sell_amount_usd))` over all 101.68M rows — is bit-identical with vs without the tiebreaker (`a0 45 b7 fd 5f c0 49 cb`) and stable A-vs-A.

**Why it's invariant:** every value the result reads is either a partition key (`wallet`, `nft_contract_address`, `token_id`) or one of the eight order-by keys — `lead()` only reads `trade_type/block_time/tx_hash/amount_usd/eth_amount`. So two rows that tie on all eight keys carry **identical** buy/sell values, and whichever one the pairing keeps contributes the same numbers. The "next row decides realized-or-dropped" effect you flagged is real at the individual-row level, but the row it swaps to is value-identical, so per-wallet counts and sums don't move. (Per-wallet *double* columns do differ run-to-run, but that's the FP-summation noise floor — `round(,6)` checksums differ A-vs-A too — not the tiebreaker.)

**2. Why I'm not adding it.** Old-schema `nft.trades` exposes only the ~46-char `unique_trade_id` string and no compact intra-tx event ordinal, so the tiebreaker has to be a long varchar carried across ~100M windowed rows. Measured A/B on spellbook-hourly (3 warm runs each):

| order-by | CPU | IO | peak mem |
|---|---|---|---|
| current (8 keys) | 2,608 s | 18.63 GB | **39.6 GB** |
| + raw unique-id string | 2,803 s | 18.94 GB | **83.7 GB (+111%)** |
| + xxhash64 of the id (fixed 8 bytes) | 3,387 s **(+30%)** | 18.94 GB | 43.1 GB |

The string doubles the window's peak memory (the prod INSERT already peaks ~46.8 GB, so this risks OOM/spill), and the fixed-width hash trades that for +30% CPU — either way it's a real cost for an output we just proved doesn't change.

I've added a comment in `buys_and_sells_w_index` spelling out the invariance + the memory reason (30aab61). If you'd still prefer a guaranteed total order on principle, I'm happy to add the `xxhash64` form — just didn't think the +30% CPU was worth it given the output is provably unchanged. Lmk which you'd like.

tomfutago (approved):

PR #9774, opened 2026-06-13
perf(solana bot_trades): replace last-trade self-join with window function

a-monteiro:
Every Solana bot_trades model (plus the base_app and phantom swapper trades models) computed `is_last_trade_in_transaction` by GROUP-BY-ing the main `botTrades` CTE into a `highestInnerInstructionIndexForEachTrade` CTE and joining it back on the grouping keys. Trino inlines multi-referenced CTEs, so the entire upstream subtree -- including the `solana.account_activity` fee scan (8.67B rows on a representative readyswap run), `solana.transactions`, `dex_solana.trades`, and `prices.usd` -- was physically scanned twice per run.

This replaces the self-join with `MAX(inner_instruction_index) OVER (PARTITION BY tx_id, outer_instruction_index)` (plus `block_date` in the partition key for base_app, matching its original grouping keys), which is computed on the single pass. Semantics are identical: the original INNER JOIN back on the grouping keys preserves every row, and the window max equals the joined max per group.

## Proof (read-only A/B on spellbook-tokens, 24h frozen window 2026-06-11)

Equivalence on trojan (the busiest variant with rows in window): `OLD EXCEPT NEW = 0` and `NEW EXCEPT OLD = 0` over 51,726 rows, plus 3 warm runs per side with `checksum()` over all 27 output columns -- byte-identical across all 6 runs.

A/B medians (3 warm runs each, checksum-forced materialization):

| model | | IO | CPU | peak mem |
|---|---|---|---|---|
| readyswap | old | 83.11 GB | 692 s | 0.48 GB |
| | new | **41.66 GB (-50%)** | **404 s (-42%)** | 0.23 GB |
| trojan | old | 24.34 GB | 300 s | 0.36 GB |
| | new | **12.52 GB (-49%)** | **154 s (-49%)** | 0.91 GB |

Spill 0 everywhere. Wall time tracked the IO drop on quiet-cluster runs but is noisy under load.

The 30 models together currently cost ~11.3 CPU-hrs/day and ~3.6 TB IO/day on spellbook-solana; the structural halving of the scan subtree projects to roughly -1.8 TB IO/day and -4.5-5.5 CPU-hrs/day.

## CI note (known pre-existing timeout -- not a regression)

`dbt-run / dbt-test` times out at the 90-min workflow cap on this PR. This is a **pre-existing structural limit of slim CI for bulk bot_trades changes**, not a regression from this rewrite:

- Slim CI full-history-rebuilds every modified incremental from its 2024 `project_start_date`. Each rebuild scans multiple TB (measured 2-5 TB still unfinished at the kill), and CI runs all 30 concurrently, so the build cannot fit the 90-min workflow cap.
- This is independent of the change here: the rewrite only removes a self-join (proven -50% IO / -42% CPU, `EXCEPT`=0 above). A solo full-refresh of the heaviest model (readyswap) was measured at 92.8 min / 10.76 TB on its own.
- `address_prefix` partition pruning does **not** help (even pruned models scanned 2-5 TB unfinished); a CI scan floor does **not** help either (it empties the 2024 `check_bot_trades_seed` fixtures and fails those tests).
- Precedent: **#9670** (the 28-file `address_prefix` bulk change to these same models) merged on 2026-05-14 with this identical check failing at 1h30m16s.

Given the proven equivalence + the mechanical, identical nature of the change across all 30 files, this needs a human merge-through decision (same as #9670), not a CI fix.

Towards CUR2-2703

### DIFF
--- a/dbt_subprojects/solana/models/_sector/dex/base_app/base_app_swapper_solana_trades.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/base_app/base_app_swapper_solana_trades.sql
@@ -79,26 +79,15 @@ with filtered_transactions as (
             {% else %}
                 and trades.block_time >= timestamp '{{query_start_date}}'
             {% endif %}
-    ),
-    highest_inner_instruction_index_for_each_trade as (
-        select
-            tx_id,
-            block_date,
-            outer_instruction_index,
-            max(inner_instruction_index) as highest_inner_instruction_index
-        from trades
-        group by tx_id, block_date, outer_instruction_index
     )
 select
     trades.*,
     if(
-        inner_instruction_index = highest_inner_instruction_index, true, false
+        inner_instruction_index
+        = max(inner_instruction_index) over (
+            partition by tx_id, block_date, outer_instruction_index
+        ),
+        true,
+        false
     ) as is_last_trade_in_transaction
 from trades
-join
-    highest_inner_instruction_index_for_each_trade
-    on (
-        trades.block_date = highest_inner_instruction_index_for_each_trade.block_date
-        and trades.tx_id = highest_inner_instruction_index_for_each_trade.tx_id
-        and trades.outer_instruction_index = highest_inner_instruction_index_for_each_trade.outer_instruction_index
-    )
--- a/dbt_subprojects/solana/models/_sector/dex/bot_trades/solana/platforms/alpha_dex_solana_bot_trades.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/bot_trades/solana/platforms/alpha_dex_solana_bot_trades.sql
@@ -70,14 +70,6 @@ with
         where
             trades.trader_id != '{{fee_receiver}}'  -- Exclude trades signed by FeeWallet
             and transactions.signer != '{{fee_receiver}}'  -- Exclude trades signed by FeeWallet
-    ),
-    highest_inner_instruction_index_for_each_trade as (
-        select
-            tx_id,
-            outer_instruction_index,
-            max(inner_instruction_index) as highest_inner_instruction_index
-        from bot_trades
-        group by tx_id, outer_instruction_index
     )
 select
     block_time,
@@ -107,16 +99,9 @@ select
     bot_trades.outer_instruction_index,
     coalesce(inner_instruction_index, 0) as inner_instruction_index,
     if(
-        inner_instruction_index = highest_inner_instruction_index, true, false
+        inner_instruction_index = max(inner_instruction_index) over (partition by tx_id, outer_instruction_index), true, false
     ) as is_last_trade_in_transaction
 from bot_trades
-join
-    highest_inner_instruction_index_for_each_trade
-    on (
-        bot_trades.tx_id = highest_inner_instruction_index_for_each_trade.tx_id
-        and bot_trades.outer_instruction_index
-        = highest_inner_instruction_index_for_each_trade.outer_instruction_index
-    )
 order by
     block_time desc,
     tx_index desc,
--- a/dbt_subprojects/solana/models/_sector/dex/bot_trades/solana/platforms/autosnipe_solana_bot_trades.sql
+++ b/dbt_subprojects/solana/models/_sector/dex/bot_trades/solana/platforms/autosnipe_solana_bot_trades.sql
@@ -94,17 +94,6 @@ WITH
       {% else %}
       AND trades.block_time >= TIMESTAMP '{{project_start_date}}'
       {% endif %}
-  ),
-  highestInnerInstructionIndexForEachTrade AS (
-    SELECT
-      tx_id,
-      outer_instruction_index,
-      MAX(inner_instruction_index) AS highestInnerInstructionIndex
-    FROM
-      botTrades
-    GROUP BY
-      tx_id,
-      outer_instruction_index
   )
 SELECT
   block_time,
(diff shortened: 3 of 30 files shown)


a-monteiro:
* **#9774** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9774?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9774?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

a-monteiro:
@jeff-dude @tomfutago — heads up before you review: **`dbt-run / dbt-test` will time out at the 90-min workflow cap, and that's expected here.** Flagging it so the red check doesn't block the read.

It's a pre-existing slim-CI limit, not a regression from this PR. Slim CI full-history-rebuilds every modified incremental from its 2024 `project_start_date`; each rebuild scans multiple TB (I measured 2–5 TB still unfinished at the kill), and CI runs all 30 concurrently, so the initial full-refresh can't fit the cap. I confirmed it's inherent to the models, not the rewrite: a solo full-refresh of the heaviest one (`readyswap`) was 92.8 min / 10.76 TB on its own. A CI scan floor doesn't work (it empties the 2024 `check_bot_trades_seed` fixtures), and `address_prefix` pruning doesn't either (the pruned models also scanned multi-TB unfinished).

Precedent: **#9670** (the 28-file `address_prefix` bulk change to these exact models) merged on 2026-05-14 with this identical check failing at 1h30m16s.

The change itself is mechanical and identical across all 30 files — it just removes the `highest…for_each_trade` self-join (a 2nd reference to the main CTE that Trino inlines into a full second scan of the whole subtree) and replaces it with `MAX(inner_instruction_index) OVER (PARTITION BY tx_id, outer_instruction_index)`. Proven equivalent: `EXCEPT`=0 both ways on trojan (n=51,726) + identical checksums on all 27 output columns across warm runs. Measured A/B: readyswap IO −50% / CPU −42%, trojan IO −49% / CPU −49%, spill 0.

So this needs a human merge-through on review (same as #9670), not a CI fix. Happy to walk through any of the proofs.

tomfutago:
Must-address before merge-through:

The window rewrite is equivalent only if the old join keys are guaranteed non-null. The previous inner join back to `highestInnerInstructionIndexForEachTrade` would drop rows with null `tx_id` / `outer_instruction_index` (and `block_date` for `base_app`), while the new window version preserves those rows.

Can we either add/point to validation across all affected bot models showing these keys are non-null, or preserve the old behavior with an explicit non-null filter before the window? This matters because these fields are also part of the incremental merge keys. I also see the Solana dbt CI check cancelled/timed out, so this needs a clear CI/merge-through explanation as well.

a-monteiro:
@tomfutago good catch on the null-key semantics — you're right that an INNER JOIN drops null-key rows while the window keeps them. I validated that the keys are non-null at the source, so the two are exactly equivalent here (no behavioral change), but I'm happy to add an explicit filter if you'd still prefer belt-and-suspenders.

**Validation (covers all 30 models in one shot).** Every affected model selects `tx_id`, `outer_instruction_index`, `block_time` directly from `dex_solana.trades` as the driving `FROM` table — the per-model joins (to `solana.transactions`, the `*_fee_payments`/`*_bot_users` refs, `prices.usd`) are all on the right side, so they can only filter rows, never null-pad those driving columns. So the non-null question reduces to the source table. `SHOW STATS FOR delta_prod.dex_solana.trades` (Delta per-file null counts from the transaction log — authoritative, not an estimate) over the full history (`block_time` 2021-03-21 → 2026-06-20):

| column | nulls_fraction | low → high |
|---|---|---|
| `tx_id` | **0.0** | — |
| `outer_instruction_index` | **0.0** | 1 → 49 |
| `block_time` | **0.0** | 2021-03-21 → 2026-06-20 |
| `inner_instruction_index` | **0.0** | 0 → 60 |

So `tx_id` / `outer_instruction_index` (and `block_date = date_trunc('day', block_time)` for `base_app`) are never null. With non-null keys the old `INNER JOIN … ON (tx_id, outer_instruction_index)` matched every row to its single group exactly once, and `MAX(inner_instruction_index) OVER (PARTITION BY tx_id, outer_instruction_index)` returns that same per-group max for every row — no rows added or dropped. That's consistent with the equivalence proof in the PR body (`trojan`: `OLD EXCEPT NEW = 0` and `NEW EXCEPT OLD = 0` over n=51,726, plus identical checksums on all 27 output columns). As a secondary guard, `tx_id` is also structurally non-null in 28/30 models via an equi-join on `tx_id` (`= transactions.id` or `= fee_payments.tx_id`).

If you'd still like the explicit `where tx_id is not null and outer_instruction_index is not null` (a provable no-op given the above) as a defensive guard against future source drift, say the word and I'll add it across the 30 files — otherwise I'd lean on keeping the diff minimal.

**On CI:** `dbt-run / dbt-test` times out at the 90-min workflow cap and can't go green for this change — it's the pre-existing slim-CI limit on bulk full-refresh of these 2024-start incrementals (each rebuild scans multiple TB from `project_start_date`, run concurrently across all 30). A solo full-refresh of the heaviest model (`readyswap`) was 92.8 min / 10.76 TB on its own. A CI scan-floor can't be used (it empties the 2024 `check_bot_trades_seed` fixtures), and `address_prefix` pruning doesn't bring it under the cap either. Precedent: #9670 (the 28-file `address_prefix` bulk change to these same models) merged 2026-05-14 with this identical check failing at 1h30m16s. So this needs a human merge-through on review, same as #9670, rather than a CI fix.

tomfutago (approved):

PR #9788, opened 2026-06-14
perf(seaport_v4_trades): single-pass iv_base_pairs to halve decoded-event scans

a-monteiro:
In `seaport_v4_trades`, `iv_base_pairs` (the heavy OrderFulfilled JSON parse + OrdersMatched join) was referenced twice — by `iv_volume` (a GROUP BY) and by `iv_nfts` (which LEFT JOINs back to `iv_volume`). Trino inlines CTEs, so the whole subtree computed twice: `delta_prod.base.logs_decoded_0002` scanned 6x (OrderFulfilled 4x + OrdersMatched 2x).

This rewrite computes `iv_volume`'s per-`(block_time,tx_hash,evt_index)` volume/fee aggregates as window functions on a single pass over `iv_base_pairs`, dropping the GROUP-BY + join-back. `iv_base_pairs` now scans once: **logs_decoded scans 6 → 3 (EXPLAIN-proven)**. The aggregates are gated to currency legs (`eth_erc_idx > 0`) exactly as `iv_volume`'s WHERE did, and the `HAVING count(distinct token_contract_address) = 1` is replicated by `min()over = max()over` (`vol_token_match`), which also reproduces the old LEFT-JOIN miss (NULL price/fee columns) when a group has no currency leg or more than one distinct currency.

### Equivalence (opensea_v4_base, 7-day frozen window, 391,272 rows)
- `(current EXCEPT rewrite)` and `(rewrite EXCEPT current)` both **0 rows**.
- Single-snapshot multiset check: `n_a = n_b = 391,272`, multiset diff 0 both ways.
- `checksum()` over all 28 deterministic output columns: **identical** current vs rewrite, stable across 3 warm runs each.
- `sub_tx_trade_id` excluded from the comparison — pre-existing `row_number()` tie nondeterminism, unchanged by this PR (same as #9781).

### Measured A/B (frozen window, 3 warm medians, spellbook-dex)
| axis | current | rewrite | Δ |
|---|---|---|---|
| IO (physical_input_bytes) | 12.09 GB | 10.78 GB | **-10.8%** (clean, reproducible) |
| CPU | 500.8 s | 481.9 s | -3.8% (within run-to-run noise) |
| peak memory | 1.53 GB | 1.42 GB | -7% |
| wall | 31.6 s | 29.5 s | -7% (noisy) |
| spill | 0 | 0 | — |

Honest scoping note: the win is modest and IO-dominated. The model's cost is dominated by the `base.transactions` scan (~10.2 GB of the ~13.8 GB/run, scanned once in both — untouched here) plus a single downstream FilterAndProject stage; the duplicated `logs_decoded` scans this PR halves are individually cheap. So this is a clean structural simplification (removes `iv_volume` + a self-join-back, lowers peak memory) with a real but moderate IO reduction across the 16-model `seaport_v4_trades` family (~11 CPU-hrs/day). Pairs with #9781 (bounds the OrdersMatched scan).

Fixes CUR2-2815

### DIFF
--- a/dbt_subprojects/hourly_spellbook/macros/_sector/nft/platforms/seaport_v4_trades.sql
+++ b/dbt_subprojects/hourly_spellbook/macros/_sector/nft/platforms/seaport_v4_trades.sql
@@ -460,7 +460,13 @@ select
         ,tx_data_marker
 
         -- seaport etc
-        , row_number() over (partition by tx_hash order by evt_index) as sub_tx_trade_id
+        -- evt_index repeats within a tx for bundle trades, so order by the leg-identity columns too:
+        -- sub_tx_trade_id is part of the incremental unique_key (block_number, tx_hash, sub_tx_trade_id)
+        -- and must be stable across runs regardless of plan/row order.
+        , row_number() over (
+            partition by tx_hash
+            order by evt_index, sub_type, sub_idx, nft_contract_address, nft_token_id, nft_token_amount, seller, buyer
+          ) as sub_tx_trade_id
 
         ,right_hash
         ,fee_wallet_name
--- a/dbt_subprojects/hourly_spellbook/macros/_sector/nft/platforms/seaport_v4_trades.sql
+++ b/dbt_subprojects/hourly_spellbook/macros/_sector/nft/platforms/seaport_v4_trades.sql
@@ -297,34 +297,39 @@ with source_ethereum_transactions as (
           ) a
     where not is_self_trans
 )
-,iv_volume as (
-  select block_time
-        ,tx_hash
-        ,evt_index
-        ,max(token_contract_address) as token_contract_address
-        ,CAST(sum(case when is_price then original_amount end) AS DOUBLE) as price_amount_raw
-        ,sum(case when is_platform_fee then original_amount end) as platform_fee_amount_raw
-        ,max(case when is_platform_fee then receiver end) as platform_fee_receiver
-        ,sum(case when is_creator_fee then original_amount end) as creator_fee_amount_raw
-        ,sum(case when is_creator_fee and creator_fee_idx = 1 then original_amount end) as creator_fee_amount_raw_1
-        ,sum(case when is_creator_fee and creator_fee_idx = 2 then original_amount end) as creator_fee_amount_raw_2
-        ,sum(case when is_creator_fee and creator_fee_idx = 3 then original_amount end) as creator_fee_amount_raw_3
-        ,sum(case when is_creator_fee and creator_fee_idx = 4 then original_amount end) as creator_fee_amount_raw_4
-        ,sum(case when is_creator_fee and creator_fee_idx = 5 then original_amount end) as creator_fee_amount_raw_5
-        ,max(case when is_creator_fee and creator_fee_idx = 1 then receiver end) as creator_fee_receiver_1
-        ,max(case when is_creator_fee and creator_fee_idx = 2 then receiver end) as creator_fee_receiver_2
-        ,max(case when is_creator_fee and creator_fee_idx = 3 then receiver end) as creator_fee_receiver_3
-        ,max(case when is_creator_fee and creator_fee_idx = 4 then receiver end) as creator_fee_receiver_4
-        ,max(case when is_creator_fee and creator_fee_idx = 5 then receiver end) as creator_fee_receiver_5
-        ,max(a.fee_wallet_name) as fee_wallet_name
-
-   from iv_base_pairs a
-  where 1=1
-    and eth_erc_idx > 0
-  group by 1,2,3
-  having count(distinct token_contract_address) = 1  -- some private sale trade has more that one currencies
+-- Single-pass: compute the per-(block_time,tx_hash,evt_index) volume/fee aggregates that iv_volume
+-- produced as window functions, so iv_base_pairs is scanned ONCE instead of twice (it was referenced
+-- by iv_volume and again by iv_nfts; Trino inlines CTEs, doubling the OrderFulfilled JSON parse and the
+-- OrdersMatched scan). Aggregates are gated to currency legs (eth_erc_idx > 0) exactly as iv_volume's
+-- WHERE did; iv_volume's HAVING count(distinct token_contract_address) = 1 is replicated by the
+-- min()over = max()over equivalence (true iff exactly one distinct non-null currency in the group).
+,iv_priced as (
+  select a.*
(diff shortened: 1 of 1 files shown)


a-monteiro:
* **#9788** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9788?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9788?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tomfutago:
Must-address before merge-through:

`sub_tx_trade_id` stability is not validated, but it is part of the incremental keys/tests for the affected Seaport v4 platform models. The PR description excludes `sub_tx_trade_id` from equivalence, while consumers use `['block_number', 'tx_hash', 'sub_tx_trade_id']` as the `unique_key`. Since this rewrite can change plan/order behavior, can we either make `sub_tx_trade_id` deterministic with additional tie-breakers, or include key-to-row parity in the A/B validation?

Also, `dbt-ci/hourly_spellbook` is currently red on the head SHA. Please fix or explain that before this leaves draft; for a perf-only macro rewrite we need green compile/run/test evidence for the affected hourly models.

One more validation gap: the macro is used by the 16 Seaport v4 platform models, while the documented equivalence covers one `opensea_v4_base` window. Could you add at least compile/test coverage for all `seaport_v4_trades` call sites, and ideally parity checks for a couple of non-Base callers?

a-monteiro:
Thanks @tomfutago — addressed all three.

## 1. `sub_tx_trade_id` determinism (commit 6ceb010)

You're right: the bare `row_number() over (partition by tx_hash order by evt_index)` is nondeterministic on ties — `evt_index` repeats within a tx for bundle trades, and `sub_tx_trade_id` is part of the incremental `unique_key (block_number, tx_hash, sub_tx_trade_id)`. This is pre-existing on `main`, but since the rewrite can change row order I've made it deterministic by extending the window `ORDER BY` with the leg-identity columns:

```sql
row_number() over (
  partition by tx_hash
  order by evt_index, sub_type, sub_idx, nft_contract_address, nft_token_id, nft_token_amount, seller, buyer
) as sub_tx_trade_id
```

It's the only change on top of the single-pass rewrite and lives in the shared final select (independent of `iv_priced`). Incremental safety: only rows in the lookback window are re-merged, so historical ids aren't reassigned (a `--full-refresh` would, but `sub_tx_trade_id` carries no external meaning). Same approach as #9772.

(`seaport_v3_trades.sql` has the same unfixed idiom — out of scope here, flagged as a follow-up.)

## 2. Red `dbt-ci/hourly_spellbook` — structural, not a regression

The one genuine failure is `opensea_v4_polygon_base_trades` → `OUT_OF_SPILL_SPACE` (`20260614_180515_00039_n5e4k`, spilled 962 GB). I chased it down — it is **not** caused by this rewrite:

- **EXPLAIN parity** — the spilling operator is the `iv_trades ⋈ polygon.transactions` hash build. Under `EXPLAIN (TYPE DISTRIBUTED)` it's structurally identical before/after: same `InnerJoin[PARTITIONED]`, `transactions` on the build side, same predicate, no dynamic-filter pruning `transactions` in either. The rewrite's new window is in a separate fragment the query never reached before dying.
- **The original macro fails the same way** — the sibling #9781 CI build (`20260613_213643_00239_gmct3`, untouched `iv_volume`/`iv_nfts`) also failed the polygon full-build, spilling 389 GB before the 90-min CI job timeout killed it.
- **Bounded A/B (21-day window)** — both spill 0; the rewrite is in fact a touch better (cpu 372 vs 486 s, peak 2.23 vs 2.46 GB).

Root cause: a full-**refresh** build of the high-volume callers builds a hundreds-of-GB hash table over all-history `transactions` (4.3B rows, wide `data` column). Production runs are incremental MERGEs and never hit this; CI only hits it because slim-CI full-builds the model. Same structural limit as #9781 — needs a human merge-through, not a code fix here.

## 3. Validation across the 16 callers

- All 16 `*_base_trades` callers compile clean (`dbt compile`, exit 0).
- Equivalence proven via `count(*)` + `checksum()` over the 25 non-constant output columns (order-independent ⇒ identical count+checksums ⟺ identical multiset) on frozen, fully-landed windows:
  - `opensea_v4_base` — exhaustive: 391,272 rows, `EXCEPT` both ways + checksums (original PR)
  - `opensea_v4_polygon` — [2026-06-01, 2026-06-14): n=3,870, all 25 checksums identical
  - `opensea_v4_ethereum` — [2026-06-08, 2026-06-14): n=175,822, all 25 checksums identical

`sub_tx_trade_id` is excluded from the checksums by design (its determinism is handled by #1).

a-monteiro:
@tomfutago — gentle nudge: all three of your 06-18 points were addressed in [my 06-20 reply](https://github.com/duneanalytics/spellbook/pull/9788#issuecomment-4758945073), so I think this is ready for a re-review whenever you have a moment.

Recap of the resolution:
- **`sub_tx_trade_id` determinism (commit 6ceb010).** Extended the final-select `row_number()` `ORDER BY` with the leg-identity columns: `order by evt_index, sub_type, sub_idx, nft_contract_address, nft_token_id, nft_token_amount, seller, buyer`. It's the only change on top of the single-pass rewrite, in the shared final select (independent of `iv_priced`).
- **Cross-caller parity.** All 16 callers compile clean; parity proven beyond the `opensea_v4_base` window (391,272 rows, EXCEPT both ways + 28 checksums) on **polygon** `[2026-06-01,06-14)` n=3,870 + **ethereum** `[2026-06-08,06-14)` n=175,822 — all 25 non-constant cols checksum-identical.
- **Red CI explained (structural, not a regression).** The sole real failure is `opensea_v4_polygon_base_trades` → OUT_OF_SPILL on the full-refresh `iv_trades ⋈ polygon.transactions` hash build. `EXPLAIN (TYPE DISTRIBUTED)` shows that join is structurally identical original-vs-rewrite (transactions on the build side, no dynamic-filter pruning); the **original** macro hits the same wall (sibling #9781's CI build also spilled before the 90-min timeout). Bounded-21d A/B: both spill 0, rewrite slightly better (cpu 372 vs 486 s, peak 2.23 vs 2.46 GB). Prod is incremental MERGE and never hits this path.

This is the modest IO-only lever (base.transactions dominates and is untouched) that pairs with #9781. Happy to mark it ready for merge on your go.

0xRobin (approved):

PR #9750, opened 2026-06-10
perf(uniswap_tvl_daily): bound forward-fill calendar to anchor window

a-monteiro:
`uniswap_tvl_daily` is the single most expensive model on `spellbook-dex` (~32.9 CPU-hrs/day, ~7% of the cluster), and its cost is a nested-loop row explosion rather than IO. The `tvl_daily` CTE forward-fills daily liquidity events into a per-day series by range-joining `daily_cum` against the entire `utils.days` calendar (`c.block_date <= d.day and d.day < c.next_day`). That calendar spans ~6,368 days (2009-today), so Trino cross-joins every anchor row against all of history before filtering. `analyze_query` on a representative run (`20260608_174910_19828_fdhqm`) showed 21.5M input rows amplified to 137.2B (6366x) in a `NestedLoopJoinOperator`, with that single stage carrying 65% of query CPU (skew 22.85x, peak ~21.8 GB).

Bounding the calendar to `timestamp >= (select min(block_date) from daily_events_final)` removes only days strictly before the earliest anchor — days that can never satisfy `block_date <= day`, so they contribute zero join rows. Output is therefore byte-identical and the nested loop collapses from a 6.4k-day cross-join to the few days in the forward-fill window. The same bound is applied to the full-refresh branch (`daily_events`).

Proven equivalent and measured on prod data (read-only, spellbook-hourly):

`(current EXCEPT rewrite)` and `(rewrite EXCEPT current)` both returned `0` rows over the live 3-day window (optimism+polygon+arbitrum slice, 593,718 rows each side); checksums over every structural output column were byte-identical. Full-model A/B, checksum-forced over all 16 output columns, warm runs:

| metric | current | bounded | delta |
|---|---|---|---|
| CPU | 3603 s | 459 s | -87% (7.9x) |
| wall | 152 s | 25 s | -84% (6.1x) |
| IO scanned | 2.53 GB | 2.55 GB | flat (CPU-bound) |
| peak memory | 8.2 GB | 8.7 GB | flat |
| output rows | 85.71M | 85.71M | identical |

The build SELECT is ~7.9x cheaper. Since the eliminated nested loop was 65% of the production run's CPU, the end-to-end daily model drops on the order of ~3x — roughly 32.9 -> ~11 CPU-hrs/day, freeing ~20 CPU-hrs/day (~4-5% of the dex cluster). No read-side change: same rows, same values.

Fixes CUR2-2707

PR #9790, opened 2026-06-14
perf(pancakeswap_tvl_daily): bound forward-fill calendar range-join to anchor window

a-monteiro:
Same fix as #9750 (uniswap_tvl_daily), applied to `pancakeswap_tvl_daily` — the second-biggest model on spellbook-dex and the largest unshipped `tvl_daily` sibling.

The `tvl_daily` CTE forward-fills `daily_cum` across the full ~6.4k-day `utils.days` calendar via an inequality range-join (`c.block_date <= d.day and d.day < c.next_day`), which Trino executes as a NestedLoopJoin — cross-joining every anchor row against all of history. Bounding `days` to `timestamp >= (select min(block_date) from <anchor>)` removes only calendar days that can never satisfy `c.block_date <= d.day` (they fall below every anchor's `block_date`), so the output is identical by construction. Applied to both branches (incremental anchor `daily_events_final`, full-refresh anchor `daily_events`), plus a CI-only 14-day floor on the full-refresh scan so the initial CI build finishes under the 90-min timeout and the `unique_combination_of_columns` test runs (inert in prod).

### Correctness evidence is the prod A/B below — not CI

The equivalence proof was run on **prod data on a prod cluster (spellbook-tokens), over the full cumulative history** that the prod-cadence incremental window forward-fills — it does **not** rely on the CI relation in any way. The CI-only 14-day floor is purely a buildability escape hatch for the initial slim-CI full build; it intentionally yields a 14-day slice whose cumulative balances are **not** prod-equivalent, and it is **not** part of the correctness argument here (the only CI assertion is `unique_combination_of_columns`, a key-uniqueness check, not value parity). The `days`-bound change itself is `target.name`-independent and is what the prod proof exercises.

Proof (spellbook-tokens, prod data, `SET TIME ZONE 'UTC'`, frozen 3-day incremental window = prod cadence):

- Equivalence at the calendar-join boundary, OLD (unbounded) vs NEW (bounded) in one query so `current_timestamp` cannot drift: `old EXCEPT new = 0`, `new EXCEPT old = 0`, identical totals (13,010,668 rows/side).
- Full incremental SELECT A/B, `checksum()` over all 16 output columns, 3 warm interleaved runs: identical count (10,412,641) and identical checksums on all 14 structural columns — **this includes the cumulative balance columns** `token0_balance_raw`, `token1_balance_raw`, `token0_balance`, `token1_balance` (bit-exact OLD vs NEW). The only 2 columns that vary are the price-derived USD valuations (`*_balance ... _usd`), and that variance is pre-existing `max_by(price, minute)` price-data nondeterminism — it varies OLD-vs-OLD too and is independent of this change.
- Cost (medians of 3 warm runs): CPU **479.3 → 62.4 s (−87%, 7.7×)**; wall 21.3 → 8.1 s (−62%); IO 1.04 → 1.05 GB (flat — the nested loop is CPU-bound, not IO); peak 0.51 → 0.66 GB; spill 0.

Projected end-to-end: ~3.99 → ~1.2 CPU-hrs/day on spellbook-dex. Generalises P11 (bound a forward-fill calendar range-join to its anchor window) across the `tvl_daily` family.

Fixes CUR2-2826
Towards CUR2-2707

### DIFF
--- a/dbt_subprojects/dex/models/_projects/pancakeswap/pancakeswap_tvl_daily.sql
+++ b/dbt_subprojects/dex/models/_projects/pancakeswap/pancakeswap_tvl_daily.sql
@@ -97,6 +97,10 @@ days as (
         timestamp as day 
     from 
     {{ source('utils','days') }}
+    -- bound the calendar to the forward-fill window: the join below only matches days >= the earliest
+    -- anchor (block_date), so earlier days can never join. Without this, ~6.4k days of full history get
+    -- cross-joined against every anchor row, exploding into a nested-loop scan (the model's top cost).
+    where timestamp >= (select min(block_date) from daily_events_final)
 ),
 
 tvl_daily as (
@@ -215,6 +219,11 @@ daily_events as (
         , amount1
     from 
     {{ ref('pancakeswap_daily_agg_liquidity_events') }}
+    {% if target.name == 'ci' %}
+    -- bound the CI full-refresh scan so the initial build completes under the 90-min timeout and the
+    -- unique_combination_of_columns test runs on real CI data; prod renders without this.
+    where block_date >= current_date - interval '14' day
+    {% endif %}
 ),
 
 daily_cum as (
@@ -234,6 +243,9 @@ days as (
         timestamp as day 
     from 
     {{ source('utils','days') }}
+    -- bound the calendar to the forward-fill window (see incremental branch); prunes the same
+    -- full-history cross-join on a full refresh.
+    where timestamp >= (select min(block_date) from daily_events)
 ),
 
 tvl_daily as (


a-monteiro:
* **#9790** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9790?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9790?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tomfutago:

The CI-only 14-day cap is fine as a buildability escape hatch, but for cumulative TVL it means the CI relation is not production-equivalent for balances/values. Can we make sure the PR description links or clearly records the prod A/B validation for Pancake specifically, and avoid treating CI value parity as evidence here?

a-monteiro:
Agreed on both points, and to be clear I'm not leaning on CI value parity at all — the CI 14-day cap is purely a buildability escape hatch for the initial slim-CI full build, and you're right that the resulting 14-day slice is not prod-equivalent for cumulative balances/values (CI only asserts `unique_combination_of_columns`, i.e. key uniqueness, never value parity).

The correctness evidence is the **prod A/B**, run on prod data on a prod cluster (spellbook-tokens) over the full cumulative history the prod-cadence incremental window forward-fills — I've now called this out explicitly in the description under "Correctness evidence is the prod A/B — not CI". For Pancake specifically:

- Calendar-join boundary, OLD (unbounded `days`) vs NEW (bounded `days`) in one drift-cancelling query: `old EXCEPT new = 0` and `new EXCEPT old = 0`, identical totals (13,010,668 rows/side).
- Full-SELECT A/B, `checksum()` over all 16 output cols, 3 warm runs: identical count (10,412,641) and identical checksums on all 14 structural columns — including the cumulative balance columns `token0_balance_raw` / `token1_balance_raw` / `token0_balance` / `token1_balance` (bit-exact OLD vs NEW). The only two columns that differ are the price-derived USD valuations, and that's pre-existing `max_by(price, minute)` nondeterminism upstream — it varies OLD-vs-OLD too, independent of this change.

So the cumulative TVL balances are provably unchanged on prod; the bound only drops calendar days below every anchor's `block_date`, which can never satisfy `c.block_date <= d.day`. Same proof shape as the merged-pattern sibling #9750.

tomfutago (approved):

PR #9794, opened 2026-06-15
perf(curvefi_ethereum_pools_balances): bounded-lookback hybrid seed (~3.0 CPU-hrs/day, 2.84 TB/day)

a-monteiro:
Replaces the full-history forward-fill seed in the shared `balances_incremental_subset_daily_legacy` macro with a bounded-lookback hybrid, opted into by `curvefi_ethereum_pools_balances` only.

## Problem
`curvefi_ethereum_pools_balances` (dbt `incremental` + `merge`) flows through `dex_pools_balances` -> `balances_incremental_subset_daily_legacy`. To seed the daily forward-fill, the macro's `changed_balances` UNION arm re-derives every key's last-known balance from **all** pre-window history of `tokens_ethereum.balances_daily_agg_base` via `max_by(balance, day)`. Measured on spellbook-dex (24h): **24 hourly MERGE builds, each scanning a constant ~118.5 GB** (3.315B rows, ~94% of the query's CPU) to emit ~1 changed row -> **3.04 CPU-hrs/day, 2.84 TB/day IO**.

## Fix
Add an opt-in `self_seed_relation` param (+ `self_seed_lookback_days`, default 21). When set, the prior-state seed is the union of:
- the **recent** pre-window source window (`not incremental_predicate('day')` AND `day >= now() - 21d`) — so late-arriving / restated balances are still self-corrected, and
- **deep settled history carried forward from the model's own latest pre-lookback partition** (`self_seed_relation`), instead of re-scanning the full source history.

`curvefi_ethereum_pools_balances` opts in with `self_seed_relation = this`. Default (param unset) renders byte-identically to today, so the 14 other callers of the macro are unchanged.

### Why 21 days is safe
`balances_daily_agg_base` re-merges only its last 3 days (`incremental_predicate('block_time')`), so a `(key, day)` row freezes after ~3 days. Recon-vs-recon on prod data shows the live snapshot vs a fresh all-history re-derivation: diverges at a 6-day-old seed (~0.045%, dust stablecoin positions zeroed by late-landing source rows), **0/0 at a 14-day-old seed and a 34-day-old seed**. 21d sits comfortably past convergence with margin; the recent re-read catches exactly those late arrivals the deep snapshot would miss.

## Proof (read-only, spellbook-dex, UTC, window = now-3d)
Reconstructed the macro's forward-fill output both ways (old all-history seed vs hybrid):
- **EXCEPT both ways = 0/0**, 26917 rows each side, **byte-identical `checksum()` over all 9 output columns** across 3 OLD + 3 HYBRID runs (also confirms determinism).

A/B (3 warm runs, medians):

| metric | old (all-history) | hybrid | Δ |
|---|---|---|---|
| IO | 118.45 GB | 1.17 GB | **-99.0% / 101x** |
| CPU | 504.8 s | 5.6 s | **-98.9% / 90x** |
| wall | 13.9 s | 1.2 s | -91% |
| peak | 3.62 GB | 0.039 GB | -99% |
| spill | 0 | 0 | - |

Projected model: ~3.04 -> ~0.15 CPU-hrs/day, ~2.84 -> ~0.03 TB/day. No `--full-refresh` needed (the existing table already holds full forward-filled history; the deep seed reads it directly).

Commit 2 adds a `target.name=='ci'` 30-day floor on the source scan (the shared-macro change marks all callers `state:modified`, so CI full-refresh-builds them) — renders only under `--target ci` (verified dev render byte-identical), prod unaffected.

Fixes CUR2-2833

### DIFF
--- a/dbt_macros/shared/balances_incremental_subset_daily.sql
+++ b/dbt_macros/shared/balances_incremental_subset_daily.sql
@@ -19,7 +19,9 @@
         start_date,
         address_list = none,
         token_list = none,
-        address_token_list = none
+        address_token_list = none,
+        self_seed_relation = none,
+        self_seed_lookback_days = 21
     )
 %}
 
@@ -96,9 +98,34 @@ filtered_daily_agg_balances as (
             ,token_standard
             ,token_id
             ,max_by(balance, day) as balance
-        from filtered_daily_agg_balances
-        where day >= cast('{{start_date}}' as date)
-        and not {{ incremental_predicate('day') }}
+        from (
+            {% if self_seed_relation is none %}
+            -- legacy: re-derive the prior state from all pre-window source history
+            select
+                blockchain, day, address, token_symbol, token_address, token_standard, token_id, balance
+            from filtered_daily_agg_balances
+            where day >= cast('{{start_date}}' as date)
+            and not {{ incremental_predicate('day') }}
+            {% else %}
+            -- bounded-lookback hybrid: re-read only the recent pre-window source so late-arriving /
+            -- restated balances within the lookback are still self-corrected ...
+            select
+                blockchain, day, address, token_symbol, token_address, token_standard, token_id, balance
+            from filtered_daily_agg_balances
+            where not {{ incremental_predicate('day') }}
+            and day >= cast(date_trunc('day', now() - interval '{{ self_seed_lookback_days }}' day) as date)
+            union all
+            -- ... and carry deep (settled) history forward from the model's own latest pre-lookback
+            -- partition instead of re-scanning the full source history every run
+            select
+                blockchain, last_updated as day, address, token_symbol, token_address, token_standard, token_id, balance
+            from {{ self_seed_relation }}
+            where day = (
+                select max(day) from {{ self_seed_relation }}
+                where day < cast(date_trunc('day', now() - interval '{{ self_seed_lookback_days }}' day) as date)
+            )
+            {% endif %}
+        )
         group by 1,3,4,5,6,7
         )
     {% endif %}
--- a/dbt_macros/shared/balances_incremental_subset_daily.sql
+++ b/dbt_macros/shared/balances_incremental_subset_daily.sql
@@ -114,21 +114,29 @@ filtered_daily_agg_balances as (
             and not {{ incremental_predicate('day') }}
             {% else %}
             -- bounded-lookback hybrid: re-read only the recent pre-window source so late-arriving /
-            -- restated balances within the lookback are still self-corrected ...
+            -- restated balances within the lookback are still self-corrected (these rows carry the
+            -- CURRENT token metadata via filtered_daily_agg_balances) ...
             select
                 blockchain, day, address, token_symbol, token_address, token_standard, token_id, balance
             from filtered_daily_agg_balances
             where not {{ incremental_predicate('day') }}
             and day >= cast(date_trunc('day', now() - interval '{{ self_seed_lookback_days }}' day) as date)
             union all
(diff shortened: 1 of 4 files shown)


tomfutago:

The hybrid self-seed snapshot is chosen from `day < now() - self_seed_lookback_days`, independently of `incremental_predicate('day')`. Defaults are safe today (`3d` incremental, `21d` self-seed lookback), but if someone runs a wider incremental repair/backfill, the self-seed row can overlap the in-window source branch and compete with fresh source rows for the same key/day.

Can we either anchor the deep self-seed before the incremental window, or assert/document that `self_seed_lookback_days >= DBT_ENV_INCREMENTAL_TIME` must hold?

A second thing to double-check: the deep self-seed reuses stored `token_symbol` and adjusted `balance` from the target table, whereas the legacy path re-enriches from current token metadata every run. Are we comfortable freezing those fields for old positions, especially if token metadata/decimals are corrected later?

a-monteiro:
* **#9794** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9794?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9794?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

a-monteiro:
Thanks @tomfutago — both good catches.

**1. Deep self-seed vs. a wider incremental window (overlap).** Fixed in `8ad2607` by anchoring the deep partition *strictly before the in-window arm*, not just before the lookback floor:

```sql
where day = (
    select max(day) from {{ self_seed_relation }}
    where day < cast(date_trunc('day', now() - interval '{{ self_seed_lookback_days }}' day) as date)
    and not {{ incremental_predicate('day') }}   -- <- added
)
```

Now the carried partition can never fall inside the incremental window regardless of how `self_seed_lookback_days` compares to `DBT_ENV_INCREMENTAL_TIME`, so the deep arm and the in-window source branch can't both emit the same key/day. It's a provable no-op for the default config (given `day < now()-21d`, the extra `day < now()-3d` is redundant) — verified on prod data, the chosen anchor is identical with and without it (`2026-05-29` either way), so the EXCEPT=0 equivalence and the A/B numbers in the description are unchanged. I went with the structural anchor rather than just asserting `lookback >= DBT_ENV_INCREMENTAL_TIME` so a future repair run can't silently violate the invariant.

**2. Frozen `token_symbol`/`balance` on the deep-carried rows.** Real trade-off, but the blast radius is narrow and I've documented it in the macro. Two things bound it:

- An incremental run never rewrites pre-window historical output rows in *either* path (the final select is filtered to the in-window predicate), so legacy doesn't retro-correct old rows on a normal run either — the only delta is the *in-window* rows of a position that hasn't changed in >lookback days.
- The recent-source arm (`[now-lookback, window_start)`) still re-enriches any *active* position from current `erc20` metadata every run; only positions dormant for >21d carry the stored value, and only until they next change or a `--full-refresh`.

`erc20.decimals` is effectively immutable on-chain (corrections are rare mis-ingestion fixes, usually followed by a `--full-refresh` of dependents), and this is the same self-seed trade-off already shipped on the sibling P7 stablecoins balances (#9727/#9730). I can't fully un-freeze it without `{{this}}` exposing `balance_raw` (it only stores the divided `balance`), so re-enriching just the symbol would be inconsistent. Happy to gate a metadata-sensitive caller off the hybrid if you'd prefer, but for curvefi pool balances I think the freeze is acceptable.

tomfutago (approved):

PR #9797, opened 2026-06-15
perf(angstrom_ethereum_base_trades): materialize bundle transactions staging

a-monteiro:
`angstrom_ethereum_base_trades` is the single largest IO consumer on the spellbook-dex cluster (24h reality-check: 20.17 CPU-hrs/day, 18.52 TB/day scanned, ~1055 GB per build for 1 merged output row). The bundle-calldata source scan lives in the `angstrom_tx_data` macro, which filters `ethereum.transactions` by `to = <angstrom>` AND a non-pushable `varbinary_substring(data, 1, 4) = 0x09c5eabe` selector. That CTE is referenced ~11x across the recursive bundle decoder (`angstrom_bundle_decoding_recursive_offset`) and the nested order macros, and Trino inlines each reference back into an independent scan.

`EXPLAIN ANALYZE` on a full build shows the inlining expands into **~300 `ethereum.transactions` scan operator instances** totalling **2.57B rows / 606 GB physical input — 85% of the model's IO**. Each instance re-reads the full transactions window (~8M rows / ~2 GB) only to re-derive the same ~6k bundle rows, because the selector predicate is filtered locally rather than pushed to the connector.

This PR extracts the scan once into a shared staging model, `angstrom_ethereum_bundle_transactions`, and points the `ethereum` branch of `angstrom_tx_data` at it. The ~300 transactions scans collapse to ~300 reads of a tiny (~6.3k-row / 4.3 MB) pre-filtered table plus one staging build.

## Change

- New macro `angstrom_tx_data_raw` — the original raw `ethereum.transactions` scan body, verbatim.
- New staging model `angstrom_ethereum_bundle_transactions` (incremental/merge, partitioned by `block_month`, `unique_key = [block_month, tx_hash]`) — wraps `angstrom_tx_data_raw` for Ethereum. Carries a CI-only 14-day floor so the `state:new` model doesn't full-refresh ~1yr of calldata and blow the 90-min CI budget; prod backfill/incremental runs are unaffected.
- `angstrom_tx_data` now reads the staging model for `ethereum` (re-applying the incremental window) and falls back to `angstrom_tx_data_raw` for any other chain. Pure source-swap — no decoding logic changed.

## Proof (read-only, spellbook-hourly, UTC)

Equivalence — the staging definition vs the original inline filter, projected to the 6 emitted columns, over a 3-day window:

- `(original EXCEPT staging)` = **0 rows**, `(staging EXCEPT original)` = **0 rows**.
- Merge-key uniqueness over [06-11, 06-14): `rows = distinct(tx_hash) = distinct(block_month, tx_hash) = 4123` — lossless.

A/B (median of 3 warm runs, `checksum()` over all output columns, never bare `count(*)`):

| metric | before | after (projected) | Δ |
|---|---|---|---|
| physical input / build | ~717 GB | ~114 GB | **-84%** |
| ...of which `ethereum.transactions` | 606 GB (300 scans) | ~3.6 GB (1 build + ~300 tiny reads) | -99% |
| reported scan / day | ~18.5 TB | ~3.1 TB | **-84%** |
| cpu | ~3331 cpu-s | ~1550 cpu-s | **~-47%** |
| peak memory | 9.6 GB | unchanged / better | — |
| spill | 0 | 0 | — |

The staging table is tiny and pre-filtered, so the ~300 inlined references now read kilobytes instead of re-scanning the full transactions window. IO is the headline; the CPU drop follows from eliminating 299 redundant scan+selector passes. The recursive decode, `pool_info` log scan, and block joins are untouched, so model output is identical (boundary EXCEPT = 0 both ways).

Realized whole-model numbers will be measured post-merge via the follow-up recipe (48h soak).

Towards CUR2-2709

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/angstrom/utils/angstrom_tx_data.sql
+++ b/dbt_subprojects/dex/macros/models/_project/angstrom/utils/angstrom_tx_data.sql
@@ -6,21 +6,26 @@
     )
 %}
 
-SELECT 
+{% if blockchain == 'ethereum' %}
+-- Read the shared, pre-materialized bundle-transactions staging model instead of re-scanning
+-- the raw transactions table. The varbinary_substring(data,1,4) selector filter is non-pushable,
+-- so each inlined copy re-scanned the full transactions window: the recursive bundle decoder and
+-- nested order macros expand this CTE into ~300 transactions scan operators per build
+-- (~85% of the model's physical IO). See angstrom_ethereum_bundle_transactions / CUR2-2709.
+SELECT
     block_number,
     block_time,
-    hash AS tx_hash,
-    index AS tx_index,
-    to AS angstrom_address,
-    data AS tx_data
-FROM {{ source(blockchain, 'transactions') }}
-WHERE 
-    success = true AND
-    block_number >= {{ earliest_block }} AND
-    to = {{ angstrom_contract_addr }} AND 
-    varbinary_substring(data, 1, 4) = 0x09c5eabe
+    tx_hash,
+    tx_index,
+    angstrom_address,
+    tx_data
+FROM {{ ref('angstrom_ethereum_bundle_transactions') }}
 {% if is_incremental() %}
-    AND {{ incremental_predicate('block_time') }}
+WHERE {{ incremental_predicate('block_time') }}
+{% endif %}
+
+{% else %}
+{{ angstrom_tx_data_raw(angstrom_contract_addr, earliest_block, blockchain) }}
 {% endif %}
 
 
--- a/dbt_subprojects/dex/macros/models/_project/angstrom/utils/angstrom_tx_data_raw.sql
+++ b/dbt_subprojects/dex/macros/models/_project/angstrom/utils/angstrom_tx_data_raw.sql
@@ -1,0 +1,27 @@
+{% macro
+    angstrom_tx_data_raw(
+        angstrom_contract_addr,
+        earliest_block,
+        blockchain
+    )
+%}
+
+SELECT 
+    block_number,
+    block_time,
+    hash AS tx_hash,
+    index AS tx_index,
+    to AS angstrom_address,
+    data AS tx_data
+FROM {{ source(blockchain, 'transactions') }}
+WHERE 
+    success = true AND
+    block_number >= {{ earliest_block }} AND
+    to = {{ angstrom_contract_addr }} AND 
+    varbinary_substring(data, 1, 4) = 0x09c5eabe
+{% if is_incremental() %}
+    AND {{ incremental_predicate('block_time') }}
+{% endif %}
+
+
+{% endmacro %}
--- a/dbt_subprojects/dex/models/_projects/angstrom/ethereum/_schema.yml
+++ b/dbt_subprojects/dex/models/_projects/angstrom/ethereum/_schema.yml
@@ -1,6 +1,40 @@
 version: 2
 
 models:
+  - name: angstrom_ethereum_bundle_transactions
+    meta:
+      blockchain: ethereum
+      sector: dex
+      contributors: jnoorchashm37
+    config:
+      tags: ['ethereum', 'angstrom', 'trades']
+    description: >
+      Pre-materialized Angstrom bundle transactions on Ethereum, extracted once from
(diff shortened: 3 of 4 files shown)


a-monteiro:
* **#9797** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9797?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9797?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

tomfutago:
Schema/key/test follow-up:

Please add `not_null` tests on both `block_month` and `tx_hash` for `angstrom_ethereum_bundle_transactions`. The uniqueness test matches the model `unique_key`, but Spellbook incremental models also need explicit non-null coverage for every merge-key column.

Validation follow-up: since the staging model is CI-limited to 14 days and the existing Angstrom base-trades seed rows are historical, can we confirm whether CI actually selected and tested `angstrom_ethereum_base_trades` after the macro swap? If not, please add/link a targeted Dune regression for downstream `angstrom_ethereum_base_trades` parity, ideally including a month boundary and the seed rows.

a-monteiro:
@tomfutago thanks — addressed both:

**1. `not_null` on merge keys** — added `not_null` tests on `block_month` and `tx_hash` for `angstrom_ethereum_bundle_transactions` (commit d6da7af).

**2. Did CI actually build + test `angstrom_ethereum_base_trades`?** Yes. The macro swap marks it `state:modified`, so slim-CI selected it. From the dex CI run on the first commit, base_trades was built **both ways** and all three tests passed:

```
OK created sql incremental model ...angstrom_ethereum_bundle_transactions [CREATE TABLE (32,559 rows)]
OK created sql incremental model ...angstrom_ethereum_base_trades         [CREATE TABLE (251,609 rows)]   # full-refresh
...
OK created sql incremental model ...angstrom_ethereum_bundle_transactions [MERGE (6,346 rows)]
OK created sql incremental model ...angstrom_ethereum_base_trades         [MERGE (14,729 rows)]            # incremental
...
PASS check_dex_base_trades_seed_angstrom_ethereum_base_trades_...
PASS dbt_utils_unique_combination_of_columns_angstrom_ethereum_base_trades_tx_hash__evt_index
PASS dbt_utils_unique_combination_of_columns_angstrom_ethereum_bundle_transactions_block_month__tx_hash
```

So the macro→staging plumbing is exercised end-to-end (full-refresh + incremental) in CI.

**But you're right to flag the seed rows specifically** — two things you should know:

- The 3 seed rows **are native bundle txs**, not composable Uni-v4 trades. I checked them on prod: all three have `to = 0x0000000aa232009084Bd71A5797d089AA4Edfad4` (the Angstrom contract) and selector `0x09c5eabe`, block_time Aug-2025. So they flow through exactly the path this PR swaps.
- They're from **Aug-2025**, which is outside the staging model's `--target ci` 14-day floor, so CI doesn't reproduce them. And independently, `check_dex_base_trades_seed` is a **no-op as currently configured**: its `filter: version: 1` doesn't match the seed rows (the model emits `version = '4'`), so it evaluates 0 rows. That's pre-existing — it isn't guarding parity before or after this PR.

So CI proves the plumbing + the last-14-days slice, but **not** the historical seed-row parity. That part is instead proven read-only against prod, since the change is a pure source-swap (the staging model's SELECT is byte-identical to the old inline scan):

- `(current inline scan EXCEPT staging def)` = **0** and `(staging def EXCEPT current)` = **0** over the 6 emitted columns, with identical checksums, on a representative window;
- merge-key uniqueness `rows = distinct(tx_hash) = distinct(block_month, tx_hash) = 4123` over `[2025-06-11, 2025-06-14)`;
- downstream decode logic is unchanged.

**To turn this into an active CI regression on the seed rows + month boundary** (what you're asking for), I'd: (a) extend the staging model's CI-only floor to also include the Aug-2025 seed week (bounded, CI-only, cheap), and (b) fix the seed test's `filter` to `version: 4` so it actually checks those rows. I left it out of this commit to avoid activating a previously-dead test inside a perf PR (if prod data for those txs has drifted it could surface unrelated red). Happy to add it here if you'd like, or file it as a follow-up since the no-op filter pre-dates this change — your call.

tomfutago (approved):

PR #9815, opened 2026-06-22
Fix Uniswap V4 sqrt price ordering

tomfutago:
## Summary
- Replace Uniswap V4 `block_index_sum` ordering with a deterministic key using block number, transaction index, and event index.
- Use decoded `evt_tx_index` when populated, and fall back to the canonical chain `transactions.index` when decoded event rows have null transaction indexes.
- Apply the corrected ordering consistently to sqrt price history and Uniswap/PancakeSwap base liquidity event interval joins.

## Root Cause
This was likely not caused by a SQL logic change in today's deployed PRs, but was surfaced by #9785 causing the affected macro file to rerun.

The failing model, `uniswap_v4_arbitrum_sqrtpricex96`, was not itself changed today. The relevant deployed PR was #9785 (`0c70be4f5`), which changed `dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql`. That file contains both the downstream liquidity macro and the `sqrtpricex96` macro, so dbt state selection can rerun `uniswap_v4_arbitrum_sqrtpricex96` even though the model wrapper did not change.

The actual root cause is that the previous merge key was too weak: `['id', 'blockchain', 'block_index_sum']`, where `block_index_sum = evt_block_number + evt_index / 1e6`. On Arbitrum, `evt_index` can repeat across different transactions in the same block, so this value is not a true event-ordering key.

Investigation evidence from the last 3-day incremental window:
- Previous source rows: `105,411`
- Distinct previous merge keys: `105,409`
- Distinct keys when transaction identity is included: `105,411`

Concrete collision found:
- Pool: `0x864abca0a6202dba5b8868772308da953ff125b0f95015adbf89aaf579e903a8`
- Block: `476185398`, `evt_index = 5`
- Two different swap transaction hashes shared the same previous merge key.
- The target already had two rows for that key, so a later incremental merge could hit `MERGE_TARGET_ROW_MULTIPLE_MATCHES`.

A hash-derived fallback was considered and briefly tested, but it is not valid here because `block_index_sum` is also the sqrt-price interval ordering boundary. Hash order is deterministic but not chain transaction order. Follow-up diagnostics on BNB showed hash order disagreed with real transaction order for many same-pool/same-block groups, so this PR uses `transactions.index` instead.

## Test Plan
- Dune diagnostic: corrected Arbitrum 3-day source key check returned `rows = 106300`, `corrected_key_count = 106300`, `duplicate_keys = 0`.
- Dune diagnostic: previous hash-order validation showed the reviewer finding was valid: Uniswap BNB 7d had `665,864` rows with hash order different from tx order across `280,528` pool-blocks; PancakeSwap BNB 7d had `2,860,102` rows different across `1,019,332` pool-blocks.
- Dune diagnostic: transaction-index ordering returned `null_order_keys = 0` and `duplicate_keys = 0` for Uniswap BNB, Uniswap Ethereum, PancakeSwap BNB, Uniswap Optimism, and Uniswap Polygon 3-day windows.
- Dune diagnostic: all 14 checked V4-compatible chains have `transactions` tables, and all 32 V4-compatible model wrappers now pass a `transactions` relation.
- CI: pending for latest commit `95326542f`.

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/pancakeswap_compatible_liquidity.sql
+++ b/dbt_subprojects/dex/macros/models/_project/pancakeswap_compatible_liquidity.sql
@@ -25,7 +25,7 @@ get_pools as (
 get_events as (
     select 
         *,
-        evt_block_number + evt_index/1e6 as block_index_sum
+        {{ uniswap_compatible_v4_block_index_sum('evt_block_number', 'evt_tx_index', 'evt_tx_hash', 'evt_index') }} as block_index_sum
 
     from 
     {{ PoolManager_evt_ModifyLiquidity }}
@@ -192,6 +192,7 @@ modify_liquidity_events as (
             evt_block_number,
             evt_block_date,
             evt_index,
+            block_index_sum,
             id,                -- pool id lives here
             sender,            -- caller/sender (useful metadata)
             tickLower,
@@ -211,7 +212,7 @@ modify_liquidity_events as (
             , e.evt_block_number as block_number 
             , e.evt_tx_hash as tx_hash
             , e.evt_index
-            , e.evt_block_number + e.evt_index/1e6 as block_index_sum
+            , e.block_index_sum
             , 'modify_liquidity' as event_type 
             , cd.currency0 as token0 
             , cd.currency1 as token1
--- a/dbt_subprojects/dex/macros/models/_project/pancakeswap_compatible_liquidity.sql
+++ b/dbt_subprojects/dex/macros/models/_project/pancakeswap_compatible_liquidity.sql
@@ -7,6 +7,7 @@
     , PoolManager_call_ModifyLiquidity = null
     , liquidity_pools = null
     , liquidity_sqrtpricex96 = null
+    , transactions = null
     )
 %}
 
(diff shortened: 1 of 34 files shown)


tomfutago:
Note on why the V4-compatible models now pass and join the chain `transactions` relation:

`block_index_sum` is used for two separate purposes here: the incremental merge key and the chronological ordering boundary for sqrt-price/liquidity intervals. When `evt_tx_index` is null, defaulting it to a constant such as `0` would not preserve either property. Multiple events in the same pool/block can still collide on the merge key, and all null-index events would be ordered as if they happened at the start of the block, which can assign the wrong sqrt-price interval.

Using `coalesce(evt_tx_index, tx.index)` recovers the actual chain transaction order for decoded event rows where `evt_tx_index` is missing, while keeping the existing decoded value when available.

Recommended follow-up: optimize the macro shape so the `transactions` join only applies to rows where `evt_tx_index is null` (for example via separate null/non-null branches). The current change prioritizes correctness of event ordering and merge-key uniqueness; the join can likely be narrowed for performance once CI is green.

0xRobin (approved):

PR #9820, opened 2026-06-24
chore(ci): remove disabled workflows

0xRobin:
Removes three GitHub Actions workflows disabled in the Actions UI. These workflows are no longer active automation and clutter the workflow list.


Pluies (approved):
🙏

tomfutago (approved):

PR #9823, opened 2026-06-24
fix(dex): re-key 0x Settler aggregator rows on real trace_address (datashare collision, CUR2-2901)

thevaizman:
## What

Fixes the `dex_aggregator.trades` Snowflake-datashare regression introduced by #9800 (CUR2-2844). The deterministic 0x Settler decode emits `evt_index = rn` + a hardcoded `trace_address = [-1]`. The datashare MERGE keys on **`(blockchain, tx_hash, evt_index, trace_address)`** — *narrower* than the dbt unique key (which also has `project, version`) — so those settler rows collide with other aggregators that also key on `[-1]`, breaking the MERGE with **error 100090** ("Duplicate row detected during DML action").

Full-history scan (2019→2026): **17,239 colliding `(blockchain, tx_hash, evt_index, trace_address)` groups, every one involving `0x API/settler`** — `cow_protocol/1` (17,060), `lifi/2` (173), `bitget_dex_aggregator/2` (6). All three partners key on `trace_address = [-1]`.

## Fix

Emit the **real on-chain settler-call `trace_address`** instead of the `[-1]` sentinel, for all settler rows:
- `zeroex_settler_txs_cte.sql` — surface `trace_address` (already computed upstream as `tr.trace_address`) in the final staging SELECT.
- `zeroex_settler_agg.sql` — thread it `settler → calls → trades → results` (passes through `add_amount_usd`'s `bt.*`) and emit `CAST(trace_address AS array<bigint>) AS trace_address`.

A real trace path is never `[-1]`, so settler rows can no longer collide with the `[-1]`-keyed decoders. `evt_index = rn` is unchanged; **both** the `0x API` row and its partner row are preserved (no volume change) — this is a key fix, not a re-attribution.

## Why re-key (not exclude)

The originally-considered `WHERE cow_rn IS NULL` exclude would have removed only the 17,060 CoW collisions and **left the 179 lifi/bitget collisions** (those are non-CoW settler fills) — the datashare would still break. Re-keying every settler row off `[-1]` fixes all 17,239 at once, and restores the pre-#9800 behaviour (the old decoder used a real maker-transfer log index for `evt_index`, so these rows never collided).

## Validation

- **Resolves all 3 pairings — code-proven**: `cow_protocol` (11 chain models), `lifi` (`lifi_v2_optimism_trades.sql:43`) and `bitget` (`bitget_dex_aggregator_trades.sql:54`) all hardcode `ARRAY[-1]`; a real trace ≠ `[-1]`.
[...]

PR #9825, opened 2026-06-24
feat(dex): deterministically decode 0x Settler sell side (taker_token, CUR2-2903)

thevaizman:
## Context

Follow-up to the deterministic 0x Settler aggregator decode ([#9800](https://github.com/duneanalytics/spellbook/pull/9800) / CUR2-2844), which reliably captured the **buy** side but left the **sell** side (`taker_token` / `token_sold`) `NULL` on **~52%** of ethereum settler rows. Root cause: `AllowedSlippage` has no sell-token field, so the agg *guessed* the sell token via a transfer pivot that fails on native-ETH sells, Permit2/AllowanceHolder pulls where `Transfer.from ≠ taker`, and ambiguous multi-outflow cases.

This makes the sell side deterministic from calldata + native value, like the buy side. Linear: CUR2-2903.

## What changed (2 macros; value-only on the sell columns)

**`zeroex_settler_txs_cte.sql` (staging)** — CASE-guarded decode of the **first** settler action from the `execute`/`executeMetaTxn` calldata (`B1 = OA + rel0 + 69`, with `TRY_CAST` + length/bounds guards so malformed calldata → NULL, never an overflow). Surfaces three new columns:
- `sell_token_decoded` / `sell_amount_decoded` — permit-class take/VIP actions (16-selector whitelist; arg word1 = taker `PermitTransferFrom`, left-padded address in word1's last 20 bytes).
- `native_value_decoded` — `NATIVE_CHECK.msgValue` (robust to nested / AllowanceHolder / ERC-4337 calls where the top-level `tx.value` is 0 but the native input rides an internal call).

**`zeroex_settler_agg.sql` (agg)** — consumes them:
- Per-chain wrapped-native map (17 chains) for **both** legs. Native buys now pivot-match the real wrapped receipt off-mainnet too (today they map to *mainnet* WETH everywhere → never match off-mainnet → floor). **Ethereum is byte-identical** (`0xC02aaA39…`).
- Adds `tx_value`; `native_amt = COALESCE(NATIVE_CHECK.msgValue, tx.value)`.
- Sell-leg precedence: **CoW → NULL** (out of scope) → **pivot (`sell_n=1`, unchanged)** → **calldata** → **native (wrapped-native + `native_amt`)**.

`evt_index` (-1), `trace_address` (the real settler-trace path from #9823), `rn`, and all unique keys are **unchanged**.

## Validation (dune-analyst, on-chain)

- **Realized impact (ethereum, 30d, 395,111 rows): `taker_token` NULL rate 53.1% → 0.51%.** Remainder ≈ 0.50% CoW (NULL by design) + 7 residual rows (0.0018%; nested-`BASIC` native + 1 plain-RFQ handled by `zeroex_settler_rfq`).
- The `NATIVE_CHECK.msgValue` path is load-bearing: it recovers ~5.05% of rows (nested-call native sells with `tx.value=0`) that would otherwise stay NULL.
- 16 selectors derived from `ISettlerActions` via keccak-256 and cross-checked; **0 decoder mismatches** across all 17 settler chains (90d). All 17 wrapped-native addresses verified priced in `prices.usd_with_native` + present in `tokens.erc20` (decimals 18).
- Examples: `0x4e869ecb…` → 0.5 ETH sold (native → wrapped); `0x409afa39…` → GRT 6,704.145 (calldata, ambiguous `sell_n=2` resolved).

## ⚠️ Deploy / full-refresh

Staging gains columns → a **coordinated full-refresh** is required to backfill history (`on_schema_change='sync_all_columns'` only adds them going forward): `zeroex_v2_<chain>_settler_txs` → `zeroex_v2_<chain>_trades` → `zeroex_api_fills_deduped` → `zeroex_trades` → `dex_aggregator_trades` (the last has a 7-day predicate; full-refresh re-materializes all aggregator projects). Please coordinate at deploy.

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_agg.sql
+++ b/dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_agg.sql
@@ -2,7 +2,8 @@
 {%- if target.name == 'ci' -%}
     {%- set start_date = (modules.datetime.date.today() - modules.datetime.timedelta(days=14)).strftime('%Y-%m-%d') -%}
 {%- endif -%}
-{%- set weth = '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2' -%}
+{#- Native legs (CUR2-2903) are represented with the chain's native token address from
+    source('dune','blockchains') — see the native_token CTE below (not a hardcoded wrapped-token map). -#}
 {%- set native_tokens = '(0x0000000000000000000000000000000000000000, 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee)' -%}
 {%- set erc20_transfer_topic = '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef' -%}
 {#- 2^128. minAmountOut at/above this is a "no minimum" sentinel (callers scatter across the top of the
@@ -18,15 +19,21 @@
 --     (executeMetaTxn) — verified to be the true buyToken recipient, NOT the calldata `recipient` (which is
 --     often an intermediate routing hop). Anchoring on a single calldata-named token to the verified user
 --     means it cannot mis-bind to internal routing hops the way the heuristic did.
---   token_sold = best-effort: the unique Transfer(from=receiver, token != buyToken).
+--   token_sold = deterministic (CUR2-2903), precedence per row: the unique Transfer(from=receiver) when
+--     resolvable (sell_n=1, unchanged); else the calldata permit-class decode (sell_token_decoded from the
+--     first take/VIP action); else native (wrapped-native token + NATIVE_CHECK.msgValue, or tx.value). CoW
+--     fills (cow_rn) keep a NULL sell leg (out of scope). The permit amount is the signed sell-order size.
 --   volume_usd = priced via either leg (add_amount_usd), so an unpriced buyToken still values off the sell leg.
--- NOTE: native-ETH buyToken is mapped to mainnet WETH for pricing, so it only resolves a price on ethereum;
--- per-chain wrapped-native pricing for native buys is a tracked follow-up (off-mainnet native buys -> NULL volume).
+-- NOTE: native legs (buy and sell) are represented with the chain's native token address from
+-- source('dune','blockchains') (the canonical Dune native representation), which prices in
+-- prices.usd_with_native on all settler chains. Native BUYS can't transfer-pivot (no ERC20 outflow of that
+-- address) so they take the minAmountOut floor; native SELLS use NATIVE_CHECK.msgValue / tx.value.
 
 WITH settler AS (
     SELECT
         tx_hash, block_time, block_number, method_id, settler_address, zid, tag, rn, cow_rn,
-        buy_token, min_amount_out, settler_msgsender, trace_address
+        buy_token, min_amount_out, settler_msgsender, trace_address,
+        sell_token_decoded, sell_amount_decoded, native_value_decoded
     FROM {{ ref('zeroex_v2_' ~ blockchain ~ '_settler_txs') }}
     -- exclude the sentinel zid (non-trade fills), matching the zeroex_v2 pipeline this replaces
     WHERE zid != 0xa00000000000000000000000
@@ -37,6 +44,14 @@ WITH settler AS (
     {% endif %}
 ),
 
+-- Per-chain native token (address/symbol/decimals) from dune.blockchains — the canonical representation
+-- for native legs (per review). One row per chain; LEFT JOIN ON true / UNION'd into token_metadata below.
+native_token AS (
+    SELECT token_address, token_symbol, token_decimals
+    FROM {{ source('dune', 'blockchains') }}
+    WHERE name = '{{ blockchain }}'
+),
+
 -- Distinct partition key of the settler txs, so the transactions/logs scans prune by partition
 -- (block_time + block_number + tx_hash) rather than a tx_hash-only post-scan filter, per zeroex_v2.sql's norm.
 settler_tx_keys AS (
@@ -45,7 +60,7 @@ settler_tx_keys AS (
 
 -- Tx-level sender/recipient (the settler trace's own `from` is an intermediary for nested/Relay-routed calls).
 txs AS (
-    SELECT t.hash AS tx_hash, t."from" AS tx_from, t."to" AS tx_to
+    SELECT t.hash AS tx_hash, t."from" AS tx_from, t."to" AS tx_to, t.value AS tx_value
     FROM {{ source(blockchain, 'transactions') }} t
     JOIN settler_tx_keys k
         ON  k.block_time = t.block_time
@@ -67,11 +82,21 @@ calls AS (
         t.tx_from, t.tx_to,
         -- receiver (the user): execute -> tx-level sender; executeMetaTxn -> msgSender (relayer pays gas).
         CASE WHEN s.method_id = 0xfd3ad6d4 THEN s.settler_msgsender ELSE t.tx_from END AS receiver,
-        -- native sentinels represented/priced as WETH (see header note)
-        CASE WHEN s.buy_token IN {{ native_tokens }} THEN {{ weth }} ELSE s.buy_token END AS buy_token,
-        s.trace_address
+        -- native sentinels -> the chain's native token address from dune.blockchains (canonical repr).
+        -- NOTE: a native buyToken no longer transfer-pivots (the native address has no ERC20 Transfer), so
+        -- native buys take the minAmountOut floor — accepted tradeoff for the canonical representation.
+        CASE WHEN s.buy_token IN {{ native_tokens }} THEN nt.token_address ELSE s.buy_token END AS buy_token,
+        nt.token_address AS native_addr,
+        s.trace_address,
+        s.sell_token_decoded,
+        s.sell_amount_decoded,
+        -- native sell amount: NATIVE_CHECK.msgValue (calldata; robust to nested/AllowanceHolder/4337 calls
+        -- where the top-level tx.value is 0), else the top-level tx.value. NULLIF so a decoded 0 (e.g. a
+        -- zero-msgValue NATIVE_CHECK) falls through to tx.value instead of sticking at 0.
+        COALESCE(NULLIF(s.native_value_decoded, UINT256 '0'), t.tx_value) AS native_amt
     FROM settler s
     LEFT JOIN txs t ON t.tx_hash = s.tx_hash
+    LEFT JOIN native_token nt ON true
 ),
 
 transfers AS (
@@ -126,6 +151,12 @@ token_metadata AS (
     SELECT contract_address, symbol, decimals
     FROM {{ source('tokens', 'erc20') }}
     WHERE blockchain = '{{ blockchain }}'
+    UNION ALL
+    -- ensure the native token address (dune.blockchains) always resolves symbol/decimals, even where it's
+    -- absent from tokens.erc20 (e.g. mode). NOT IN keeps token_metadata unique per address (no join fan-out).
+    SELECT token_address, token_symbol, token_decimals
+    FROM native_token
+    WHERE token_address NOT IN (SELECT contract_address FROM {{ source('tokens', 'erc20') }} WHERE blockchain = '{{ blockchain }}')
 ),
 
 trades AS (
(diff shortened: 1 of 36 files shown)


dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_agg.sql
```diff
@@ -67,9 +94,15 @@ calls AS (
...
-        s.trace_address
+        -- native sentinels mapped to the chain's wrapped-native (priced on all chains; this is ALSO the
+        -- buy-leg transfer-pivot match token, so a wrapped receipt resolves the exact buy amount).
+        CASE WHEN s.buy_token IN {{ native_tokens }} THEN {{ wrapped_native }} ELSE s.buy_token END AS buy_token,
+        s.trace_address,
+        s.sell_token_decoded,
+        s.sell_amount_decoded,
+        -- native sell amount: NATIVE_CHECK.msgValue (calldata; robust to nested/AllowanceHolder/4337 calls
+        -- where the top-level tx.value is 0), else the top-level tx.value.
+        COALESCE(s.native_value_decoded, t.tx_value) AS native_amt
```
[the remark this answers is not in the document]
thevaizman: Good catch — fixing. `native_amt` now uses `COALESCE(NULLIF(native_value_decoded, UINT256 '0'), tx_value)`, so a decoded `0` (incl. a zero-`msgValue` `NATIVE_CHECK`) falls through to the top-level `tx.value` instead of sticking at 0. In practice `NATIVE_CHECK` enforces `msg.value == msgValue` so a 0 there means `tx.value` is 0 too, but the `NULLIF` makes the fallback robust regardless. (in the next push)

dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_agg.sql
```diff
@@ -146,8 +179,20 @@ trades AS (
         CASE WHEN c.cow_rn IS NOT NULL THEN c.min_amount_out
              WHEN l.buy_n = 1 THEN l.buy_amount
              ELSE c.min_amount_out END AS token_bought_amount_raw,
-        CASE WHEN c.cow_rn IS NULL AND l.sell_n = 1 THEN l.sell_token END AS token_sold_address,
-        CASE WHEN c.cow_rn IS NULL AND l.sell_n = 1 THEN l.sell_amount END AS token_sold_amount_raw,
```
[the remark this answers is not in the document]
thevaizman: Looked into this — not a real-world miss, so holding off on a code change here:

- The `token <> buy_token` exclusion only drops an outflow when the **sold** token equals the (mapped) buy token — i.e. the user sells wrapped-native **and** buys native, a WETH→ETH unwrap. 0x API does not route unwraps through Settler as swaps, so this does not occur as an aggregator trade.
- Even if it did, the sell side is backstopped by priority-2: selling wrapped-native goes through a `TRANSFER_FROM` permit, so `sell_token_decoded` = the wrapped token and the calldata branch resolves it. The pivot exclusion only affects priority-1.
- For a normal native buy (buy native, sell some ERC20 X), the pivot excludes only the wrapped buy-leg and still finds X → `sell_n=1`, unchanged/correct. Excluding the received leg from the sell side is the intended behaviour (prevents the bought token being miscounted as a sell).
- ethereum is unchanged (wrapped == mainnet WETH there already); the only behavioural delta is on the non-existent off-mainnet unwrap case above.

dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_agg.sql
```diff
@@ -3,6 +3,29 @@
...
+    'scroll':      '0x5300000000000000000000000000000000000004',
+    'linea':       '0xe5d7c2a44ffddf6b295a15c148167daaaf5cf34f',
+    'polygon':     '0x0d500b1d8e8ef31e21c99d1db9a6444d3adf1270',
+    'bnb':         '0xbb4cdb9cbd36b01bd1cbaebf2de08d9173bc095c',
+    'avalanche_c': '0xb31f66aa3c1e785363f0875a1b74e27b85fd66c7',
+    'mantle':      '0x78c1b0c915c4faa5fffa6cabf0219da63d7f4cb8',
+    'berachain':   '0x6969696969696969696969696969696969696969',
+    'monad':       '0x3bd359c1119da7da1d913d1c4d2b7c461115433a'
+} -%}
+{%- set wrapped_native = wrapped_native_map.get(blockchain, weth) -%}
```
0xRobin: Ideally we don't hardcode this

dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_agg.sql
```diff
@@ -18,15 +41,19 @@
...
---   token_sold = best-effort: the unique Transfer(from=receiver, token != buyToken).
+--   token_sold = deterministic (CUR2-2903), precedence per row: the unique Transfer(from=receiver) when
+--     resolvable (sell_n=1, unchanged); else the calldata permit-class decode (sell_token_decoded from the
+--     first take/VIP action); else native (wrapped-native token + NATIVE_CHECK.msgValue, or tx.value). CoW
+--     fills (cow_rn) keep a NULL sell leg (out of scope). The permit amount is the signed sell-order size.
 --   volume_usd = priced via either leg (add_amount_usd), so an unpriced buyToken still values off the sell leg.
--- NOTE: native-ETH buyToken is mapped to mainnet WETH for pricing, so it only resolves a price on ethereum;
--- per-chain wrapped-native pricing for native buys is a tracked follow-up (off-mainnet native buys -> NULL volume).
+-- NOTE: native legs (buy and sell) are represented as the chain's wrapped-native ERC20 (per wrapped_native_map),
+-- which is priced in prices.usd_with_native on all 17 settler chains, so native trades value on every chain.
```
0xRobin: Native legs should be represented with the contract address that's present in `dune.blockchains` for that chain. Not the wrapped version.

dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_txs_cte.sql
```diff
@@ -82,6 +89,63 @@ settler_trace_data AS (
         ))
 ),
 
+-- Step 5b: Decode the FIRST settler action's byte offset from the execute/executeMetaTxn calldata.
```
0xRobin, line 92: Why do we need to manually decode this here?
We should rely on the decoding pipeline and decoded call tables if possible.

0xRobin (changes requested):
left a few comments.

dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_agg.sql
thevaizman: Done in 40e21cb97 — removed the hardcoded per-chain map. Native legs now resolve the address from a `native_token` CTE (`SELECT token_address, token_symbol, token_decimals FROM source('dune','blockchains') WHERE name = '<blockchain>'`), so nothing is hardcoded.

dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_agg.sql
thevaizman: Done in 40e21cb97 — native legs (buy and sell) now use the chain's native token address from `dune.blockchains` (via the `native_token` CTE), not the wrapped version; `token_metadata` UNIONs it so symbol/decimals resolve on every chain.

One consequence to flag: a native buyToken no longer transfer-pivots (the dune.blockchains native address has no ERC20 Transfer), so native buys fall to the `minAmountOut` floor. If you'd rather keep accurate native-buy amounts, I can keep a wrapped match-token just for the buy-leg pivot while still emitting the dune.blockchains address — lmk.

dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_txs_cte.sql
thevaizman: The settler `actions[]` are an opaque `bytes[]` array — the decoding pipeline surfaces the top-level `execute`/`executeMetaTxn` args (incl. `actions` as raw bytes) but doesn't decode the individual actions or their nested Permit2 struct, so the sell token/amount have to be parsed from the action calldata. This is the same reason the existing `buy_token` / `min_amount_out` / RFQ decode (from #9800) reads fixed calldata offsets rather than a decoded table.

The parse here is bounds-guarded (`TRY_CAST` + length checks) and gated on the take/VIP selector whitelist. If there's a decoded call/actions table I've missed that unpacks the per-action permit, I'm happy to switch — pointers welcome. Leaving this unresolved for your call.

thevaizman:
Thanks for the review @0xRobin — addressed in 40e21cb97:

- **Hardcoding / native representation (resolved):** dropped the hardcoded per-chain wrapped map. Native legs (buy + sell) now use the chain's native token address from `source('dune','blockchains')` via a new `native_token` CTE, not the wrapped version. `token_metadata` UNIONs it so symbol/decimals resolve on every chain. Tradeoff noted on the thread: a native buyToken no longer transfer-pivots, so native buys fall to the `minAmountOut` floor — happy to keep a wrapped match-token just for the buy-leg pivot if you'd prefer accurate native-buy amounts.
- **Manual decode (left open for your call):** the settler `actions[]` are an opaque `bytes[]` the decode pipeline doesn't unpack, so the per-action Permit2 must be parsed from calldata (same as the existing `buy_token`/RFQ decode from #9800). Bounds-guarded + selector-gated. Glad to switch if there's a decoded actions/call table I missed.

Also fixed an issue unrelated to your comments that the CI surfaced: the 17 `zeroex_v2_<chain>_trades` models keyed their incremental MERGE + unique test on `(block_month, block_date, tx_hash, evt_index)`, which isn't unique for multi-settler-call txs (→ `MERGE_TARGET_ROW_MULTIPLE_MATCHES` + a failing unique test). Added `trace_address` to both the `unique_key` and the `unique_combination_of_columns` test on all 17 (matches #9823's datashare key; verified 0 dup-groups on the 5-col key in prod).

Re-requesting your review 🙏

0xRobin:
I'm confused by this @thevaizman 

> Tradeoff noted on the thread: a native buyToken no longer transfer-pivots, so native buys fall to the minAmountOut floor — happy to keep a wrapped match-token just for the buy-leg pivot if you'd prefer accurate native-buy amounts.

Does the settler handle in native tokens or in wrapped native tokens? If it handles in native tokens directly, there wouldn't be a wrapped erc20 transfer to pivot on either right?

thevaizman:
You're right, and the data backs it — the settler handles the user's native leg in native tokens directly, so there's no wrapped ERC20 Transfer to pivot on either way:

- **Native sells:** the user sends `msg.value` (native ETH) — there's no ERC20 Transfer *from* the taker (that's exactly why the old transfer-pivot left these NULL). The amount comes from `NATIVE_CHECK.msgValue` / `tx.value`, not a transfer.
- **Native buys:** the taker receives raw ETH, not WETH. Over the last 14d only **1,437 / 59,565 (2.4%)** ethereum native buys — and **1,979 / 149,451 (1.3%)** on base — had *any* WETH→taker transfer. ~98% deliver native directly with nothing to pivot on.

So you're correct that there's no wrapped pivot to lose — scratch my "keep a wrapped match-token" aside, it was based on a wrong assumption. The `dune.blockchains` native address loses nothing vs the wrapped version: native sells are exact (`msg.value`), and native buys take the `minAmountOut` floor — which they'd take under *any* representation, since there's no transfer to pivot on. (Of the ~2% that do receive WETH, only ~39/14d on eth have a non-zero floor, and there the floor ≈ the actual amount, median 1.0x — not worth a hybrid.) Native pricing is also unaffected: the `dune.blockchains` native address is priced in `prices.usd_with_native` on all 17 chains (native price ≈ wrapped within 0.07%).

dbt_subprojects/dex/macros/models/_project/zeroex/zeroex_settler_txs_cte.sql
thevaizman: Confirmed there are no decoded call/actions tables we can use here — the settler `actions[]` are an opaque `bytes[]` and the per-action Permit2 isn't unpacked by the decoding pipeline, so we keep the guarded manual parse (consistent with the existing `buy_token` / `min_amount_out` / RFQ decode from #9800). Resolving — happy to revisit if a decoded actions table lands.

0xRobin (approved):
Thanks for the clarifications @thevaizman, looks good!

PR #9827, opened 2026-06-24
docs: converge Spellbook agent instructions

0xRobin:
Adds AGENTS.md as the shared source of truth for AI agent instructions, with CLAUDE.md importing it for Claude Code. Moves shared skills from .cursor/skills to .claude/skills and removes Cursor-specific rules after promoting their guidance into AGENTS.md.


tomfutago:
hmm, so claude is the boss now? what if someone still wants to use cursor?

0xRobin:
@tomfutago The goal is that all of the guidance is available to all of the agents. 

cursor automatically reads skills from `.claude/skills` for example and also reads `agents.md`
I would have liked to put the skills in `.agents/skills` but that's not supported (yet) by any of the providers.

The reason for removing the cursor rules (and moving them into the agents.md file) is that other harnesses did not adopt this syntax and thus can't leverage them properly.

tomfutago (approved):

PR #9828, opened 2026-06-25
fix(solana): use block_time lineage for dex microbatches

0xRobin:
Updates remaining Solana DEX microbatch lineage inputs to use block_time event-time metadata instead of manual batch window filters. This removes the leftover Pumpswap and Orca Whirlpool SQL workarounds and switches tokens_solana.transfers source lineage from block_date to block_time for CUR2-1597.


jeff-dude (approved):
safe to deploy with override to skip modified ✅

PR #9829, opened 2026-06-25
fix(solana): align daily balances merge grain

0xRobin:
Fixes the incremental merge key for solana_utils.daily_balances so it matches the model's address/day output grain. Adds a dbt uniqueness test for that grain; existing duplicate rows will still require a full refresh/backfill of the table.

### DIFF
--- a/dbt_subprojects/solana/models/solana_utils/_schema.yml
+++ b/dbt_subprojects/solana/models/solana_utils/_schema.yml
@@ -48,6 +48,12 @@ models:
       tags: ['solana','balances']
     description: >
         get the daily balances for each address, from which we can materialize a latest balance later on.
+    data_tests:
+      - dbt_utils.unique_combination_of_columns:
+          arguments:
+            combination_of_columns:
+              - day
+              - address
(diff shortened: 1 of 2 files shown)


2bwfkjpr25-lab:
Thanks for addressing the incremental merge-grain mismatch!
This is only a contract clarification, not a request to expand this PR.
The current SQL produces one row per (address, day), and this PR aligns the merge key and uniqueness test with that implementation. However, the public Dune documentation describes each row as an end-of-day balance for a specific (address, token, day).
Could a maintainer confirm which grain is the intended public contract before this PR is finalized?

2bwfkjpr25-lab:
Related follow-up: #9965 documents a separate state-semantics issue.

When the final balance-changing activity is lamport-only and its transaction metadata contains no token snapshot, `coalesce(post_token_balance, 0)` can store an unobserved token state as a known zero.

This is not a request to expand this PR. #9829 remains focused on aligning the incremental merge grain; historical duplicate rows still require the proposed full refresh/backfill.

PR #9830, opened 2026-06-25
CUR2-1640: add Solana account activity prefix join

0xRobin:
Adds a redundant address_prefix predicate where solana.account_activity.address is already joined to stake_call_Merge.account_sourceStakeAccount. This is output-preserving and intended to enable partition pruning for the merge branch of staking_solana.stake_actions.

Validation: ran dbt deps and dbt compile --select staking_solana_stake_actions.


0xRobin:
Validation query for the output-preserving `address_prefix` predicate added in this PR. It compares the old merge-branch join against the new join on a tight window; both result counts should be `0`.

```sql
with
old_merge as (
    select
        aa.address,
        aa.block_slot,
        aa.tx_id,
        aa.balance_change,
        m.account_sourceStakeAccount,
        m.call_block_slot,
        m.call_tx_id,
        m.call_outer_instruction_index,
        m.call_inner_instruction_index
    from delta_prod.solana.account_activity aa
    join delta_prod.stake_program_solana.stake_call_Merge m
        on aa.address = m.account_sourceStakeAccount
        and aa.block_slot = m.call_block_slot
        and aa.tx_id = m.call_tx_id
    where aa.writable = true
        and aa.balance_change != 0
        and aa.token_mint_address is null
        and aa.tx_success
        and aa.block_time >= timestamp '2024-01-01'
        and aa.block_time < timestamp '2024-01-02'
        and m.call_block_time >= timestamp '2024-01-01'
        and m.call_block_time < timestamp '2024-01-02'
),
new_merge as (
    select
        aa.address,
        aa.block_slot,
        aa.tx_id,
        aa.balance_change,
        m.account_sourceStakeAccount,
        m.call_block_slot,
        m.call_tx_id,
        m.call_outer_instruction_index,
        m.call_inner_instruction_index
    from delta_prod.solana.account_activity aa
    join delta_prod.stake_program_solana.stake_call_Merge m
        on aa.address = m.account_sourceStakeAccount
        and aa.address_prefix = substring(m.account_sourceStakeAccount, 1, 2)
        and aa.block_slot = m.call_block_slot
        and aa.tx_id = m.call_tx_id
    where aa.writable = true
        and aa.balance_change != 0
        and aa.token_mint_address is null
        and aa.tx_success
        and aa.block_time >= timestamp '2024-01-01'
        and aa.block_time < timestamp '2024-01-02'
        and m.call_block_time >= timestamp '2024-01-01'
        and m.call_block_time < timestamp '2024-01-02'
)
select 'old_minus_new' as check_name, count(*) as diff_rows from (select * from old_merge except select * from new_merge)
union all
select 'new_minus_old' as check_name, count(*) as diff_rows from (select * from new_merge except select * from old_merge);
```

0xRobin:
Validation is now saved and executed on Dune: https://dune.com/queries/7842856

Result on 2024-01-01 window:

| check_name | diff_rows |
|---|---:|
| old_minus_new | 0 |
| new_minus_old | 0 |

This supersedes my earlier raw SQL-only validation comment.

a-monteiro (approved):

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_stake_actions.sql
```diff
@@ -45,6 +46,7 @@ with
         FROM aa
         JOIN {{ source('stake_program_solana', 'stake_call_Merge') }} m ON 1=1 
             AND aa.address = m.account_sourceStakeAccount --the source table gets completely merged so this is safest to join on
+            AND aa.address_prefix = substring(m.account_sourceStakeAccount, 1, 2)
```
a-monteiro, line 49: **Perf win is dynamic-filter-dependent and unproven by the PR.**

The established `address_prefix` idiom (e.g. `bloom_solana_bot_trades.sql:35`) pins the prefix to a compile-time **literal** (`address_prefix = '{{ fee_receiver_1[:2] }}'`), which gives static partition pruning. Here the prefix is compared to `substring(m.account_sourceStakeAccount, 1, 2)` — a **join-derived** value — so pruning on `account_activity` relies on a dynamic filter from the `stake_call_Merge` build side reaching the scan at runtime, which can silently fail to fire.

The change is output-preserving regardless (the predicate is algebraically implied by `aa.address = m.account_sourceStakeAccount`, so no merge-branch rows can be added or dropped). But since the entire point is performance, it's worth confirming the prune actually triggers rather than assuming it: an A/B of `physical_input_bytes` (and partitions/files read) for the merge branch with vs without this line, via `EXPLAIN ANALYZE` on the CI table. `stake_call_Merge` is small, so the distinct-prefix set should be tiny and the dynamic filter should complete — likely a real win, just unverified by anything currently in the PR. If bytes scanned don't drop, this predicate is a no-op.

jeff-dude (approved):

dbt_subprojects/solana/models/_sector/staking/solana/staking_solana_stake_actions.sql
0xRobin: Ran the requested A/B `EXPLAIN ANALYZE` on the merge branch for a tight `2024-01-01` window.

Results on `solana.account_activity` scan:

| variant | physical input | execution | address_prefix dynamic filter |
|---|---:|---:|---|
| old join | 494.68GB | 2.63m | none |
| new join | 473.66GB | 1.29m | present, collected in 1.92s |

So this does trigger a dynamic filter on the `address_prefix` partition key for this window. The scan reduction is modest (~21GB / ~4.3%), but runtime improved materially in this run. Agreed this is dynamic-filter-dependent rather than equivalent to the literal-prefix pattern.

PR #9834, opened 2026-06-26
Add PumpSwap BuyExactQuoteIn trades

0xRobin:
Adds PumpSwap `BuyExactQuoteIn` trades to the existing `pumpswap_solana_base_trades_backfill` path, sourced from `BuyEvent`. `BuyExactQuoteIn` (`0xc62e1552b4d9e870`) is present on-chain in raw instruction data but is not decoded into `pump_amm_call_buy`; the emitted `BuyEvent` carries the actual amounts.

The model anti-joins existing `pump_amm_call_buy` rows to avoid double-counting legacy `Buy` coverage while preserving the existing call-backed trade path.

## Impact

Since `2025-11-01`, this covers an additional `208,873,603` PumpSwap trades and `$11.24B` in volume.

That is a `+42.05%` trade-count uplift and `+37.42%` USD-volume uplift over current same-period PumpSwap coverage. Relative to current all-time PumpSwap coverage, it adds `+16.83%` more trades and `+11.85%` more USD volume.

CUR2-1699


jeff-dude (approved):
proactive approval with revert needed
{# TEMP CI scope: revert begin to protocol_begin before merge. #} 

this one will take awhile to run on deploy

PR #9842, opened 2026-06-29
Add Kraken withdrawal credentials to staking entities

dnstaked:
### Description:

Adds Kraken validator withdrawal credentials to `staking_ethereum.entities_withdrawal_credentials` so deposits using these credentials are tagged as Kraken / CEX.

This normalizes 26 Kraken withdrawal addresses into full `0x01` and `0x02` withdrawal credentials and replaces the placeholder row in the static mapping.

`entity_unique_name` values are numbered `Kraken 3` through `Kraken 28`, continuing after the existing `Kraken 1` and `Kraken 2` depositor-address labels. The `0x01` and `0x02` credential rows for the same withdrawal address use the same unique name.

Validation:
- Verified 52 unique 32-byte credentials
- Verified 26 `0x01` credentials and 26 `0x02` credentials
- Verified 26 numbered `entity_unique_name` values, paired across credential types
- Ran `git diff --check`


dnstaked:

jeff-dude:
## Validation: deposit labeling impact (Dune MCP)

Validated using the CI lookup table from this PR:

`dune.dune_spellbook_ci__tmp_pr9842_28466761779_1.staking_ethereum_entities_withdrawal_credentials`

### CI lookup table

| metric | value |
|--------|-------|
| Kraken rows | 52 |
| Distinct credentials | 52 (`0x01` + `0x02` per address) |
| Distinct wallet names | 26 (`Kraken 3`–`Kraken 28`) |

[Query](https://dune.com/queries/7851365)

### Prod vs PR on `staking_ethereum.deposits`

Joined prod `staking_ethereum.deposits` to the CI credentials table on `withdrawal_credentials` to simulate the new label path (without rebuilding `deposits`).

**All history** ([query](https://dune.com/queries/7851369)):

| bucket | deposit rows | ETH staked |
|--------|-------------|------------|
| Already `Kraken` in prod | 93,190 | (most of ~3.57M ETH) |
| **Unlabeled in prod → Kraken with PR** | **835** | **~347K ETH** |
| Labeled other entity in prod | 5,383 | ~172K ETH |

Most matching deposits are already Kraken via **depositor address** (`Kraken 1`/`Kraken 2`). The main win from this PR is the **835 currently-null rows** — recent examples from wallets like `Kraken 21`, `Kraken 23`, etc.

**Sample before/after** ([query](https://dune.com/queries/7851368)):

| block_time | amount_staked | prod_entity | pr_entity | pr_entity_unique_name |
|------------|---------------|-------------|-----------|----------------------|
| 2026-06-30 | 283 ETH | `null` | Kraken | Kraken 23 |
| 2026-06-28 | 258 ETH | `null` | Kraken | Kraken 21 |
| 2026-06-26 | 1,920 ETH | `null` | Kraken | Kraken 21 |

**Unlabeled ETH by wallet** ([query](https://dune.com/queries/7851374)): top wallets are Kraken 21 (~114K ETH), Kraken 23 (~78K), Kraken 25 (~57K).

### Simulated `deposits` entity join

Replaced the entities side with `staking_ethereum.entities` **UNION ALL** CI credential rows, mirroring the `LEFT JOIN` OR logic in `staking_ethereum_deposits.sql` ([query](https://dune.com/queries/7851371)) — June 2026 deposits with these credentials:

| metric | count |
|--------|-------|
| `null` → Kraken | 199 |
| entity changed | 273 |
| `entity_unique_name` changed | 273 |

**Watch-out:** 5,383 historical rows are labeled **Figment 1** in prod (depositor-address match) but use Kraken withdrawal credentials ([query](https://dune.com/queries/7851370)). The OR-join can produce ambiguous matches when multiple identifier paths hit different entities.

### Post-merge expectation

After merge + **`deposits` full refresh**, expect:
- **835 deposits** to flip from unlabeled → `Kraken` / `CEX` with granular `Kraken 3`–`Kraken 28` names
- **`flows`** to inherit those labels on the deposit side (withdrawals already pick up entity via validator pubkey → deposits join)

CI only built the lookup view, not `entities`/`deposits` — so this simulation is the right way to preview impact before a downstream rebuild.

---

**Open question:** For the ~5,383 deposits currently labeled **Figment 1** via depositor address but carrying these Kraken withdrawal credentials — is that the intended attribution today, or should withdrawal credentials take precedence once this PR lands? The `deposits` join uses OR across identifier types with no explicit priority, so I want to confirm whether those rows should stay Figment or flip to the matching `Kraken N` wallet.

jeff-dude:
## Resolved: deterministic entity attribution when identifier types conflict

Answering the open question above — pushed `staking_ethereum_deposits.sql` changes to make this deterministic.

**Decision:** `withdrawal_credentials` (beneficial owner) now takes priority over `depositor_address`/`tx_from` (operator), which takes priority over `pubkey`. Rationale: withdrawal credentials identify who actually controls/receives the stake; depositor/tx_from addresses often just identify who *operated* the deposit on someone else's behalf (e.g. Figment depositing for a Kraken-controlled validator).

**Root cause:** the `ete`/`etes` joins in `staking_ethereum_deposits.sql` OR-matched across all 4 identifier types with no priority. Since `staking_ethereum_entities` has one row per identifier type, a deposit matching two different identifier types (e.g. depositor_address → Figment, withdrawal_credentials → Kraken) could match two different entity rows — producing duplicate output rows / non-deterministic attribution. That's exactly the ~5,383-row Figment/Kraken ambiguity from the validation above.

**Fix:** replaced the OR-join with a ranked, deduped join — `ROW_NUMBER() OVER (PARTITION BY tx_hash, evt_index ORDER BY <priority>)`, keeping only the top-ranked match per deposit. Same priority order applied to the `sub_entity` join for consistency.

**Validated on Dune:**
- Refactored model compiles and runs cleanly against prod tables (sample rows attached, no duplicates per `tx_hash`/`evt_index`).
- Confirmed the fix flips a real ambiguous row: deposit `0xd34f720b...` currently labeled `Figment 1` (via depositor_address) and using the new `Kraken 27` withdrawal credential now resolves to `Kraken`/`Kraken 27` under the new logic — 3 sample rows checked, all flipped as expected.

Once this + the Kraken credentials land and `deposits` gets a full refresh, expect the ~5,383 Figment-labeled rows using these Kraken withdrawal credentials to flip to the matching `Kraken N` entity, on top of the ~835 previously-unlabeled rows already covered by the credential data itself.

Known follow-up (not addressed here, flagging for visibility): `staking_ethereum_flows.sql`'s `MAX(entity)` aggregation per validator pubkey has a similar (much rarer) arbitrary tie-break if the same validator ever receives deposits attributed to different entities across multiple deposit events. Left out of scope here since it's a different bug class (temporal, not within-deposit ambiguity) and no known real-world case surfaced it.

jeff-dude (approved):

PR #9848, opened 2026-06-30
perf(solana): forward-fill stake_account_epochs SOL balance (drop full-history as-of scan)

a-monteiro:
`staking_solana_validator_stake_account_epochs` joined each recent stake-epoch to `solana_utils_daily_balances` via an as-of predicate (`bal.day <= date_trunc('day', b.epoch_time)` + `max_by(sol_balance, day)`). Because `daily_balances` is activity-only (not forward-filled), the as-of join had no time bound on the balance side, so every six-hourly MERGE re-scanned the full ~5.5B-row native-SOL history (~700 GB), fanned out to ~2.1B rows, then collapsed with `max_by` — ~16.8 CPU-hrs/day and ~2.83 TB IO/day, 70% of it in the LookupJoin.

This rewrites the incremental path as a forward-fill: the last-known balance is carried forward from the most recent prior-epoch snapshot already in `{{ this }}` (which is itself a per-epoch forward-filled snapshot of every delegated stake account), updated only with recent `daily_balances` activity (a bounded ~15-day window). The seed read is cheap because the table is naturally epoch-clustered, so no repartition/backfill is needed. The full-refresh branch keeps the original all-history as-of join (there is no prior snapshot to seed from on a fresh build) behind a CI-only scan floor so the modified-model CI build completes under the workflow cap.

## Equivalence (read-only, prod, spellbook-hourly, UTC)

Reconstructed the forward-fill for epochs 990-994 and diffed against the live table (which holds the current all-history as-of output; `sol_balance` is a single division, so no FP-accumulation drift — it is a clean oracle):

- rows: **67,155,563 both sides**
- `new EXCEPT oracle` = **5**, `oracle EXCEPT new` = **5**
- The 5 are 3 brand-new-this-window stake accounts (`has_seed=false, has_delta=false`) that were funded before the seed epoch and have had no balance activity since, so seed and recent activity are both absent → `sol_balance` null vs the old funding balance. **= 99.9999926% identical.** This is the only divergence; such accounts self-resolve on their next balance change. (Drift-free run with epochs pinned by fixed timestamp; an earlier run showed a 13.4M one-sided diff that was purely `now()`-window drift dropping epoch 990 from one side.)

Sizing of the corner over the live window: 13.45M window accounts, of which only **40,587 (0.30%) are new-this-window**; 40,457 of those are covered by the recent-activity branch, leaving **130 (0.001%)** with neither seed nor recent activity, of which only the handful with a pre-seed balance diverge.

## A/B magnitude (checksum-forced over all 7 output columns, 3 warm runs, medians)

| Axis | OLD (prod MERGE) | NEW (forward-fill select) | Δ |
|---|---|---|---|
| IO | 708 GB | 12.6 GB | ~56x (−98%) |
| CPU | 4.27 CPU-hr | ~0.075 CPU-hr | ~57x |
| peak mem | 17.8 GB | ~4.1 GB | −77% |
| spill | 0 | 0 | — |

Per day (4 heavy MERGEs): ~16.8 → ~0.7 CPU-hrs/day, ~2.83 TB → ~60 GB/day IO. NEW checksums were identical across all 3 runs (deterministic).

## Notes

- Pure SQL change — no `partition_by`/config change, no `--full-refresh` backfill required. Explicit `partition_by=['epoch']` would harden the seed read into a contract (instead of relying on natural file-clustering) but needs a backfill; left as optional follow-up.
- `_schema.yml` unchanged: the `unique_combination_of_columns` (epoch, epoch_time, stake_account, vote_account) grain is preserved (key columns come from the raw model unchanged).
- The ≤3-accounts/run `sol_balance` null on brand-new-anciently-funded accounts is the one behavior change; closing it exactly requires a forward-filled balance snapshot (the CUR2-2740 territory) and is out of scope here.

Fixes CUR2-2742


0xRobin (commented):
While we're in here, can we check if we have the option to lower the hardcoded lookback of 10 days in this model?
We've had some wider lookback periods for solana models based on epochs as they can span 2-3 days and it's often important to have the full epoch finalized inside an incremental run even in degraded situations (data delayed for ~48h). 
But I believe this isn't that important here given that we're only looking for the latest state within an epoch, this doesn't necessarily need the full epochs data available in the incremental window. 

Best case, we can use the standard `incremental_predicate(..)` filters here, but we need to double check that would be safe.

a-monteiro:
@0xRobin good call to check — I dug into it, and the 10-day lookback turns out to be load-bearing here rather than over-broad, so I'd keep it (and I added an in-code comment, 1ae8e97, explaining why).

**Why the standard `incremental_predicate()` macro isn't safe here:** it expands to `DBT_ENV_INCREMENTAL_TIME` / `DBT_ENV_INCREMENTAL_TIME_UNIT`, which default to `1 day`. The filter is on `epoch_time`, and `epoch_time` is the epoch *start*. Live cadence right now: epochs are ~48h apart, the just-finalized epoch 994 started **51h** ago, the in-flight epoch ~3h ago. So a 1-day predicate would re-emit **only the in-flight epoch** and drop even the just-finalized one — and in a degraded/48h-delayed situation it would drop a lot more.

**Why it can't simply be lowered, either:** this model MERGEs on `(epoch, stake_account)`, and its re-emit window is coupled to the upstream `_raw` model, which itself re-emits new `(epoch, stake_account)` rows across a 10-day window to "absorb late-arriving delegations." If this model's window is narrower than raw's, a delegation that raw picks up late for an epoch in the `(N, 10]`-day range would never propagate here — it'd be a silent missing row until a full refresh. So the safe floor for this model is `>= raw's` window, i.e. 10 days as long as raw is 10 days.

You're right that the *balance* side doesn't need the full epoch in-window (we only want the latest state per epoch) — and that's exactly what the forward-fill now exploits: the balance recency is handled by the bounded `balance_changes` delta, not by the re-emit window. So the wide lookback no longer drives any meaningful scan cost (that was the whole 56x win); narrowing it now would only trim a bit of MERGE write-amplification, against a real late-delegation-completeness risk.

If we do want to tighten it, I think the right move is a **coordinated `_raw` + this-model** lookback reduction, gated on a late-arrival profile of `staking_solana_stake_account_delegations` (how long after an epoch starts do delegations still land?) — happy to file that as a separate follow-up if you'd like, but it's out of scope for this cost PR.

0xRobin:
Ack, if there's multiple models involved in the lineage it's better to handle in a follow-up PR.

0xRobin (approved):

PR #9853, opened 2026-06-30
perf(dex): bound full-history self-scan in uniswap_compatible v4 sqrtpricex96 builder

a-monteiro:
Bounds the full-history `{{this}}` self-scan in the uniswap-compatible v4 `sqrtpricex96` price builder. The incremental seed (`get_latest_active_pools`) re-scanned all of each model's own history every run — ~387M rows on `pancakeswap_infinity_cl_bnb` — to recover one prior price per active pool, across 15 active chain models (~8.5 CPU-hrs/day, ~1.9 TB/day IO, 240 runs/day each).

`sqrtpricex96` is sparse / event-grain (one row per price change), not a forward-filled snapshot, so a naive time-bound would drop the seed for dormant-then-active pools and corrupt the downstream price gap-fill in `base_liquidity_events`. The seed must therefore reach back to each active pool's last price change, however old.

### Design (per review): window-anchored monthly-latest rollup

Each chain gets a sparse `sqrtpricex96_monthly` rollup — **one row per pool per active month**, holding that month's latest price event (`max_by` over `block_index_sum`), merged on `(blockchain, id, block_month)`. The builder reconstructs each active pool's previous price from:

- **complete prior months** — the rollup, `block_month < date_trunc('month', W)`, and
- **the current partial month** — a bounded `{{this}}` re-read, `block_time >= date_trunc('month', W) and block_time < W`,

where `W` is the model's own incremental window start (`uniswap_compatible_v4_window_start()` = `date_trunc('day', now() - interval '3' day)`, i.e. the exact lower bound of `incremental_predicate`). The existing inner-join-active-pools + anti-join-`base_events` + `max_by`-per-pool then collapses this to one latest-prior row per active pool.

Because `W` is derived from the **data window, not `now()`**, the seed `(< W)` and `base_events (>= W)` partition the timeline disjointly and the build is idempotent at any window size. A month-boundary crossing is handled by anchoring on `date_trunc('month', W)`: the rollup buckets every event by `date_trunc('month', block_time)`, so a window straddling month end simply reads one fewer month from the rollup and one more from the current-month self-read.

The rollup builds from the raw event sources (not the builder's output), so the two models do not form a dbt `ref` cycle, and it is self-scan-free: a forward-moving reload window means a pool's window-latest event *is* its month-latest event, so `merge` converges each month's row to that month's true latest and then leaves it frozen.

### Why a rollup and not a full-history `latest` snapshot

The earlier revision of this PR used a full-history "latest price per pool" companion. As flagged in review, that silently breaks idempotency on arbitrary-window / adhoc runs. Quantified on `pancakeswap_infinity_cl_bnb` at a widened 45-day window: the full-history companion reproduces only **330 of 4,579** required seeds (misses **93%**), because its aged horizon is anchored to `now()` and cannot supply a seed for the extra history the wider window exposes. The monthly rollup reproduces **4,579 / 4,579** (`EXCEPT` both ways = 0) at the same window.

### Equivalence

New seed proven equal to the full-history seed read-only on prod — `(current EXCEPT rewrite)` and `(rewrite EXCEPT current)` both 0 rows — at the default 3-day window on `pancakeswap_infinity_cl_bnb` (1,505 seeds / 2,619 active pools) and `uniswap_v4_base` (13,278 seeds / 36,387 active pools), and at a 45-day window on bnb (4,579 seeds, 0/0). Model output is unchanged: only the seed derivation differs and everything downstream is deterministic, so `unique_combination_of_columns` is unaffected.

### Rollup size (sparse vs forward-filled)

Sparse (one row per pool per **active** month) is what keeps the rollup small on spammy chains. On `uniswap_v4_base` (14.99M distinct pools, 18 months) sparse = **15.6M** rows total vs a forward-filled pool×month snapshot = **103M** (and re-emitting all 15M pools every month buys no read benefit — one forward-filled partition ≈ the whole sparse table). bnb: sparse 105.7k vs forward-filled 356k.

### Backfill

`sqrtpricex96_monthly` self-backfills on first prod build (full-refresh branch groups all history into month rows once), then goes incremental. The builder `ref`s it, so dbt builds the rollup first and the builder reads the freshly-populated table in the same invocation — no manual step. A `target.name == 'ci'` 14-day floor keeps the forced CI full-refresh bounded.

Realized whole-family CPU/IO numbers to follow post-merge soak.

Fixes CUR2-2835

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
+++ b/dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
@@ -4,6 +4,16 @@ cast({{ block_number }} as decimal(38, 0)) * decimal '1000000000000'
     + cast({{ evt_index }} as decimal(38, 0))
 {%- endmacro %}
 
+{#-
+    Lower bound of this subproject's incremental reload window: the same expression
+    incremental_predicate() compares against, so a model can anchor seed / lookback
+    bounds to the data window rather than to now(). Keeping seed bounds window-relative
+    (not now()-anchored) is what makes arbitrary-window / adhoc builds reproducible.
+-#}
+{% macro uniswap_compatible_v4_window_start() -%}
+date_trunc('{{ var("DBT_ENV_INCREMENTAL_TIME_UNIT") }}', now() - interval '{{ var("DBT_ENV_INCREMENTAL_TIME") }}' {{ var("DBT_ENV_INCREMENTAL_TIME_UNIT") }})
+{%- endmacro %}
+
 {% macro uniswap_compatible_v4_liquidity_sqrtpricex96(
     blockchain = null
     , project = 'uniswap'
@@ -11,14 +21,15 @@ cast({{ block_number }} as decimal(38, 0)) * decimal '1000000000000'
     , PoolManager_evt_Initialize = null
     , PoolManager_evt_Swap = null 
     , transactions = null
-    , latest_relation = none
-    , reread_lookback_days = 14
+    , monthly_relation = none
     )
 %}
 
 {% if is_incremental() %}
 
 -- force price reload
+
+{%- set window_start = uniswap_compatible_v4_window_start() %}
 
 with
 
@@ -85,21 +96,31 @@ get_latest_active_pools as (
             th.*,
             be.block_index_sum as base_block_index_sum
         from 
-        {%- if latest_relation is none %}
+        {%- if monthly_relation is none %}
         {{this}} th 
         {%- else %}
-        -- bounded recent self-read UNION the aged deep-tail companion: reproduces the
-        -- full-history max_by-per-pool seed while scanning only the last
-        -- {{ reread_lookback_days }} days of {{this}} plus ~one row per pool.
+        -- Seed each active pool's previous price from the latest event strictly before
+        -- the reload window, WITHOUT self-scanning full history:
+        --   * complete prior months come from the sparse monthly-latest rollup
+        --     ({{ monthly_relation }}, one row per pool per active month), and
+        --   * the current partial month [month(window_start), window_start) comes from a
+        --     bounded {{this}} re-read.
+        -- window_start is this model's own incremental window lower bound, so the seed
+        -- (< window_start) and base_events (>= window_start) partition the timeline
+        -- disjointly and the max_by-per-pool below collapses to exactly one latest-prior
+        -- row per active pool. Anchoring on the data window (not now()) keeps
+        -- arbitrary-window / adhoc builds reproducible.
         (
             select id, block_time, block_number, evt_index, block_index_sum, sqrtpricex96
-            from {{this}}
-            where block_time >= current_date - interval '{{ reread_lookback_days }}' day
+            from {{ monthly_relation }}
+            where block_month < cast(date_trunc('month', {{ window_start }}) as date)
 
             union all
 
             select id, block_time, block_number, evt_index, block_index_sum, sqrtpricex96
-            from {{ latest_relation }}
+            from {{this}}
+            where block_time >= date_trunc('month', {{ window_start }})
+            and block_time < {{ window_start }}
         ) th 
         {%- endif %}
         inner join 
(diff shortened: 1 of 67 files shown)


a-monteiro:
* **#9853** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

a-monteiro:
Independent supporting proof for the companion's source-side `block_index_sum` reproduction (the `aged_events` ingestion recomputes bis from source via the `transactions` join; the consumer later joins on that exact bis, so any drift would break the price lookup).

Ran read-only on prod spellbook-dex over a bounded settled slice `[now-10d, now-3d)` for `pancakeswap_infinity_cl_bnb`, comparing source-recomputed rows (id, block_number, evt_index, **block_index_sum**, sqrtpricex96) vs the builder's stored `hive.pancakeswap_infinity_cl_bnb.sqrtpricex96`:

```
src_rows | this_rows | src_minus_this | this_minus_src
10,015,890 | 10,015,890 |        0       |       0
```

Byte-identical at 10M-row scale — confirms the source-recomputed bis equals the stored join key (the tx join is deterministic: hash is unique per tx, so coalesce(evt_tx_index, tx.index) is unambiguous). Complements the seed-level EXCEPT=0 proofs already in the description.

(Context: I picked up CUR2-2835 in a separate session and built a parallel 3-chain version, #9854, before noticing this PR already existed and covered all 15 chains with a cleaner seed — #9854 is closed as a duplicate. This proof is the one reusable artifact from that detour.)

0xRobin (changes requested):
Relying on a full historical latest model will block us from correctly running these models with arbitrary time windows and should be considered an anti-pattern that can silently break idempotency on adhoc runs.

A better solution for reducing these frequent full historical scans is to build a monthly latest rollup model that is read.
That reduces the scan to:
+ read latest month from before the current incremental window from the rollup model
+ read current month from `self`

Should be very doable and still provide a solid performance improvement. Do put some care in handling the case where an incremental window crosses over the month boundary.

a-monteiro:
Thanks — reworked to the monthly-latest rollup exactly as suggested. Each chain now has a sparse `sqrtpricex96_monthly` model (one row per pool per active month, `max_by` price, merge key `(blockchain, id, block_month)`), and the builder seeds from:

- latest months before the window → rollup, `block_month < date_trunc('month', W)`
- current month → `self`, `block_time >= date_trunc('month', W) and block_time < W`

`W = uniswap_compatible_v4_window_start()` renders to `date_trunc('day', now() - interval '3' day)` — the exact lower bound of `incremental_predicate`, so the seed and `base_events` partition the timeline disjointly. **Month-boundary crossing** is handled by anchoring everything on `date_trunc('month', W)`: the rollup buckets by `date_trunc('month', block_time)`, so a window straddling month-end just reads one fewer rollup month and picks the remainder up from the current-month self-read.

On the idempotency concern you raised, quantified at a widened 45-day window on `pancakeswap_infinity_cl_bnb`: the old full-history companion reproduced only **330 / 4,579** required seeds (**-93%**, since its horizon was `now()`-anchored), whereas the rollup reproduces **4,579 / 4,579** (`EXCEPT` both ways = 0). Equivalence also holds at the default window on bnb and `uniswap_v4_base`.

Sparse (active months only) rather than forward-filled keeps it small on spammy chains — `uniswap_v4_base` is 15.6M rollup rows vs 103M if forward-filled, with no read benefit. The rollup reads the raw event sources (not the builder) so there's no `ref` cycle, and it's self-scan-free (forward-moving window ⇒ window-latest = month-latest per pool, so `merge` freezes each completed month).

Note this is deliberately different from the curvefi `self_seed_relation` (#9794): that seeds from a forward-filled **daily** snapshot that's value-idempotent and already window-anchored via `not incremental_predicate(day)`, so its `now()`-relative lookback is just a late-arrival constant. `sqrtpricex96` is sparse/event-grain, so the seed here has to be explicitly window-anchored — hence the rollup. Full design + numbers in the updated PR description. PTAL.

0xRobin (approved):

PR #9854, opened 2026-06-30
perf(dex): seed uniswap v4 sqrtpricex96 from a latest-price companion (drop full-history self-scan, 3 chains)

a-monteiro:
## What

The incremental `sqrtpricex96` builder macro (`uniswap_compatible_v4_liquidity_sqrtpricex96`) recovers each active pool's latest *prior* price (to seed the `previous_block_index_sum` lag) by re-scanning **all of `{{this}}` every run** in `get_latest_active_pools`. That is an unbounded self-scan: on `pancakeswap_infinity_cl_bnb` the rep run read **386M+ rows / ~9.2 GB** and processed ~400M rows to emit ~1 useful row per pool (poor-selectivity 2.5e-9), with the downstream self-join being the dominant CPU cost.

This adds an opt-in `latest_price_relation` parameter. When set to a per-chain `sqrtpricex96_latest` companion — a forward-filled **latest SETTLED price per pool** (1 row/pool), maintained incrementally from source with a bounded lagged slice + self-carry — the builder seeds from that ~85k-row companion instead of self-scanning full history. Default (`none`) renders byte-identical SQL, so unopted chains are unchanged.

Opts in the 3 highest-cost chains (71% of the family's ~8.6 CPU-hrs/day, ~1.86 TB/day):
- `pancakeswap_infinity_cl_bnb` (3.27 CPU-hrs/day)
- `uniswap_v4_bnb` (1.42)
- `uniswap_v4_base` (1.41)

Mirrors the proven curvefi self-seed pattern (#9794 / CUR2-2833). Event-grain sqrtpricex96 can't use curvefi's "latest snapshot partition" trick (no single partition holds every pool's latest price), so it needs a separate forward-filled companion to *be* that deep-carry source.

## Why a companion (not a naive lookback)

A pool active now can have its previous price arbitrarily far back (measured max seed age = **192 days**; a 21-day lookback would corrupt ~18% of seeded pools). The self-carry in the companion retains every pool's latest indefinitely, so the per-run source fold only needs a narrow lagged slice to catch each row once as it settles. `block_index_sum` is monotone in block_number, so `max_by/max` over (self-carry ∪ newly-settled slice) is idempotent.

## Proofs (read-only, prod spellbook-dex, current window)

1. **Builder output equivalence** — reconstructed the builder incremental output both ways (current full-`{{this}}` anti-join seed vs. latest-settled-per-pool seed): `rows_old = rows_new = 4,663,139`, `old_minus_new = 0`, `new_minus_old = 0`. Byte-identical.
2. **bis-match (companion source-fold == stored output)** — over a bounded 7-day settled slice, source-recomputed rows vs. stored `{{this}}` rows: `src = this = 10,015,890`, `src_minus_this = 0`, `this_minus_src = 0`. The `block_index_sum` join key (consumed by `base_liquidity_events` for the price lookup) is reproduced exactly, so the companion's source fold writes exactly what the builder stored.

## Measured cost (component A/B, pancakeswap_infinity_cl_bnb)

| | IO | CPU | rows |
|---|---|---|---|
| Removed: old seed self-scan of `{{this}}` | 9.22 GB | 54.5 CPU-s (+ ~252 CPU-s downstream self-join in the rep) | 386.9M |
| Added: companion 1-day lagged source fold | ~0.4–1 GB | ~10–13 CPU-s | ~11–25M |

The companion add-back is ~10% of what's removed on both axes — a net win on CPU and IO. (The add-back is dominated by the `transactions` join the companion must redo to recompute `block_index_sum`; a narrow 1-day slice minimizes it since the self-carry covers retention.) Whole-model realized numbers will be measured post-merge via `/query-doctor followup`.

## Selective opt-in

The companion's benefit scales with the eliminated self-scan size; its cost scales with the chain's `transactions` volume. So opt-in must be selective — only high-activity chains where the self-scan dominates. Top-3 clearly qualify; the remaining 13 chains are tracked in a follow-up for per-chain assessment.

## Backfill (one-time, prod)

Each companion's first incremental run needs a complete seed. To avoid a heavy full-history `transactions` join, pre-seed each table once from the existing builder output (which already stores `block_index_sum`) — the exact SQL is in a comment at the top of each companion model file. After seeding, incremental runs maintain it from a narrow source slice. The companion's non-incremental branch (full source history) carries a `target.name == 'ci'` 30-day floor so CI's full-refresh build completes under the timeout.

Towards CUR2-2835

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
+++ b/dbt_subprojects/dex/macros/models/_project/uniswap_compatible_liquidity.sql
@@ -11,8 +11,20 @@ cast({{ block_number }} as decimal(38, 0)) * decimal '1000000000000'
     , PoolManager_evt_Initialize = null
     , PoolManager_evt_Swap = null 
     , transactions = null
+    , latest_price_relation = none
     )
 %}
+
+{#
+    latest_price_relation: opt-in perf lever (default none = unchanged, byte-identical output).
+    When set to a ref() of this model's `sqrtpricex96_latest` companion, the incremental seed that
+    recovers each active pool's latest PRIOR price reads that 1-row-per-pool companion instead of
+    re-scanning ALL of {{this}} every run. The companion holds the latest SETTLED price per pool
+    (block_time < the incremental window_start), which is provably equal to the legacy full-history
+    anti-join seed (rows match, EXCEPT=0 both ways). Only worthwhile on high-activity chains where the
+    eliminated self-scan dominates the companion's maintenance cost. See macro
+    uniswap_compatible_v4_liquidity_sqrtpricex96_latest below and CUR2-2835.
+#}
 
 {% if is_incremental() %}
 
@@ -70,6 +82,7 @@ get_active_pools as ( -- get only the pools that were active on incremental run
 ),
 
 get_latest_active_pools as (
+{% if latest_price_relation is none %}
     select 
         th.id
         , 'exclude' as check_filter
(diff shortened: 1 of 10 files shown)


a-monteiro:
Closing as a duplicate of #9853, which already addresses CUR2-2835 more comprehensively (all 15 chains vs my 3) and with a better seed design: it unions a bounded recent {{this}} re-read (which already has block_index_sum stored, so no transactions re-join) with an aged-tail companion, whereas this PR's companion-only seed re-incurs the transactions join for all settled rows. Deferring to #9853. I missed #9853 in the resume state before building this — my mistake.

PR #9856, opened 2026-06-30
perf(chainlink): incrementalize vrf_v1_random_request_logs + bound fulfilled scans in vrf_request_fulfilled_daily (x5)

a-monteiro:
Family follow-up of merged #9766 (which fixed `chainlink_polygon_vrf_request_fulfilled_daily`). Replicates that exact pattern across the remaining 5 chains: arbitrum, avalanche_c, bnb, ethereum, fantom.

Each `chainlink_<chain>_vrf_request_fulfilled_daily` aggregates an upstream `vrf_request_fulfilled` view that decodes three raw `<chain>.logs` sources (a `RandomnessRequest` request log plus the v1 and v2 fulfilled logs). The daily model's incremental `WHERE` sits above that view's per-tx GROUP BY, so the bound never reaches the log scans — every incremental run full-scans all three log streams over all history (bnb ~1,180 GB/day) to merge ~1 row/day.

The fix has two halves, mirroring #9766:

1. **`chainlink_<chain>_vrf_v1_random_request_logs` view → incremental table.** The fulfilled side can reference an arbitrarily old request, so the request key set must stay all-history; materializing it (merge on `[tx_hash, index]`, `block_time` file-skipping on the incremental harvest) means the daily reads a compact accumulated table instead of re-scanning the full `logs` for `RandomnessRequest` every run.
2. **Inline the `vrf_request_fulfilled` view into each daily model** and push `incremental_predicate(...block_time)` below the per-tx GROUP BY onto the v1/v2 *fulfilled* view scans (these are bounded — a fulfillment's `date_start` derives from its own `block_time`). The outer `incremental_predicate('evt_block_time')` is kept authoritative.

Soundness: the request set is all-history (no dropped late fulfillments), and each fulfilled scan is bounded by its own block_time. Fantom's v2 arm additionally `LEFT JOIN`s `fantom.transactions` for `tx_from`; that join is bounded by `tx.block_time` too, which is sound because `tx.hash = fulfilled.tx_hash` ⇒ same block ⇒ same `block_time`, so the window drops no matched row.

Proof (read-only, spellbook-sqlmesh; old inner-unbounded path vs new inner-bounded path under the same frozen outer window, which isolates the inline+push-down from the model's own MERGE/late-arrival behavior):

- `(old EXCEPT new)` and `(new EXCEPT old)` = 0 both ways — bnb (738 rows, validates the 4 uniform chains) and fantom (62 rows, validates the bounded `transactions` join). Token sums match at the floating-point floor (bnb 1.07e-14; fantom exact 0.0).
- A/B cost (bnb, 3 warm-run medians): request-table incremental build 0.565 GB + bounded v1/v2 fulfilled scans 1.607 GB + ~435k-row table read ≈ **~2.2 GB/run** vs the current **~1,180 GB/day** — roughly **~530× IO** and **~90× CPU**. The larger ratio vs the ccip/ocr/fm siblings is because vrf eliminates three full all-history log scans, not one.

Family baseline ~1,711 GB/day + ~1.64 CPU-hrs/day across the 5 chains, expected to drop to ~15-20 GB/day. Prod check: all 5 daily tables have 0 duplicate keys and 0 null `operator_address` (no null-key hazard). The v1/v2 fulfilled-log models stay views.

Towards CUR2-2973

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/chainlink/arbitrum/chainlink_arbitrum_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/chainlink/arbitrum/chainlink_arbitrum_schema.yml
@@ -701,8 +701,6 @@ models:
       - dbt_utils.unique_combination_of_columns:
           arguments:
             combination_of_columns:
-              - blockchain
-              - block_number
               - tx_hash
               - index
     columns:
@@ -714,10 +712,16 @@ models:
       - *topic1
       - *topic2
       - *topic3
-      - *tx_hash
+      - name: tx_hash
+        description: "Transaction Hash"
+        data_tests:
+          - not_null
       - *block_number
       - *block_time
-      - *index
+      - name: index
+        description: "Index"
+        data_tests:
+          - not_null
       - *tx_index
 
   - name: chainlink_arbitrum_vrf_v2_random_fulfilled_logs
--- a/dbt_subprojects/daily_spellbook/models/chainlink/arbitrum/chainlink_arbitrum_vrf_request_fulfilled_daily.sql
+++ b/dbt_subprojects/daily_spellbook/models/chainlink/arbitrum/chainlink_arbitrum_vrf_request_fulfilled_daily.sql
@@ -11,6 +11,52 @@
   )
 }}
 
+-- CUR2-2973: the upstream view chainlink_arbitrum_vrf_request_fulfilled computes
+-- evt_block_time = MAX(block_time) above a GROUP BY, so filtering its output with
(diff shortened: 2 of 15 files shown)


a-monteiro:
* **#9856** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9856?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

0xRobin (approved):

PR #9778, opened 2026-06-13
perf(lido_accounting): single-pass prices.usd scan (7x -> 1x, -83% IO)

a-monteiro:
`lido_accounting` (hourly, `materialized='table'`, full rebuild ~24x/day) scanned `prices.usd` **7 times per build** — the `eth_prices` CTE (WETH daily close, a self-join) was inlined across the three `tokens_prices` UNION branches, and none of the `DATE_TRUNC('day', minute) = ...` / `EXTRACT(hour)=23` predicates push down to the `usd_0003` fact, so every reference re-read the full ~4.7B-row fact. Measured build: **33.0B rows / 212 GB scanned, ~5.1 TB/day**.

This rewrites the prices section to a **single pass**: one scan of `prices.usd` for the token set (8 ethereum tokens + stSOL) at the daily close, deriving the WETH close per day with a `max(...) OVER (PARTITION BY period)` window instead of the self-join, and emitting the "today = yesterday's close" relabel rows (ethereum tokens) via a `CROSS JOIN UNNEST` fan-out instead of the two extra UNION branches. Same P13 family as #9756 (`lido_liquidity`): `minute` is the only prunable column on the fact.

The downstream account UNION + `LEFT JOIN tokens_prices` + `GROUP BY` is unchanged; `tokens_prices` keeps the same columns.

**Equivalence (read-only, prod `prices.usd`, UTC, full history 2020-12-01 -> live today, same query so `now()` is consistent):** old vs new `tokens_prices` = **16,623 rows each, `EXCEPT` 0 both ways** on (period, token, symbol, decimals, price, eth_usd_price) and on `round(token_eth_price, 9)`; the 8 today-relabel rows match. Independently, `checksum()` over **all 7 output columns** is **byte-identical** old-vs-new across 3+3 warm runs.

**A/B magnitude (isolated prices section, 3 warm-run medians):**

| dim | old (7 scans) | new (1 scan) | Δ |
|---|---|---|---|
| IO | 212.4 GB / 33.0B rows | **36.8 GB / 6.17B rows** | **-82.7% (5.8x)** |
| CPU | 697.6 cpu-s | 677.6 cpu-s | -2.9% (flat) |
| wall | 7.1 s | 7.5 s | flat (noisy) |
| peak mem | 6.44 GB | 7.14 GB | +10.9% |
| spill | 0 | 0 | — |

[...]

PR #9870, opened 2026-07-06
perf(lido_accounting): full-refresh -> bounded-lookback incremental

0xRobin:
## What

Convert `lido.accounting` from full-refresh `table` to `merge`-incremental. Bound the lookback via the project's `incremental_predicate` macro on both the target merge and the two upstream scans (`daily_close_prices` on `prices.usd`, and the outer feeder aggregation); full-refresh path stays unbounded from 2020-12-01.

## Why lookback (not clean append)

Two restatement sources make append unsafe:

1. **Provisional-price rollover.** `tokens_prices` UNNEST emits `ARRAY[period, today]` for `period=yesterday`, copying yesterday's 23:59 close as a placeholder for today's unclosed period. When today rolls over, that period gets its own real close and the most-recent period's `value_usd` / `value_eth` / `token_price` restate exactly once.
2. **Late-arriving feeder tx** on the 16 `lido_accounting_ethereum_*` feeders. 15 of 16 are already incremental with the same `incremental_predicate` window; only `lido_accounting_ethereum_revenue` remains `table` (deferred to CUR2-2762 — this PR is independent).

## Patch (2 files, +39/-3)

`lido_accounting.sql`
- `materialized='incremental'`, `incremental_strategy='merge'`
- `unique_key = (period, hash, primary_label, secondary_label, account, category, base_token_address)` — 7 cols, empirically unique after cross-branch analysis of the ~34 UNION-ALL branches (stETH→WETH OR-join branches always carry distinct `primary_label`/`secondary_label`/`account`/`category` tuples, so no collision on shared `base_token_address`).
- `incremental_predicates = [incremental_predicate('DBT_INTERNAL_DEST.period')]` — target-side lookback via the project macro (`hourly_spellbook` defaults: `now() - 3 day`).
- `daily_close_prices` source WHERE: `{{ incremental_predicate('minute') }}` on incremental, unbounded from 2020-12-01 on full-refresh.
- Outer `WHERE {{ incremental_predicate('accounts.period') }}` inserted before `GROUP BY` on incremental runs only.

`lido_schema.yml`
- `dbt_utils.unique_combination_of_columns` on the 7-col unique_key.
- `not_null` on each of the 7 key columns (AGENTS.md mandate for materialized incremental models).

## Proof

Parity measured on a fully-settled 10-day window `[2026-05-01, 2026-05-11)` — no in-flight writes. The same guarantee holds for the 3-day operational window because it still strictly contains the ≤1-day upstream restatement horizon, and 15/16 feeders already settle inside that same 3-day macro window.

| Check | Result |
|---|---|
| Rounded EXCEPT (`ROUND(*, 8)` on `value_base_token`, `value_usd`, `value_eth`, `token_price`, `token_eth_price`), both directions, single-query single-snapshot | `fwd_count = 0`, `rev_count = 0` |
| Raw `FULL OUTER JOIN` with per-column diff (`compare.sql`) | 16 rows differ, **all USDC**, all sub-machine-epsilon: `max |Δ usd| = 2.91e-11` on values in the -$44k to -$95k range (relative ≈ 3e-16); `max |Δ eth| = 1.42e-14` on 28-60 ETH values (relative ≈ 5e-16). Deterministic columns (`token_price`, `token_eth_price`, `value_base_token`) all `diff = 0`. |
| Root cause of raw noise | Trino parallel-SUM aggregation-order non-determinism on USDC groups with many summands. Pre-existing property; would appear between two consecutive runs of the same FULL query when Trino picks a different plan. Not introduced by the incremental logic. |
| Row counts on 10d slice | full = incr (equal counts across all tested windows). |

Rounded tolerance `1e-8` = 1 microcent USD / 10 gwei — ~1000x above the machine-epsilon noise floor while still catching any real semantic gap.

## Selectivity

`prices.usd` 23:59 close scan (ethereum-token + stSOL filter), 10-day bounded vs unbounded:

| dim | full (from 2020-12-01) | incr (10d bounded, measured) | Δ |
|---|---|---|---|
| rows scanned | 16,822 | 279 | **-98.3% (60.3x)** |

Actual operating window under the macro is `now() - 3 day` (~3.3x tighter than the measured 10-day window), so real IO reduction is strictly larger. Row-count based; byte-scan reduction on columnar Delta with partition pruning expected higher.

Beyond IO, the collapse of the full-history UNION-ALL + GROUP BY + window from ~2000 days to ~3 days should deliver the real **CPU cut** (unlike #9778's IO-only fix which left CPU flat at ~5 cpu-hrs/day).

## Backfill

Post-merge: one full-refresh run to seed the incremental table (matches current cost). All subsequent hourly runs process only the macro's lookback slice.

## Related

- Builds on #9778 (CUR2-2793, single-pass `prices.usd`, -82.7% IO).
- Feeder `lido_accounting_ethereum_revenue` still `table`; incrementalizing it is tracked in CUR2-2762.
- Closes CUR2-2811.


dbt_subprojects/hourly_spellbook/models/_project/lido/lido_accounting.sql
```diff
@@ -22,6 +24,15 @@ with tokens AS (
...
         ) as tokens(address)
 ),
+{% if is_incremental() -%}
+-- Trailing-restatement lookback: covers today's provisional-price rollover
+-- (tokens_prices copies yesterday's 23:59 close into today until the day closes)
+-- plus a conservative buffer for late-arriving feeder tx.
+lookback AS (
+        SELECT CAST(max(period) - interval '10' day AS timestamp) AS since
+        FROM {{ this }}
+),
```
0xRobin: we should use the established incremental_predicate macro's

0xRobin: Done in cbf34a3bb — `incremental_predicate('DBT_INTERNAL_DEST.period')` target-side + `incremental_predicate('minute' / 'accounts.period')` source-side. Macro default (`now() - 3 day`) is safe here: all 16 upstream base spells in the lineage already use the same window, and this is a terminal spell (no downstream refs).

tomfutago (approved):
lgtm

PR #9872, opened 2026-07-07
fix(thorchain): self-healing incremental window for frozen gold event models

dennisp42:
## Summary

- 13 thorchain event models froze because their `incremental_predicate` filters the source on `block_time >= now() - 3 day`, while the raw source lands events **6-45 days** after block_time (ingestion lag). Every late arrival falls outside the window and never merges.
- Fix: filter the source on `_ingested_at` instead of block_time, via the existing `incremental_predicate` macro. Ingestion time is recent even when block_time is old, so a rolling window on `_ingested_at` catches late arrivals, and because it keys on arrival it is also robust to out-of-order arrivals. No new macro.
- `incremental_predicates` dropped from all touched models: an ingestion-time source bound paired with the existing `now()`-based block_time merge predicate is the asymmetry behind the #9421 duplicate-row bug. With no dest predicate the merge matches on `unique_key` across the whole table, dup-safe on these small, sparse tables.

### Models

- **Raw-sourced (5):** `node_account_status`, `active_vault`, `inactive_vault`, `gov_slash`, `rune_pool_deposit`. Gold-only change (node reads its silver view; the rest read the raw source).
- **Silver-view-sourced (2):** `stake`, `withdraw`. Silver is a `view` over fresh raw, so gold-only change (filter on the silver-carried ingestion timestamp).
- **Silver-incremental-sourced (3):** `loan_open`, `loan_repayment`, `pending_liquidity`. The silver is `incremental` and froze the same way, so both layers change: silver filters raw on `_ingested_at` and carries it forward, gold filters on it.

## How we know it's right

- Ingestion lag 6-45 days across all sources (min 6d, so none arrive inside the 3-day block_time window): dune.com/queries/7904761
- Out-of-order arrival window is 0 (arrivals are late but strictly in ingestion order), so a rolling `_ingested_at` window catches every row: dune.com/queries/7904768
- The first 5 were recovered via an incremental 120d run and QA'd exact, row-for-row vs raw source: dune.com/queries/7904657

## Notes

Minimal stopgap. Anchoring on `_ingested_at` and dropping the merge predicate keeps this small and defers the general incremental solution to the compute-optimization project. `loan_open`/`loan_repayment` raw sources have been quiet since 2025-10 (little to recover), included for consistency; `pending_liquidity` is live. After merge, one incremental backfill with a wide `_ingested_at` lookback recovers the silver-set history like the first 5.

CUR2-3130


dennisp42:
Checked the out-of-order arrival concern (does a late event ever land with a block_time below the frontier already in the table, which `max(block_time) - buffer` would miss). Across all 5 sources: 0 out-of-order batches out of 43-62 each, min forward gap ~3 days. Arrivals are late (10-45d ingestion lag) but strictly monotonic in block_time, so anchoring on `max(block_time) - DBT_ENV_INCREMENTAL_TIME` catches every one and the buffer is sufficient for these sources.

If a source ever delivered out of order beyond the buffer, the fallback is anchoring on `_ingested_at` instead of block_time.

Out-of-order check: https://dune.com/queries/7904768
Ingestion-lag verification: https://dune.com/queries/7904761

dbt_subprojects/daily_spellbook/models/thorchain/silver/thorchain_silver_loan_open_events.sql
```diff
@@ -47,6 +47,7 @@ base AS (
         DATE(cast(from_unixtime(cast(block_timestamp / 1e9 as bigint)) as timestamp)) AS block_date,
         date_trunc('month', cast(from_unixtime(cast(block_timestamp / 1e9 as bigint)) as timestamp)) AS block_month,
         date_trunc('hour', cast(from_unixtime(cast(block_timestamp / 1e9 as bigint)) as timestamp)) AS block_hour,
+        _ingested_at,
         current_timestamp AS _inserted_timestamp
```
0xRobin: No need to propagate the _ingested_at column further if we already have a _inserted_timestamp column that downstream consumers can use for incremental loading.
This is the case for all silver models.

0xRobin (changes requested):

dbt_subprojects/daily_spellbook/models/thorchain/silver/thorchain_silver_loan_open_events.sql
dennisp42: Good call, dropped the _ingested_at propagation. The silvers now only use _ingested_at in their raw-source WHERE (arrival filter), and the gold filters on the existing _inserted_timestamp. Raw-sourced gold (no silver layer) still filter the raw _ingested_at directly since there's no _inserted_timestamp to reuse there. Pushed in 8afe456.

0xRobin (approved):

PR #9873, opened 2026-07-07
fix(thorchain): pending_liquidity silver adds _inserted_timestamp on incremental (unblocks prod cadence)

dennisp42:
## Summary

- `thorchain_silver_pending_liquidity_events` gets `on_schema_change = 'append_new_columns'`. This fixes the prod cadence failure `COLUMN_NOT_FOUND "_inserted_timestamp"` on `thorchain_defi_pending_liquidity_events` (dbt Cloud run 70471887325976, query_id 20260707_152523_09882_ba5wi).

## Root cause

#9872 added `current_timestamp AS _inserted_timestamp` to `thorchain_silver_pending_liquidity_events` and repointed the `thorchain_defi_pending_liquidity_events` incremental predicate to that column. The silver is an incremental model with `on_schema_change` unset, so the default `ignore` applied: the new column was never added to the existing prod table. The gold's incremental run then filtered on `silver._inserted_timestamp`, which doesn't exist on the un-refreshed table, and errored.

CI didn't catch it because CI builds fresh (`is_incremental()` is false), so the silver table is created with the column every time and the gold resolves. The failure only appears on the incremental path against a pre-existing table.

pending_liquidity is the only one of the #9872 models where a new column was added to an incremental silver. The other silvers (loan_open, loan_repayment, stake, withdraw, update_node) already emitted `_inserted_timestamp` before #9872, so their tables already carry it.

## Fix

`append_new_columns` lets the next incremental silver run ALTER the table to add `_inserted_timestamp` (NULL on pre-existing rows, `current_timestamp` on newly merged rows). The daily/hourly cadence clones main at run time, so the next scheduled cadence self-heals once this merges. A `--full-refresh` of the silver (as the pending CUR2-3130 deploy does) also resolves it.

## Verification

- `dbt run --select thorchain_silver_pending_liquidity_events thorchain_defi_pending_liquidity_events` on the incremental path resolves `_inserted_timestamp` after the silver adds the column.

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/thorchain/silver/thorchain_silver_pending_liquidity_events.sql
+++ b/dbt_subprojects/daily_spellbook/models/thorchain/silver/thorchain_silver_pending_liquidity_events.sql
@@ -6,6 +6,7 @@
     incremental_strategy = 'merge',
     unique_key = ['block_month', 'event_id'],
     partition_by = ['block_month'],
+    on_schema_change = 'append_new_columns',
     tags = ['thorchain', 'liquidity', 'pending']
 ) }}
 


dennisp42:
@jeff-dude this is the fix for the `_inserted_timestamp` COLUMN_NOT_FOUND you flagged on the daily cadence (run 70471887325976). Root cause: #9872 added a new column to the pending_liquidity silver (incremental, on_schema_change unset -> ignore), so it never landed on the existing table and the gold's incremental predicate couldn't resolve it. This sets append_new_columns so the next cadence self-heals (cadence clones main at run time). Detail in the PR body.

dennisp42:
@jeff-dude closing this. append_new_columns would add the column going forward but leave _inserted_timestamp NULL across history, and the gold filters on it, so that's a band-aid, not the right fix. #9872 added the column = a breaking change that needs a --full-refresh rebuild so _inserted_timestamp is populated over history. Doing the full-refresh instead (this is the deploy that should have run for #9872 rather than a forward-only incremental).

PR #9885, opened 2026-07-10
feat(optimize_spell): throttle OPTIMIZE to ~1/day per table on hourly cadence

a-monteiro:
On the ~hourly cadence subprojects (dex, tokens, hourly_spellbook) each hot table was re-compacted on every run (~26x/day). Because each `OPTIMIZE` rewrites files that the next run's small MERGE immediately re-fragments, the repeated compaction is wasted work — measured ~30k redundant OPTIMIZEs/day, ~54 wall-hours and ~8 TB rewritten on spellbook-cadence alone, with OPTIMIZE accounting for ~14% of dbt wall-time there.

This gates OPTIMIZE behind a stateless per-model daily slot: `md5(relation)[:8] % 24` compared against the run's UTC hour, so each model compacts at most once per UTC day and the load is smeared across the 24 hourly runs rather than spiked. It mirrors the throttle already running in spellbook-sqlmesh. `daily_spellbook` (1 run/day) and `solana` (~5/day) are intentionally excluded — an hourly slot would rarely be hit for them. The 1000-row incremental gate from #9679 still applies underneath, and `table` models still always compact on their slot run.

The throttled-subproject scope is hardcoded in the macro rather than opted into via each `dbt_project.yml` on purpose: `optimize_spell` is a project-level post-hook, so any edit under `dbt_subprojects/<p>/` would select that subproject in CI and `state:modified.macros` would then flag 100% of its models (verified: 664/664 tokens models), forcing a full-subproject rebuild for a change that never alters model output. Keeping the change confined to `dbt_macros/dune/**` avoids that cascade (that path is CI-build-exempt by design). See CUR2-3180 for the full write-up of that CI behavior.

Verified via `dbt run-operation` against the tokens and daily_spellbook subprojects: `project_name` resolves in post-hook context, throttled projects skip off-slot and fire on-slot, `OPTIMIZE_SLOT_BYPASS=true` forces compaction, and non-throttled projects are unchanged. Slot distribution is even across all 24 hours. Rollout is monitor-only: no model data changes, effective on the next prod cadence run; will confirm the per-table OPTIMIZE drop (~26x -> ~1x) and watch MERGE wall-time via query_history over 48h.

### DIFF
--- a/dbt_macros/dune/optimize_spell.sql
+++ b/dbt_macros/dune/optimize_spell.sql
@@ -1,7 +1,7 @@
 {#-
   optimize_spell: post-hook that runs Delta-Lake OPTIMIZE on a materialized
-  model, conditional on the most recent main statement having affected enough
-  rows to make compaction worthwhile.
+  model, conditional on frequency and on the most recent main statement having
+  affected enough rows to make compaction worthwhile.
 
   Background: prior versions ran `ALTER TABLE ... EXECUTE optimize` on every
   prod incremental model regardless of whether the MERGE wrote any rows. About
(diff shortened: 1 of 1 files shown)


a-monteiro:
* **#9885** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9885?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9885?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

a-monteiro:
Good catch — fixed, though a bit differently than the autofix.

The autofix only patched the incremental branch (`is_throttled or rows_affected >= min_rows`). I refactored the macro into a single decision + single emit so the throttled/table/incremental paths don't each duplicate the `ALTER TABLE ... EXECUTE optimize` block. Net behavior matches the intent: on a throttled project a model only reaches the macro on its once-daily slot run, and there it now compacts **unconditionally** (row gate bypassed), so a low-volume slot hour can no longer starve the table of compaction. The row gate still applies to unthrottled incrementals. Shared `optimize_throttled_project()` helper keeps the project list in one place.

a-monteiro:
Superseded by #9889 — we're disabling the inline OPTIMIZE entirely (DuneCP health-check owns compaction) rather than throttling it, so the per-day throttle is moot.

PR #9889, opened 2026-07-11
chore(optimize_spell): disable inline OPTIMIZE (DuneCP health-check owns compaction)

a-monteiro:
Turns off the inline Delta OPTIMIZE post-hook across all spellbook subprojects. The `optimize_spell` macro is now a no-op; DuneCP's autonomous table health-check scan owns compaction of the trino-owned spell tables, so running OPTIMIZE synchronously in the dbt post-hook is redundant and only adds latency to each model build's critical path.

Reversible off-switch: the macro stays wired into every subproject post-hook, it just emits nothing. We'll evaluate for ~1 week, then either delete the macro and its post-hook calls for good or revert.

Towards CUR2-3191.


a-monteiro:
* **#9889** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9889?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9889?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jeff-dude (commented):
shouldn't we remove from each project config?
https://github.com/duneanalytics/spellbook/blob/main/dbt_subprojects/daily_spellbook/dbt_project.yml#L67-L68
or we're certain this avoids the post hook as-is, even if no-op?

edit: i see the description now. i'll approve with that caveat.

jeff-dude (approved):
approving with previous comment acknowledge

a-monteiro:
@jeff-dude correct, it's safe as-is: the hook still fires but the macro renders to an empty string, and dbt skips hooks whose compiled SQL is empty — so zero statements reach Trino (verified via a run-operation probe when this shipped).

The reason we didn't pull it from the five `dbt_project.yml` files now: any subproject config edit flags every model in that subproject via `state:modified.macros`, forcing full-subproject CI rebuilds (dex + tokens + hourly ≈ thousands of models). A `dbt_macros/dune/**`-only edit runs zero dbt CI, deploys on the next cadence run, and is instantly revertible if the DuneCP evaluation surprises us.

Removing the post-hook entries + deleting the macro is the follow-up in CUR2-3191 once DuneCP compaction coverage is verified over the ~1-week window.

PR #9891, opened 2026-07-16
feat(dex): add Origin ARM swaps to dex.trades

naddison36:
## Summary

Adds Origin ARM swaps on Ethereum to `dex.trades`.

The model includes swaps from:

- Lido ARM
- Ether.fi ARM
- Ethena ARM

Both `swapExactTokensForTokens` and `swapTokensForExactTokens` are supported.

## Methodology

The model uses decoded successful ARM calls to determine:

- Tokens bought and sold
- Raw input and output amounts
- Taker
- ARM contract
- Transaction and block metadata

Some older ARM implementations do not return an ABI-encoded amounts array. For these calls, the missing amount is recovered from the corresponding ERC-20 `transfer` or `transferFrom` call nested beneath the swap trace.

A deterministic `evt_index` is assigned using the call trace order, supporting transactions containing multiple ARM swaps.

## Changes

- Adds `origin_arm_ethereum_base_trades`
- Registers the model in `dex_ethereum_base_trades`
- Declares the six Origin Protocol decoded call sources
- Adds Origin ARM to `dex.info`
- Adds uniqueness and non-null tests
- Adds regression seeds covering:
  - Legacy exact-input trace recovery
  - Legacy exact-output trace recovery
  - Ether.fi ARM
  - Ethena ARM

## Validation

- Successfully executed the rendered model against recent Dune data
- Verified the expected 16-column DEX base-trades schema
- Verified all four regression transactions and raw amounts against Dune
- Validated YAML files
- `git diff --check` passes

## Project metadata

- `project`: `origin_arm`
- `version`: `1`
- `blockchain`: `ethereum`


naddison36:

jeff-dude (changes requested):
thank you for raising. some feedback:

**before merge**
1. **Guard the recovery `SUM()`** — 49% of swaps (33k of 68k) have NULL `output_amounts`, ongoing today, so this is your hot path. A second qualifying transfer under a swap trace (fee, refund, sweep) silently inflates amounts. Restrict to direct children (`CARDINALITY(traces.trace_address) = CARDINALITY(swaps.trace_address) + 1`) or use `MAX` + `COUNT(*) = 1` guard.
2. **Add a multi-swap seed row** — all 4 rows are `evt_index = 1`; the ROW_NUMBER logic is untested.

**Strongly encouraged:**
3. Add `AND traces.block_date = swaps.block_date` and `AND traces.call_type = 'call'` to the traces join ON — enables partition pruning (full refresh currently scans ~1.7 TB) and blocks delegatecall double-counting.
4. Ether.fi NULLs start exactly 2025-10-31 — looks like an implementation ABI missing from decoding. Re-submitting it would shrink your recovery path.

**Questions:**
- Does the ARM guarantee exactly one transfer per leg per swap?

naddison36:
Thanks for the detailed review.

I’ve addressed the feedback in separate commits:
1. 36398d4a9 — replaced the recovery SUM() with MAX(...) HAVING COUNT(*) = 1. Recovery now fails closed when more than one qualifying transfer exists. I initially tested the direct-child approach, but Ether.fi’s transfer occurs beneath its implementation delegatecall, so that approach excluded valid swaps.

3. 0b6ceb6be — added two distinct rows from a real multi-swap transaction, covering evt_index 1 and 2 and testing the ROW_NUMBER() ordering.

3. 07a5a9e4f — added traces.block_date = swaps.block_date and traces.call_type = 'call' to improve partition pruning and exclude delegatecalls from recovery.

4. I resubmitted the Ether.fi ARM proxy (0xfb0a3cf9b019bfd8827443d131b235b3e0fc58d2) to Dune for decoding. I also checked an affected transaction: both its proxy call and implementation delegatecall successfully return zero bytes, so output_amounts may correctly remain NULL after re-decoding. The guarded recovery path remains necessary.

Regarding the transfer guarantee: current ARM implementations perform one input transfer and one output transfer per swap, but this is not an immutable guarantee across upgrades. The MAX + COUNT(*) = 1 guard ensures recovery only accepts exactly one qualifying transfer and excludes ambiguous swaps.

jeff-dude (approved):
thank you, looks good to go ✅

PR #9892, opened 2026-07-16
Enable Delta Change Data Feed globally for all spellbook tables

shahil-bhikha:
Add delta.enableChangeDataFeed = true to the create_table_properties macro to enable CDC for all Delta tables created by dbt, both in production and CI.

This applies globally at CREATE TABLE time via the WITH clause for all models using file_format = 'delta' (100+ models across all sub-projects).

Property is rendered via dune_properties() macro into valid Trino SQL and applies to prod, dev, and CI equally. CDC is required for downstream datashare replication pipelines.


shahil-bhikha:

a-monteiro (changes requested):
We shouldn't do a blanket CDF enable for all tables

PR #9909, opened 2026-07-30
feat(macros): add optional filtering_columns model config

a-monteiro:
Spells get their Data Explorer filtering hint from the catalog service, which derives it from the table's partition columns. Views and unpartitioned tables have no partition columns, so they show no hint at all -- `tokens_evm.transfers`, `nft.trades`, `gas.fees` and every other cross-chain sector view included. Models can now declare `filtering_columns=['block_month', 'blockchain']` and the post-hook macros emit `dune.data_explorer.filtering_columns`.

Every property-emitting macro applies it, not just `expose_spells`. On the table path, `ALTER TABLE SET PROPERTIES` flows through `replaceTable()` -> `CreateTable`, which rebuilds the whole data explorer metadata struct from the properties in that single statement, so whichever macro emits last has to carry the key or it gets dropped and silently refilled from the partition columns. Views go through `alter_view_properties`, which upserts.

`none` and `[]` mean different things. An absent config omits the property; an explicit empty list emits `[]`, which is the only way to withdraw a hint already published on a view, since view updates upsert only the keys they send. On a table `[]` is not a true clear -- it falls back to the partition-derived default, because the catalog refills empty filtering columns from the partition columns. Both paths are documented.

The value has to reach the catalog as a JSON array, so a malformed config is a prod-only failure inside a post-hook. The guard rejects anything that isn't a list of strings, using the same `is sequence` / `is not mapping` / `is not string` combination as `trino__get_merge_sql` in `incremental.sql` -- a dict is a sequence to Jinja and would otherwise serialize as a JSON object.

The post-hooks only render when `target.name == 'prod'`, so no dbt CI job ever executes them. `dbt_macros/tests/test_property_macros.py` renders all six macros against a stubbed dbt context and asserts the emitted property map: present/absent/explicitly-empty, view and table statements, the other properties surviving alongside, nothing emitted off-prod, and the guard against a bare string, a non-string member, a tuple, a dict, an int and a bool. 44 assertions, no Trino needed. #9911 wires it into the pytest workflow and explains why that has to be a separate PR.

No model sets the config yet, so this is inert on its own; the models are in #9910.

### DIFF
--- a/dbt_macros/dune/config_trino_properties.sql
+++ b/dbt_macros/dune/config_trino_properties.sql
@@ -7,6 +7,22 @@
   ])
 {%- endmacro -%}
 
+{#
+  The catalog service only derives filtering columns from a table's partition columns, so views
+  and unpartitioned tables get no Data Explorer filtering hint unless the spell sets one. Every
+  macro that emits properties has to apply this: on the table path, ALTER TABLE SET PROPERTIES
+  replaces the whole data explorer metadata struct, so the last statement of a run must carry it.
+#}
+{%- macro apply_filtering_columns(properties) -%}
+  {%- set columns = model.config.get('filtering_columns', none) -%}
+  {%- if columns -%}
+    {%- if columns is not sequence or columns is mapping or columns is string or columns | reject('string') | list | length > 0 -%}
+      {%- do exceptions.raise_compiler_error("Invalid filtering_columns '%s'. Must be a list of column names." % columns) -%}
+    {%- endif -%}
+    {%- do properties.update({'dune.data_explorer.filtering_columns': tojson(columns)}) -%}
+  {%- endif -%}
+{%- endmacro -%}
+
 {% macro expose_spells(blockchains, spell_type, spell_name, contributors) %}
   {%- set validated_contributors = tojson(fromjson(contributors | as_text)) -%}
   {%- if ("%s" % validated_contributors) == "null" -%}
@@ -25,6 +41,7 @@
             'dune.data_explorer.freshness': var('freshness'),
             'dune.vacuum': '{"enabled":true}'
           } -%}
+    {%- do apply_filtering_columns(properties) -%}
     {%- if model.config.materialized == "view" -%}
       CALL {{ model.database }}._internal.alter_view_properties('{{ model.schema }}', '{{ model.alias }}',
         {{ trino_properties(properties) }}
@@ -45,6 +62,7 @@
             'dune.data_explorer.category': 'abstraction',
             'dune.vacuum': '{"enabled":true}'
           } -%}
+    {%- do apply_filtering_columns(properties) -%}
     {%- if model.config.materialized == "view" -%}
       CALL {{ model.database }}._internal.alter_view_properties('{{ model.schema }}', '{{ model.alias }}',
         {{ trino_properties(properties) }}
--- a/dbt_macros/dune/config_trino_properties.sql
+++ b/dbt_macros/dune/config_trino_properties.sql
@@ -12,12 +12,15 @@
   and unpartitioned tables get no Data Explorer filtering hint unless the spell sets one. Every
   macro that emits properties has to apply this: on the table path, ALTER TABLE SET PROPERTIES
   replaces the whole data explorer metadata struct, so the last statement of a run must carry it.
+
+  An explicit empty list is emitted rather than skipped. View property updates only upsert the
+  keys they send, so `[]` is the only way to withdraw a hint that is already published.
 #}
 {%- macro apply_filtering_columns(properties) -%}
   {%- set columns = model.config.get('filtering_columns', none) -%}
-  {%- if columns -%}
+  {%- if columns is not none -%}
     {%- if columns is not sequence or columns is mapping or columns is string or columns | reject('string') | list | length > 0 -%}
-      {%- do exceptions.raise_compiler_error("Invalid filtering_columns '%s'. Must be a list of column names." % columns) -%}
+      {%- do exceptions.raise_compiler_error("Invalid filtering_columns '" ~ columns ~ "'. Must be a list of column names.") -%}
     {%- endif -%}
     {%- do properties.update({'dune.data_explorer.filtering_columns': tojson(columns)}) -%}
   {%- endif -%}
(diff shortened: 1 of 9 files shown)


a-monteiro:
* **#9909** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9909?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9909?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dbt_macros/dune/config_trino_properties.sql
```diff
@@ -7,6 +7,22 @@
   ])
 {%- endmacro -%}
 
+{#
+  The catalog service only derives filtering columns from a table's partition columns, so views
+  and unpartitioned tables get no Data Explorer filtering hint unless the spell sets one. Every
+  macro that emits properties has to apply this: on the table path, ALTER TABLE SET PROPERTIES
+  replaces the whole data explorer metadata struct, so the last statement of a run must carry it.
+#}
+{%- macro apply_filtering_columns(properties) -%}
```
jeff-dude, line 19: `if columns` treats an explicit empty list as absent. Core supports `dune.data_explorer.filtering_columns = []` to clear the field, while view property updates preserve omitted keys, so a deployed view hint cannot currently be rolled back. Please distinguish `none` from `[]` (for example, `if columns is not none`), emit `tojson([])`, and document `[]` as the explicit clear path.

jeff-dude (dismissed):
Two blockers before merge:

1. Support explicit clearing of persisted filtering columns. Core accepts `[]` as the clear operation, but the current truthiness guard omits the property.
2. Ensure CI exercises the changed macro. The current project detection only treats `dbt_macros/shared/` as shared code, so #9909 skipped dbt execution, and #9910 is based on `main` rather than this branch. Stack #9910 on #9909 or update CI/add a prod-target render check that covers the actual macro.

The property key, JSON serialization, catalog fallback, and coverage of all production metadata emitters otherwise look sound.

a-monteiro:
Both blockers addressed in b2bbc33.

**1. Explicit clearing.** You're right, and I verified it against core before changing anything: `unmarshalStringArray` only rejects `null` and malformed JSON, so `[]` unmarshals to an empty non-nil slice and `applyProperty` assigns `FilteringColumns = []` (`trino_upsert_delete_table_properties.go:200-208`, `:99-104`). The guard is now `columns is not none`, so an absent config omits the property and `[]` emits `[]`.

One caveat worth writing down, which I've put in the docs: `[]` is a true clear only on the view path. On the table path the statement goes through `replaceTable()` -> `CreateTable`, and `BuildFromProto` refills empty filtering columns from the partition columns (`tablemetadata.go:181-188`), so on a partitioned table `[]` means "revert to the partition-derived default" rather than "no hint". Nothing to fix, but it would be a surprise later.

**2. CI coverage.** Added `dbt_macros/tests/test_property_macros.py` -- prod-target render checks over all six property macros, asserting the emitted property map for present/absent/explicitly-empty config, the view vs table statement, the other properties surviving alongside the new key, nothing emitted off-prod, and the guard against a bare string, a non-string member, a tuple, a dict, an int and a bool. 44 assertions, runs in 0.6s with no Trino. #9911 points the existing pytest workflow at it and triggers on `dbt_macros/**`.

I went that route rather than stacking #9910, and I'd push back on stacking specifically. Every model depends on `mark_as_spell` (614 in tokens alone, ~7400 across the five subprojects), so once a macro diff is in the branch, `state:modified.macros` selects all of them and CI attempts a full-repo rebuild. Same reason #9911 is separate: both `.github/workflows/**` and `scripts/**` match the shared-code branch in `dbt_full_run.yml`, so putting the workflow change here would mark every subproject modified while the macro diff is present, with the same result. Landing them apart keeps each PR's selection empty or tiny, and once #9911 is in, every future macro edit is covered.

Worth noting the current state is not "CI passed", it's "CI ran nothing" -- the green board on this PR is five `Skipped - no changes` statuses. #9911 fixes that class of gap, not just this PR.

dbt_macros/dune/config_trino_properties.sql
```diff
@@ -7,6 +7,24 @@
...
+  {%- set columns = model.config.get('filtering_columns', none) -%}
+  {%- if columns -%}
+    {%- if columns is string or columns | reject('string') | list | length > 0 -%}
+      {%- do exceptions.raise_compiler_error("Invalid filtering_columns '%s'. Must be a list of column names." % columns) -%}
```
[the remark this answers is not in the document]
a-monteiro: Half right, and the half that's wrong is the part this describes.

A list does not raise here: `%` only unpacks tuples, so `"...'%s'..." % ['block_month', 3]` formats to `Invalid filtering_columns '['block_month', 3]'`. The existing test for that path asserts a compiler error and passes, which is what prompted me to check rather than take the report at face value.

A **tuple** does raise, though, which the report didn't identify: `"...'%s'..." % ('block_month', 3)` -> `TypeError: not all arguments converted during string formatting`. Reachable if someone writes `filtering_columns = ('a', 3)`.

Fixed in b2bbc33 by dropping `%`-formatting for `~` concatenation, which is type-proof, with the tuple case added to the parametrized rejection test.

dbt_macros/dune/config_trino_properties.sql
a-monteiro: Done in b2bbc33 -- `columns is not none`, `[]` emitted as `[]`, and documented as the explicit clear path in `model_config_block.md`, `AGENTS.md` and the macro docstring.

Verified the core side first: `unmarshalStringArray` only rejects `null`/malformed, so `[]` gives an empty non-nil slice and `applyProperty` assigns it (`trino_upsert_delete_table_properties.go:200-208`). One wrinkle I documented: `[]` clears on views, but on the table path `BuildFromProto` refills empty filtering columns from the partition columns (`tablemetadata.go:181-188`), so there it means "back to the partition default".

jeff-dude (approved):
i will look to deploy with skip state modified, to be extra cautious.

PR #9900, opened 2026-07-21
Hide curated spells marked for removal from Data Explorer

jeff-dude:
## What & why
Flip 19 spells from `expose_spells` to `hide_spells` per the Data Explorer cleanup, covering the categories marked "removed entirely": staking, lending, CEX flows, bridges, rollup economics, labels/identity, other, plus `nft.mints`.
These tables disappear from Data Explorer and search on each model's next scheduled prod run; the underlying tables and data are untouched and remain queryable.
Keep-list spells in this repo (dex, tokens, nft, utils, thorchain/TON/aptos/sui/near) were audited and already correctly exposed.

## Architecture
- 19 model files: post-hook swap only, no SQL body or config changes — `staking_ethereum.*`, `lending.*`, `cex.*`, `bridges_evms.*`, `rollup_economics_ethereum.*`, `labels.ens`, `safe.safes_all`, `addresses.stats`, `dune.blockchains`, `nft.mints`
- Tested: `dbt parse` clean on hourly_spellbook and daily_spellbook; post-hook-only changes won't trigger CI model builds (`state:modified.body`)
- Needs human eyes: confirm the 19-table list matches the final cleanup doc before merge
- Blast radius: Data Explorer/search visibility metadata only; queries referencing these tables keep working

<details><summary>Agent context</summary>

- Source of truth: CUR2-3341 comment (table list from core `table_overrides.go` review)
- All 107 blockchain-section keep tables (thorchain/TON/aptos/sui/near) verified as `expose_spells` — no changes
- `labels.ens` has no explicit `schema=` in config (inherits `labels`); matched by alias
- Repro: `grep -c hide_spells` on the 19 files; `uv run dbt parse --project-dir dbt_subprojects/{hourly,daily}_spellbook/`
- No agent review yet
</details>

Towards CUR2-3341

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #9910, opened 2026-07-30
feat(spells): declare filtering_columns on cross-chain sector views

a-monteiro:
Needs #9909 to land for the property to actually be emitted; the config is inert without it, so merge order doesn't matter.

The Data Explorer filtering hint comes from the catalog service, which derives it from a table's partition columns. Views have none, so every cross-chain sector view currently shows no hint at all — including `tokens.transfers`, the second most queried spell after `dex.trades` (which does have hints, via its partitioning). Each value mirrors the parent's partition order, plus `blockchain` where the view unions per-chain tables: `blockchain` is constant within each upstream table, so it still skips files via Delta statistics.

| spell | value |
| --- | --- |
| `tokens.transfers`, `tokens_evm.transfers` | `['block_month', 'blockchain']` |
| `nft.trades` | `['blockchain', 'project', 'block_month']` |
| `nft.transfers` | `['block_month', 'blockchain']` |
| `nft.wash_trades` | `['blockchain']` |
| `gas.fees`, `gas_evm.fees` | `['block_month', 'blockchain']` |
| `gas_solana.fees` | `['block_date']` |
| `dex_evm.trades` | `['block_month', 'blockchain', 'project']` |
| `bitget_dex_aggregator.trades` | `['block_month', 'blockchain']` |

Two are not the obvious value. `gas_solana.fees` uses `block_date` because its quarterly backfill tables are partitioned by `block_date`, not `block_month`. `nft.wash_trades` gets `blockchain` alone because the view never selects `block_month` even though its upstream tables are partitioned by it — exposing `block_month` there would be an additive schema change, worth doing separately.

`dex_evm.trades` and `gas_evm.fees` come from @jeff-dude (ed6b9a3), who also aligned `nft.trades` with its parent partition order and trimmed `nft.wash_trades`. Both are in core's curated overrides (`core-go/syncsense2/models/table_overrides.go`) and were missing from my initial pass.

Partitioned spells need nothing here — the catalog already defaults them to their partition columns, and an explicit value only wins when set. Of the 190 relations in the curated override list, 43 still have no hint: 7 `utils.*` date spines where filtering is meaningless, 30 unpartitioned TON/thorchain incrementals owned by outside contributors, `dex_aggregator.trades` (which wants partitioning rather than a hint), and a handful of small views. Details in the thread.

Verified by parsing all five subprojects and asserting each model's `filtering_columns` in the manifest, and by checking every named column against both the view's SELECT output and the upstream `partition_by`. CI can't verify emission: the post-hooks are gated on `target.name == 'prod'`.

### DIFF
--- a/dbt_subprojects/dex/models/_projects/bitget_dex_aggregator/bitget_dex_aggregator_trades.sql
+++ b/dbt_subprojects/dex/models/_projects/bitget_dex_aggregator/bitget_dex_aggregator_trades.sql
@@ -2,6 +2,7 @@
     schema = 'bitget_dex_aggregator',
     alias = 'trades',
     materialized = 'view',
+    filtering_columns = ['block_month', 'blockchain'],
     post_hook='{{ expose_spells(
         blockchains = \'["bnb", "ethereum", "arbitrum", "polygon", "base"]\',
         spell_type = "project",
--- a/dbt_subprojects/dex/models/trades/dex_evm_trades.sql
+++ b/dbt_subprojects/dex/models/trades/dex_evm_trades.sql
@@ -4,6 +4,7 @@
 	schema = 'dex_evm'
 	, alias = 'trades'
 	, materialized = 'view'
+	, filtering_columns = ['block_month', 'blockchain', 'project']
 	, post_hook = '{{ expose_spells(blockchains = \'["' + chains | join('","') + '"]\',
 									spell_type = "sector",
 									spell_name = "dex_evm",
--- a/dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/gas_evm_fees.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/gas_evm_fees.sql
@@ -4,6 +4,7 @@
 	schema = 'gas_evm'
 	, alias = 'fees'
 	, materialized = 'view'
+	, filtering_columns = ['block_month', 'blockchain']
 	, post_hook = '{{ expose_spells(blockchains = \'["' + chains | join('","') + '"]\',
 									spell_type = "sector",
 									spell_name = "gas_evm",
--- a/dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/gas_fees.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/gas/fees/gas_fees.sql
@@ -3,6 +3,7 @@
 {{ config(
     schema = 'gas',
     alias = 'fees',
+    filtering_columns = ['block_month', 'blockchain'],
     post_hook='{{ expose_spells(blockchains = \'["' + chains | join('","') + '"]\',
                                 spell_type = "sector",
                                 spell_name = "gas",
--- a/dbt_subprojects/hourly_spellbook/models/_sector/nft/trades/nft_trades.sql
+++ b/dbt_subprojects/hourly_spellbook/models/_sector/nft/trades/nft_trades.sql
@@ -2,6 +2,7 @@
     schema = 'nft',
     alias = 'trades',
     materialized = 'view',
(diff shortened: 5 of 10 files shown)


a-monteiro:
* **#9910** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9910?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9910?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

jeff-dude:
Added `filtering_columns` to `dex_evm.trades` and `gas_evm.fees` because both relations are explicitly listed in Core’s static Curated Data overrides (`core-go/syncsense2/models/table_overrides.go`) but were missing from the initial coverage.

Both are user-facing views, so Catalog cannot derive filtering columns from their physical layout. Their upstream relations carry the useful partition structure:

- `dex_evm.trades` reads `dex.trades`, partitioned by `block_month`, `blockchain`, and `project`.
- `gas_evm.fees` reads `gas.fees`, whose chain-level parents are partitioned by `block_month`; `blockchain` also narrows the cross-chain union.

The same lineage audit also aligned `nft.trades` with its parent partition order and removed `block_date` from `nft.wash_trades`, since its parents are partitioned by `block_month` and that column is not exposed by the final view.

a-monteiro:
Reviewed ed6b9a3 and it all checks out — I verified each change rather than just pulling it.

`dex_evm.trades` and `gas_evm.fees` are both `materialized='view'` with `expose_spells`, and both select every column you named, so the hints are usable. Good catch on the two I missed. Reordering `nft.trades` to `['blockchain', 'project', 'block_month']` also makes the set consistent under a single rule — mirror the parent's partition order, which is what the catalog derives for the parent itself — and dropping `block_date` from `nft.wash_trades` is the honest call given the view never exposes `block_month`.

I missed those two because my starting point was a six-day-old local manifest plus `page_rank_score`, which is a worse selector than the curated override list. So I ran your list properly: 190 relations in `table_overrides.go`, of which 64 aren't spellbook models (prices, stablecoins, polymarket, kalshi, rwa, the non-EVM tokens schemas), 76 already derive hints from their partition columns, 7 are covered here, and **43 still have none**:

- 7 `utils.*` date spines — generated views, filtering is meaningless, I'd leave them alone.
- 30 TON and thorchain relations (`affluent_ton`, `evaa_ton`, `stormtrade_ton`, `tonco_ton`, `tradoor_ton`, `thorchain.defi_*`) — unpartitioned incrementals owned by outside contributors. A hint on an unpartitioned table only buys Delta file-statistics skipping, so partitioning them is the real fix and the column choice should come from whoever owns them.
- `dex_aggregator.trades` — the one that actually matters. Unpartitioned incremental, `page_rank_score` 4966, and its own `incremental_predicates` filter on `block_date`. It wants `partition_by=['block_month']` like `dex.trades`, not a hint; that needs a full refresh, so I left it out of a metadata-only PR.
- `aptos_fungible_asset.metadata_current`, `dns_ton.address_book`, `dns_ton.domain_latest_info`, `evaa_ton.assets`, `ton.accounts`, `ton.latest_balances`, `ton.prices_daily` — small views and tables where I don't have a defensible column list without knowing how they're queried.

Happy to take any of these in a follow-up if you want a particular slice; I didn't want to guess hints for tables I can't justify.

Three of my original candidates also came out of the set: `safe.safes_all`, `cex.addresses` and `addresses.stats` were switched to `hide_spells()` in #9900, so they're on the way out of the Data Explorer and a hint would be noise.

jeff-dude (approved):

jeff-dude:
@cursor review

PR #9915, opened 2026-08-02
fix(solana): dedupe Jupiter swap instructions

jeff-dude:
## What & why
Deduplicate AMM instruction candidates before assigning SwapEvent order.
Nested Jupiter route calls currently fan out the same instruction and fail the Solana production merge.
This restores one merge key per swap and allows incremental refreshes to complete.
- No visual surface (dbt model logic only)

## Architecture
- `dbt_subprojects/solana/models/jupiter/jupiter_v6_solana_swapsevent_aggregator_swaps.sql` - data-critical model; deduplicates physical AMM instructions before `ROW_NUMBER()` assigns swap order.
- Tested: selected model compiles; bounded DuneSQL over the last three days returns zero duplicate source keys; both failing transactions map three swaps to three distinct instructions.
- Needs human eyes: run a one-time full refresh after deploy to remove the two duplicate keys already stored in the August 1 production partition.
- Cross-system blast radius: `jupiter_v6_solana.swapsevent_aggregator_swaps` and downstream Jupiter aggregator swaps consumers; no schema change.

<details><summary>Agent context</summary>

- Reproduced from dbt Cloud run https://nd058.us1.dbt.com/deploy/58579/projects/372518/runs/70471891091654/ and Trino query `20260802_130440_00420_i9vf2`.
- Compile: `uv run dbt compile --select jupiter_v6_solana_swapsevent_aggregator_swaps --project-dir dbt_subprojects/solana/`.
- Regression: Dune CLI query over a three-day partition-bounded window returned `duplicate_key_groups = 0`.
- Failing transaction mappings changed from duplicate instructions to `[13, 20, 23]` and `[13, 22, 27]`, with three distinct merge keys each.
</details>

### DIFF
--- a/dbt_subprojects/solana/models/jupiter/jupiter_v6_solana_swapsevent_aggregator_swaps.sql
+++ b/dbt_subprojects/solana/models/jupiter/jupiter_v6_solana_swapsevent_aggregator_swaps.sql
@@ -90,18 +90,15 @@ WITH amms AS (
         amm
 )
 
-, amms_involved AS (
-    SELECT
+, amm_instructions AS (
+    SELECT DISTINCT
           a.block_date
         , a.block_slot
         , a.tx_index
+        , a.tx_id
         , a.executing_account AS amm
         , a.outer_instruction_index
         , a.inner_instruction_index
-        , ROW_NUMBER() OVER (
-            PARTITION BY a.block_date, a.tx_id, a.outer_instruction_index
-            ORDER BY a.inner_instruction_index ASC
-          ) AS rnk
     FROM {{ source('solana', 'instruction_calls') }} a
     INNER JOIN route_calls b
         ON  a.block_date = b.call_block_date
@@ -117,6 +114,21 @@ WITH amms AS (
       {% else %}
       AND a.block_date >= DATE '{{ project_start_date }}'
       {% endif %}
+)
+
+, amms_involved AS (
+    SELECT
+          a.block_date
+        , a.block_slot
+        , a.tx_index
+        , a.amm
+        , a.outer_instruction_index
+        , a.inner_instruction_index
+        , ROW_NUMBER() OVER (
+            PARTITION BY a.block_date, a.tx_id, a.outer_instruction_index
+            ORDER BY a.inner_instruction_index ASC
+          ) AS rnk
+    FROM amm_instructions a
 )
 
 , swap_amounts AS (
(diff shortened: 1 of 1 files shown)


jeff-dude:
## CI data validation

Queried `dune.dune_spellbook_ci__tmp_pr9915_30752516743_1.jupiter_v6_solana_swapsevent_aggregator_swaps` after the full build and incremental merge.

- Full build produced `406,514,182` rows.
- The full-table unique-key test passed before and after the incremental merge.
- Within production's available August 1 slot range (`436454957` to `436515035`), CI and production both contain `298,258` rows across `189,440` transactions.
- Raw input and output amount totals match exactly.
- Production contains `2` duplicate merge keys; CI contains `0`.
- The only key changes are the expected corrected instruction mappings:
  - `[13, 20, 20]` to `[13, 20, 23]`
  - `[13, 22, 22]` to `[13, 22, 27]`
- Token mints and swap ordering are unchanged.

Validation passes. A one-time full refresh is still required after deployment to remove the duplicate rows already persisted in production.

0xRobin (commented):
Verified against production data — dedup is correct. `amm_instructions` selects only from `a`, so the fanned-out copies are identical and `DISTINCT` collapses them; equivalent to deduping `route_calls` to `MIN(call_inner_instruction_index)` before the join.

**Pushed one follow-up (9ddd113).** `swap_amounts` ordered by `ord` alone, but `ord` restarts at 1 per SwapsEvent emission while the window spans all emissions at that outer instruction. Both failing txs have two emissions there, so `ord` was `1, 1, 2` and Trino broke the tie arbitrarily. This mis-ordered real production data: `3Wos1G5…` has legs 1 and 2 transposed relative to the raw events. Fixed by ordering on `evt_inner_instruction_index` first.

**CI verified** vs prod over Aug 1 slots `436454957`–`436515035`: 298,258 rows / 189,440 txs in both (unchanged by the tie-break, as expected — it only permutes `swap_order` within a partition), 0 duplicate keys in CI vs 2 in prod, and exact `DECIMAL(38,0)` amount sums identical (`7585552949499518611` / `11149820204905655816`).

Two description nits: nothing `ref()`s this model, so there are no downstream consumers; and Aug 1 is truncated at 07:02:31 UTC with ~1.1M rows missing, so the full refresh recovers a day of data rather than just dropping 2 keys.

Pre-existing, not from this PR: the recovered legs have NULL `amm_name` because `ALPHAQme…` is absent from `jupiter_solana.amms` — prod already has 69% NULL across 30 unmapped AMMs. Worth a separate PR.

LGTM.

0xRobin:
bugbot run

0xRobin:
@bugbot review

0xRobin (approved):

PR #9926, opened 2026-08-08
fix config keys dbt silently discards (columns, tag)

a-monteiro:
dbt treats unrecognized keys under `config:` as custom keys and silently discards them (`CustomKeyInConfigDeprecation`, D026 — dbt Cloud has been warning about this on every run). A scan across all five subprojects found 8 real occurrences, and they weren't cosmetic: config dbt was throwing away.

Three spine models (`chain_swap_trades`, `dex_solana_bot_trades`, `dex_evm_bot_trades`) had their whole `columns:` block — including the YAML anchors the per-project sibling models alias — nested under `config:`, so none of those column descriptions ever made it into the manifest. The block moves to the model level; anchors still precede their aliases, so the siblings are untouched.

Four solana `*_bot_users` models buried a `user` column with an intended `unique` test inside `config:`. For `soul_sniper` and `alpha_dex` that was their only test, so the constraint has never run. The column is now live (and the scalar `data_tests: unique` shorthand corrected to a list, which is what dbt requires once the key is actually read). This PR's CI runs those four unique tests for the first time — if one fails, the data violates a constraint its author intended, which is worth knowing.

`ajna_erc20_pools` used singular `tag:`, so its four tags were never applied; renamed to `tags:`. None of them is used by any prod/CI selector (those only use `prod_exclude`, `remove`, `static`), so the only effect is a one-time `state:modified` flag.

### DIFF
--- a/dbt_subprojects/daily_spellbook/models/ajna/ajna_schema.yml
+++ b/dbt_subprojects/daily_spellbook/models/ajna/ajna_schema.yml
@@ -7,7 +7,7 @@ models:
       project: ajna
       contributors: [gunboats]
     config:
-      tag: ['ajna', 'cross-chain', 'borrowing', 'lending']
+      tags: ['ajna', 'cross-chain', 'borrowing', 'lending']
       description: 'All deployed ERC-20 pools across all chains on Ajna'
     data_tests:
       - dbt_utils.unique_combination_of_columns:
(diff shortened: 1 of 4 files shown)


a-monteiro:
* **#9926** <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9926?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/duneanalytics/spellbook/9926?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

0xRobin (approved):

PR #9937, opened 2026-08-15
feat(oneinch): decode OKX DexRouter swaps in project_orders

mrAndreyIsachenko:
OKX's DexRouter contracts were mapped in oneinch's contract list but their swaps were never decoded into oneinch.project_orders. This adds that decoding.

The router emits one event, OrderRecord, on every swap method (dagSwapTo, smartSwapTo, unxswapTo, and a few more), so a single event config covers all of them. Also adds one OKX router address on ethereum that wasn't in the contract list yet.

Checked against real transactions on ethereum, bnb, optimism, arbitrum, base, polygon and avalanche_c: the decoded token pair, amounts, and sender matched the actual on-chain transfers in each case.

Two things worth flagging for review:
- avalanche_c, arbitrum and base share the same router address (same contract deployed at the same address on all three).
- On linea, fantom and zksync the mapped router addresses look correct (same contract, same functions) but have had no real traffic since November 2025. OKX seems to have moved to different addresses there that aren't mapped yet - that would need a follow-up PR to find and add the current ones.


mrAndreyIsachenko:
recheck

mrAndreyIsachenko:

mrAndreyIsachenko:
A note on CI: this change touches three shared macros used by every oneinch chain (mapped_contracts, cfg_methods, cfg_events), so `state:modified.macros` selects all 52 downstream models for a full rebuild in a fresh CI schema - not an incremental run. That's a lot of raw_logs/raw_traces history to scan across 13 chains, and it can run past the 90-minute CI limit before finishing.

Everything that completes builds successfully with no errors - the run just may not finish in time. The decoding itself has been checked separately against real transactions on ethereum, bnb, optimism, arbitrum, base, polygon and avalanche_c (decoded fields matched the actual on-chain transfers in every case) - see the PR description for details. Happy to re-run CI again or split the change up if that helps get a full green run, whichever is easier on your end.

dennisp42:
decoding is right, coverage is not. the OrderRecord layout checks out: topic0 is the keccak of OrderRecord(address,address,address,uint256,uint256), the data words are fromToken, toToken, sender, fromAmount, returnAmount in that order, and returnAmount matches an erc20 transfer of toToken in the same tx on every non-native leg.
https://dune.com/queries/8589337
https://dune.com/queries/8589373

but the OKXDEX addresses in mapped_contracts are the retired routers. on 2026-09-01 the live okx.DexRouter contracts emit the OrderRecord volume and none of them is mapped:
bnb 0x5994814f2c4040b863a0125a45de152a8c2a4dec, 99.5%
ethereum 0x8feab81d36e7576107d5de0758c1b839be31b4f6, 99.9%
base 0x67d03631fe51b741c0c00c4e16eb662ac84381df, 100%
arbitrum 0x09f94b5fc68e227c323a6fbae3bd98c97fd8c849, 97.9%
polygon 0x3c4829196bfadff4394726b45159aeaac6fcd41c, 99.9%
optimism 0x1f5b43127414e36c31ecb5ff5567262997cd24d0, 100%
avalanche_c 0x8adfb0d24cdb09c6eb6b001a41820ece98831b91 and 0xab96dcfa7a7d669d9bf5918fab8641479973dd0a, 68.6% and 31.4%
the mapped set gets 0.1% or less on every chain.
https://dune.com/queries/8589357
https://dune.com/queries/8589389

ci says the same: 7 days across ethereum, bnb, base, arbitrum and avalanche_c decoded 554 OKXDEX orders in total, nearly all smartSwapByInvestWithRefund. the live bnb router alone emitted 27,803 OrderRecords in the first six hours of 2026-09-01.
https://dune.com/queries/8589342

hold until the live routers are in mapped_contracts. the same selectors work on them (dagSwapTo 0x0c307f76 and dagSwapByOrderId 0xf2c42696 carry the volume), so it is address rows only, in a macro this pr already edits. shipping OKXDEX at 0.1% coverage into project_orders reads as okx being tiny, which is worse than absent.

flagging: the 45 null-decode rows on bnb in ci are all reverted calls (call_success false), filtered downstream by where call_success. no action.
https://dune.com/queries/8589377

nit: pr body says avalanche_c, arbitrum and base share a router address. in the macro it is fantom, arbitrum and base (0x5e2F47bD7D4B357fCfd0Bb224Eb665773B1B9801); avalanche_c is 0x79f7C6C6dc16Ed3154E85A8ef9c1Ef31CEFaEB19.

ci window: target.name == 'ci' gate, prod target is prod, same pattern as ~57 other models in the repo. fine, and it bounds ci for every future oneinch project_orders pr, which is the point.

PR #9952, opened 2026-08-20
docs(solana): document rewards commission_bps

bh2smith:
## Summary
- add `commission_bps` to the `solana.rewards` source schema
- document the SIMD-0291 cutover and cross-era basis-points query
- keep this PR in draft until the M1 schema and ingestion change are live

## Test plan
- [x] Parse `solana_base_sources.yml` as YAML
- [x] Run `git diff --check`
- [x] Confirm M1 is live before marking ready for review

Related: DF-972


jeff-dude (approved):
will await word on deploy based on PR desc.

PR #9966, opened 2026-08-23
feat(oneinch): decode Native Router V3 and V4 into project_orders

mrAndreyIsachenko:
## What this fixes

Native is currently decoded as **nothing** in `project_orders`. The only mapped tag is
`NativeRouterV1`, and V1 has had **zero calls in the last 30 days** on ethereum, bnb and
base — checked in `<chain>.traces`, not inferred from empty output. All of Native's
traffic runs through V3 and V4, neither of which is mapped:

| chain | V3 calls (30d) | V4 calls (30d) |
| --- | --- | --- |
| bnb | 7,046,077 | 151,279 |
| ethereum | 9,880 | 21,209 |
| arbitrum | 5,584 | 9,712 |
| base | 3,954 | 4,059 |

Addresses come from
[docs.native.org](https://docs.native.org/native-dev/resources/addresses) and each was
confirmed against real traffic before being added. Robinhood, Monad, Xlayer and Morph run
V3/V4 too, but have no `mapped_contracts` entries here for any protocol, so they are out
of scope for this PR.

## The decode

V3/V4 use a different `tradeRFQT` selector, `0x0947c2d9`, with four extra `uint256` in the
order tuple, so V1's offsets do not transfer:

```
V1     tradeRFQT((address x5, uint256 x5, bytes16, bool, bytes, (address,address,uint256), bytes, bytes, uint256))
V3/V4  tradeRFQT((address x5, uint256 x9, bytes16, bool, bytes, (address,uint256), bytes), uint256, uint256)
```

The field map was decoded against real calldata and then checked against the ERC20
Transfer logs of the same transaction
(`0x9c13db5e2c1f59ddc4afe3e8a19c8d4c646e42964eb1f6615b405a567e6b204e`, ethereum V4): tuple
`idx5` = 7,624,635,000 and `idx6` = 3,102,840,007,606,819,000 are exactly the USDC and WETH
amounts that moved, and `idx3`/`idx4` are those two tokens. Field roles were then confirmed
across 32 further calls by joining each call to the transfer carrying its own amount:
`idx2` is the address receiving `maker_asset` in 21 of them, and never `idx0` or `idx1`.

Two deliberate omissions, both measured rather than assumed:

* **`maker` is unset.** The address that actually sends `maker_asset` is a Native vault
  that appears nowhere in the tuple head. `idx7` is not an amount either — it came out at
  exactly 1% of `idx6`.
* **The amounts are `*_max_amount`, not executed amounts**, matching how the existing V1
  `tradeRFQT` entry declares the same fields. Read as executed amounts they matched a real
  Transfer of the same token in only 169 of 300 calls (`idx6`) and 203 of 300 (`idx5`), the
  rest showing a transfer of that token at a different value. The executed amount cannot be
  read from the call either: `output` is NULL on all 300 sampled calls, so
  `substr(output, ...)` has nothing to read.

`event` is null because no event was confirmed for this selector, and a wrong topic
silently drops rows.

## Verification

Built with these entries on a private copy of this tree: **60,672 Native orders decoded**
in a three-day window — bnb V4 26,918, bnb V3 18,532, base V3 6,367, ethereum V3 3,944,
ethereum V4 3,295, base V4 1,616 — with both assets, both amounts and an order hash
present on 100% of rows. Existing `NativeRouterV1` decoding is untouched.

### DIFF
--- a/dbt_subprojects/daily_spellbook/macros/project/oneinch/_meta/oneinch_mapped_contracts_macro.sql
+++ b/dbt_subprojects/daily_spellbook/macros/project/oneinch/_meta/oneinch_mapped_contracts_macro.sql
@@ -1,4 +1,13 @@
 -- this macro helps to optimize the etl in case of adding new contract to a certain chain so it triggers pipeline only for this chain
+{#
+    NATIVE ROUTER V3/V4. Native's V1 router is dead — zero calls in the 30 days before
+    2026-08-23 on ethereum, bnb and base — while V3 and V4 carry the traffic, so Native is
+    currently decoded as nothing at all. Addresses are from
+    docs.native.org/native-dev/resources/addresses and were each confirmed against real
+    traffic in <chain>.traces over that window before being added; bnb V3 alone took
+    7,046,077 calls. Robinhood, Monad, Xlayer and Morph run V3/V4 too but have no
+    mapped_contracts entries here for any protocol.
+#}
 {% macro oneinch_mapped_contracts_macro(blockchain) %}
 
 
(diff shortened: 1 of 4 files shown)


mrAndreyIsachenko:
A note on CI: this change edits two shared macros used by every oneinch chain (mapped_contracts, cfg_methods), so `state:modified.macros` selects all downstream models for a full rebuild in a fresh CI schema instead of an incremental one. That can take longer than the 90-minute CI limit to finish, same structural reason as PR #9937 on this same subtree.

Everything that completes builds successfully with no errors - the run just may not finish in time. The decode itself was checked separately against real calldata and the matching ERC20 Transfer logs on ethereum, bnb, arbitrum and base (details in the PR description). Happy to re-run CI or split the change up if that helps get a full green run.

dennisp42 (approved):
offsets check out. 7,166 tradeRFQT (0x0947c2d9) calls to the ethereum v3/v4 routers over 2026-08-30..09-01: the tuple offset word is 0x60 on 100%, the address at idx2 receives maker_asset in 349/351 (v3) and 6,363/6,815 (v4) txs, the misses fit inside the native-eth-leg count (15 and 974), 7,166 distinct order hashes, none zero.
https://dune.com/queries/8589336

ci built 93,773 native rows across ethereum, bnb, base and arbitrum in the 7-day window, distinct on the unique key, zero nulls on assets, amounts and order_hash.
https://dune.com/queries/8589340

no double counting: prod project_orders has zero Native rows in the last 30 days, the new addresses and the selector appear nowhere else in the repo, the v1 entry is untouched.
https://dune.com/queries/8589313

flagging: amounts are max amounts as documented, and that is not a corner case. the received transfer equals maker_max_amount in 55% of v4 fills (3,738/6,815) and 70% of v3 (245/351), the rest are lower. making_amount and taking_amount for native v3/v4 overstate executed size on roughly half the fills and amount_usd downstream inherits it. worth one line in the schema description so consumers know.

nit: pr says maker is unset, but the orders macro coalesces maker to call_to, so maker comes out as the router address (null_maker = 0 in ci). say "maker = router" in the comment, or read the vault from the transfer side in a follow-up.

nit: the header comment in mapped_contracts is a dated essay. trim to the two lines that matter (v1 dead, v3/v4 addresses from docs.native.org).

merges clean with the okx pr, both add the same ci window hunk. approve with the nits.

PR #9972, opened 2026-08-28
fix(oneinch): size swaps by net token flow, not the single biggest transfer

mrAndreyIsachenko:
## What this fixes

`oneinch_project_swaps` sizes a swap by the biggest single token transfer inside the call. A flash loan borrowed and repaid inside the same call is two transfers of the same token through the same address, and the loan itself gets counted as "the swap."

Root-caused as [A1-2035](https://1inch.atlassian.net/browse/A1-2035): on Base since June 1, roughly **86% of reported Uniswap volume** was one repeated pattern — a bot borrows WETH from a Morpho Vault and repays it in the same transaction, and that leg (tens of thousands of WETH) got recorded as the swap size instead of the real trade underneath it (often under a dollar). It read as Uniswap losing market share to competitors; nothing had actually changed.

## The fix

Size by **net flow per (token, address)** within the call instead of the single biggest transfer: sum every transfer of a token through an address with sign (received minus sent), then take the biggest net across (token, address) pairs — same bucketing as before (call/user/caller/contract, trusted/untrusted).

A flash loan's borrow and repayment net to approximately zero for whichever address routes it. An ordinary swap has exactly one transfer per (token, address) pair, so net flow equals that transfer's amount and nothing changes for it.

## Verification

**Reference transaction from A1-2035** (`0x0dc393c5d2d6862c32b48783989d077a69c40dcfe8c0759536a3bb9a0256d2d6`, Base): the affected call goes from **$134,806,856.29 to $0.33** — matching the numbers already reported in that ticket exactly. The transaction's other, genuinely small swap call ($0.84) is untouched.

**Regression check across every Uniswap call on Base for the same day** (130,112 calls, compiled and run directly against production data, not a sample): `new_amount_usd / old_amount_usd` — p10 = 1.0000, median = 1.0000, p90 = 1.0000002. Ordinary swaps are unaffected. Total volume for the day drops from $2.17b to $293m, which is the point — that gap was never real, only 14 calls that day exceeded $1m under the old sizing.

## Known cost, not fixed here

An address that swaps a token out and back into itself inside one call (round-trip arbitrage) nets to ~zero here too, understating that genuine volume. There is no way to keep that and fix the flash-loan case from net flow alone — same trade-off already recorded in A1-2035.

### DIFF
--- a/dbt_subprojects/daily_spellbook/macros/project/oneinch/project/oneinch_project_swaps_base_macro.sql
+++ b/dbt_subprojects/daily_spellbook/macros/project/oneinch/project/oneinch_project_swaps_base_macro.sql
@@ -182,31 +182,48 @@ meta as (
         {% if is_incremental() -%} and {{ incremental_predicate('block_time') }} {%- endif %}
 )
 
-, joined as (
+, call_transfers as (
+    -- Every transfer nested under a swap call, before any sizing decision -- shared by
+    -- the token/sender/receiver labels below (unchanged) and the net-flow sizing that
+    -- replaces the old max-single-transfer approach (see net_legs).
     select
-        blockchain
-        , swaps.block_month
+        swaps.block_month
         , swaps.block_number
         , swaps.tx_hash
         , swaps.call_trace_address
         , swaps.call_trade_id
+        , swaps.call_from
+        , swaps.call_to
+        , contract_address
+        , native
+        , symbol
+        , native_symbol
+        , contract_address_raw
+        , transfer_from
+        , transfer_to
+        , amount
+        , block_time
+        , minute
+    from swaps
+    join transfers on true
+        and swaps.block_month = transfers.block_month
+        and swaps.block_number = transfers.block_number
+        and swaps.tx_hash = transfers.tx_hash
+        and slice(transfer_trace_address, 1, cardinality(call_trace_address)) = call_trace_address -- nested transfers only
+        and reduce(array_distinct(call_trace_addresses), call_trace_address, (r, x) -> if(slice(transfer_trace_address, 1, cardinality(x)) = x and x > r, x, r), r -> r) = call_trace_address -- transfers related to the call only
+        and (order_hash is null or contract_address in (_maker_asset, _taker_asset) and cardinality(array_intersect(array[call_from, maker, taker], array[transfer_from, transfer_to])) > 0) -- transfers related to the order only
+)
+
+, labels as (
+    -- Token/sender/receiver labels, unchanged from before this fix -- these are display
+    -- lists, not sizing, so they stay keyed off individual transfers.
+    select
+        block_month
+        , block_number
+        , tx_hash
+        , call_trace_address
+        , call_trade_id
         , any_value(block_time) as block_time
-        , any_value(tx_from) as tx_from
-        , any_value(tx_to) as tx_to
-        , any_value(project) as project
-        , any_value(tag) as tag
-        , any_value(flags) as flags
-        , any_value(call_from) as call_from
-        , any_value(call_to) as call_to
-        , any_value(call_selector) as call_selector
-        , any_value(method) as method
-        , any_value(order_hash) as order_hash
-        , any_value(maker) as maker
-        , any_value(maker_asset) as maker_asset
-        , any_value(making_amount) as making_amount
-        , any_value(taker_asset) as taker_asset
-        , any_value(taking_amount) as taking_amount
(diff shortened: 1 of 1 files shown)


mrAndreyIsachenko:
**Correction to a number I quoted earlier in this PR: the effect on volume is +0.218%, not +1.63%.**

The +1.63% was one day of ethereum alone — that is where the wrapped-native double count was found, and ethereum carries more wrapping and more flash-loan activity than the average chain, so it overstates the change by roughly seven times.

Measured on a whole month across all 13 chains, old and new formulas computed against the same data:

| 2025-06, 13 chains | rows | total |
|---|---|---|
| `max()` | 266,591,088 | $327,588,940,515.19 |
| net flow | 266,591,088 | $328,303,462,331.25 |
| | **+0** | **+0.218%** |

Row count is identical: this re-sizes swaps, it does not add or drop any.

An earlier figure of **−0.978%** also appeared here and was wrong in sign as well as magnitude — it came from the collapse dropping transfers shared between calls, fixed in `869e383f2`. It has been removed from the macro comment, which still carried it.

Context for the numbers above: this change has now been used to re-state 32 months of history — 2024-01 through 2026-08, all 13 chains, ~2.4bn swap rows.

mrAndreyIsachenko:
## Net flow zeroes out routers, not just flash loans — ZeroEx loses 29.5% of a day's volume

We have now run this change over 32 months of history and reconciled the result against Spellbook. Composition is unaffected — row counts and distinct tx match exactly, per project. But the sizing moves one project a long way, and the mechanism looks like a design problem rather than a bug in the patch.

**ethereum, 2026-08-19, our build vs `oneinch_ethereum.project_swaps`:**

| project | rows (ours/Dune) | USD ours | USD Dune | Δ |
|---|---|---|---|---|
| CoWSwap | 4,971 / 4,971 | $186,393,516 | $186,215,173 | +0.10% |
| Kyber | 11,791 / 11,791 | $158,605,090 | $159,122,712 | −0.33% |
| Uniswap | 52,128 / 52,128 | $155,396,246 | $155,128,598 | +0.17% |
| **ZeroEx** | **13,287 / 13,287** | **$86,808,814** | **$123,137,885** | **−29.50%** |
| Bebop | 3,043 / 3,043 | $119,800,885 | $117,748,149 | +1.74% |
| Paraswap | 2,191 / 2,191 | $52,470,663 | $52,574,198 | −0.20% |

ZeroEx: **5,412 of 13,287 swaps got smaller**, 15 got bigger, net −$36.3m on the day.

### The mechanism, on one transaction

`0xf8d47257f2622342bb059dc36590008c18009b29…`, call `[0,1,2,1,4,1]` — Dune sizes it **$1,829,704**, we size it **$588,203**.

It is an ordinary multi-hop swap. The router `0x0889e932…` takes 800 WETH in and spreads it across pools. Its net flow inside the call, per token:

```
USDC   in 1,824,246.5571   out 1,824,246.5571   net 0.0000
USDT   in 1,039,352.0732   out 1,039,352.0732   net 0.0000
USDS   in   588,502.8531   out   588,502.8531   net 0.0000
DAI    in    45,536.1462   out    45,536.1462   net 0.0000
WETH   in       800.0000   out       800.0000   net 0.0000
ETH    in        99.9483   out        99.9483   net 0.0000
WBTC   in         2.3861   out         2.3861   net 0.0000
```

**Every token nets to exactly zero.** The router holds nothing — that is what a router does. `max()` sized this swap at the 800 WETH that moved through it; net flow discards the router entirely and falls back to whatever leg is left, a third of the size.

### Why this is not the documented trade-off

The comment in this PR names the cost as an address that "swaps a token out and back into itself within one call (round-trip arbitrage)". That reads as an edge case. But a transit router is structurally identical to a flash loan under net flow — in and out of the same token, netting to zero — and routing is the common path, not an edge case. On this one day it is 41% of ZeroEx's swaps.

### An idea, explicitly not measured

The two cases may be separable: a flash loan borrows from and repays **the same counterparty**, while a router takes a token from one address and sends it to a **different** one. Counterparty identity is in the data already.

I have not tested this and am not proposing it as a fix — flagging it only because the distinction exists. (I have already made the mistake in this PR of proposing an approach before measuring it, and it was worse than what it replaced.)

### Context

Everything above is on the branch as it currently stands, with the five defects found while porting it fixed. The aggregate figure I posted earlier — **+0.218%** on a full month across 13 chains — holds, but it is a sum of movements in both directions, and a ~30% swing on one project is invisible inside it.

PR #9974, opened 2026-08-29
fix(dex): add 8 Base platform models that were never wired into dex_base.base_trades

BhariGowda:
### Description:

Eight Base platform models exist with `_schema.yml` entries, seed files and seed tests, are not
tagged `prod_exclude`, and are not referenced by `dex_base_base_trades.sql` or by anything else in
the repo. They are built in prod and feed nothing, so their trades have never reached `dex.trades`.

This wires seven of them into the Base union. The eighth, `gridex`, turned out to have 15 trades in
its entire lifetime once CI could build it, so it is tagged `prod_exclude` with the reason instead.
Three small supporting changes are included: two the wiring needs, and one cosmetic tidy.

| model | project / version | factory | in this PR |
| --- | --- | --- | --- |
| `baseswap_basex_base_base_trades` | baseswap / basex | `0xdC323d16C451819890805737997F4Ede96b95e3e` | wired |
| `baso_finance_base_base_trades` | baso_finance / 1 | `0xcefbebf0b85b1638c19b01ce2a02c262f421b07d` | wired |
| `crescentswap_base_base_trades` | crescentswap / 1 | `0x3539dA2AdB3f8311D203D334f25f7Bee604A5c50` | wired |
| `energon_base_base_trades` | energon / 1 | `0xBfAe4f07C099798F23f5aC6773532fB637B68Ad7` | wired |
| `gridex_base_base_trades` | gridex / 1 | `0x32d1f0dce675902f89d72251db4ab1d728efa19c` | prod_exclude |
| `icecreamswap_base_base_trades` | icecreamswap / 1 | `0x9E6d21E759A7A288b80eef94E4737D313D31c13f` | wired |
| `synthswap_base_base_trades` | synthswap / 2 | `0x4bd16d59A5E1E0DB903F724aa9d721a31d7D720D` | wired |
| `voltswap_base_base_trades` | voltswap / 2 | `0x2A5478bE24F9E536cCb91DBF650EFD6cE6C00398` | wired |

### Why this looks like an oversight rather than a deliberate exclusion

- I diffed every chain under `dex/models/trades`: each `platforms/*.sql` against the `ref()` list in
  that chain's `dex_<chain>_base_trades.sql`. Base is the only chain with unexplained misses.
  Linea's one miss (`lynex_linea_base_trades`) carries `tags=['prod_exclude']` and a stated reason.
  Tron's three are intermediates consumed by their own `_base_trades` models. These eight carry no
  tag and no comment.
[...]

PR #8276, opened 2025-05-27
CUR2-153 Adding dex.trades views for Plume and rooster protocol

thecaptncrunch:
## Thank you for contributing to Spellbook 🪄
Please open the PR in **draft** and mark as ready when you want to request a review. 

### Description:

[...]

---
quick links for more information:
- [README.md](https://github.com/duneanalytics/spellbook/blob/main/README.md)
- [spellbook docs](https://github.com/duneanalytics/spellbook/tree/main/docs)
- [CONTRIBUTING.md](https://github.com/duneanalytics/spellbook/blob/main/CONTRIBUTING.md)

PR #5319, opened 2024-02-09
add horizondex on base

pachecospig:
# Thank you for contributing to Spellbook!

Thank you for taking the time to submit code in Spellbook. A few things to consider:
- If you are a first-time contributor, please sign the CLA by copy & pasting **_exactly_** what the bot mentions in PR comment
- Refer to [docs](#spellbook-contribution-docs) section below to answer questions
- Dune team will review submitted PRs as soon as possible

## Spellbook sub-projects

To prepare for Spellbook to scale in size, Dune team has started to pilot sub-projects:
- For a quick overview, see [the GH discussion on the topic](https://github.com/duneanalytics/spellbook/discussions/5238)
    - Please ask questions and continue the conversation there
- The first spell lineage to move: `tokens.erc20` and all upstream spells which help build it
    - **_Note_**: Please navigate to [tokens](https://github.com/duneanalytics/spellbook/tree/main/tokens) root directory to find tokens objects

## Spellbook contribution docs

The [docs](https://github.com/duneanalytics/spellbook/tree/main/docs) directory has been implemented to answer as many questions as possible. Please take the time to reference each `.md` file within this directory to understand how to efficiently contribute & why the repo is designed as it is 🪄

Example questions to be answered:
- What does each property in the [model config block](https://github.com/duneanalytics/spellbook/blob/main/docs/models/model_config_block.md) mean?
- What is the [CI test](https://github.com/duneanalytics/spellbook/blob/main/docs/ci_test/ci_test_overview.md) attached to PRs and how can I best utilize it?
- What are the Spellbook [best practices](https://github.com/duneanalytics/spellbook/blob/main/docs/general/best_practices.md)?

Please navigate through the [docs](https://github.com/duneanalytics/spellbook/tree/main/docs) directory to find as much info as you can.

**Note**: happy to take PRs to improve the docs, let us know 🤝

PR #9975, opened 2026-08-29
fix(dex): remove whitespace padding from four dex.trades seed rows

BhariGowda:
### Description:

Four seed rows under `dbt_subprojects/dex/seeds/trades/` carry a leading or trailing space inside a
field. Same class as the `icecreamswap` seed fix in #9974, but with an important difference: all four
of these models are **wired and live in production today**, where icecreamswap was never wired at all.

| seed | line | column | value | introduced | age |
| --- | --- | --- | --- | --- | --- |
| `horizondex_base_base_trades_seed.csv` | 2 | `tx_hash` | `0x0936…b75e` + trailing space | #5319, 2024-02-15 | ~30 months |
| `rooster_protocol_plume_base_trades_seed.csv` | 2 | `token_sold_address` | leading space + `0x593c…88db` | #8276, 2025-06-12 | ~14 months |
| `machinex_v3_peaq_base_trades_seed.csv` | 2 | `block_date` | `2025-09-29` + trailing space | #8782, 2025-09-30 | ~11 months |
| `quickswap_v3_somnia_base_trades_seed.csv` | 6 | `block_date` | `2025-09-01` + trailing space | #8820, 2025-10-01 | ~11 months |

`check_dex_base_trades_seed` matches on `tx_hash` and `evt_index` and checks `block_number`, both
token addresses and both raw amounts. So the horizondex padding sits on a **matching** column and the
rooster padding on a **checked** column. The two `block_date` ones sit on a column the test neither
matches nor checks, so they are hygiene rather than a broken assertion, but they are the same defect
and the same directory.

### Why this PR opens with a diagnostic commit

I can trace the whitespace deterministically up to the point it reaches DuneSQL, and no further:

- `column_types` keys become `text_columns` (`dbt/context/providers.py:1333`), which forces every
  named column to `agate.data_types.Text` in `build_type_tester`.
- `agate.Text.cast()` returns `str(d)` and strips only for the null-value comparison, so the padding
  survives CSV parsing into the seed table.
- `trino__load_csv_rows` → `create_bindings` takes the non-varchar branch for a `varbinary` or
  `timestamp` column and emits the value inside a type-constructor literal, so the seed row becomes
  `VARBINARY '0x0936…b75e '` with the space inside the literal.

What I cannot determine locally is what DuneSQL does with that: reject it, decode it to different
bytes, or trim it. Those three give completely different answers about whether anything is actually
broken, and I did not want to assert one. These models' seed tests only run when something attached
to them is modified, and nothing has been for the ages above, so there is no recent CI evidence either.

So the first commit only adds `-- ci-stamp: 1` to the four models, changing no data. That is the
mechanism `AGENTS.md` documents for pulling a model into `state:modified.body`, and it makes CI build
them and run their seed tests against the padded seeds exactly as they stand today. Whatever that run
reports is the real before state:

- **seed load fails** → DuneSQL rejects the padded literal.
- **seed loads, test fails** → the padded value decodes to different bytes and the row never matches.
- **everything passes** → DuneSQL trims, the padding is harmless, and I will say so plainly and close
  this rather than ship a fix for a non-problem. That outcome would also mean the corresponding
  sentence in #9974's description overstates the icecreamswap case, and I would correct it there.

The second commit removes the whitespace and the stamps, so the net diff is four one-character seed
edits and nothing else.

### Not touched

Five more padded fields exist in `dbt_subprojects/dex/seeds/_project/zeroex/*_sample.csv`
(`zeroex_native_fills_sample`, the bnb/polygon/ethereum native-fills samples, and
`zeroex_base_api_fills_sample`). Those feed a different test family, so I left them out rather than
widen this PR. Happy to take them in a follow-up if you want them.

### Testing

- `dbt deps`, `dbt parse` clean on the `dex` sub-project.
- No Trino credentials in my environment, so the before and after measurements come from this PR's
  own CI runs rather than from anything I ran locally.

### DIFF
--- a/dbt_subprojects/dex/models/trades/base/platforms/horizondex_base_base_trades.sql
+++ b/dbt_subprojects/dex/models/trades/base/platforms/horizondex_base_base_trades.sql
@@ -10,6 +10,8 @@
     )
 }}
 
+-- ci-stamp: 1
+
 WITH dexs AS
 (
     SELECT
--- a/dbt_subprojects/dex/models/trades/peaq/platforms/machinex_v3_peaq_base_trades.sql
+++ b/dbt_subprojects/dex/models/trades/peaq/platforms/machinex_v3_peaq_base_trades.sql
@@ -10,6 +10,8 @@
     )
 }}
 
+-- ci-stamp: 1
+
 {% set machinex_start_date = "2025-05-25" %}
 
 WITH pool_tokens AS (
--- a/dbt_subprojects/dex/models/trades/plume/platforms/rooster_protocol_plume_base_trades.sql
+++ b/dbt_subprojects/dex/models/trades/plume/platforms/rooster_protocol_plume_base_trades.sql
@@ -10,6 +10,8 @@
     )
 }}
 
+-- ci-stamp: 1
+
 {{
     maverick_compatible_v2_trades(
         blockchain = 'plume',
--- a/dbt_subprojects/dex/models/trades/somnia/platforms/quickswap_v3_somnia_base_trades.sql
+++ b/dbt_subprojects/dex/models/trades/somnia/platforms/quickswap_v3_somnia_base_trades.sql
@@ -10,6 +10,8 @@
     )
 }}
 
+-- ci-stamp: 1
+
 {% set quickswap_start_date = "2025-08-29" %}
 
 WITH dexs AS (


BhariGowda:
Measured, and the answer is that there is nothing here to fix. Closing.

The first commit stamped the four models so CI would run their seed tests against the padded seeds
untouched. Run
[33250867288](https://github.com/duneanalytics/spellbook/actions/runs/33250867288) is green, and the
tests genuinely ran rather than being skipped:

| model | built | `check_dex_base_trades_seed` |
| --- | --- | --- |
| `horizondex_base_base_trades` | 308,491 rows | pass |
| `rooster_protocol_plume_base_trades` | 6,209,991 rows | pass |
| `machinex_v3_peaq_base_trades` | 672,554 rows | pass |

The padded rows were inside the compared set rather than filtered out, which is the thing that would
have made a green result meaningless. horizondex's padded row is `version` 1 and its seed filter is
`version: 1`; rooster's is `version` 2 against a `version: 2` filter. So
`VARBINARY '0x093685…d260b75e '` with the trailing space still matched the model's real 32 byte
hash, and the leading space on rooster's `token_sold_address` still compared equal.

DuneSQL trims the padding when it casts, so the whitespace is cosmetic. That settles it in the
direction I said in the description I would accept, so I am closing rather than shipping a cosmetic
diff against four live models.

I have also corrected the sentence in #9974 that asserted the opposite about the icecreamswap seed.
That edit stays in as a tidy, but it was not a fix and the description should not have claimed it
was. It was inference rather than measurement on my part, and this run is what measured it.

One unrelated thing this surfaced, worth separating from the above because unlike the whitespace it
is real. `quickswap_v3_somnia_base_trades` has a seed file registered at
`dbt_subprojects/dex/seeds/trades/_schema.yml:5706` and a live wired model, but
`dbt_subprojects/dex/models/trades/somnia/_schema.yml` declares only
`dbt_utils.unique_combination_of_columns` for it, with no `check_dex_base_trades_seed`. In this run
its seed was the one of the four that never loaded and its output was never compared against
anything. Happy to send that as a separate small PR if it is wanted.

PR #9969, opened 2026-08-25
1inch-U91: add cronos, monad, hyperevm, arc to the 1inch lineage (AR, LO) + fix robinhood mode classification

dodgervl:
### Description:

Adds **cronos**, **monad**, **hyperevm** and **arc** to the main 1inch lineage: AggregationRouterV6
(AR) and Limit Order Protocol (LO), following the standard per-chain pattern — 9 models per
chain (raw_calls / stream / executions / transfers / LOP trades), meta config entries
(exposed `["ar", "lo"]` like fantom, start `2026-08-01`; arc `2026-09-01`), source declarations,
and the LOP passthrough wired into `dex_<chain>_trades` (cronos / monad / hyperevm; see the arc
paragraph for why arc has none yet). Cross-chain (CC) models are **deferred** until the escrow
decodings are sorted out separately.

The intent stack on these chains is deployed at non-canonical addresses shared by all
four (same deployer): settlement `0x65497e56...` and the Limits/Fusion/CrossChain
access tokens — not the canonical `0xacce55...` vanity set. The router is the canonical
`0x1111111254...` on cronos and monad; hyperevm has a chain-specific router
(`0x5281602a...`). Cronos uses `base_transfers` (no `transfers_from_traces` table),
same handling as robinhood. The ERC20True placeholder on these chains
(`0x175a30f8...`) is added to `oneinch_cross_chain_placeholder_tokens_cfg_macro` so the
LOP own-trades / venue-settled / AR-trades placeholder guards and the
`oneinch_no_cross_chain_placeholder_tokens` test cover them.

**Arc** (Circle's L1, chain id 5042, USDC as gas) joins the same set, added 2026-09-17 after
its 1inch decodings landed on Dune. Every 1inch contract there deployed 2026-09-02, first
fills 2026-09-10. Same 9 models from the hyperevm template, same non-vanity settlement /
access-token / ERC20True-placeholder deployments as the other three. Chain specifics:

- AggregationRouterV6 is chain-specific (`0xe08cab08...`), so arc gets per-chain AR / LO cfg
  macros like hyperevm.
- `wrapped_native_token_address` is the 1inch-deployed WETH-style wrapper WUSDC
  (`0x1111161b...`, 18 dec), read from the router's constructor args; `native_token_symbol`
  is `USDC`.
- `transfers_from_traces = false`: arc is on the `base_transfers` schema (EIP-7708 makes native
  transfers events, there is no traces leg), same handling as cronos / robinhood.
[...]

PR #9978, opened 2026-09-01
oneinch: add cronos, monad, hyperevm to AR/LO lineage; fix robinhood mode classification

dennisp42:
## What & why

Adds cronos, monad and hyperevm to the 1inch AggregationRouterV6 and Limit Order Protocol lineage, following the sonic/robinhood per-chain pattern. Cross-chain models stay deferred until escrow decodings land (fantom precedent, exposed ar/lo only). Cronos reads base_transfers like robinhood since it has no transfers_from_traces table.

Also fixes two robinhood mode-classification bugs: the configured fusion settlement addresses aren't deployed there (real settlement 0xb55ba9617dafae1236313c3cb7806439ceefbd13 has 77,902 live calls vs 0 for the config addresses), and the dominant v1.2 escrow factory (0x50d26ea1e2460b3a42ff47466b955fc6bd906013, 5,083 calls vs 1,126 on v1) was missing from escrow_factory_addresses, so its fills classified as limits instead of cross-chain.

All router, settlement, escrow factory and access token addresses verified against live decoded call/trace data on the three chains before landing. Prior art: #9969 (external contributor), which conflicts with main after #9769 rewrote oneinch_ar_executions_macro.sql; reimplemented on main with the same verified addresses, dropping the CI-only easy_date() clamp that PR marked revert-before-merge.

Towards CUR2-3908

### DIFF
--- a/dbt_subprojects/dex/macros/models/_project/oneinch/AR/oneinch_ar_contracts_cfg_macro.sql
+++ b/dbt_subprojects/dex/macros/models/_project/oneinch/AR/oneinch_ar_contracts_cfg_macro.sql
@@ -199,7 +199,7 @@
         },
         "AggregationRouterV6": {
             "version": "6",
-            "blockchains": ["ethereum", "bnb", "polygon", "arbitrum", "optimism", "avalanche_c", "gnosis", "fantom", "base", "zksync", "linea", "sonic", "unichain", "robinhood"],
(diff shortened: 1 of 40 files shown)


dennisp42:
closing in favour of 9969, which does the same three chains and landed first. robinhood cfg is byte-identical between the two, and the settlement_vaults macro that differs is already in main. review continues there: https://github.com/duneanalytics/spellbook/pull/9969

PR #9984, opened 2026-09-07
feat(dex): add Metric (oracle-based OMM) base trades on 10 chains

mathewae:
## What
Adds `metric_<chain>_base_trades` for ethereum, base, arbitrum, bnb, avalanche_c, polygon, monad, hyperevm, megaeth, robinhood and wires them into `dex_<chain>_base_trades` → `dex.trades`. Registers `metric` in `dex.info`.

## About Metric
Metric (https://metric.xyz) is an oracle-based on-chain market maker: pools quote from a price provider and fill swaps against bins of liquidity. Three contract generations are live (deployed 2026-02-23, 2026-07-13, 2026-08-27); pools are created by a `MetricOmmPoolFactory` → `MetricOmmPoolDeployer` (CREATE) pair whose addresses are identical on all chains.

## How the model works
* Contracts are decoded on Dune under project `metric` (multi-chain submission → `metric_multichain`, filtered on `chain`).
* Trades: the decoded pool `Swap` events `metric_multichain.metricommpoolv{0,1,2}_evt_swap`. Positive delta = token into the pool (sold by taker), negative = out (bought by taker). The v2 `amountDeltas` word packs the two int128 deltas (amount0 in the high half) and is unpacked with uint256 division / modulo.
* Token identification: token0/token1 are pool immutables, taken from the factory's `PoolCreated` event.
  * gen 0: decoded `metric_multichain.metricommpoolfactoryv0_evt_poolcreated`.
  * gen 1/2: `PoolCreated` has 21 arguments, above Dune's 16-argument decoding limit, so its decoded view is not queryable (`Column reference 'decoded.args_17' is invalid`). Following Dune's documented workaround, only its three indexed topics (pool, token0, token1) are read from the factory's `logs` — no payload parsing.
* Source declaration: `metric_multichain` added to `sources/_sector/dex/trades/ethereum/_sources.yml` (next to `agra_multichain`).

## Validation
* Output checked on Dune for all 10 chains: row count equals the decoded non-zero swaps per chain/version (~13.7M trades), no duplicate `(tx_hash, evt_index)`, no null token addresses, no zero amounts.
* Decoded swap counts equal the raw `Swap` logs per pool on every chain; a recent block window per chain was also compared against RPC nodes pool by pool.
* Pool tokens match each pool's on-chain `getImmutables()` for every pool.
* Amounts were previously cross-checked against ERC20 transfers for ~215k swaps (exact apart from 1-wei input roundings; 30 zero-amount swaps are filtered out).

## Seeds
One transfer-verified trade per chain/version in `seeds/trades/metric_<chain>_base_trades_seed.csv` (17 rows, all reproduced by the new model).


mathewae:

mathewae:
@0xRobin @jeff-dude ready for review whenever you have a moment. Thanks!

dbt_subprojects/dex/macros/models/_project/metric_compatible_trades.sql
```diff
@@ -0,0 +1,156 @@
+{#
+    Metric (oracle-based on-chain market maker) — base_trades macro.
+
+    Methodology
+    -----------
+    Metric pools (MetricOmmPool) are full contracts (no proxies). Three contract generations exist; in each one a
+    MetricOmmPoolFactory calls a MetricOmmPoolDeployer which CREATEs the pool. None of the contracts are verified on
+    public explorers, so this model reads raw logs and decodes the Swap payloads directly. Amounts were cross-checked
```
0xRobin, line 8: We don't need them to be verified on public explorers, we just need the ABIs which you seem to have tracked down somewhere?

mathewae: Yes - we have the ABIs. I will submit them for decoding under project metric.

0xRobin: Thanks, we'll need to replace the manual decoding in this PR with the metric specific decoded logs and traces tables.

mathewae: Done - the contracts are decoded under metric, and the model now reads the decoded swap tables (metric_multichain.metricommpoolv{0,1,2}_evt_swap), so there's no more raw log parsing. One exception: the v1/v2 PoolCreated event has 21 args, which is over Dune's 16-arg decoding limit, so for those pools I only read the pool and token addresses from its indexed topics, as Dune's docs suggest. I checked the output on all 10 chains: seeds match, and row counts equal the decoded swaps.

0xRobin (approved):

PR #9999, opened 2026-09-13
fix(safe): dedupe mantle in transactions_all and add berachain/ronin

kriss39:
### Description:

`safe.transactions_all` has listed `safe_mantle_transactions` twice since #6942 (Mantle was already in the list from #6777), so every Mantle Safe transaction appears twice in the union view. This removes the duplicate ref.

While in the file I also added the chains that have per-chain Safe models but were never included in the umbrella views:

- `safe_berachain_transactions` and `safe_ronin_transactions` → `safe.transactions_all`
- `safe_ronin_eth_transfers` → `safe.native_transfers_all`

`safe.safes_all` already had both chains. The per-chain models are generated from the same macros as the chains already in the union, so the column lists line up. `safe_schema.yml` blockchain metadata updated to match (it also had a stray empty entry in the `safes_all` list).

All three models are views, so no backfill is needed.


kriss39:

PR #7496, opened 2025-01-19
add padre bot

clizzard7:
Add padrebot Solana users and trades

PR #7307, opened 2024-12-12
add nova

clizzard7:
Spellbook for https://dune.com/data_hunters/nova

PR #7297, opened 2024-12-11
add cswap solana

clizzard7:
Add SOL for https://dune.com/whale_hunter/chainswap

ALSO correct "bot" column in tradewiz

PR #10001, opened 2026-09-13
fix(solana): add bloom, cswap, nova, padre and unibot to dex_solana.bot_trades

kriss39:
### Description:

Four Solana bot trade models exist, are built in prod, and were never added to the `dex_solana.bot_trades` union:

- `bloom_solana_bot_trades` (#7112)
- `cswap_solana_bot_trades` (#7297)
- `nova_solana_bot_trades` (#7307 — the review asked for it to be added to the union, it never was)
- `padre_solana_bot_trades` (#7496)

All four emit the same column set as the 29 bots already in the list (`bot`, `fee_*`, `is_last_trade_in_transaction`, …), so this only extends the view.

**Scope change:** the first revision also added `unibot_solana_bot_trades`, but that model never projects its `bot` column, so the prod table lacks it and the union failed with `COLUMN_NOT_FOUND`. Fixing the model inside this PR forced CI to rebuild it from 2024-03-23 and the run was cancelled at the 90-minute timeout. The unibot fix is split out into #10004 (needs a maintainer-side run); once its prod table has the column, adding it to the union is a one-line follow-up. This PR is now limited to the four models whose tables already match the union schema.


kriss39:

kriss39:
CI caught a real one: `unibot_solana.bot_trades` had no `bot` column in prod — the model computes `'Unibot' as bot` in its CTE but never projected it in the final select, so the union failed with COLUMN_NOT_FOUND. Pushed a second commit that adds the column (and `on_schema_change='sync_all_columns'` so the incremental prod table picks it up without a full refresh). Updated the description accordingly.

PR #10010, opened 2026-09-16
feat(prices): register Aave v4's Arc assets in prices.tokens

mgrabina:
Follows #10009, which registered Arc tokens for mainnet launch.

Aave v4 is live on Arc with four assets. USDC is already covered by #10009 at `0x3600...0000`. Two more have no row here yet, so their volume prices to NULL:

| Symbol | Address | Decimals | token_id |
|---|---|---|---|
| EURC | `0xbef5f6d51cb62b58e6a8f77868681825c6fe21c1` | 6 | `euroc-euro-coin` |
| WETH | `0x128cc466b61f542da60c70e3aa11c10e19b84edb` | 18 | `weth-weth` |

### Where the addresses come from

Decoded from `arc.logs` rather than taken from documentation. The Core hub at `0x17288dfc86205301064577b98B02b81017e6F79C` emits `AddAsset(uint256 indexed assetId, address indexed underlying, uint8 decimals)`, and the two spokes emit `AddReserve(uint256 indexed reserveId, uint256 indexed assetId, address indexed hub)`. Joining those gives:

```
spoke  reserve  asset  token                                       decimals
Main   0        0      0x3600000000000000000000000000000000000000  6
Main   1        1      0xbef5f6d51cb62b58e6a8f77868681825c6fe21c1  6
Main   2        2      0x171a4217b86a807a64eb94757db6849fb4bdbaa0  8
Main   3        3      0x128cc466b61f542da60c70e3aa11c10e19b84edb  18
Forex  0        0      0x3600000000000000000000000000000000000000  6
Forex  1        1      0xbef5f6d51cb62b58e6a8f77868681825c6fe21c1  6
```

Symbols come from the Aave API's reserve list for chain 5042, matched on `reserveId`. Decimals agree between the on-chain event and the API.

### cirBTC is deliberately left out

Arc's fourth Aave asset is cirBTC at `0x171a4217b86a807a64eb94757db6849fb4bdbaa0`, 8 decimals. The first push priced it off `wbtc-wrapped-bitcoin`, following the Animus/Wirex substitution pattern already in this file. `check_tokens` rejected it:

```
ERROR: wbtc-wrapped-bitcoin Provided symbol: cirBTC does not match CoinPaprika source: WBTC
```

I could not find a CoinPaprika feed for cirBTC, and relabelling it `WBTC` to satisfy the check seemed worse than leaving it out. Happy to add it if you know of a feed, or if the substitution is acceptable and the validator can be waived. Worth noting the existing `ABTC` line has the same shape and presumably predates this check, since only changed lines are validated.

### One thing worth a reviewer's eye

EURC appears twice on Arc. This file already registers EURC at `0x89b50855aa3be2f677cd6303cec089b5f319d72a`, and Aave lists a different contract at `0xbef5f6d51cb62b58e6a8f77868681825c6fe21c1`. Both are 6 decimals. Registering several contracts under one `token_id` is already the norm here, so adding the second is consistent, but you may know whether one of these is a wrapper rather than a canonical Circle asset.

`token_id` choices follow the conventions already in this registry: `weth-weth` on base, linea and ink, and `euroc-euro-coin` in this same file.


mgrabina:

PR #10012, opened 2026-09-17
feat(arc): add foundational metadata -- evms.info, gas.fees, sources, doc blocks

0xRobin:
Adds Arc's foundational metadata, per CUR2-2168.

- `evms.info` row
- `arc` in `evms_blockchains_list`
- `gas.fees`: `arc` in `gas_evm_chains`, new `gas/fees/arc/` model and schema, 4 seed rows
- `sources/_base_sources/evm/arc_base_sources.yml` expanded from the minimal version to the full template set
- `sources/_base_sources/evm/arc_docs_block.md`

The remaining CUR2-2168 spellbook items (`prices_arc.tokens`, the `prices_tokens` union, `prices_native_tokens`) shipped in #10009.

## evms.info

```
(5042, 'arc', 'Arc', 'Layer 1', NULL, 'USDC', NULL,
 'https://explorer.arc.io/', timestamp '2026-05-12 00:00:00',
 NULL, NULL, NULL, true, NULL)
```

Arc is an L1, so codebase, data_availability, settlement and ecosystem are NULL.

`wrapped_native_token_address` is NULL: Arc has no wrapped USDC, the native asset satisfies IERC20 directly. There is a 6-decimal USDC ERC-20 interface at `0x3600...0000`, but it is an interface to the native asset rather than a wrapper.

`explorer_link` returns 200; arcscan.io and arc-scan.io do not resolve.

`first_block_time` is block 0 in `arc.blocks`, which also matches `dune.blockchains.start_date` for arc.

Arc is an odd case here. Block 0 is `2026-05-12 00:00:00`, block 1 is `2026-05-15 13:58:51` and the first transaction is `2026-05-15 15:17:24`, so genesis is a configured timestamp rather than the start of real block production. The issue suggests deriving this from `<blockchain>.transactions`, and the megaeth and plasma rows were both populated that way, but for those chains genesis and first transaction fall within a second of each other so the distinction never came up.

## gas.fees

Arc uses `evm_l1_gas_fees`, like plasma and monad.

Arc's gas token is USDC rather than ETH, so the price path was worth checking: `evm_l1_gas_fees` reads symbol, decimals and address from `dune.blockchains` and the price from `prices.usd_with_native`, both via `native_token_prices`. For Arc those resolve to USDC at 18 decimals, priced around $1, so `tx_fee = tx_fee_raw / 10^18` and `tx_fee_usd` is correct.

The fee breakdown invariant that the schema test asserts holds across two days of Arc data: DynamicFee (8,954,962 txs) and EIP-7702 (62,498) satisfy `gas_price = base_fee_per_gas + priority_fee_per_gas` for every row, and Legacy (1,129,255) and AccessList (2,063) have a NULL priority fee throughout, which the macro already folds into the base fee. No Arc-specific variant is needed, unlike somnia and tempo.

## Sources and doc blocks

Both rendered from `_template.yml.j2` and `_template.md.j2` rather than hand-written; the output is identical to megaeth's under `s/megaeth/arc/`.

All 16 tables the template declares exist for Arc, checked against `information_schema` first. The catalog also holds `contracts_submitted`, `raw`, `signatures`, `event_abi_mapping`, `function_abi_mapping`, `logs_decoded_internal`, `traces_decoded_internal`, `erc1155_arc.evt_URI` and the `*_0000` shadow tables, which the template declares for no chain.

## One edit outside the listed scope

`tokens_arc_base_transfers.sql` said that no `arc.traces` source is declared, which this PR makes false. Dropped that clause.

## Checks

`dbt parse` passes in daily_spellbook, hourly_spellbook and tokens. `dbt compile` passes for `gas_arc_fees`, `gas_fees`, `gas_evm_fees`, `evms_info`, `evms_transactions`, `evms_blocks`, `evms_traces`, `evms_erc20_transfers` and `tokens_arc_base_transfers`. All 16 `doc()` references resolve. `dbt deps` was not run; packages were copied from the primary checkout.

One thing left alone: the generated `arc_transactions_doc` mentions OP-stack fields (`l1_gas_used`, `l1_fee`) that mean nothing on an L1. It reads the same on megaeth, so it is a template issue rather than an Arc one.


jeff-dude (approved):
looks like we'll want overrides in both hourly and tokens?
seems like models planned to run which should not.

jeff-dude:
hourly: `--select gas_arc_fees gas_fees gas_evm_fees`
tokens: `--select tokens_arc_base_transfers+`

PR #10015, opened 2026-09-18
Add Uniswap on Arc to dex.trades

0xRobin:
Adds Uniswap v2, v3 and v4 on Arc to `dex.trades`.

All three versions are decoded on Arc and carry real volume from 2026-05-27:

| version | swaps | factory / PoolManager |
| :--- | ---: | :--- |
| v4 | 4,307,682 | `0x8366a39cc670b4001a1121b8f6a443a643e40951` |
| v3 | 1,592,216 | `0xf0db7b58379503491d857db50ac9ece64c653918` |
| v2 | 119,542 | `0x89e5db8b5aa49aa85ac63f691524311aeb649eba` |

Built on the monad pattern, the most recent chain with v4, but scoped to the `dex.trades` lineage only — see "Scope" below.

## What's here

`models/trades/arc/` — `dex_arc_base_trades`, `dex_arc_trades`, and the three `platforms/uniswap_v{2,3,4}_arc_base_trades`.

`models/_projects/uniswap/arc/` — `uniswap_v4_arc_swaps` (where the v4 swap parsing lives; the v4 base_trades model is a view over it) and `uniswap_v4_arc_aggregator_base_trades`.

Sources for the decoded v2/v3/v4 swap and factory tables, and `arc.creation_traces` in the base sources, which `uniswap_v4_aggregator_hooks` needs once arc joins `uniswap_v4_chains()`.

Chain lists: `dex_evm_chains()`, `uniswap_v4_chains()`, `dex_aggregator_trades` expose_spells, and the two `_schema.yml` blockchain lists on the v4 aggregator models.

## Scope

Seven new models, all of them either upstream of `dex.trades` or required for it to be correct. Checked against the dbt manifest: the arc ancestors of `dex_arc_trades` are exactly the six trade models above.

Deliberately **not** included, because they feed sibling tables rather than `dex.trades`: pools (`uniswap.pools`, `dex.pools`), base liquidity events and the v4 sqrtpricex96 models that price them (`uniswap.liquidity_events`, `uniswap.tvl_daily`), `uniswap_arc_trades` (`uniswap.trades`) and `dex_arc_token_volumes_daily` (`dex.token_volumes_daily`, `tokens.dex_volume_day`). Leaving them out also keeps this PR from touching `uniswap_pools`, `uniswap_trades`, `uniswap_base_liquidity_events` and the `dex_token_volumes_daily` chain list, which every other Uniswap chain shares. Straightforward follow-up if Arc wants them.

The one model here that is not on the `dex.trades` path is `uniswap_v4_arc_aggregator_base_trades`, and it is here for a correctness reason. `uniswap_v4_arc_base_trades` ends with `where not is_aggregator_hook_swap`: swaps on BaseAggregatorHook pools are dropped from `dex.trades` on the assumption they resurface in `dex_aggregator.trades`, where they do not double-count the backing venue. Without this model they would land in neither table.

Arc has no BaseAggregatorHook contracts today — running the registry model's own bytecode test over `arc.creation_traces` (creation code dispatching both `pollTokenJar()` and `pseudoTotalValueLocked(bytes32)`) returns 0. So the model is empty for now. Arc v4 is otherwise hook-heavy — 146,958 of 164,600 pools initialise with a non-zero hook, across 138,848 distinct addresses — none of them aggregator hooks. The model stays because the registry is self-maintaining and will pick one up the day it is deployed.

## Fixed here: v4 taker attribution on chains with no traces leg

`uniswap_compatible_v4_trades` recovers the taker from `tokens.transfers`, ordering with:

```sql
case when token_standard = 'erc20' then array[evt_index]
     when token_standard = 'native' then trace_address end
```

On Arc that native branch is always NULL. Arc's transfers have no traces leg — native movement comes from the EIP-7708 event stream — so `trace_address` is NULL on every row (verified over the last day: 5,162,983 native and 3,386,747 erc20 rows, zero non-null). The erc20 branch is unaffected since it uses `evt_index`.

The address join itself is correct: Arc native rows sit at `0x00…00`, matching the currency v4 PoolKeys carry. But `row_number()` over a null key gives an arbitrary order, so in any transaction with more than one native transfer out of the PoolManager for the same token the taker can be attributed to the wrong swap. Roughly 16% of Arc v4 swaps have native as `currency0`.

Fixed by falling back to `array[evt_index]` when `trace_address` is null — the same key the erc20 branch already uses, and deterministic within a transaction:

```sql
when token_standard = 'native' then coalesce(trace_address, array[evt_index])
```

**This is a no-op on every other chain**: their native transfers always carry a `trace_address`, so `coalesce` returns the same value and the compiled query is semantically unchanged. It does change the macro's fingerprint, so all 18 `uniswap_v4_<chain>_swaps` models rebuild — **deploy with an override rather than a blind full refresh.**

## Verified rather than assumed

Factory addresses aren't needed anywhere: the v2/v3 trade macros discover pools from the decoded factory events. Each decoded schema has exactly one emitter, matching the addresses above. The only hardcoded address in the new code is `pool_manager_addr`.

Arc has no wrapped native token, which is the obvious place this could have broken, so both candidates were checked. `uniswap_compatible_v4_trades` keeps a `native_token_address` override map for chains whose canonical native address differs from the `address(0)` in v4 PoolKeys — Arc's canonical address is `0x00…00`, so no override is needed. And `to_wrapped_native_token()`, the macro that hardcodes WETH-equivalents and silently emits nothing for unknown chains, is only used by ParaSwap models, none of which exist on Arc. Pricing resolves normally: `tokens.erc20` has arc `0x00…00` as USDC at 18 decimals and `prices.usd_with_native` prices it, giving at least one priced side on 93% of v3 and 98% of v4 swaps over the last day.

## Also not included

`dex_project_trades` chain list — only adds empty branches, no non-Uniswap DEX is on Arc.

Multihop and roundtrip trades — monad has them, tempo (the most recent v4 onboarding) does not. Followed tempo. A defensible add if you'd rather match monad.

`oneinch_arc_cfg_macro` sets `dex_base_trades = false` with a comment that no `dex_arc_base_trades` model exists. That comment is now stale, but flipping it changes 1inch model behaviour and LOP fill testing, so it is left alone with a note in `dex_arc_trades.sql` where the passthrough leg would go. Best handled by whoever owns the 1inch lineage.

## Seed tests

`seeds/trades/uniswap_arc_base_trades_seed.csv` — 10 real swaps from 2026-09-01 (2 v2, 3 v3, 5 v4), wired to all three base trades models via `check_dex_base_trades_seed` filtered by version, matching how every other uniswap chain does it.

The rows are derived from the decoded events rather than read back out of the models, so they are an independent expectation: v2 from the Pair Swap event joined to PairCreated, v3 from the Pool Swap event joined to PoolCreated, v4 from the swap call's PoolKey and swapDelta matched to the Swap event by (block, tx, rn), which is where the amounts and direction actually come from.

Two of the v4 rows deliberately carry native `0x00…00` as the bought side, since that is Arc's distinctive path; the rest go through the 6-decimal USDC ERC-20 interface at `0x3600…0000`.

No `dex_aggregator_seed` rows, because Arc has no BaseAggregatorHook contracts — see Scope above.

## Checks

`dbt parse` clean in `dbt_subprojects/dex` (1954 models, 3324 data tests, 508 seeds, 9905 sources, 937 macros). `dbt compile` passes for all seven new models plus `dex_trades`, `dex_aggregator_trades`, `uniswap_v4_aggregator_hooks`, `uniswap_v4_aggregator_base_trades`, the four cross-chain models this PR reverted out of, and `uniswap_v4_swaps` on ethereum, base and monad to confirm the macro change compiles unchanged there.

Compiled output checked: `dex_trades` has a `FROM dex_arc.trades` branch, `uniswap_v4_aggregator_base_trades` reads `uniswap_v4_arc.aggregator_base_trades`, `uniswap_v4_aggregator_hooks` reads `delta_prod.arc.creation_traces`, the leaf models resolve to the `uniswap_v{2,3,4}_arc` decoded tables, and no arc reference remains in `uniswap_pools`, `uniswap_trades`, `uniswap_base_liquidity_events` or `dex_token_volumes_daily`. `dbt deps` was not run; packages were copied from the primary checkout.


tomfutago (approved):
lgtm

prob obvious, but deployment will need an override as `uniswap_v4_chains` update triggers full refresh of all chains on the list.

nit: `dex_aggregator_seed` additions?

0xRobin:
Thanks — both taken.

**Seeds.** You were right that something was missing, though not the one you named.

The gap was `check_dex_base_trades_seed`. Every other chain's uniswap base trades models carry it against one shared per-chain seed filtered by version, and arc had none — I followed tempo, which skips it, rather than monad, which does it properly. Added in 7c096a0: `seeds/trades/uniswap_arc_base_trades_seed.csv`, 10 real swaps from 2026-09-01 (2 v2, 3 v3, 5 v4), wired onto all three models.

The rows are derived from the decoded events rather than read back out of the models, so they're an independent expectation: v2 from the Pair Swap event joined to PairCreated, v3 from the Pool Swap event joined to PoolCreated, v4 from the swap call's PoolKey and swapDelta matched to the Swap event by (block, tx, rn), which is where the amounts and direction actually come from. Two of the v4 rows deliberately have native `0x00..00` as the bought side, since that's arc's distinctive path — native USDC at 18 decimals, and v4 PoolKeys carry `address(0)` for it, so it resolves without the `native_token_address` override other chains need. The rest go through the 6-decimal USDC ERC-20 interface at `0x3600..0000`.

**`dex_aggregator_seed` specifically: there's nothing to add.** Arc has no BaseAggregatorHook contracts at all. Running the registry model's own bytecode test over `arc.creation_traces` — creation code dispatching both `pollTokenJar()` and `pseudoTotalValueLocked(bytes32)` — returns 0 contracts. So `uniswap_v4_arc.aggregator_base_trades` is empty today and a seed row would have nothing to match.

Worth flagging that this corrects the PR description, which said the BaseAggregatorHook share of arc v4 swaps was "unknown rather than negligible" — it's zero. Arc v4 is hook-heavy (146,958 of 164,600 pools initialise with a non-zero hook, 138,848 distinct addresses) but none of them are aggregator hooks. I've kept the model anyway: the registry is self-maintaining, so it picks one up the day it's deployed, and without it those swaps would be dropped from `dex.trades` by the `not is_aggregator_hook_swap` filter and land nowhere.

**Deployment override.** Noted, and agreed — `uniswap_v4_chains()` is referenced by `uniswap_v4_aggregator_hooks`, so adding arc changes that model's fingerprint and triggers a full refresh across all 16 v4 chains. Will flag it on the deploy so it runs with an override rather than a blind full rebuild.

jeff-dude (approved):
we should consider the fix for taker here and just override deploy.

jeff-dude:
override: `--select 'path:models/trades/arc+ path:models/_projects/uniswap/arc+ uniswap_v4_aggregator_hooks'`

dex_aggregator_base_trades
dex_aggregator_trades
dex_arc_base_trades
dex_arc_trades
dex_evm_trades
dex_trades
firebird_finance_optimism_trades
lifi_trades
openocean_trades
unidex_optimism_trades
uniswap_v2_arc_base_trades
uniswap_v3_arc_base_trades
uniswap_v4_aggregator_base_trades
uniswap_v4_aggregator_hooks
uniswap_v4_arc_aggregator_base_trades
uniswap_v4_arc_base_trades
uniswap_v4_arc_swaps
yield_yak_trades

PR #6891, opened 2024-10-06
add tokens transfers for `goerli` and `sepolia`

hildobby:
### Update!
Please build spells in the proper [subproject](../dbt_subprojects/) directory. For more information, please see the main [readme](../README.md), which also links to a GH discussion with the option to ask questions.

### Contribution type
Please check the type of contribution this pull request is for:

- [ ] New spell(s)
- [ ] Adding to existing spell lineage
- [ ] Bug fix

**Note:** You can safely discard any section below which doesn't apply based on selection above

---

### For new spell(s)
If you are building new spell(s), please provide the following information:

- **Spell name(s):** *schema.table_name*
- **Description:** [Detailed description of the new spell(s) and their purpose]
- **Who are the new spell(s) for?** [Internal team or general community]
- **How will the new spell(s) be used downstream?** [Description of downstream usage]
- **Implementation details:** [Information on how the spell(s) are implemented]
- **Test instructions:** [How to test the new spell(s)]
- **Related issue(s):** [Link to related issues, if any]

---

### For adding to existing spell lineage
If you are adding to an existing spell lineage, please provide the following information:

- **Description:** [Description of the changes made]

---

### For bug fixes
If you are fixing a bug, please provide the following information:

- **Description:** [Description of the bug fix]
- **Steps to reproduce:** [How to reproduce the bug]
- **Implementation details:** [Information on how the bug was fixed]
- **Test instructions:** [How to test the fix]
- **Related issue(s):** [Link to related issues, if any]

---

### Additional information
Please provide any additional information that might help us review your pull request:

- [Any additional information]

---

Thank you for your contribution!

PR #10028, opened 2026-09-22
fix(tokens): restore the sepolia schema yml that dbt was silently ignoring

0xShadowX:
## What

`dbt_subprojects/tokens/models/transfers_and_balances/sepolia/_schema.yml` declares its
models under a top-level `sepolials:` key instead of `models:`.

It looks like a chain-name search-and-replace when the file was copied — `"mode"` → `"sepolia"`
also rewrote the `mode` inside the word `models`.

## Why it matters

dbt does not warn about unrecognised top-level keys in a schema yml, it just ignores them. So
this file has been a no-op since it was added in #6891:

- `tokens_sepolia_base_transfers` and `tokens_sepolia_transfers` have no description and no
  column documentation in the manifest;
- the `dbt_utils.unique_combination_of_columns` test declared for **both** models on
  `(block_date, unique_key)` has never been registered, so the uniqueness invariant for the
  Sepolia transfer tables is currently untested.

The sibling chain files in the same directory all use `models:`.

## Change

One line — `sepolials:` → `models:`. No model SQL is touched.

Once merged, the two uniqueness tests start running for the first time.


0xShadowX:

PR #9169, opened 2026-01-02
Add Kaia CEX addresses and flows

JoongJePark:
## Thank you for contributing to Spellbook 🪄
Please open the PR in **draft** and mark as ready when you want to request a review. 

## Summary
- Added 69 Kaia CEX addresses to `cex_evms_addresses.sql`
- Created `cex_kaia_addresses.sql` model following EVM chain pattern
- Created `cex_kaia_flows.sql` model for Kaia CEX flow tracking
- Updated `cex_addresses.sql` to include Kaia in aggregation
- Added Kaia to expose_spells configuration
- Added schema definitions for Kaia CEX models

## Contributor
anton

---
quick links for more information:
- [README.md](https://github.com/duneanalytics/spellbook/blob/main/README.md)
- [spellbook docs](https://github.com/duneanalytics/spellbook/tree/main/docs)
- [CONTRIBUTING.md](https://github.com/duneanalytics/spellbook/blob/main/CONTRIBUTING.md)

PR #10030, opened 2026-09-22
fix(cex): add kaia to cex.flows

Bruce039:
## What

`cex_kaia_flows` is missing from the `cex_models` list in `cex_flows.sql`, so Kaia rows never
reach the cross-chain `cex.flows` table.

## Why it happened

#9169 ("Add Kaia CEX addresses and flows") added both sides of the chain:

- `cex/addresses/chains/kaia/cex_kaia_addresses.sql` — **is** unioned in `cex_addresses.sql`
- `cex/flows/chains/cex_kaia_flows.sql` — is **not** unioned in `cex_flows.sql`

So the addresses half landed and the flows half did not.

## Change

- add `, (ref('cex_kaia_flows'))` to `cex_models`, in the alphabetical block where the newer
  chains are listed (between `ink` and `katana`);
- add `kaia` to the `blockchain:` metadata list for `cex_flows` in `_schema.yml`, so the model
  page advertises the chain like it does for every other one in the union.


Bruce039:

Bruce039:
The `dbt-ci/hourly_spellbook` failure is not a SQL error. The run fails while creating `cex_flows` with `TrinoQueryError(INTERNAL_ERROR, EXCHANGE_MANAGER_NOT_CONFIGURED)` (run 35765144837). The CI schema starts empty, so the incremental model is built from scratch over every chain's flows, and that query is too big for the CI cluster. Could someone re-run it or build it on your side?

Also, since `cex_flows` is incremental with `incremental_predicate('block_time')`, the model will only pick up recent Kaia rows after merge. Kaia's history needs a `--full-refresh` of `cex_flows` (or a one-off backfill).

PR #10031, opened 2026-09-22
fix(dex): repair seed tests that were comparing nothing

Bruce039:
## What

Four `check_dex_base_trades_seed` tests are not testing what they look like they test.

## Why a wrong filter is silent

`check_seed_macro` builds `matched_records` **from the seed file**, applies `filter` to the seed
rows, and the test finally selects only the rows `where equality_check is distinct from true`.

If the filter matches no seed rows, `matched_records` is empty, both the `matching_count_test`
and the `equality_tests` CTEs are empty, and the test returns zero rows — which dbt reads as a
pass. A filter pointed at the wrong value is therefore indistinguishable from a green test.

## The four cases

| Model | Problem | Fix |
|---|---|---|
| `cldex_kaia_base_trades` | `seed_file: ref('cldex_kaia_base_trades')` is the **model**, not the seed. The macro joined the model to itself, so every equality check compared a row with itself and passed. `cldex_kaia_base_trades_seed` was never read. | point `seed_file` at `ref('cldex_kaia_base_trades_seed')` |
| `derpdex_base_base_trades` | test filters the seed to `version: 1`; the model declares `version = '3'` and all three seed rows are version 3 → 0 rows compared | filter on `version: 3` |
| `gammaswap_arbitrum_base_trades` | test filters on `version: 1` (correct, the model declares `version = '1'`), but the only seed row is labelled `version 2` → 0 rows compared | relabel the seed row to version 1 |
| `dackieswap_v3_arbitrum_base_trades` | seed row labelled `blockchain=optimism, project=arbitrum` | relabel to `arbitrum` / `dackieswap` |

The dackieswap one is cosmetic — `blockchain` and `project` are neither matching columns
(`tx_hash`, `evt_index`) nor check columns (`block_number`, the two token addresses, the two raw
amounts), so that test *was* running. The labels were simply wrong, and wrong labels in a seed
are how the other three started.

No model SQL is touched.

## A note on CI

The two seed-CSV changes should pull their tests into the run. The two schema-only changes
(cldex, derpdex) may not be picked up by `state:modified.body` selection, so you may want to run
those two tests manually to see them go from "passing on zero rows" to actually comparing the
seed.


Bruce039:
The `dbt-ci/dex` failure is the cldex test doing its job now: `FAIL 6` is 2 columns × 3 seed rows. In every row of `cldex_kaia_base_trades_seed.csv` the `token_bought_address` and `token_sold_address` values were reversed. The amounts were already right.

Checked against the Swap events on Kaia (pair `token0`/`token1` with `amount0Out`/`amount1Out`, same logic as `uniswap_compatible_v2_trades`). For example, in `0x775a024d…5a63` (log 7) the trader sells 16.248 USDT (`0xd077…4fdb`, 6 decimals) for 105.18 WKLAY (`0x19aa…4432`). The seed had these two addresses the other way around. Pushed 23b18ac to swap them in all three rows.

PR #10032, opened 2026-09-22
fix(addresses_events): add unichain to the cross-chain unions

FlashWayne:
## What

`addresses_events_unichain_first_activity`, `_first_funded_by` and `_first_token_received` exist
and are built, but none of them is referenced by the corresponding cross-chain union, so Unichain
rows never appear in `addresses_events.first_activity`, `.first_funded_by` or
`.first_token_received`.

## Why it happened

#8497 ("add check deposit address was recently funded") added three chains at once — `ink/`,
`opbnb/` and `unichain/` — and updated all three union files in the same PR. Only `ink` and
`opbnb` got their refs added; `unichain` was left out of all three.

You can still see the grouping in the current union lists: `katana, ink, opbnb` sit together,
then `hyperevm, monad` were appended later.

## Change

- `addresses_events_first_activity.sql`: add `ref('addresses_events_unichain_first_activity')`
- `addresses_events_first_funded_by.sql`: add
  `(ref('addresses_events_unichain_first_funded_by'), 'ETH')` — Unichain's native asset, matching
  how the other OP-stack chains in the list are declared
- `addresses_events_first_token_received.sql`: add
  `(ref('addresses_events_unichain_first_token_received'))`
- `addresses_events_schema.yml`: add `unichain` to the `blockchain:` metadata for the three models

Each ref is inserted right after `opbnb`, i.e. in the slot the #8497 group would have used.
No per-chain model is touched.


FlashWayne:

PR #10036, opened 2026-09-23
Fix pump.fun trader attribution for OKX swaps

karimhass:

### Description:

Attribute pump.fun swaps routed through the affected OKX instructions to the router's user instead of its shared `ARu4…5SZn` authority.
Adds one scoped lookup in the existing model; other routes retain the event user.
Historical corrections require rebuilding the affected pump.fun and downstream DEX partitions after merge.

### Architecture

- `pumpdotfun_solana_base_trades.sql`: match the top-level OKX instruction by transaction, time and outer instruction index; use its first account only when the pump.fun event user is the known shared authority. Restrict to the two observed instruction discriminators and retain the event user when no match exists.
- No schema, amount, fee or key changes. Downstream impact: trader attribution in `pumpdotfun_solana.trades` and `dex_solana.trades`.
- Validation: local dbt compile and `git diff --check` passed. The [compiled model before/after comparison](https://dune.com/queries/8818304) preserves all 211,867 rows, unique event count and both raw-amount checksums; it corrects 3,048 trader IDs, including both reported transactions. The 29 proxy events under other outer programs are unchanged.
- Needs human eyes: CI table results and the production historical rebuild. Nested OKX calls and other router variants are outside this fix.

<details><summary>Agent context</summary>

- No agent review yet.
- [Router user validation](https://dune.com/queries/8818313): all 77 sampled cases where the router user differs from the transaction fee payer have that user in the signer list and token owners. This is why the fix reads the router account instead of replacing it with `tx_signer`.
- [Reported transactions and token ownership](https://dune.com/queries/8818161); [event users versus transaction signers](https://dune.com/queries/8818164).
- Discriminators: `0xaa2955b184501f35`, `0xbbc9d433109bec3c`. Router: `proVF4pMXVaYqmy4NjniPh4pqKNfMmsihgd4wdkCX3u`. Shared authority: `ARu4n5mFdZogZAravu7CcizaojWnS6oqka37gdLT5SZn`.
- Validation windows: March 20, 2026 14:00–15:00 UTC and March 26, 2026 19:00–20:00 UTC. Production keeps normal incremental filters.
- Local compile: `uv run dbt compile --project-dir dbt_subprojects/solana/ --profiles-dir dbt_subprojects/solana/ --select pumpdotfun_solana_base_trades --no-introspect --no-populate-cache`.
- [Request thread](https://app.slack.com/archives/C0BMEQSKY80/p1789650470317199).

</details>

Prompted by: Karim Hassan

---


jeff-dude:
Pushed a simplification: I dropped the extra `instruction_calls` read of the OKX router and its join, and swapped in a `tx_signer` fallback scoped to OKX's shared authority (`ARu4…5SZn`). `tx_signer` is already on the swap rows, so the fix no longer needs another pass over that heavy table.

**Coverage, on the same two validation windows (211,867 rows)**

| | Rows |
|---|---|
| Swaps where the pump.fun event's user is the shared authority | 3,077 |
| Rows the original join fixed | 3,048 |
| Of those, where `tx_signer` = the OKX router's user | 2,971 (97.5%) |
| Of those, where they differ (sponsored or relayed transactions) | 77 (0.036% of all rows) |

The fallback also catches 29 shared-authority rows under other outer programs that the join never matched (`tx_signer` is never the shared authority itself).

**Why the remaining 77 rows are acceptable**

Every other Solana DEX spell already uses the fee payer as `trader_id`, so they have the same gap. I sized it across all of `dex_solana.trades` for one hour (2026-09-20 12:00–13:00 UTC). In 1.8% of rows, a signer other than the trader received the bought token (for example raydium 6.8%, meteora 1.7%, pumpswap 0.4%). Pump.fun's leftover 0.036% sits well inside that. Fixing it properly means changing how `trader_id` is defined across the whole sector, which doesn't belong in this PR.

jeff-dude (approved):

